PR #7, opened 2017-06-03
remove anything that breaks if resolvconf service is referenced

georgejhunt:
This is an oversight. dhcpcd replaces resolvconf in the current iiab playbook


holta:
Just a Footnote as to how this showstopper bug arose a month ago, while making an essential fix to shutdown/bootup times: (which had been painfully slow over sev minutes!)
https://github.com/XSCE/xsce/pull/913

PR #14, opened 2017-06-04
Kiwix

georgejhunt:
smoke tested in CentOS, and raspbian.
upgrade from centos xsce-6.2 required deletion of /etc/xsce/xsce.env and /library/zims/library.xml
on pi I deleted /library/zims/library.xml and it worked
It does not look like there is a i386 version available -- don't think win32 version cuts it.

### DIFF
--- a/roles/kiwix/tasks/main.yml
+++ b/roles/kiwix/tasks/main.yml
@@ -13,7 +13,7 @@
 
 - name: Set kiwix source file name armv7l
   set_fact:
-     kiwix_src_file: "kiwix-tools_armhf_2017-06-02.tar-gz"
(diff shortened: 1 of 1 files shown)


tim-moody:
is the intel http://download.kiwix.org/nightly/2017-06-04/kiwix-tools_linux64_2017-06-04.tar.gz?

holta:
@tim-moody who are you asking about http://download.kiwix.org/nightly/2017-06-04/kiwix-tools_linux64_2017-06-04.tar.gz -- are you asking @georgejhunt ?

Or is this a matter of just gambling and seeing if it works on x86_64 ?

tim-moody:
sorry read a comment before looking at the play. name is OK

Issue #8, opened 2017-06-03
box and box.lan need to work on "all" IIAB's, regardless of local_vars.html & Admin Console


holta:
http://box does not show up until forced thru local_vars.html or similar.  Hence all documentations's is flawed, as all our documentation assumes newbies instantly have access to http://box/admin (etc).

So let's make sure "box" and "box.lan" work truly instantly, no matter what (on Raspbian especially :)

holta:
Given that this bug renders most all documentation nearly useless to newbies (who seek to use http://box/admin etc) I'd suggest we change Line 46 of https://github.com/iiab/iiab/blob/master/vars/default_vars.yml as immediately as possible.

Line 46 currently reads:
xsce_hostname: schoolserver

It should read:
xsce_hostname: box

Or is there a better way?

holta:
@georgejhunt @tim-moody @jvonau

In the past these 3 files seemed to work:
https://github.com/iiab/iiab/blob/master/roles/network/templates/named/school.external.zone.db
https://github.com/iiab/iiab/blob/master/roles/network/templates/named/school.internal.zone.db
https://github.com/iiab/iiab/blob/master/roles/network/templates/named/school.local.zone.db

As documented in FAQ #13 @ http://FAQ.IIAB.IO
("Can I name my server something other than http://BOX.LAN ?")

However I can reconfirm this is not working on CentOS and Raspbian today (newbies cannot access http://box/admin etc, with a strong guarantee that this will work in all conditions, regardless of local_vars.yml).  Changing Line 46 of default_vars.yml would seem to be an immediate fix (to "xsce_hostname: box") but others may know a better way?

jvonau:
How are you accessing the webserver? From a remote client browser or from the server itself?

holta:
Clarif for @jvonau :

box and box.lan fail to work from the server itself :/

jvonau:
Yea I suspect that /etc/resolv.conf is wrong

jvonau:
bit of background reading: https://serverfault.com/questions/513271/why-dont-my-server-query-the-2nd-entry-in-my-resolv-conf 

The only nameserver that can resolve "box" is local so one would need to have it's ip address(I like to use 127.0.0.1 but 172.18.96.1 should work if br0 is present).

holta:
[...]

PR #1, opened 2017-05-30
changes to move admin console to its own repo

tim-moody:

PR #15, opened 2017-06-04
change the default servername from schoolserver to box

georgejhunt:
I manually changed the /etc/hosts file 172.18.96.1   schoolserver.lan   schoolserver to 172.18.96.1 box.lan box and confirmed Adam's observation that on the local console, this is sufficient to fix the problem.

Associated to the wifi, box resolves even without this change.

### DIFF
--- a/vars/default_vars.yml
+++ b/vars/default_vars.yml
@@ -43,7 +43,7 @@ local_tz: "{{lookup ('env','TZ') }}"
 
 # Network Parameters
 
-xsce_hostname: schoolserver
+xsce_hostname: box
 xsce_domain: lan
 lan_ip: 172.18.96.1
 lan_netmask: 255.255.224.0


holta:
Profound Thanks @georgejhunt 

Hopefully @jvonau and @tim-moody can comment or confirm this is an acceptable (or sufficient for now) way to resolve #8 (which has been making newbies' lives impossible).

jvonau:
I'm not part of this repository so "@" does not work. The name resolution stack is broken as mentioned in #8. This is a hack around not having local dns working correctly but will work for the short term. Be warned that others gremlins might occur elsewhere.

jvonau:
One could also just add 'box' as an alias in /etc/hosts.

holta:
Ref @jvonau's #16 idea, possibly updating /etc/hosts with an alias (also a great idea, so newbies who override their hostname don't suddenly lose access to http://box/info documentation!)

tim-moody:
I have also made this change in #1

PR #16, opened 2017-06-04
Update hosts.yml

jvonau:
add 'box' as a short alias name to ensure the web docs are always available at http://box/info


jvonau:
The advantage of changing just /etc/hosts with this PR is that the fix could be implemented by running xsce-network in place of a full runansible that PR #15 would need.

georgejhunt:
thanks for your PR. Turns out that Tim had already included this change

jvonau:
implemented by RP #48 not sure why there would not be conflicts now... the code s/b there....

PR #18, opened 2017-06-05
Fixes

georgejhunt:
Document what I have found running runansible, and iiab-admin/install. Still running ICO, and finding things

### DIFF
--- a/roles/2-common/tasks/packages.yml
+++ b/roles/2-common/tasks/packages.yml
@@ -76,6 +76,7 @@
    - wget
    - openssl   #FC 18 does not supply, but pear requires
    - gawk
+   - curl
   tags:
     - download
 
--- a/roles/network/tasks/debian.yml
+++ b/roles/network/tasks/debian.yml
@@ -25,7 +25,7 @@
   ignore_errors: True
 
 - name: Get the dhcp client daemon used in recent raspbian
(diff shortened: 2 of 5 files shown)


tim-moody:
hope there are not too many of these things that were sitting in my PR that may have been done elsewhere

georgejhunt:
After ICO, on debian VM, the following were checked and worked:
nextcloud, kiwix, kalite, sugarizer, munin
failed: awstats (probably not enough time to accumulate log)
          moodle -- we need a solution
Tim do these changes look ok to  you? (master is currently broken and I'd like to fix as soon as possible)

tim-moody:
I took all the is_OS: false out of the OS.yml and put in default_vars rather than having them spread all over the place.

tim-moody:
you need to reboot before awstats works

tim-moody:
how is master broken (not rhetorical)?

tim-moody:
for some reason default_vars has is_rpi:true. should be false

georgejhunt:
So I discovered an is_ubuntu, and an is_redhat missing, when I was running
iiab-admin.

I guess I need to let you fix it the way you want to.

georgejhunt:
These changes demonstrate the brokenness of master. My fault -- for not better testing what I pulled into master after this morning's call

PR #26, opened 2017-06-12
Osm

georgejhunt:
smoke tests on debian VM, and brings up the search window on a remote client. Not tested with tiles.
iiab->maps


tim-moody:
in  roles/osm/templates/defaults.ini does it make sense to remove the sections other than [OSM]?

tim-moody:
this will affect fix osm in factor. maybe should just pull that into ansible

holta:
Clarif: @tim-moody is referring to the fix-osm script in https://github.com/georgejhunt/iiab-factory/tree/master/content/osm-fixes

@georgejhunt, @floydianslips & all agreed this was a good idea, during this morning's meeting (http://minutes.iiab.io)

ASIDE: ./fix-osm should be removed from [make-content-ready.sh](https://github.com/iiab/iiab-factory/blob/master/box/generic/make-content-ready.sh)

PR #27, opened 2017-06-12
WordPress 4.8 instead of 4.7.3

holta:
WordPress 4.8 was released 4 days ago and I tested this on Raspbian, based on:

   http://download.iiab.io/packages/wordpress-4.8.tar.gz

If folks care about testing this on Debian, CentOS, Ubuntu, etc please do that promptly in coming days so this can be pushed/merged into master, Thanks!


tim-moody:
no objection, but didn't try it

PR #28, opened 2017-06-12
on/off vpn

georgejhunt:
smoke tested and tested for function of a pi


holta:
Incredible, Profound Thanks @georgejhunt !

I'll test this as soon as I can when I get home later tonight.

PR #29, opened 2017-06-13
remove menu from iiab

georgejhunt:


tim-moody:
looks OK to me, but didn't try it.

Issue #9, opened 2017-06-03
http://box/info does not show offline docs on new installs


holta:
http://box/info does not show our great offline documentation on new installs.

That is until I manually run "cd /opt/schoolserver/iiab/scripts; ./refresh-wiki-docs.sh"

How should we enforce [programmatically, even if just socially for now?] that this is run during all installs?

PS box/info IT/Support docs originated in IIAB/XSCE 6.2.  This issue is slightly exacerbated by Issue #8 (box and box.lan don't appear on many fresh installations, per https://github.com/iiab/iiab/issues/8) but really that is separate!

georgejhunt:
proposal to move /scripts/refresh-wiki-docs to roles/3-base-server/templates, run it during install, and copy it to /usr/bin/xs-refresh-docs for later end user update.
maybe all the xs-* binaries should be replaced with iiab-*

holta:
Either xs-refresh-docs or iiab-refresh-docs are A-Ok with me!

As @tim-moody outlined Thursday, there are many cases where "xs" and "xsce" should eventually be replaced wholesale with "iiab" in a methodical way, rather than in a piecemeal fashion?

PR #31, opened 2017-06-13
Delete legacy scripts/refresh-wiki-docs.sh

holta:
A new version of this exists here:
/usr/bin/xs-refresh-wiki-docs

In future it should move here:
/usr/bin/iiab-refresh-wiki-docs

Related to larger issue of http://box/info (offline docs) being refined in coming days/weeks: #9


tim-moody:
how does the new version get there?

holta:
@georgejhunt : can you help answer @tim-moody 's question as to how the new version /usr/bin/xs-refresh-wiki-docs gets there?

(Then we can push this deletion of the legacy script, so we don't have duplicate copies...)

georgejhunt:
Please don't delete this model until I have another script that works.
It's nice to look back and compare what was working with what is not yet
working

PR #40, opened 2017-06-16
Logkalite

georgejhunt:
Not yet smoke tested -- but worked by hand. It shadows lots of items at the root of the document root, so I'm not sure if it is a good idea.
Tim -- your thoughts?


georgejhunt:
please ignore the osm crap at the end of the diff. I'll fix before I do the smoke test

tim-moody:
not too enthusiastic about having things like management locked into kalite. maybe try introducing a more significant part of the url like /kalite, but it may require url rewriting on the page output.

georgejhunt:
replaced by #42

PR #42, opened 2017-06-17
Logkal

georgejhunt:
I think this addresses Tim's concerns. But I don't know if it works yet

### DIFF
--- a/roles/kalite/tasks/install.yml
+++ b/roles/kalite/tasks/install.yml
@@ -66,3 +66,14 @@
   with_items:
     - { src: 'kalite-serve.service.j2', dest: '/etc/systemd/system/kalite-serve.service', mode: '0644'}
     - { src: 'kalite.sh.j2', dest: '/etc/profile.d/kalite.sh', mode: '0644'}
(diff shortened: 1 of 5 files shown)


roles/kalite/tasks/install.yml
```diff
@@ -66,3 +66,14 @@
   with_items:
     - { src: 'kalite-serve.service.j2', dest: '/etc/systemd/system/kalite-serve.service', mode: '0644'}
     - { src: 'kalite.sh.j2', dest: '/etc/profile.d/kalite.sh', mode: '0644'}
+    - { src: 'kalite.config', dest: '/etc/{{ apache_config_dir}}', mode: '0644'}
+
```
jvonau, line 70: should that be kalite.conf to match the template file name where kalite.config is?

jvonau (changes requested):

roles/kalite/tasks/install.yml
holta: @georgejhunt shouldn't .config be .conf as jvonau suggests ?

Plz confirm Y/N!

Issue #46, opened 2017-06-19
The use of ansible_lsb may cause a failure on CentOS & OS.yml file naming


jvonau:
The use of ansible_lsb can cause a failure on CentOS or Fedora if the redhat-lsb-core rpm is not installed(1). This optional rpm(2) is not generally installed by default(3) on server class installs but can be drawn into the mix by other packages that still list it as a dependency(4). Now with the move to systemd replacing initscripts this is more of a backwards compatibility package like net-tools providing the legacy ifconfig command which has been replaced with the newer ip addr usage. If one wishes to ensure this ansible module is available adding this rpm as an ansible dependency to scripts/ansible makes sense unless one's goal is to be a x11 free install.

1. https://github.com/v1v/ansible-role-chef-client/issues/5
2. http://mirror.centos.org/centos/7/os/x86_64/repodata/d918936f5019be3fb66e9981a28cb2a41477a2963d741d454f79377a22214f43-c7-x86_64-comps.xml
3. https://serverfault.com/questions/821874/how-to-install-a-minimal-version-of-lsb-release-on-centos
4. https://blogs.wcode.org/2015/01/fake-lpr-or-how-to-install-redhat-lsb-core-without-cups-and-x11/

tim-moody:
assuming that ansible_lsb uses it. on debian with no lsb_release command available, ansible_lsb still has the correct values. I'm pretty sure I checked this on centos as well, and the values are there even though the package is not.

jvonau:
Well nothing like not guessing and going to the source(1). It appears like it first looks for the lsb_release binary if found uses it, then falls back to parsing /etc/lsb-release if present but is not present on redhat based products. Why the change to iiab.yml going back to using ansible_distribution in aa107858736f9e1da8523e87e22844088b934d96 ?

1. https://github.com/ansible/ansible/blob/stable-2.2/lib/ansible/module_utils/facts.py

tim-moody:
an excellent question. I hadn't noticed the commit.

jvonau:
Just trying to future proof the code, FWIW Debain 9 is moving away from lsb(1).
1. https://www.debian.org/releases/stretch/amd64/release-notes/ch-information.en.html#reduced-lsb-support

georgejhunt:
I'm resonsible for
https://github.com/iiab/iiab/commit/aa107858736f9e1da8523e87e22844088b934d96,
because I was tryiing to do the massive xsce->iiab change, and ansible_lsb
[...]

PR #58, opened 2017-06-23
fix for centos

jvonau:
#46


tim-moody:
small point, but the dotted notation is easier ( ansible_local.local_facts.os )

jvonau:
The whole stanza can go away with the other fix. close this PR when this stanza is removed.

jvonau:
#65

PR #67, opened 2017-06-27
disable, and turn off apparmor

georgejhunt:
treat in a manner similar to selinux

### DIFF
--- a/roles/1-prep/tasks/main.yml
+++ b/roles/1-prep/tasks/main.yml
@@ -23,6 +23,10 @@
             dest=/etc/chrony.conf
             src=chrony.conf.j2
 
+- name: Disable apparmor -- on by default in ubuntu
+  service: name=apparmor enabled=False state=stopped
+  when: is_ubuntu
+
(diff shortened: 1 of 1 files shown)


holta:
@jvonau & all -- FYI those that want to keep AppArmor (Ubuntu installs this by default) can use the following 3 lines to make dhcpd & bind9 run:

1) I got dhcpd to run by adding this to /etc/apparmor.d/local/usr.sbin.dhcpd :

   /etc/dhcpd-xs.conf r,

2) I got bind9 to run by adding this to /etc/apparmor.d/local/usr.sbin.named :

   /etc/named-xs.conf r,
   /var/named-xs/** rw,

holta:
PS let's all try to better understand Ubuntu's security model (and the wider consequences of disabling AppArmor, so we don't endanger schools in the developing world especially) and then document these security recommendations @ http://wiki.laptop.org/go/IIAB/Security

(Comparing Raspbian & CentOS security models as well!)

Issue #69, opened 2017-06-28
Use of /etc/php.ini


jvonau:
The variable local_tz can be blank if TZ has not been exported to the environment first.

jvonau:
From a session with TK no TZ present:
        "ansible_env": {
            "ANSIBLE_LOG_PATH": "/opt/iiab/iiab/iiab-install.log", 
            "DISPLAY": ":0", 
            "HOME": "/root", 
            "LANG": "en_US.UTF-8", 
            "LANGUAGE": "en_US.UTF-8", 
            "LOGNAME": "root", 
[...]

PR #70, opened 2017-06-28
Timezone handling

jvonau:
Clean up the use of local_tz #69


holta:
@georgejhunt can we review this & related stuff during Monday's http://minutes.iiab.io call, in case it might be contributing to Elgg 2.3.4 not installing?

(@jvonau suggests it might possibly be related to #330, or at least offering us a line of attack to get Elgg solved for Mike & TK ideally ;)

PR #73, opened 2017-06-28
Notags

georgejhunt:
This supplants PR 72 -- not yet tested


jvonau:
Removing DL DL2 tags that work really doesn't gain anything and removes the ability to pre-seed, but I like the internet_available change

PR #76, opened 2017-06-28
treat ubuntu dhcpd similarly to debian

georgejhunt:
Even when there is only one LAN interface, put it under the bridge, and the dhcpd always attaches to the bridge.

### DIFF
--- a/roles/1-prep/tasks/detected_network.yml
+++ b/roles/1-prep/tasks/detected_network.yml
@@ -250,7 +250,7 @@
 - name: for debian, always use bridging
   set_fact:
      iiab_lan_iface: br0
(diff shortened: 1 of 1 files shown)


jvonau:
Hand tested with TK it works please push

jvonau:
think this is a dup of https://github.com/iiab/iiab/pull/73/commits/e9059199be05f7bd12da3ea47b1e07a0d6b0d451

jvonau:
PR #84 has this change in a different branch but this change is not in master so the problem still exists on master

georgejhunt:
Seems well tested by me and TK with jerry's help.

PR #83, opened 2017-06-30
Recognize OS early in bash, and use it everywhere

georgejhunt:
Smoke tested on ubuntu, debian, centos


tim-moody:
The real question is chicken and egg. Adding yet another script to calculate something to discover in iiab.env seems an unnecessary complication.  What all will use this env variable? Anything besides iiab.env?

My preference was to have iiab.env do the calc and get sourced elsewhere, including local facts, but Jerry took the approach of letting local_facts calc the os and then set iiab.env for use elsewhere. To me the question is do we need anything before ansible runs? If not, I have come around to Jerry's approach. (#58, #65 )

I should add that I did the same os calc in curl-me which does precede ansible, but neither iiab.env nor local_facts are in place yet (nor would first.sh be) and I think this is a reasonable redundancy.

georgejhunt:
On Fri, Jun 30, 2017 at 11:35 AM, Tim Moody <notifications@github.com>
wrote:

> The real question is chicken and egg. Adding yet another script to
> calculate something to discover in iiab.env seems an unnecessary
> complication. What all will use this env variable? Anything besides
> iiab.env?
>
> My preference was to have iiab.env do the calc and get sourced elsewhere,
> including local facts, but Jerry took the approach of letting local_facts
> calc the os and then set iiab.env for use elsewhere. To me the question is
> do we need anything before ansible runs? If not, I have come around to
> Jerry's approach. (#58 <https://github.com/iiab/iiab/pull/58>, #65
> <https://github.com/iiab/iiab/pull/65> )
>
> I should add that I did the same os calc in curl-me which does precede
> ansible, but neither iiab.env nor local_facts are in place yet (nor would
> first.sh be) and I think this is a reasonable redundancy.
>
I think you are saying that you came to the same conclusion that I have
come to, when you wrote curl-me,: that local_facts are are discovered by a
script that is placed in /etc/ansible/local_facts, which in turn requires
execution of a playbook, which comes after the vars/files at the top level.
So if we want to have a separate vars file for raspbian, or debian-9, we
need something before ansible.

Am I correctly interpreting your comment?

I debated putting the same 2 lines in each starter bash script, but one
place is more supportable long term.

tim-moody:
No. Jerry contended and I confirmed that local_facts are gathered by ansible even before vars in the top level, so it is in fact the first script run and its variables may be used to select os.yml.

The only messiness, shared by this approach, is that we use ansible to put it in place. That makes me wonder if it belongs in factory along with the installation of ansible. It should never have any variable substitution in it.

georgejhunt:
Ok, I'll move local_facts to factory, and remove it from ansible.

I'm assuming that you mean no substitution in local_facts. There will be
substitution in iiab.env from ansible variables

PR #85, opened 2017-07-01
Startup

georgejhunt:
not smoke tested. But just checking with Tim before I waste time on it. 

I'm not very good at interpreting suggestions. 
I understand from Tim's comment, that Jerry has determined that this will work.


tim-moody:
looks right to me, but hard to keep it all straight

you might want to return os_release = to DISTRIB_RELEASE in case we want it for os-release.yml

tim-moody:
looks like the last commit did this

runansible
```diff
@@ -33,6 +33,11 @@ then
  exit
 fi
 
+if [ ! -f /etc/ansible/facts.d/local_facts.fact ]; then
+   mkdir -p /etc/ansible/facts.d
+   cp ./scripts/local_facts.fact /etc/ansible/facts.d/local_facts.fact
+fi
+
```
jvonau: I would have the cp ./scripts part after the fi to insure the latest version local_facts.fact is always being used.

jvonau:
The uuidgen program would need to be installed at the same time as ansible in scripts/ansible.

tim-moody:
# enable this install config to be tracked
  if [ ! -f /etc/iiab/uuid ]; then
    uuidgen > /etc/iiab/uuid
  fi
  UUID=`cat /etc/iiab/uuid`

should be moved somewhere else. setting data should not be a by-product of getting facts.

jvonau:
with the moving of local_facts.fact now roles/1-prep/tasks/prep.yml will fail 
- name: local facts script
  template: src=local_facts.fact.j2
            dest=/etc/ansible/facts.d/local_facts.fact
            owner=root
            group=root
            mode=0700

and the next one is not needed now 
- name: re-read facts
  setup: filter=ansible_local

jvonau:
- name: Set XO model and preload
  set_fact:
    phplib_dir: '{{ ansible_local["local_facts"]["phplib_dir"] }}'
    xo_model: '{{ ansible_local["local_facts"]["xo_model"] }}'
    iiab_preload: '{{ ansible_local["local_facts"]["iiab_preload"] }}'

remove iiab_preload from all the playbooks now

should phplib_dir move to the OS.yml files too? That will be useful with php_version that is already present there.

jvonau:
I would have the cp ./scripts part after the fi to insure the latest version local_facts.fact is always being used.
Will /usr/lib/php5 always be php5?

jvonau:
FWIW I like the centos-7 debain-8 naming in PR #83, would make debain9 much easier in the long run.

tim-moody:
I also find os-release a bit more readable than osrelease.

georgejhunt:
me too

jvonau:
looks OK to me, but like I said above, needs the same treatment as the you used elsewhere 
remove: 
-     iiab_base_ver: '{{ ansible_local["local_facts"]["iiab_base_ver"] }}'
-     iiab_preload: '{{ ansible_local["local_facts"]["iiab_preload"] }}'
-
-- name: Defaulting iiab_base_ver
-  set_fact:
-     iiab_base_ver: 0
-  when: iiab_base_ver == ""
-
-- name: Defaulting iiab_prepped
-  set_fact:
-     iiab_prepped: False

jvonau:
FWIW I'd push this one then #91 right after as it has the fix for the above noted issue. Don't add anything to this branch with another commit, that will make your job harder to rebase.

tim-moody:
shouldn't the case be on $OS_VER

georgejhunt:
yes, thanks

tim-moody:
what about  failed_when: os_ver == "Not sup ..." on message task instead of pause

georgejhunt:
I'm planning to test these recent changes along with all the rest of the PR's this week

PR #94, opened 2017-07-04
failed_when

georgejhunt:
Thanks Tim


tim-moody:
message needs to show os and version or you could get 'ubuntu is not supported' for ubuntu 17.4

would it be too much to set an is_os_supported flag to false rather than testing for a literal?

you would still need the string and an os.yml that corresponds to it, like unsupported.yml (which could contain is_os_supported: False with  is_os_supported: True in default vars).

(keeps from overloading the variable with both a state and a message)

PR #100, opened 2017-07-05
Stop runansible sooner if we don't have a matching os.localvars 

jvonau:
Allows for future fine grain tuning by way of os.localvars. Subset of #98

### DIFF
--- a/runansible
+++ b/runansible
@@ -8,13 +8,32 @@ fi
 
 # if vars/local_vars.yml is missing, put a default one in place
 if [ ! -f ./vars/local_vars.yml ]; then
-	OS=`grep ^ID= /etc/*release|cut -d= -f2`
-	OS=${OS//\"/}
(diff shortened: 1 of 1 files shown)


georgejhunt:
So it's a quibble: 
but I prefer:
  case $OS in
        OLPC | fedora)
        cp ./vars/olpc.localvars ./vars/local_vars.yml
        ;;
        centos)
        cp ./vars/centos.localvars ./vars/local_vars.yml
        ;;
        debian | ubuntu | raspbian)
        cp ./vars/debian.localvars ./vars/local_vars.yml
         ;;
        *)
        echo "Unknown OS -- no matching os.localvars -- exiting now..."
        exit 0
        ;;
esac
But I'm ok with it the way you have it. Whichever way, I plan to run it, before I merge.

georgejhunt:
Seems ok

PR #101, opened 2017-07-05
use ansible_date_time.tz for local_tz

jvonau:
Think this might be a better starting point, should always be present. Would indicate that /etc/localtime has been set if not blank.


georgejhunt:
Looks ok to me

PR #104, opened 2017-07-06
Dash should be underscore in default_vars.yml (activity_server_install)

holta:


tim-moody:
I stuck this in the poweroff PR

PR #105, opened 2017-07-06
Poweroff

tim-moody:


holta:
Huge Thanks @tim-moody !

Am cc'ing @deldesir (Blondel) in Haiti so he can hopefully learn a lot here, and assist in any ongoing development helping teachers & all get this right in coming months/semesters.

e.g. if the PIN-code-to-Power-Off later needs to be changed by teachers & schools etc (not just by Linux-fluent operators) or wherever this takes us in future.

tim-moody:
this is service only, no ui

PR #108, opened 2017-07-06
disable creation of online groups

jvonau:
The some files have references to schoolserver might work with the out of the way.


georgejhunt:
If we have someone willing to test, this seems like a way to get started with a more recent ejabberd.  There seems to be a current way to establish groups.  But I'm not sure it makes any sense on Pi 0

PR #110, opened 2017-07-07
services should have hyphen

georgejhunt:
The Gui uses xo-services_enabled. Hook it up again

### DIFF
--- a/roles/activity-server/tasks/main.yml
+++ b/roles/activity-server/tasks/main.yml
@@ -129,5 +129,5 @@
     - option: path
       value: /activities
     - option: enabled
-      value: "{{ xo_services_enabled }}"
+      value: "{{ xo-services_enabled }}"
             
--- a/roles/idmgr/tasks/main.yml
+++ b/roles/idmgr/tasks/main.yml
@@ -44,7 +44,7 @@
   with_items:
     - idmgr
     - xinetd
-  when: xo_services_enabled
+  when: xo-services_enabled
 
 - name: Disable idmgr service
   service: name={{ item }}
@@ -53,7 +53,7 @@
   with_items:
     - idmgr
     - xinetd
-  when: not xo_services_enabled
+  when: not xo-services_enabled
 
 #idmgr needs an extra step
 - name: Enable ejabberd service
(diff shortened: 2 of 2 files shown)


jvonau:
You are barking up the wrong tree with this one, ansible variables must have a _ and not a - or ansible pukes. The real problem was discovered by [Adam](https://github.com/iiab/iiab/commit/c3d2a8e7e66ad9c3c72dca2f978043d2e9469fd6) and fixed by Tim in PR #105 please withdraw this PR. The console should change to match what is already in place. Thanks

tim-moody:
just added test-hyphen-var: True to local_vars and did runtags 1-prep. nothing puked. (you already raised an issue about activity-server in local vars, which means it is an acceptible name, just not the one that controls the install.

we have had problems with hyphens, maybe in iiab.env because bash chokes. I can't remember. But ansible is OK and we have one in iiab.ini and I think javascript is OK.

holta:
Maybe we're all getting tripped up with "should" versus "must"

Underscores sound like a very good idea in a future version (across all Ansible variables anyway?)

tim-moody:
I am open to a discussion on variable naming conventions, though I personally prefer hyphens when possible as it is one less key press. My top candidate for change is _enabled as its meaning is really _enable (like _install), but its name sounds like a state.

georgejhunt:
I remember a discussion, now a few years ago, when we agreed that filenames
would use hyphen. Because services have systemd files, and because we
inherited same from olpc, we continued to use hyphens (postgres-xs,
xo-services, named-xs).

At the end of 2016, because of errors between _ and -, I started moving
toward _ everywhere (bash really complains with -, and I was doing bash
scripting). But at this point, the console is separated out, and Tim
correctly reminds me of the convention I agreed to.

Regarding service_enabled:, it's probably just a few minutes to "sed -i -e
/(.*_enable)d:/\1:/g", and we've been doing similar things recently.

jvonau:
If I recall the reason for using _ correctly it was because there was a
ansible crash when the variable had - and writing to the iiab.ini file to
record the dis/enabled state of the service. However you are free to do as
you please.

georgejhunt:
ansible claims that - is not a legitimate character in a variable name

PR #117, opened 2017-07-09
lockout xo service on debian family

jvonau:
Until the xo-services doesn't crash or otherwise upset the user experience lets hide them from being discovered by the end-user.


georgejhunt:
I don't want to do this, when we've been doing similar with default_vars.yml successfully for so long

georgejhunt:
Ok, I'l reopen this, and we can discuss it during our call tomorrow.

georgejhunt:
use var files per today's discussion

PR #121, opened 2017-07-11
Osverfix

georgejhunt:


georgejhunt:
centos, ubuntu, and debian 8 all passed the smoke test with this change. Not tested raspbian!

PR #123, opened 2017-07-11
for discussion Phpxml

jvonau:


tim-moody:
why are these not in os.yml?

tim-moody:
isn't when: is_debuntu and ansible_local.local_facts.os_ver == "debian-9" redundant since debian-9 is debuntu

jvonau:
posted this as it looks like php-xml-parser is what you want for the mysql part
is_debian-9 in the os.yml file might make that cleaner

tim-moody:
I feel we have php package logic spread over too many places. I would consolidate it into php.yml either as a separate role or under httpd. always installing the few packages used only by one or a couple of services is a small price to pay for simplicity. purists can declare php.yml in the meta for services that need it.

having is_<os_ver> variables set in os.yml file is a good idea to simplify the syntax.

jvonau:
php-xml-parser needed for raspbian

PR #124, opened 2017-07-12
F18

georgejhunt:
miscellaneous fixes for f18 VM smoke test


jvonau:
You made it farther than TK, who ran into the os.yml issue.
push it

Issue #36, opened 2017-06-14
can we delete /library/dokuwiki-2014-09-29 ?


holta:
Can we please delete dokuwiki-2014-09-29 ?

Two (apparently identical) copies of  the dokuwiki file tree within /library are plainly excessive~ no other package requires this!

PS Modifying https://github.com/iiab/iiab/blob/master/roles/dokuwiki/tasks/install.yml is likely the way to make this happen.

jvonau:
Think this is a result of multiple ansible runs, timestamps on the directories would be useful. The first pass renames the directory but a second pass extracts the tar file but the target directory already exists so it doesn't move it to the 'new' name. Would creating a symlink from /dokuwiki-2014-09-29 to /dokuwiki be OK? PR #39

holta:
#39 is already merged!

But see also #50 for another (related?) dokuwiki filename fix?

jvonau:
yea {{ dokuwiki_version }} went from a file name with a extension to an actual version string with an assumed file type tacked on later. Remember I did ask if that was acceptable and was for discussion. Should make testing a newer version(wow almost 3 years now) easier in the future by just moving the symlink to the newer directory.

jvonau:
@holta issue resolved?

holta:
I don't know!  As every machine I've touched for the past couple years has had this garbage directory "/library/dokuwiki-2014-09-29" so I'd like confirmation on Raspbian (if not other OS's) if indeed it's hopefully gone now :)

holta:
@georgejhunt is hoping to try to figure out why this happens on all OS's to some people but not others... (who merely get a symbolic link, rather than a duplicate directory)

holta:
Fixed.

PR #136, opened 2017-07-17
bump DokuWiki to current stable version

georgejhunt:
addresses issue #36. not smoke tested, not for 6.3


holta:
This dokuwiki playbook does not have an active+healthy community in 2017, so we should not get distracted investing resources in it at this point in time; I strongly agree with @tim-moody on this general point - that we stay focused on field communities' actual proven needs.

Be that as it may, a small clarification, in case someone chooses to revive this sub-community's energy:
2017-02-19b (from March 9) is not in fact the latest stable version at this time.
2017-02-19e (from August 23) is currently the latest stable version.

PR #138, opened 2017-07-17
mysql - php-xml  moodle php-zip for raspbian-9 BUGFIX

jvonau:
moodle or mysql might have issues on raspbian-8 due to [missing dependencies](https://github.com/iiab/iiab/search?utf8=%E2%9C%93&q=is_debian_8&type=)

### DIFF
--- a/roles/moodle/tasks/main.yml
+++ b/roles/moodle/tasks/main.yml
@@ -14,18 +14,17 @@
     - python-psycopg2
     - php{{ php_version }}-pgsql
     - php{{ php_version }}-curl
-#    - php{{ php_version }}-zip
     - php{{ php_version }}-gd
 #    - php{{ php_version }}-mbstring 
 # mbstring is now included in php-cli
     - php{{ php_version }}-cli 
   when: is_debuntu
 
-- name: php-zip debian 8-9 changed name
+- name: php-php_version-zip is-debuntu except debian-8
   package: name=php{{ php_version }}-zip
-  when: is_debian_9 or is_ubuntu
+  when: is_debuntu and not is_debian_8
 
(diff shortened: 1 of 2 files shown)


jvonau:
[debian-9 dependencies](https://github.com/iiab/iiab/search?utf8=%E2%9C%93&q=is_debian_9&type=)

tim-moody:
I agree you have clearly shown the places where debian 9 packages are selected. Only testing will show if they are needed on raspbian stretch.
(is there a reason why we need usbmount 0.0.14 instead of 0.0.22?)

jvonau:
I suspect the distinction made in mysql for php-xml is important as mysql is part of the base install. No idea about the bump in usbmount but if you what that usbkey as optional storage option it's needed.

tim-moody:
I'll concede php-xml, though the other way to do it would be in raspbian-9.yml.

Looks to me like the usbmount conditional preserves a hack for the time when stretch was not complete and this change would perpetuate it on raspbian-9 as well. if I apt-get usbmount on raspbian-9 I get 0.0.22.

jvonau:
The that was a stop gap measure are far as I can recall.

tim-moody:
usbmount play fails after os upgrade on debian as package is upgraded from 0.0.14 to 0.0.22

jvonau:
I'm not dealing with usbmount in this PR, that is a totally different issue from what is being addressed. Open an issue and wait or summit a PR to remove the usbmount hack.

jvonau:
reworked for just the package naming discrepancies with raspbian

holta:
ASIDE: usbmount 0.0.14 vs 0.0.22 issues differ on each OS, but are moving forward well in #195

PR #146, opened 2017-07-20
use python_path for activity-server

jvonau:
should fix non-redhat


jvonau:
I'd like to keep activity-server around as this playbook has the last know working example of loading "content" on demand from usbkeys.

PR #148, opened 2017-07-21
runansible straight through

georgejhunt:
Adam's fast track, for Guatemala

### DIFF
--- a/roles/1-prep/tasks/raspberry_pi_2.yml
+++ b/roles/1-prep/tasks/raspberry_pi_2.yml
@@ -55,8 +55,3 @@
 - name: Enable rootfs resizing service
   service: name=iiab-rpi-root-resize
            enabled=yes
-- name: Reboot if the config.txt was changed
-  command: /sbin/reboot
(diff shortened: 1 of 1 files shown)


holta:
Huge Thanks @georgejhunt...this massively helps as all 6.4 master installs (on Raspbian Lite anyway, possibly elsewhere too) were rebooting ~4min into runansible, at this line:

TASK [1-prep : Reboot if the config.txt was changed] ***************************

tim-moody:
This will mean that none of the benefits of the reboot will be active during the remainder of the run

including a working rtc and better memory management.

holta:
Good to know.  An important compromise as @georgejhunt & I talked about this by phone: a reboot remains very important when installation is done(*) for the reasons you mention and also to get host name right.

(*) a reboot is far too disruptive during installation.  Unless the process can continue automatically after reboot.

ASIDE: reboot after installation is an easy way to guarantee /etc/iiab/handle is disseminated properly (for those running openvpn@xscenet)

PR #150, opened 2017-07-21
Osm path

jvonau:


holta:
I smoke-tested this exhaustively today, in 3 diff ways on Raspbian Lite, and will now commit it, so that ansible can run to completion when local_vars.tml contains "osm_install: False".

Feedback from other OS's much appreciated.

PR #155, opened 2017-08-22
runansible succeeded

georgejhunt:


tim-moody:
forgot to mention that cups_install: True does not seem appropriate here or in raspbian-8 from which is was copied. I could see a False here, but a True should be in local_vars.

PR #159, opened 2017-08-23
escape the dollar sign in password hash

georgejhunt:


tim-moody:
not sure why this is necessary as this has been there for a long time. $ is the delimeter

georgejhunt:
I agree. But on testing raspbian-9, pixel-lite, the warning about
iiab-admin default password was not working, and the reason was that the
character match was failing, because dollar sign was interpreted as a
variable, rather than a literal.

tim-moody:
if you change the hash you change the password

PR #160, opened 2017-08-23
Ifnames

georgejhunt:
Don't waste users time (raspbian-9), when the networking breaks due to interface name change, and will work after a reboot- but user might just give up

### DIFF
--- a/roles/1-prep/tasks/raspberry_pi_2.yml
+++ b/roles/1-prep/tasks/raspberry_pi_2.yml
@@ -41,7 +41,7 @@
               dest=/etc/dphys-swapfile
   when: is_debuntu
 
-- name: restart the swqp service
+- name: restart the swap service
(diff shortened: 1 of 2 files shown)


georgejhunt:
Smoke tested on raspbian-9. 
Debian-9 includes a new "feature" called "predictable ifnames". The default on raspbian is to include the mac address prefixed with "enm" as the interface name. 
Unfortunately this predictably will fail when an image is transferred to another pi so we MUST disable this new feature -- which can best be done by adding net.ifnames=0 to the kernel command line. This PR puts the kernel command line change at the earliest possible point in the install process.

jvonau:
http://downloads.raspberrypi.org/raspbian/release_notes.txt 
Please revert this PR as upstream moved to disabling predictable network interface names for Ethernet devices and there is a question of the source of the firmware. #230

PR #162, opened 2017-08-24
Kalite venv

jvonau:

PR #163, opened 2017-08-24
OSM-venv

jvonau:


tim-moody:
what amount of addition storage does using a venv require?

jvonau:
There are two ways for the venv to be populated, files totally isolated or with symlinks pointing to what is already present on the system. That is virtualenv_site_packages= in ansible.
This fully isolated type I used du shows:
66M	/usr/local/osm/
4.0M /usr/local/osm/lib/python2.7/site-packages/iiab
22M	/usr/local/osm/lib/python2.7/site-packages/babel

While linking can save some space, the amount saved would depend on what is already installed. Nice part is all the dependencies are in /usr/local/osm/

tim-moody:
for whoosh I see virtualenv_site_packages=no, which I believe is the default.
for osm I was expecting virtualenv_site_packages=yes

jvonau:
I opted for the fully isolated method, wasn't sure what other frozen dependencies whoosh might have. That would be the place to toggle sharing with the system installed libraries thou. I have not tested not having virtualenv_site_packages= the same in the two stanzas. I've tried yes first but did not note the installed footprint.

georgejhunt:
So I don't think <100MB is too much penalty for the benefits going forward

jvonau:
It's less than 100M actually, 66 for the full install - 28 for what is installed anyway= 38M maximum space saving by sharing, didn't think it was worth it but sharing could be explored.  When updating OSM install just blow away usr/local/osm and re-install,  same hold true for kalite.

holta:
FYI to help others understand the underlying motives here...a citation from #162: (KA Lite venv)

georgejhunt:
I take it from the numbers regarding disk footprint, that this has successfully been smoke tested?

georgejhunt:
included in 259

Issue #78, opened 2017-06-29
Downgrading python setuptools breaks CentOS install


jvonau:
Andreas Gros 	
Jun 18
The next issue I'm running into is that the downgrade of setuptools doesn't work.
Once the uninstallation is done, pip throws an error that pkg_resources isn't present anymore:
File "/bin/pip", line 5, in <module>
from pkg_resources import load_entry_point

ImportError: No module named pkg_resources

https://github.com/iiab/iiab/blob/master/roles/kalite/tasks/install.yml
https://github.com/XSCE/xsce/blob/release-6.2/roles/kalite/tasks/install.yml

jvonau:
This [diagnostic code](https://github.com/iiab/iiab/compare/master...jvonau:strict) provides for an in the field change , first pass though ansible without flag would have no effect, and should this be encountered then one can add the flag to local_vars  for retesting. Of course after the hand installing at per "I think I fixed 6.2. But it would be great if you could help verify that. I suggest the following path:" dance. From the [notes](https://github.com/iiab/iiab/blob/master/notes)  I suspect kalite is looking for a specific version of setuptools and what is installed would be the latest as per the [revised packages.yml routine](https://github.com/iiab/iiab/blame/master/roles/2-common/tasks/packages.yml).

Should kalite now fail at a later point complaining about setuptools, the version can be fixed at an earlier point in packages.yml where the latest and greatest is being installed. As an alternate route a possible fix in #162 using venv in place of installing into the sites directory where pip/setuptools reside.

Note this use of 'dist-info' in https://github.com/pypa/pip/issues/4400 which has an example working string for downgrading but intended for a venv. I would not try that on CentOS where the path would be shared with rpms(/bin/pip) while Debian uses /usr/local/bin/pip so things are a bit different.

holta:
@jvonau @tim-moody is this something important/converging for IIAB 6.4 in coming weeks?

jvonau:
Unfixed I'll go with CentOS is broken as is, and I would not even attempt one without #164 #167 
> first pass though ansible without flag would have no effect, and should this be encountered then one can add the flag to local_vars for retesting

Note: added strict_install: True to centos-7.yml in #167 so the logic would be reversed now.

jvonau:
With no posted proof (iiab-install.log) of a successful CentOS install one has to rely on "smoketested" in #121 but I think that is for only the change, not a fresh install. On Jun 21 PR #46 
[...]

PR #167, opened 2017-08-26
Strict

jvonau:
debugging pip


holta:
@georgejhunt even if the main focus of IIAB 6.4 is Raspbian Stretch in service of the masses — this sounds like a high-priority for CentOS according to Jerry in #78 — can you investigate for possible inclusion in IIAB 6.4 ?

PR #169, opened 2017-08-26
Net wip

jvonau:
The switch away from discovered_*_iface is to allow the overrides like iiab_lan_enabled in computed_network,yml to be effective, as those set the interface to be none and taken into the calculation of iiab_network_mode.

### DIFF
--- a/roles/1-prep/tasks/raspberry_pi_2.yml
+++ b/roles/1-prep/tasks/raspberry_pi_2.yml
@@ -32,9 +32,6 @@
   with_items:
     - ntp
 
-- name: disable the renaming of eth0
-  shell: sed -i -e 's/deadline fsck/deadline  net.ifnames=0 fsck/' /boot/cmdline.txt
-
 - name: increase the swap file size (kalite pip download fails)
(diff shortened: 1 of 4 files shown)


tim-moody:
OK. I see I didn't pay attention to #156 

Still, I'm not sure patching the adapter name is the right approach, unlike what we do on every other OS.

jvonau:
I think it is better to switch to the new naming scheme and not try to force our will on the user.

tim-moody:
not clear what you mean by 'the new naming scheme' or what our 'will' constitutes. if 'new' means en... and 'will' stands for eth0, then I agree.

jvonau:
Think we are on the same page, just work with what comes out of the box.

jvonau:
What I don't like is the different placement of {% if gui_static_wan == false %} between APPLIANCE and GATEWAY, that should be consistent. I would go with APPLIANCE as being more correct as I could see 'auto' might having an effect on boot and disabling static on GATEWAY.

jvonau:
Closing contains errors but 79a2dd4 is where the hard coding lives.

PR #164, opened 2017-08-24
ansible 2.3.1

jvonau:
intended as a replacement for scripts/ansible

PR #181, opened 2017-08-31
silence pip upgrade nag

jvonau:
alternate to #164

### DIFF
--- a/scripts/ansible
+++ b/scripts/ansible
@@ -89,7 +89,7 @@ if [ ! $(which ansible-playbook) ]; then
 
   mkdir /etc/ansible/
   echo -e '[local]\nlocalhost\n' > /etc/ansible/hosts
-  pip install ansible==2.2
+  pip install ansible==2.2 --disable-pip-version-check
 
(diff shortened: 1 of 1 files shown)


holta:
I can't claim to understand this in full, but FYI @georgejhunt expressed his preference that we stick with Ansible 2.2.0 for IIAB 6.4 during this morning's call.

I did not hear any complaints (or words of support) from @tim-moody so hopefully he can weigh in as well?

FWIW Ansible 2.2.0 works for me, so we can get IIAB solid+reliable quickly in September, and then presumably upgrade Ansible for IIAB 6.5 in November or some such...

jvonau:
At least consider suppressing the pip upgrade check, people see that and
panic while it is not really required to be preformed. Would sort of help
with CentOS, but it seems that things are pip happy in the code base. I'm
starting not to care as nobody else wants to put out a quality polished
product.

holta:
Can't claim to understand pip :/ but I've added this to the agenda for Monday's call @ http://minutes.iiab.io

jvonau:
Automated testing would prefer not to be prompted for action
From [xsce-devel](https://groups.google.com/forum/#!topic/xsce-devel/gONb4q6xT3g)
>  # Installs the correct version of Ansible
      I was prompted to upgrade my pip version from 8.1.1 to 9.0.1
      Pip install --upgrade pip

PR #185, opened 2017-09-01
get default_vars in sync with published local_vars

georgejhunt:


holta:
Builds on #183

holta:
Published local_vars.yml is in both places here:

http://wiki.laptop.org/go/IIAB/local_vars.yml
http://download.iiab.io/6.4/rpi/local_vars.yml

Issue #188, opened 2017-09-01
Elgg 1.12.12 & 1.12.16 don't show anything in browser (500 Internal Server Error)


holta:
@jvonau asks if a "schoolserver" reference that he spotted in Elgg's sql dump/log might be causing this?

Full Details in PR #179 (upgrade elgg from 1.12.12 to 1.12.16) and issue #145 (sudden loss of power locks elgg/mysql).

jvonau:
Since php is involved adding #69. My 1.12.12 install was working as the correct extension_dir  is defined in /etc/php.ini for my platform. Quick test rename /etc/php.ini and restart apache2.

tim-moody:
http://iiab.me/elgg/ brings up page, though css seems missing

holta:
Close this issue if PR #202 works (Elgg upgrade from 1.12.16 LTS to 2.3.3)

holta:
#202 merged; awaits final testing so we can close this.

jvonau:
It's a dns issue, if you add the server ip address to /etc/hosts it works.

> GET 
http://10.8.0.10/elgg/ [HTTP/1.1 200 OK 317ms]
GET 
http://box/elgg/cache/1504389501/default/font-awesome/css/font-awesome.css [HTTP/1.1 200 OK 253ms]
GET 
http://box/elgg/cache/1504389501/default/elgg.css [HTTP/1.1 200 OK 311ms]
GET 
http://box/elgg/cache/1504389501/default/lightbox/elgg-colorbox-theme/colorbox.css [HTTP/1.1 304 Not Modified 216ms]
GET 
http://box/elgg/cache/1504389501/default/jquery.js [HTTP/1.1 304 Not Modified 968ms]
GET 
http://box/elgg/cache/1504389501/default/jquery-ui.js [HTTP/1.1 304 Not Modified 1239ms]
GET 
http://box/elgg/cache/1504389501/default/elgg/require_config.js [HTTP/1.1 304 Not Modified 1395ms]
GET 
http://box/elgg/cache/1504389501/default/require.js [HTTP/1.1 304 Not Modified 328ms]
GET 
http://box/elgg/cache/1504389501/default/elgg.js [HTTP/1.1 304 Not Modified 361ms]
Expected 'none', URL, or filter function but found 'progid'.  Error in parsing value for 'filter'.  Declaration dropped. colorbox.css:65:12
[...]

Issue #175, opened 2017-08-30
Nextcloud install missing 2 PHP modules (Raspbian Stretch only? Debian 9 too?)


holta:
Does anyone know if this same incomplete installation occurs on Debian 9?

In any case, I've confirmed this across 2 different installations of http://download.iiab.io/6.4/rpi/load-vpn.txt on Raspbian Stretch Lite from 2017-08-16.

Here is what appears in browser when visiting http://box/nextcloud or http://192.168.0.x/nextcloud :

    PHP module zip not installed.
    Please ask your server administrator to install the module.

    PHP module mb multibyte not installed.
    Please ask your server administrator to install the module.

    PHP modules have been installed, but they are still listed as missing?
    Please ask your server administrator to restart the web server.

jvonau:
It looks to be a [name change issue](https://github.com/iiab/iiab/search?utf8=%E2%9C%93&q=php-zip&type=) going on with php-zip while roles/nextcloud/tasks/main.yml is not using the {{ php_version }} maybe php-zip should move to the next stanza? Not sure about the other one, missing dependency maybe?

holta:
Is #69 on Ubuntu related and/or helpful?

(Certainly we experienced very much the same kind of Nextcloud PHP errors on Ubuntu 16.04 in late June!)

jvonau:
Might be the cause if all the dependencies are satisfied

holta:
Close this issue if PR #202 works ("Elgg upgrade 1.12.16 LTS to 2.3.3; Nextcloud - add php's zip mbstring")

holta:
@floydianslips can you confirm http://box.lan/nextcloud works?

(since #202 was pushed 8 days ago)

holta:
http://box.lan/nextcloud works when installed the latest IIAB 6.4 master from http://download.iiab.io/6.4/rpi/load-vpn.txt onto Raspbian Stretch Lite 2017-09-07, thanks for everyone's help !

diplodata:
So what's the solution??

PR #190, opened 2017-09-02
add mbstring for debian

georgejhunt:
smoke tested on debian 9 vm, will test on other platforms, and rpi hardware, with other branches

### DIFF
--- a/roles/nextcloud/tasks/main.yml
+++ b/roles/nextcloud/tasks/main.yml
@@ -21,6 +21,7 @@
   with_items:
     - libapache2-mod-php{{ php_version }}
     - php{{ php_version }}-mbstring
+    - php{{ php_version }}-zip
   when: is_debian
 
 - name: ubuntu and debian treat names differently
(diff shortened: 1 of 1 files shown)


holta:
Some chance this will solve Nextcloud #175 which mentions "PHP module mb multibyte not installed" ?

georgejhunt:
No I think we need to go to 2.3.3 version of nextcloud.   The older one
uses php mysql language, which has been deprecated for years, and finally
removed from php7.

holta:
> No I think we need to go to 2.3.3 version of nextcloud.

Nextcloud's current version is 12.0.2 from August 14, 2017.

You must mean Elgg 2.3.3 from May 16, 2017?

Ref #188 "Elgg 1.12.12 & 1.12.16 don't show anything in browser (500 Internal Server Error)"

georgejhunt:
smoke tested on rpi hardware. I had to attempt to log in twice, before I got the nextcloud data page. There might have been a timeout in my browser that canceled the operation.

georgejhunt:
included by accident in the elgg fixes

PR #178, opened 2017-08-30
Ubuntu install failure - BUGFIX dns dropout - BUGFIX wifi as gateway - ENHANCEMENT/REGRESSION dhcp assinged address when static - BUGFIX repsect user defined networking - ENHANCEMENT respect static addressing in local_vars - REGRESSION

jvonau:
Trying to avoid using /etc/network/interfaces at all

PR #194, opened 2017-09-02
Pi wifi gw

georgejhunt:


roles/1-prep/defaults/main.yml
```diff
@@ -46,12 +46,14 @@ ap_device: "none"
 device_gw2: ""
 
 # WiFi
-ssid: XSCE
+ssid: IIAB
```
holta, line 49: @georgejhunt what is this "ssid: IIAB" used for?

Also is there a reason it's different from the following that we have in local_vars & default_vars ?

`host_ssid: "Internet in a Box"`

tim-moody:
what does it mean to 'load rpi from wifi'?

holta:
[surfacing the question buried in ticket's "Files changed" tab]

@georgejhunt what is the "ssid: IIAB" in roles/1-prep/defaults/main.yml used for?

Also is there a reason it's different from the following that we have in [local_vars](http://wiki.laptop.org/go/IIAB/local_vars.yml) & [default_vars](https://github.com/iiab/iiab/blob/master/vars/default_vars.yml) ?

`host_ssid: "Internet in a Box"`

georgejhunt:
I wanted to explore if it was possible to do an install from scratch
without a wired internet connection. Internet connection via wifi is almost
universal in the developed world, maybe even everywhere.

So the necessity to have a wired connection seems like a shortcoming.

tim-moody:
so the ssid is because the wifi serves both as gateway and as hotspot? I believe this is possible and desirable. Or is ssid the one to connect to as opposed to the one we are exposing?
http://weworkweplay.com/play/automatically-connect-a-raspberry-pi-to-a-wifi-network/ is one approach though it may be old.

georgejhunt:
The url you pointed to might be a long term strategy we could pursue. It
looks like a good idea for an image that we provide.

I was addressing our install procedure, starting with a just downloaded
raspbian lite image.

Can we provide a path which uses available wifi internet, and gets our
software onto the pi?

georgejhunt:
The only real advantage of this suggestion is it's ability to be headless.

I reread the your comment. I'd like to hear more about your idea about
being both a gateway and a hotspot.  My experience up til now suggested
that they were mutually exclusive.

But google is wonderful

http://blog.hoxnox.com/gentoo/wifi-hotspot.html

Of course it will need to be tested. But I think it is worth shooting for

roles/1-prep/defaults/main.yml
jvonau: It's a starting point default variable, to guard against users removing SSID from *_vars and causing ansible to fail

jvonau:
This would require /etc/network/interfaces to be unmodified, which is being addressed in #178

georgejhunt:
Maybe there's convergence happening.

For this headless wifi install procedure, I just disabled all the network
re-jiggling, figuring that a bash script could later rewrite interfaces, or
br0, when the user wanted to create a hotspot

jvonau:
The only way to get on the internet with wifi is to configure the machine first by some other means to get to the load-scripts. There is more to the network switch over than just altering the network config.

georgejhunt:
I've tried to document the steps you mention to get the pi on a local wifi
internet source at:

https://github.com/georgejhunt/iiab/blob/pi_wifi_gw/roles/network/pi-wifi-
gateway.md

Tim found a path forward using wpa_supplicant rather than hostapd for AP
mode. Yes I agree that it's not easy, but I see some hope in the following:

http://blog.hoxnox.com/gentoo/wifi-hotspot.html

jvonau:
You might want to mention to the end-users that using the built-in gui-tool that is part of the desktop as an alternate to hand editing the file. Having the AP option built into rasp-config would be the preferred method, then all rpi users can benefit.

jvonau:
There is an option to bring up the rpi headless [from /boot](https://core-electronics.com.au/tutorials/raspberry-pi-zerow-headless-wifi-setup.html) where cmdline.txt is stored, that could be used with automated testing.

georgejhunt:
I did not know or remember that raspi-config had that option.  That is a
better approach

holta:
raspi-config does not have the option of connecting to a generic/nearby AP, nor the ability to enter a WPA2 password or similar.

(and certainly will never give you the ability to click on disclaimers to get online when various Captive Portals require that!)

jvonau:
I believe that is a system-tray icon in the desktop that does the wifi from the gui, saw lots of examples with a quick search. We just have to not clobber the config when we runansible.

georgejhunt:
Even if it is a system-tray function, and does not help us on the lite
version, I agree that we can make sure not to destroy the gateway
connection on wifi, if that is what the user set up through the desktop gui.

jvonau:
On Sat, Sep 2, 2017 at 2:42 PM, georgejhunt <notifications@github.com>
wrote:

> Even if it is a system-tray function, and does not help us on the lite
> version,


Then try the /boot/ wpa_supplicant.conf option.

georgejhunt:
I still would like to wait and do all the network restart at the very end of our runansible.  But for now, I don't think this is the right time

jvonau:
> I still would like to wait and do all the network restart at the very end of our runansible. But for now, I don't think this is the right time

I actually don't disagree with you on that point, one can pull the networking out of runansible's  codepath in 3/4-meta and run iiab-network standalone at the end, or move network after 8-meta.
The network would be the last to change then and might be upsetting hostname management, that is a bit of a variable but worth a shot.

Issue #193, opened 2017-09-02
Ansible fails all subsequent runs on Raspbian Stretch: Sugarizer "npm : Depends: node-gyp (>= 0.10.9)"


holta:
./runansible ran to completion using http://download.iiab.io/6.4/rpi/load-vpn.txt on Raspbian Stretch Lite on RPi3 (master, about 12 hours ago...after runansible immediately rebooted, I ran "cd /opt/iiab/iiab; ./runansible" continuing the load-vpn.txt script from there).

Sugarizer's main page http://box/sugarizer loaded correctly during spot testing.

But then all 3 subsequent runs of "./runansible" or "./runtags sugarizer" failed as follows:

TASK [sugarizer : Actually get it installed] ***********************************
fatal: [127.0.0.1]: FAILED! => {"changed": true, "cmd": ["apt", "install", "-y", "npm"], "delta": "0:00:02.159470", "end": "2017-09-02 12:28:13.043157", "failed": true, "rc": 100, "start": "2017-09-02 12:28:10.883687", "stderr": "\nWARNING: apt does not have a stable CLI interface. Use with caution in scripts.\n\nE: Unable to correct problems, you have held broken packages.", "stdout": "Reading package lists...\nBuilding dependency tree...\nReading state information...\nSome packages could not be installed. This may mean that you have\nrequested an impossible situation or if you are using the unstable\ndistribution that some required packages have not yet been created\nor been moved out of Incoming.\nThe following information may help to resolve the situation:\n\nThe following packages have unmet dependencies:\n npm : Depends: node-gyp (>= 0.10.9) but it is not going to be installed", "stdout_lines": ["Reading package lists...", "Building dependency tree...", "Reading state information...", "Some packages could not be installed. This may mean that you have", "requested an impossible situation or if you are using the unstable", "distribution that some required packages have not yet been created", "or been moved out of Incoming.", "The following information may help to resolve the situation:", "", "The following packages have unmet dependencies:", " npm : Depends: node-gyp (>= 0.10.9) but it is not going to be installed"], "warnings": []}
        to retry, use: --limit @/opt/iiab/iiab/iiab.retry

holta:
@georgejhunt LMK if you're voting to bypass this issue in IIAB 6.4, if #196 makes it too hard to get Sugarizer working at this time?

(Presumably deferring until [milestone 6.5](https://github.com/iiab/iiab/milestone/2) if so.)

georgejhunt:
[...]

PR #196, opened 2017-09-02
stretch seems to have dependency hell about the requirements for npm …

georgejhunt:
…-- give it time, and disable for now

### DIFF
--- a/vars/default_vars.yml
+++ b/vars/default_vars.yml
@@ -240,7 +240,7 @@ kalite_enabled: False
 kalite_cron_enabled: False
 
 #Sugarizer
-sugarizer_install: True
+sugarizer_install: False
 sugarizer_enabled: False
 
 # 8-MGMT-TOOLS


holta:
@georgejhunt when you say "disable for now", do you recommend we disable Sugarizer in IIAB 6.4?

(If so presumably trying to get it work in IIAB 6.5 later this year?)

Ref #193 'Ansible fails all subsequent runs on Raspbian Stretch: Sugarizer "npm : Depends: node-gyp (>= 0.10.9)"'

georgejhunt:
Smoke tested on VM

jvonau:
since npm is being used to install nodejs perhaps installing the native [nodejs](https://packages.debian.org/stretch/nodejs) package would be enough to satisfy this dependency and might be a better long term solution?  
edit: Never mind it's is flaky on [debian](https://anonscm.debian.org/cgit/collab-maint/nodejs.git/log/?id=refs/heads/master-6.x) but looks like 6.X is available but not sure of the apt magic to get it, might be the setup routine that is already enabled.

georgejhunt:
npn fails on second attempt to reinstall. We should test for installed npm, and don't try to reinstall

PR #204, opened 2017-09-03
unused/obsolete stuff

georgejhunt:
The online release notes 6.0 and 6.1 are pulled from xsce repo--so they will not be lost if they are deleted from IIAB repo, and can still be buried in the online http://box/info collection. We chose not to put the release notes 6.2 and 6.3 in the repo root. So I think the older ones should go too.

I may be wrong about the console deletions. but testing will expose that.

Not smoke tested.


georgejhunt:
This was smoke tested on VM

jvonau:
Could of reused install-console to install openvpn after altering iiab-base.yml

PR #205, opened 2017-09-03
turn on services we want to test

georgejhunt:
The runansible copies debian or centos local_vars, if into local_vars.yml, if that file is missing. For my automated testing I'd like for those files to reflect what we currently want to support.


georgejhunt:
not smoke tested

jvonau:
centos can't be correct..

georgejhunt:
garbage pr

PR #207, opened 2017-09-03
multiple ./runtags comment: X,Y,Z instead of XXX

holta:
Per @jvonau "It used to be spaces, and XXX would return a list of tags but now use ./runtags cups,elgg,dokuwiki and XXX is broken."

### DIFF
--- a/runtags
+++ b/runtags
@@ -46,7 +46,7 @@ fi
 if [ $# -ne 1 ]; then
 	echo "usage: $0 <tagname>"
 	echo
-	echo " If you would like a list of possible values, enter $0 XXX"
+	echo " If you would like a list of possible values, enter $0 X,Y,Z"
 	exit 1
 fi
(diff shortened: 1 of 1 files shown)


jvonau:
That will use tags X,Y,Z and is misleading. edit:
XXX used to return a list of possible tags to use without running anything, but now XXX runs 1-prep unconditionally.

jvonau:
Wish the dynamic available tags worked but this might be a temporary measure, but needs the docs to be present. 
"runtags uses a comma separated list of tags X,Y,Z  Please see 'http://box/path_to_local_help' for a full listing of tags."

jvonau:
XXX fixed in #209

holta:
PR #209 looks much better (even if "XXX" as a literal is very confusing, consider "list-tags" instead?)

jvonau:
Better fix #297 subset of #267

PR #172, opened 2017-08-27
Re: IIAB on Ubuntu thread BUGFIX

jvonau:
1. lets try to get sshd and openvpn installed earlier than the network.
2. remove duplicate network routine when running under runansible. (makes for cleaner logging and faster)
3. add option to suppress running of dhcpcd via flag for remote debugging.
4. account for the move of sshd and openvpn for the console.

PR #209, opened 2017-09-04
fix runtags XXX

jvonau:


holta:
@jvonau @tim-moody @georgejhunt

Can we consider "list-tags" or similar, instead of "XXX" ?

(Unfortunately I did not grasp what the error message was saying earlier: I thought XXX within "./runtags XXX" was some kind of placeholder, rather that an actual flag/command!)

tim-moody:
./runtags list-tags should already work

tim-moody:
I also do not believe we should encourage people to use runtags

holta:
@tim-moody "./runtags list-tags" starts (some kind of) very lengthy Ansible run, rather than listing possible tags as I would have expected.

jvonau:
cause you know everything

jvonau:
> I also do not believe we should encourage people to use runtags

How do you expect remote help to work when runansible crashes before getting to openvpn? In #172 moving the openvpn playbook earlier got poo poo'd. 

How does one remotely execute anything if you can't get to the web-console?

holta:
I know for a fact @tim-moody uses ./runtags to fix his home dishwasher ;) so perhaps it's fair to say that ./runtags should not be used regularly, and certainly not with multiple tags &mdash; and *definitely* not by casual readers of http://FAQ.IIAB.IO !

jvonau:
I showed you what to change to match the language already present. Do as you see fit.

PR #220, opened 2017-09-04
If local_vars.yml missing, supply a medium number of services

georgejhunt:
My automated VM testing does not do a smoke test on all that we advertise (though the default has been to install, but not enable -- so this PR may not change much)


holta:
FYI local_vars_big.yml is published to both places here:

http://wiki.laptop.org/go/IIAB/local_vars_big.yml
http://download.iiab.io/6.4/rpi/local_vars_big.yml

tim-moody:
I wouldn't mix testing and deployment. I would put my test local vars in the testing repo and cp it to vars before running the tests. I would make centos.local_vars, etc. be my best guess as to the most like configuration users of those oses would want.

holta:
Just FYI this particular "big" local_vars.yml was created for testing and exploratory implementers trying to get to know all sorts of different edu apps & the core/common LMS's &mdash; anyway that was the idea when I created this over the weekend hoping to help both (often similar!) kinds of folk.  Anyway I left a short note to that effect on the top of this page:

http://wiki.laptop.org/go/IIAB/local_vars_big.yml

georgejhunt:
maybe a single "most likely" config, as Tim suggests. I'll rework it.

holta:
@georgejhunt instead of maintaining 3 identical copies of these most common choices for local_vars.yml... (each constantly deserve grassroots-driven improvement, that's the definition of democracy!)

http://wiki.laptop.org/go/IIAB/local_vars.yml
http://download.iiab.io/6.4/rpi/local_vars.yml
https://github.com/iiab/iiab/tree/master/vars/medium.localvars

Shouldn't we pick just 1?

(With an explicit understanding that field implementers-and-their-peers are *supposed* to constantly evolve these organic community norms!)

jvonau:
Totally remove the # out playbooks there is no hope in ever being able to run them now with the plays disabled in 'meta'

holta:
In the case of http://wiki.laptop.org/go/IIAB/local_vars.yml (*) there is a strong informational + public service aspect here, such that I prefer to honor our community roots (recently deprecated services, etc) with a sincere explanation.  Currently those warnings include:

```
# Legacy XO services have been removed but are recoverable.  Please contact
# http://lists.laptop.org/pipermail/server-devel/ if you're able to help test.
...
# Similar to Calibre, but unmaintained
pathagar_install: False
pathagar_enabled: False
...
# Unmaintained
# schooltool_install: False
# schooltool_enabled: False

# Unmaintained
# debian_schooltool_install: False
# debian_schooltool_enabled: False
```

These definitely should be deleted at some point later, exactly as you say.

(*) or wherever we/they single-source this eventually?  I'm very hopeful that we find ways for grassroots implementers to strengthen their voices internationally &mdash; articulating what local_vars.yml combinations and what Content Pack remixes are Most Effective and Most Popular &mdash; so these very practical norms are encoded into single-source files like:

   https://github.com/iiab/iiab/tree/master/vars/medium.localvars (doesn't exist yet)
   http://download.iiab.io/6.4/rpi/local_vars.yml (for now!)

Ref #222 ("Embrace+Evolve grassroots norms for Sm|Md|Lg local_vars.yml + Content Pack Suites")

Issue #230, opened 2017-09-08
Raspbian Stretch 2017-09-07 fixes...firmware AP Mode & more?


holta:
@georgejhunt reports that AP mode was broken in Raspbian Stretch's 2017-08-16 firmware, until he swapped in Raspbian Jessie firmware as follows: https://github.com/iiab/iiab/commit/ee4df858b3b423380b121db256b078e9896ae796

_Hopefully today's new Raspbian Stretch (2017-09-07) fixes many things!_  In any case, here's an excerpt from [downloads.raspberrypi.org/raspbian/release_notes.txt](http://downloads.raspberrypi.org/raspbian/release_notes.txt) as we get started here, as we all seek to ruggedize+test [download.iiab.io/6.4/rpi/load.txt](http://download.iiab.io/6.4/rpi/load.txt) (and similar) together now:

```
2017-09-07:
  * Disable predictable network interface names for Ethernet devices
  * Bug fix for keyboard settings dialog in Raspberry Pi Configuration
  * Bug fix for crash on some videos and animations in Chromium
  * Bug fix for taskbar crash when running RealVNC server
  * Bug fix for reloading projects with extensions in Scratch 2
  * Bug fix for MAC address problem in Bluetooth
  * Simple mode and new icons in Thonny
  * New Japanese translations in Raspberry Pi Configuration
  * Install fonts-droid-fallback for international fonts
```

jvonau:
> reports that AP mode was broken in Raspbian Stretch's 2017-08-16 firmware

Is there an upstream bug filed anywhere? If not, open one so this issue is brought up with the primary developers? I'm concerned about downgrading because of this entry in linked release-notes.
2017-08-16 
  * Wireless firmware for Pi 3 and Pi 0W modified to address Broadpwn exploit

jvonau:
July 2017 thread https://www.raspberrypi.org/forums/viewtopic.php?f=91&t=187725 We need to know if the .txt file mentioned is present out of the box in either 08-16 or 09-07 images. [related backstory](https://github.com/RPi-Distro/repo/issues/51)

jvonau:
https://www.raspberrypi.org/forums/viewtopic.php?t=191061

jvonau:
https://www.raspberrypi.org/blog/raspbian-stretch/
> Nick says:	
18th Aug 2017 at 8:33 pm	
install of latest raspbian lite on Pi 0 W
no wireless
errors in syslog :
[...]

PR #231, opened 2017-09-08
9/7/17 version of raspbian -- reverse fixes for flawed August stretch raspbian

georgejhunt:
I can confirm that the wifi functions in AP  mode, and that removal of workarounds for August still works.
This has been tested on rpi3, using Adam's medium local_vars. Interface names returned to previous -- eth0 and wlan0.


jvonau:
push it #230

PR #235, opened 2017-09-10
xs_wan_device -> iiab_wan_device -- oops

georgejhunt:
an artifact of rebranding -- more of these???


jvonau:
more than likely....  push this

PR #233, opened 2017-09-09
remove workarounds for august stretch bugs

georgejhunt:
replacement for previous PR which disappeared when I deleted the branch it was based upon

PR #238, opened 2017-09-11
let dhcpcd do wan, make wan manual in interfaces for networking service

georgejhunt:
tested by Germain in Nicaragua. will eventually be replaced by pr178


jvonau:
With #233 reverts #156 push it.

tim-moody:
I believe:
dhcpcd does everything unless told not to. networking does nothing unless told to in interfaces.
so just leave wan out of interfaces.  if necessary use denyinterfaces  to tell dhcpcd not to mess with lan

PR #239, opened 2017-09-11
Run network playbook once during runansible

jvonau:
for illustration only ATM


tim-moody:
do networking as soon as configured, it should be possible to do services without a networking side effect.

jvonau:
revised 239 -> Run network playbook once during runansible.

jvonau:
also part of #243

jvonau:
Nothing more than eliminating the all the 'skipping' of the first round of networking that is never used that spams and bloats up the logfile. one subset of #267

holta:
@jvonau (on live call http://minutes.iiab.io) says he's repeatedly smoke-tested this over past 2 weeks.

PR #244, opened 2017-09-13
DEMO Rpi wifi gw

georgejhunt:
This is a bash script which uses the rpi wifi for upstream, and lan, simultaneously. It loads onto raspbian lite in about 30 seconds.  I started using it to load my nginx branch. But it errored out in the middle of network setup -- which I had not sufficiently disabled.
It uses dhcpcd, dnsmasq, iptables, and networking. It might be a solution for nurses who need their snapchat, and whatsapp, but also want kiwix.
There is too much intermittent latency, which I don't yet understand.  I don't think the wifi channels are full in my location.  Maybe this is a bug -- more testing required.
The script requires <ssid> and <password> modification to function out of the box.


jvonau:
I like the down and dirty, but `ip a` would be informative. Both client and ap simultaneously at the same time with`iw dev wlan0 interface add wlan0_ap type __ap`? Does that rename or create an alias of wlan0? Interesting trick, that is why I'm asking about the status of the device names.

Looks like hostapdstart start outs with AP using `ifup wlan0_ap` and would play nice with interfaces.d/access-point and `denyinterfaces wlan0_ap` for the AP side while `interface wlan0_sta` is present in /etc/dhcpcd.conf.
Which one these:
/etc/network/interfaces.d/station
auto wlan0
iface wlan0 inet manual
wpa-ssid test
or
`wpa_supplicant -Dnl80211 -iwlan0`
is providing the connection to upstream wifi?
Keep using /etc/network/interfaces.d/ as we can drop in files at will and remove those that we know can clash, just have to be aware of having duplicate device names present amongst the files present. The hostapd.conf file in git does have the logic to switch between bridge or interface, just have to stop forcing the bridge's into use in 1-prep computed_network.yml.

georgejhunt:
I'm going to work on this some more, and don't need to bother everyone with the changes

PR #245, opened 2017-09-14
Sugar

georgejhunt:
smoke tested, and QRcode function verified on a debian9 vm


georgejhunt:
This bumps the version of sugarizer to 0.9

PR #247, opened 2017-09-14
just move our config to interfaces.d

georgejhunt:
Problem this is trying to solve: Wired connection fails after boot.
dhcpcd is refusing to start because there are static, and dhcp, references in /etc/network/interfaces.
But it does not refuse to start when that same config is moved to /etc/network/interfaces.d/
smoke tested on September raspbian lite, using Adam's local_vars.yml min version.
wifi connects


jvonau:
Glad you find some inspiration from my PR's. While you have the hood up you can fix and re-enable the documented use of wan_ip  wan_netmask: wan_gateway: wan_nameserver for static networking in default_vars.yml so iiab-network or runtags network can function as expected.

georgejhunt:
smoke tested on rpi and verified as to setting static ip from the console. Required reboot for the new ip to become active. 
As a trial, I put in a restart, to change the ip on the fly, and then reverted it. I think breaking an upstream connection during an install, without halting, and issuing an explanitory message is a bad idea, and not really what most people would expect.

PR #256, opened 2017-09-17
Clean install: don't upgrade pip (for testing/debugging)

jvonau:
avoid using pip while testing.

### DIFF
--- a/roles/1-prep/defaults/main.yml
+++ b/roles/1-prep/defaults/main.yml
@@ -4,7 +4,7 @@ gui_version: 2
 
 # These entries should never be changed in this file.
 # These are defaults for boolean routines, 
-strict_install: False
+strict_install: True
 installing: False
 NUC6_firmware_needed: False
 exFAT_enabled: False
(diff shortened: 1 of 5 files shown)


holta:
Purpose of this PR, as explained by @jvonau: "workarounds that are in place for python need to go away and start with a clean system before resorting to upgrading pip right off the bat"

tim-moody:
the question of being on the latest version of packages needs to be resolved first.

jvonau:
Latest packages of 'what' are you referring too? The python stack has seen out of whack since the use of pip to handle ansible was introduced, every workaround where pip is being upgraded or downgraded was trying to address the broken python stack, that needs to stop and get a fresh baseline with this PR. Using the data gathered with this PR one can then see what the current routine wants/needs with further fixes/refinements available with PRs #162 #163 if needed or wanted.

holta:
Tested with flying colors(*) using Medium-Sized http://wiki.laptop.org/go/IIAB/local_vars.yml while fixing an OSM error when running Ansible 2.2.1.

(*) All dozen server apps were tested, covering the bases @ http://FAQ.IIAB.IO (Question/Answer 3).

Am closing this as @georgejhunt wants to consider superset #259 instead (which also includes virtual environments to isolate python within OSM & KA Lite).

PR #259, opened 2017-09-17
Python stack with virtual environments for kalite and OSM

jvonau:
For testing #256 #162 #163 concurrently


holta:
This PR has been regenerated as #265.

@jvonau does not understand why this (#259) was closed/wiped 13min ago :(

PR #262, opened 2017-09-18
try to shrink mongodb

jvonau:


holta:
This attempt to reduce MongoDB pollution fails: (when running http://download.iiab.io/6.4/rpi/load-venvtest-vpn.txt)
```
root@raspberrypi:/var/lib/mongodb/journal# ls -l
total 786440
-rw------- 1 mongodb nogroup 268435456 Sep 18 11:49 j._0
-rw------- 1 mongodb nogroup        88 Sep 18 11:49 lsn
-rw------- 1 mongodb nogroup 268435456 Sep 18 11:49 prealloc.1
-rw------- 1 mongodb nogroup 268435456 Sep 18 11:49 prealloc.2

root@raspberrypi:/var/lib/mongodb/journal# grep prealloc /etc/mongod*
/etc/mongod.conf:# Disable data file preallocation - will often hurt performance (false by default)
/etc/mongod.conf:noprealloc = true
/etc/mongodb.conf:# Disable data file preallocation.
/etc/mongodb.conf:#noprealloc = true
```
Does mongodb.conf need to be set as well as mongod.conf ?

jvonau:
From '.18' think the config file name is wrong
> ls -la /etc/mong*
-rw-r--r-- 1 root root 2143 Nov 13  2013 /etc/mongodb.conf
-rw-r--r-- 1 root root 5437 Sep 18 11:48 /etc/mongod.conf
ps -Af | grep mongo
root     13140 12461  0 12:11 pts/2    00:00:00 grep mongo
mongodb  28240     1  0 11:48 ?        00:00:13 /usr/bin/mongod --config /etc/mongodb.conf

arky:
Here is an alternative solution: You can prevent mangodb creating large files by turning on " storage.smallFiles". This makes sure all journal prealloc files are small. 

See Journaling section of MangoDB conf documentation https://docs.mongodb.com/manual/core/journaling/#journaling-journal-files

PR #260, opened 2017-09-17
MongoDB: 2 improvements to reduce DB bloat

georgejhunt:
This created a 36MB mongodb direcetory. 
I wish I could squash the squashed commits listing
smoke tested ok

PR #263, opened 2017-09-18
Mongo conf

georgejhunt:
mongo changed which name they look at. mongod.conf vs mongodb.conf. We hit both


jvonau:
um, need to fix up your git repo/branches looks the same as #260 did.

PR #268, opened 2017-09-19
cheap hack to ensure the latest usbmount in installed...

jvonau:
if .14 is not the latest

### DIFF
--- a/roles/2-common/tasks/packages.yml
+++ b/roles/2-common/tasks/packages.yml
@@ -115,6 +115,7 @@
    - libc6
    - bash
    - iptables
+   - usbmount
   when: is_debuntu
 
 # instructions state to start with a fully updated system before starting, stop using


holta:
@georgejhunt plz help review this or defer to [IIAB 6.5](https://github.com/iiab/iiab/milestone/2).

@jvonau wrote: "lets see what the debian-9 test installs...that George has on the go"

holta:
@jvonau clarifies:

> "[this PR268] would just try and update usbmount if there is a later one available from debian as we have not figured out if debian offer usbmount at all out of the box so we provide .14 as a fail-safe.  If there is a newer version available then the update would occur there if the PR is in place."

jvonau:
> The change would do no harm where it is placed, but would paper over the exception that installs .14, that should go away if debian offers .14+

holta:
I'm adding this to Thurday's http://minutes.iiab.io agenda for [IIAB 6.4](https://github.com/iiab/iiab/milestone/1) (imminent) versus [IIAB 6.5](https://github.com/iiab/iiab/milestone/2) discussion.

holta:
@jvonau let's resolve this or push it to conclusion please!

holta:
@jvonau asked to close this during our community/team call (http://minutes.iiab.io) explaining it does not make a difference in his testing.

Issue #191, opened 2017-09-02
Our Ansible messaging is unnecessarily scaring away new contributors


holta:
3 examples, that longtime Mozilla community manager @arky reminds us have very real consequences -- let's start by cleaning up false alarms & misinformation that sends people running -- towards the authentic longer-term goal that Arky outlines: (actionable Ansible messaging, quickly understandable by free software/open source veterans, with verbose flag etc if nec?)

1. Right on top “ubuntu Operating System is not supported” and “raspbian Operating System is not supported" causes people to (A) give up and/or (B) not believe anything they read from that point onwards :/  #158
2. Our installation of Ansible 2.2.0 sets off loud alarm bells in new contributors when it asks to upgrade pip from “8.1.1 to 9.0.1”.  Can “pip install ansible==2.2 --disable-pip-version-check” be added to scripts/ansible?  PR #181
3. Bright Red messaging within “ignored” sections of Ansible runs represent a flood of false alarms, that cause new contributors to Stop Believing Anything They Read!  These can easily be suppressed.  @jvonau told me there's an easy way to nail that, _to help us engage rather than scare off new contributors!_

holta:
Aside to bring Readable+Topical Messaging to new contributors of all kind:

In #204 (unused/obsolete stuff) @georgejhunt brings in Release Notes 6.0 and 6.1 from our older xsce/xsce repo, so they do not get lost, remaining a part of offline IT/implementer docs @ http://box/info

jvonau:
Point 3 would of entailed using an installing/installed marker to flag to suppress warnings expected on a first run where we know things are going to be missing, I'm going use nextcloud as the example where one is used in /opt/nextcloud/config/config.php. There was the frame work I put into place to support that but it has been removed(#85 base_ver) except for the the installing boolean. I'm going to say that it's going to be the same old song and dance in one gets proposed so I'm not interested now.

holta:
Thanks @jvonau

Few software engineers (alive or dead) ENJOY writing (nevermind maintaining!) clean documentation during their spare time...

Nevermind crisply legible explanations enabling contributors around the planet, whose 1st language is *rarely* English &mdash; a requirement we were not raised to take seriously &mdash; and yet now is firmly upon us...

Nor do many software developers yet have experience with illustrations, https://obsproject.com screen capture walk-thrus, or http://simpleshow.com cartoons that the world HAS come to expect !

_Hence the reason we all need to push ourselves forward together on this Epic Front -_

[...]

PR #275, opened 2017-09-19
Big Red Messages

jvonau:
Can be tested with just with the git instructions and `runtags prep` for working on Point 3 of #191


tim-moody:
when many lines have the same conditional I would create a new play invoked with that conditional

jvonau:
This is addressing the feedback given to the user, not the design choice
used. Yes one could of done that when the port to the Debian family was
introduced but was not, so I'm not going to try and overrule that choice
that was made then. You are free to fix issue #191  any other way you want.

jvonau:
smoketested 
[0921-revised_net_feedback.txt](https://github.com/iiab/iiab/files/1323618/0921-revised_net_feedback.txt)

PR #276, opened 2017-09-19
WordPress 4.8.1 instead of 4.8

holta:
In preparation for IIAB 6.4 release.


holta:
Ref #27 ("WordPress 4.8 instead of 4.7.3")

PR #277, opened 2017-09-19
Allow iiab-admin to be optional again

jvonau:
The iiab_admin play can be made optional, but the user would need to set iiab_admin_user in local_vars.yml to a valid sudo'er to ensure lib-usb is set correctly and grants access to the admin console. This is geared towards users who are using the install from scratch method and don't really need a second admin account on the box, seems a bit intrusive to be forced to remove something after the fact. It would be less of a pain if that play can be omitted from the beginning IMHO. I'm all for having iiab-admin available for image creation use but for deployment use I can see where it might get in the way of securing the a large number of machines. It's easier to not add than it is to subtract later. This **security feature** was in roles/xsce-admin/tasks/main.yml at [one point ](https://github.com/tim-moody/iiab/commit/2a137e83d12afd6bf4283e4390f02fa48feee408#diff-98213d45fcc0baaf3e702baaf48199a6) but was removed for some unknown reason.


holta:
@tim-moody & @georgejhunt  please review this recommendation, Thanks!

tim-moody:
I'm not clear on the objective: a) have no admin user, as the variable name suggests, or b) have a different admin user than iiab-admin.

The admin user name is already softcoded. What happens if iiab_admin_user is set to pi, for example?

jvonau:
On Wed, Sep 20, 2017 at 10:31 AM, Tim Moody <notifications@github.com>
wrote:

> I'm not clear on the objective: a) have no admin user, as the variable
> name suggests, or b) have a different admin user than iiab-admin
>

Both, {{ iiab_admin_user }} sets the name of the user in the iiab-admin
playbook and for logging into the gui. The iiab-admin playbook can be
omitted if {{ iiab_admin_user }} is set to a different vaild suder that
already exists on the system.

.
>
> The admin user name is already softcoded. What happens if iiab_admin_user
> is set to pi, for example?
>
​I'll go with unexpected changes to the primary user account, hence the
need to skip the altering of it with no_admin.
I'll give you it could be call 'no_iiab_admin' but I'm more interested in
function than what its name it. ​
​

PR #267, opened 2017-09-19
Revised net

jvonau:
replacing #243 needs smoketesting

PR #278, opened 2017-09-19
cover bad selection of install False enabled True by enduser

jvonau:
This should not occur with proper selection of options but yea never know what is modified in the field by the endusers so guard against the possibility of a bad choice.


jvonau:
Ref #267 will need rebasing if this hits master first

holta:
@tim-moody please review this squid / wondershaper rec in coming week? Thx!

PR #279, opened 2017-09-20
Avoid 1 prep's install routine post install

jvonau:
No need for the console or iiab-network to repeat the installation steps in 1-prep post-install. Subset of #267. This PR is targeted at 6.4, but could slip to 6.5, is a stepping stone to [future first run logic](https://github.com/iiab/iiab/compare/master...jvonau:shhh2) found in my shhh2 branch that builds upon #275 to address #191's point 3.


holta:
Might not make the cut as 6.4 needs to be cut off soon — smoke-testing, functional testing and integration testing increasingly critical at this point!

jvonau:
too much for 6.4

PR #280, opened 2017-09-20
First run logic

jvonau:
continuation of #275 and #279

### DIFF
--- a/iiab.yml
+++ b/iiab.yml
@@ -6,7 +6,6 @@
   - vars/default_vars.yml
   - vars/{{ ansible_local.local_facts.os_ver}}.yml
   - vars/local_vars.yml
-  - /etc/iiab/config_vars.yml
 
   roles:
       - { role: 1-prep, tags: ['prep','platform','base'] }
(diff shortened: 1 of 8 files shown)


holta:
@georgejhunt let us know when u can review this, Thx!

tim-moody:
this pr eliminates the effect of using admin console as no changes done there will ever be used by runansible

jvonau:
I don't understand the objection, runansible is intended to run **only once** to bootstrap the system, to be able in install the console afterwords, once complete should not be used again. For testing a second pass,  to simulate running the gui for the first time, should work without issue. I agree that runtags would disregard config_vars, but is being targeted at developers as a debugging tool and not enduser use, although it can be.  Don't understand why one would need to 'runansible' a second time once the gui is functioning. Please explain.

tim-moody:
OK. In that case let runansible check for the existence of config vars and stop with a notice if found, since creating it is the last thing done.

jvonau:
Think we can have that both ways, add the ability to override the check if '
runansible  --debug' is passed on the cmdline?

jvonau:
runansible --debug support added

jvonau:
smoketested 
[0921-revised_net_feedback-clean.txt](https://github.com/iiab/iiab/files/1323612/0921-revised_net_feedback-clean.txt)

PR #282, opened 2017-09-20
turn postgresql off by default and let computed vars turn on

### LINKED ISSUE #281: postgresql always installed
default vars has install true
computed vars sets install true if moodle or pathagar enabled

nothing sets to false

default vars should have false for install and enable and let computed vars turn on as needed

tim-moody:
fixes #281

### DIFF
--- a/vars/default_vars.yml
+++ b/vars/default_vars.yml
@@ -144,7 +144,7 @@ samba_install: False
 samba_enabled: False
 
 # postgresql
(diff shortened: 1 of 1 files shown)


holta:
Can someone smoke-test this on a VM at min?

tim-moody:
passed smoke test on rpi3
passed functional test on rpi3

PR #285, opened 2017-09-20
Tinynginx

georgejhunt:
This is the minimal set of changes to get nginx running, without changes to all the services that need configuration to work under nginx.


holta:
@georgejhunt plz re-open if this is needed.

Issue #293, opened 2017-09-20
it's time to stop hosting WordPress at download.iiab.io


holta:
Even if WordPress had not released an important [4.8.2 security/maintenance release](https://wordpress.org/news/2017/09/wordpress-4-8-2-security-and-maintenance-release/) yesterday, @tim-moody agrees with me it's now time for [roles/wordpress/defaults/main.yml](https://github.com/iiab/iiab/blob/master/roles/wordpress/defaults/main.yml) to use generic/dynamic URL https://wordpress.org/latest.tar.gz to install directly from WordPress.

Instead of a trail of stale/insecure installers we're caching at http://download.iiab.io/packages that wastes our time for no legit reason (subtracting value far more that adding value...)

Even if WordPress did not auto-upgrade on install (and on an ongoing basis, WordPress now auto-upgrades without manual intervention, and for good reason given sloppy sysadmin practices are a dangerous de facto norm...)

In any case: WordPress (like NextCloud) provides higher quality QA & hosting uptime than we do, so _humility on our part_ carries the day (others are far better at tracking/securing/QA'ing WordPress than we are!)  Conclusion: the transition to https://wordpress.org/latest.tar.gz now (walking away from hosting WordPress @ download.iiab.io) is the way forward, following in @floydianslips's footsteps with our [Nextcloud Ansible playbook](https://github.com/iiab/iiab/blob/master/roles/nextcloud/defaults/main.yml)'s generic installer URL: https://download.nextcloud.com/server/releases/latest-12.tar.bz2

Likewise in future we might even consider: https://download.nextcloud.com/server/releases/latest.tar.bz2

_PS let's monitor that offline instances of WordPress continue to work (and self-upgrade on their own, as is now mandatory when they briefly find Internet access...)  Of course there are integration testing/blowback risks, but these are more than worth it &mdash; everybody knows WordPress security is a PHP treadmill &mdash; so if we choose to offer the world's most popular blogging platform, we too like the rest of the offline+online world need to step up to the plate and be responsive global stewards and citizens._

holta:
FYI I very carefully checked for the possibility for WordPress 4.8.2 just 48-72hours ago(*) but security issues were likely embargoed (kept hush-hush, to contain exploits as much as possible, prior to WordPress servers upgrading themselves worldwide...)

(*) prior to and while working on #276 ("WordPress 4.8.1 instead of 4.8")

holta:
Functional testing of http://box/wordpress also completed.

[...]

PR #296, opened 2017-09-20
always download latest WordPress directly

holta:
Explained @ https://github.com/iiab/iiab/issues/293

### DIFF
--- a/roles/wordpress/defaults/main.yml
+++ b/roles/wordpress/defaults/main.yml
@@ -1,4 +1,5 @@
-wordpress_src: wordpress-4.8.1.tar.gz
+wordpress_base_url: https://wordpress.org
+wordpress_src: latest.tar.gz
 wp_db_name: iiab_wp
 wp_db_user: iiab_wp
 wp_db_user_password: changeme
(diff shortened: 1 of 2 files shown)


holta:
Surfacing my commit message so it's not so buried:

> Downloads latest WordPress directly from https://wordpress.org/latest.tar.gz whereupon Ansible's get_url renames it correctly to the actual WordPress version saved into /opt/iiab/downloads (e.g. wordpress-4.8.2.tar.gz)
> 
> Thanks @tim-moody who provided me Ansible syntax "register: wp_download_output" to capture this dynamic filename that cannot be predicted in advance month-by-month...and then moves that file into position using "unarchive: src={{ wp_download_output.dest }}  dest=/library"

jvonau:
> TASK [wordpress : Get the WordPress software] **********************************
skipping: [127.0.0.1]
TASK [wordpress : Copy it to permanent location /library] **********************
fatal: [127.0.0.1]: FAILED! => {"failed": true, "msg": "the field 'args' has an invalid value, which appears to include a variable that is undefined. The error was: 'dict object' has no attribute 'dest'\n\nThe error appears to have been in '/opt/iiab/iiab/roles/wordpress/tasks/install.yml': line 6, column 3, but may\nbe elsewhere in the file depending on the exact syntax problem.\n\nThe offending line appears to be:\n\n\n- name: Copy it to permanent location /library\n  ^ here\n"}
	to retry, use: --limit @/opt/iiab/iiab/iiab.retry
PLAY RECAP *********************************************************************
127.0.0.1                  : ok=228  changed=48   unreachable=0    failed=1   

This change fails with disregard_network set for testing which implies that off-line operation of ansible is broken.

Issue #254, opened 2017-09-16
What's causing 801MB /var/lib/mongodb ? [prealloc files nec for Sugarizer?]


holta:
Sugarizer is *not* causing this (certainly not directly) as /var/lib/mongodb is [sometimes!] only 33MB with Sugarizer 0.9 fully installed via medium-sized http://wiki.laptop.org/go/IIAB/local_vars.yml

This ballooning 801MB /var/lib/mongodb [sometimes!] arises from big-sized http://wiki.laptop.org/go/IIAB/local_vars_big.yml (on master on RPi3) and so [appeared to!] be arising as a result of one of these 10:

1. squid
2. dansguardian
3. wondershaper
4. cups
5. samba
6. phpmyadmin
7. elgg
8. dokuwiki
9. moodle
10. monit

holta:
@georgejhunt observed this 801MB "pollution" is essentially just these 3 files:
```
root@box:/var/lib/mongodb# ls -l journal/*
-rw------- 1 mongodb nogroup 268435456 Sep 16 10:52 journal/prealloc.0
-rw------- 1 mongodb nogroup 268435456 Sep 16 04:37 journal/prealloc.1
-rw------- 1 mongodb nogroup 268435456 Sep 16 04:37 journal/prealloc.2
```
Could this be some kind of buggy attempt at a Sugar Journal ?

Still baffling to us that this appears on one Sugarizer RPi3 (built w/ [big local_vars](http://wiki.laptop.org/go/IIAB/local_vars_big.yml)) but not the other (built w/ [medium local_vars](http://wiki.laptop.org/go/IIAB/local_vars.yml)).

holta:
@llaske can you confirm Y/N if Sugarizer's Journal stored in /var/lib/mongodb/journal ?

Do you know why installing Sugarizer 0.9 on RPi3 (part of IIAB 6.4, on lastest Raspbian Stretch Lite) mysteriously sometimes ends up with 769MB in here?
```
root@box:/var/lib/mongodb# ls -l journal/*
-rw------- 1 mongodb nogroup 268435456 Sep 16 10:52 journal/prealloc.0
-rw------- 1 mongodb nogroup 268435456 Sep 16 04:37 journal/prealloc.1
-rw------- 1 mongodb nogroup 268435456 Sep 16 04:37 journal/prealloc.2
```
_Thanks if you know what might be going on with the 3 mysterious files above!_

holta:
[...]

PR #299, opened 2017-09-21
smallfiles=true added to mongod.conf

holta:
Suggested by @arky in https://github.com/iiab/iiab/issues/254

### DIFF
--- a/roles/mongodb/templates/mongod.conf
+++ b/roles/mongodb/templates/mongod.conf
@@ -110,7 +110,7 @@ noprealloc = true
 #quotaFiles = 8
 
 # Use a smaller default file size (false by default)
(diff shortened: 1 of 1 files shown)


tim-moody:
and the smoke test?

holta:
Smoke tested earlier.

I forgot to mention earlier that /etc/mongo**db**.conf (for MongoDB <= 2.4) is now the same file as /etc/mongo**d**.conf (for MongoDB >= 2.6) as a result of @georgejhunt's change earlier this week, in conjunction with #260 and #262.

As you can see in lines 30 and 31 of https://github.com/iiab/iiab/blob/master/roles/mongodb/tasks/main.yml

PR #301, opened 2017-09-21
move ansible tmp file to eliminate ~

tim-moody:


tim-moody:
smoke tested on rpi
and ansible ran to completion

PR #310, opened 2017-09-25
smoke testing - correct selinux for non-redhat installs

jvonau:


holta:
Smoke-tested after layering it upon [6.4-RC4](https://github.com/iiab/iiab/releases/)

PR #311, opened 2017-09-25
missing mkdir in runansible

jvonau:


holta:
Smoke test completed on Debian 9.1 (GH) and Raspbian Stretch Lite (AH).

Issue #330, opened 2017-09-28
Elgg 2.3.4 fails to install w/ runtags or Admin Console since late Sept


holta:
When installing on Raspbian Stretch Lite using http://download.iiab.io/6.4/rpi/load-ansible2.4.0-vpn.txt or similar.  This is the kind of error that appears in browser: (when trying to visit http://box/elgg)
```
Fatal error in exception handler. Check log for Exception at time 1506558199
```
On reboot, it shows:
```
Fatal error in exception handler. Check log for Exception at time 1506558624
```
Refs: #217 #306 #307 https://github.com/iiab/iiab/pull/316#issuecomment-332694498

holta:
This failure of http://box/elgg did not appear when I re-ran a tightened up http://download.iiab.io/6.4/rpi/load-ansible2.4.0-vpn.txt on a completely new Raspbian Stretch Lite!

@jvonau did you solve it within PR #316 somehow?

jvonau:
Not that I'm aware of... Would be something to check on after a second run
thou... Didn't you mention you were going to try a sudo runansible as the
pi user at some point?

holta:
A plausible theory but it didn't quite hold:

Elgg  2.3.4 installed correctly during the 1st ./runansible on a new RPi3, but then http://box/elgg stopped working after "./runtags elgg" &mdash; then somehow ~10min later http://box/elgg seems to have fixed itself ???  Crazy.

I can't reproduce it anymore (subsequent runs of "./runtags elgg" no longer break it) but here's that (mysteriously clean) output of "./runtags elgg" that seemingly caused its working http://box/elgg to (temporarily for several minutes) break:

http://paste.ubuntu.com/25631359/

Observation 1: http://box/elgg worked (at least once anyway) when "wordpress_install: True" and "wordpress_enabled: True" were in local_vars.yml during Ansible 2.4.0's very 1st run.

Observation 2: this failure does NOT seem to happen on RPi machines built using Ansible 2.2.1 (where ".runtags elgg" can be run as often as you want, and http://box/elgg always continues working, insofar as I can tell!?)

Observation 3: when http://box/elgg mysteriously/magically fixed itself ~10min after I ran "./runtags elgg" on one RPi3... the only thing I did on that RPi3 (during those ~10min, no voodoo applied!) was run functional tests of http://box/nextcloud and http://box/wordpress as listed in Question 3 @ http://FAQ.IIAB.IO

[...]

PR #331, opened 2017-09-28
http://box/elgg fatal error: stricter syntax for Ansible 2.4.x ?

jvonau:


holta:
Possibly helps mitigate stuff like #330 (Elgg 2.3.4 regression if installing with Ansible 2.4.0) ?

Might be related to stricter syntax in Ansible 2.4.0 ?

holta:
Moving to target [IIAB 6.4](https://github.com/iiab/iiab/milestone/1) in case there's hopefully time to evaluate prior to release.

holta:
@jvonau I forgot to mention I smoke-tested this successfully, but it shows [no visible improvement](https://github.com/iiab/iiab/issues/330#issuecomment-333348079) in getting Elgg to install.

Despite multiple runs of "./runtags elgg" and many reboots for good measure, ah well back to the drawing board :)

holta:
@jvonau does this ticket have remaining value?

Or do we resolve given Elgg still won't install?

jvonau:
Need further clarification of won't install as you stated that the install is successful but a 2nd pass breaks elgg.

PR #335, opened 2017-09-28
update kiwix-tools_linux64 2017-09-12 to 2017-09-27

holta:


holta:
Reviewed with @tim-moody 

Neither of are thrilled that 2017-09-28 is missing linux64 build, but better to go with yesterday's 09-27 kiwix-tools for linux64 (now republished to http://download.iiab.io/packages/kiwix-tools_linux64_2017-09-27.tar.gz) so that we don't hold up IIAB 6.4's release.

PR #340, opened 2017-09-28
automated testing uses stock fc18 -- permit that

georgejhunt:
I don't think this is candidate for 6.4
not smoke tested


holta:
@jvonau @tim-moody do you prefer this target 6.4 or 6.5?

jvonau:
Let's make TK happy and push it.

holta:
Ref #374 where this is already checked in.

PR #345, opened 2017-09-29
try restarting with unit file after network-online.target

georgejhunt:
not smoke tested


jvonau:
> root@box:/home/iiab-admin# systemctl list-dependencies local-fs.target
local-fs.target
● ├─-.mount
● ├─boot.mount
● ├─systemd-fsck-root.service
● └─systemd-remount-fs.service

> root@box:/home/iiab-admin# systemctl --no-pager list-dependencies rc.local
rc.local.service
● ├─system.slice
● └─sysinit.target
●   ├─dev-hugepages.mount
●   ├─dev-mqueue.mount
●   ├─fake-hwclock.service
●   ├─keyboard-setup.service
●   ├─kmod-static-nodes.service
●   ├─plymouth-read-write.service
●   ├─plymouth-start.service
●   ├─proc-sys-fs-binfmt_misc.automount
●   ├─sys-fs-fuse-connections.mount
●   ├─sys-kernel-config.mount
●   ├─sys-kernel-debug.mount
●   ├─systemd-ask-password-console.path
●   ├─systemd-binfmt.service
●   ├─systemd-hwdb-update.service
●   ├─systemd-journal-flush.service
●   ├─systemd-journald.service
●   ├─systemd-machine-id-commit.service
●   ├─systemd-modules-load.service
●   ├─systemd-random-seed.service
●   ├─systemd-sysctl.service
●   ├─systemd-timesyncd.service
●   ├─systemd-tmpfiles-setup-dev.service
●   ├─systemd-tmpfiles-setup.service
●   ├─systemd-udev-trigger.service
●   ├─systemd-udevd.service
●   ├─systemd-update-utmp.service
●   ├─cryptsetup.target
●   ├─local-fs.target
●   │ ├─-.mount
●   │ ├─boot.mount
●   │ ├─systemd-fsck-root.service
●   │ └─systemd-remount-fs.service
●   └─swap.target


Might want to use after=local-fs.target or rc.local in place of =network-online.target but that is up to you. Maybe udev-reboot.service is not the best name, seems a bit misleading, maybe udev-reload.service?

georgejhunt:
Trying to think of when to have udev unavailable during the boot process is a challenge for me. I don't want udev to be unavailable during network setup. Yes, udev-reload would be better.
passes smoke test on debian9 VM

jvonau:
Pulled this from .14, wondering if a 'dirty' filesystem on the usbstick is blocking usbmount's next step of 'run-parts' that reads the commands in /etc/usbmount/mount.d/ where the '70 ln' script would be executed from.
 > Sep 27 19:49:17 box.lan systemd-udevd[28555]: Process '/usr/share/usbmount/usbmount add' failed with exit code 1.
Sep 27 19:49:17 box.lan kernel: FAT-fs (sdb1): Volume was not properly unmounted. Some data may be corrupt. Please run fsck.

holta:
I was asked to document (in http://FAQ.IIAB.IO etc) that teachers do not need to unmount their USB memory sticks.

In fact we have no options here, realistically they will never unmount USB memory sticks even if we asked them too :}

Finally, FWIW we've never had a problem over the past year or so with [Teacher's Content](http://wiki.laptop.org/go/IIAB/FAQ#Can_teachers_display_their_own_content.3F) (from USB sticks) failing, despite this very common practice (essentially universal practice) of teachers' pulling their USB sticks when they in fact need to leave in a hurry to the next class etc.

holta:
Has this (#345) been smoke-tested on Raspbian?

holta:
Smoke test succeeds but functional test doesn't fully work (as I expected anyway!)

During the 4 reboots after installing http://download.iiab.io/6.4/rpi/load-udevreboot-vpn.txt the still-inserted USB memory stick never appeared at http://box/usb (or /library/www/html/local_content) until it was fully removed from the RPi3 unit and then re-inserted.

Rebooting does not help unfortunately.  Only removing and re-inserting the stick will work (yes the USB stick was inserted during the installation of IIAB).

Not sure if this is fixable?

If not I can document this very basic workaround in http://FAQ.IIAB.IO if nec (asking folks to re-insert the USB stick, even if teachers are rarely building their own IIAB machines today, thing can change fast as our 1-line installation process gets progressively more complete every time...)

georgejhunt:
I think this observation is inherent in using udev, which only responds to
changes. I'm guessing the the first change (during discovery of the device
first time in the boot process) happens before the reload udev to make the
underlying disk be read-write rather than read-only
.
I think it is best to document it rather than spend any more time trying to
refine the timing or the udev reload.

georgejhunt:
But maybe the unit file I created could call a script, which does both the
udev reload, and call the mount script, which will do nothing if the usb
folder does not exist

holta:
@georgejhunt progress is fantastic already -- decision is up to you if more elegance is required or not -- many other fish to fry :)

PS i modified the PR to use "reload" instead of "reboot" per @jvonau's suggestion in the http://minutes.iiab.io live chat, is this ok?

georgejhunt:
or simpler:
I think a unit file can only have one ExecStart, but it can have many
ExecStartPre's -- so I modified the PR.

Can you check to see if it fixed the "discovery of USB at boot problem"?

holta:
Thx @georgejhunt this is now underway:
```
root@box:/opt/iiab/iiab# git branch
* udevreboot

root@box:/opt/iiab/iiab# git pull
remote: Counting objects: 17, done.
remote: Total 17 (delta 11), reused 11 (delta 11), pack-reused 6
Unpacking objects: 100% (17/17), done.
From https://github.com/georgejhunt/iiab
   4ba8a48..6a4c65c  udevreboot -> origin/udevreboot
Updating 4ba8a48..6a4c65c
Fast-forward
 roles/2-common/tasks/udev.yml                                     | 8 ++++----
 .../templates/{udev-reboot.service => udev-reload.service}        | 3 ++-
 2 files changed, 6 insertions(+), 5 deletions(-)
 rename roles/2-common/templates/{udev-reboot.service => udev-reload.service} (56%)

root@box:/opt/iiab/iiab# ./runansible
```
I hope this test will suffice, as I'm (trying!) to avoid yet another 2+ hour install on RPi3.

holta:
@georgejhunt the "discovery of USB at boot problem" is not fixed, per my testing described @ https://github.com/iiab/iiab/pull/345#issuecomment-333323348.  Unfortunately /library/www/html/local_content/ remains empty after multiple reboots, despite the fact USB sticks remain inserted.

It would indeed be Amazing if teachers could place USB sticks inside their IIAB's _while the server's off_ and then turn it on &mdash; so every USB stick's [Teacher's Content](http://wiki.laptop.org/go/IIAB/FAQ#Can_teachers_display_their_own_content.3F) always appears at http://box/usb without bureaucracy &mdash; if you think you're possibly close??

tim-moody:
my impression is that this used to work, at least on some platforms. In fact we had the opposite problem that the link for a usb removed while powered off remained in local_content

georgejhunt:
Adam I need to reboot 10.8.0.26. Is that ok?

holta:
> Adam I need to reboot 10.8.0.26. Is that ok?

Yes!

holta:
Clarif: I just talked to @georgejhunt by phone who will look into new strateg(ies) in a new PR to get http://box/usb working from cold boot.

(USB stick functionality isn't quite hassle-free for teachers just yet, but still a big step forward!)

Issue #326, opened 2017-09-27
debian:  [WARNING]: Consider using file module with mode rather than running chmod


arky:
Harmless warning from Ansible. Report from nightly Jenkins CI run. Please kindly close the bug if it is low priority.

==> debian: TASK [iiab-admin : edit the sudoers file--first make it editable] **************
==> debian: changed: [127.0.0.1]
==> debian:  [WARNING]: Consider using file module with mode rather than running chmod
==> debian: 
==> debian: TASK [iiab-admin : have sudo log all commands it handles] **********************
==> debian: changed: [127.0.0.1]
==> debian: 
==> debian: TASK [iiab-admin : lets wheel sudo without password] ***************************
==> debian: changed: [127.0.0.1]

PR #351, opened 2017-10-01
Using Ansible File module to fix warnings

arky:
Using Ansible file modules instead of 'chmod' to change sudoes file permissions. Fixes warning in https://github.com/iiab/iiab/issues/326


arky:
Removing stale pull request.

Issue #349, opened 2017-10-01
Debian: [WARNING]: Consider using get_url or uri module rather than running wget


arky:
Harmless warning from Ansible. Report from nightly Jenkins CI run. Please kindly close the bug if it is low priority.

> TASK [2-common : Install yum packages] ****************************************************************************************************************
> skipping: [127.0.0.1] => (item=yum-utils) 
> skipping: [127.0.0.1] => (item=createrepo) 
> skipping: [127.0.0.1] => (item=wpa_supplicant) 
> skipping: [127.0.0.1] => (item=linux-firmware) 
> skipping: [127.0.0.1] => (item=syslog) 
> skipping: [127.0.0.1] => (item=xml-common) 
> 
> TASK [2-common : Download usbmount -- not in debian-9] ************************************************************************************************
>  [WARNING]: Consider using get_url or uri module rather than running wget
> 
> changed: [127.0.0.1]
> 
> TASK [2-common : Install usbmount for debian-9] *******************************************************************************************************
> changed: [127.0.0.1]
>

jvonau:
s/b resolved problem code has been changed

Issue #327, opened 2017-09-27
debian:  [WARNING]: Consider using get_url or uri module rather than running wget


arky:
Harmless warning from Ansible. Report from nightly Jenkins CI run. Please kindly close the bug if it is low priority.

==> debian: skipping: [127.0.0.1] => (item=[]) 
==> debian: 
==> debian: TASK [2-common : Download usbmount -- not in debian-9] *************************
==> debian: changed: [127.0.0.1]
==> debian:  [WARNING]: Consider using get_url or uri module rather than running wget
==> debian: 
==> debian: TASK [2-common : Install usbmount for debian-9] ********************************
==> debian: changed: [127.0.0.1]
==> debian: 
==> debian: TASK [2-common : Install packages for Debian] **********************************

jvonau:
code merged

PR #353, opened 2017-10-01
Using Ansible get_url module to fix warnings

arky:
Using Ansible 'get_url' modules instead of 'wget' to download usbmout package. Fixes warnings in bug #327 #349


holta:
@arky is this smoke-tested and/or functionality-test?  Thanks if poss!

georgejhunt:
we migrated to wget from get_url because ansible was running out of tmp
space or memory, but wget failed much less often.

jvonau:
@holta conflicts and push

holta:
This needs testing and @georgejhunt deserves a response or clarification prior to merge:

> we migrated to wget from get_url because ansible was running out of tmp
space or memory, but wget failed much less often.

jvonau:
this one is for a small package, I'll toss it in a jenkins run (think I have in the past, can't recall)

holta:
In response to @georgejhunt's comment above, we agreed during our live call (http://minutes.iiab.io) to trial this new approach on small package (usbmount) here.

PR #309, opened 2017-09-24
Elgg - Fix permissions & handle upgrades

jvonau:

PR #355, opened 2017-10-01
fix elgg: reorder so dbuser can be created

tim-moody:


tim-moody:
runtags smoke tested on rpi
runtags tested on existing install (with no db user)
and after rm /opt/elgg, drop database elggdb and drop user Admin@local, etc.

holta:
Amazing Progress: but just to be sure, I smoke-tested & functionality-tested this PR #355 in a dozen different ways over the last 6 hours, not just with "./runtags elgg" but also with many clean installs &mdash; also retested extensively after [Jerry's amendment for Red Hat etc](https://github.com/iiab/iiab/pull/355/commits/21aa6490f948a0bbae22e5c169fbb3dccb03139c).

NOTE this bug was all too easy to miss when installing using the stock [medium-sized](http://wiki.laptop.org/go/IIAB/local_vars.yml) or [big-sized](http://wiki.laptop.org/go/IIAB/local_vars_big.yml) local_vars.yml which now contain:
```
elgg_install: True
elgg_enabled: True
```
So one test was done as above (just to check again regression) but far more importantly as below, since the original #330 bug was triggered by installs containing a local_vars.yml as follows:
```
elgg_install: True
elgg_enabled: False
```
Summary: I did several tests as above &mdash; followed by "./runtags elgg" or enabling Elgg via the Admin Console (http://box/admin -> Configure tab -> Service Enabled -> check box for "ELGG" -> Save Configuration -> Install Configured Options).

All tests worked so I will now merge!

PS above tests completed with Ansible 2.4.0 on 2 different RPi3's.  Finally "./runtags elgg" was generally run 2 or 3 times after each initial install, checking http://box/elgg all along the way, to be sure.

_Background: this bug appears to have nothing to do with Ansible 2.4.0 as earlier suspected.  It (appears) more likely it flared up from 6-day-old PR #309, because we did not test multiple runs of "./runtags elgg" back then._

PR #357, opened 2017-10-01
typo Provisioally -> Provisionally

holta:


tim-moody:
please keep in mind that in computed vars

- name: Turn on mysql if elgg or rachel enabled
  set_fact:
    mysql_install: True
    mysql_enabled: True

# we decided to enable mysql unconditionally
#  when: elgg_enabled or rachel_enabled or owncloud_enabled or phpmyadmin_enabled or wordpress_enabled or iiab_menu_install

Issue #337, opened 2017-09-28
3 red warnings when running "systemctl" (and on boot too?)


holta:
Running "systemctl" shows these 3 lines in red all RPi systems (built on Ansible 2.2.1 or 2.4.0) for now, which may or may not be a concern:

* isc-dhcp-server.service      loaded failed failed    LSB: DHCP server
* mongodb.service              loaded failed failed    High-performance, schema-free document-oriented database
* sugarizer.service            loaded failed failed    sugarizer.service

Solvable with #267 that @jvonau wants to rebase?

As mentioned in #316 (replacement for scripts/ansible & and exit codes for runansible).

holta:
@georgejhunt, ~10 days ago I thought you helped me check in code to make the red boot message "Failed to start sugarizer.service." go away?

As this pain-in-the-ass(*) message disappeared for many days over the last couple weeks (thanks to you, I thought!)  But for whatever reason, this loud/red message on startup is now back.

(*) _a huge distraction for all new implementers, as Josh K reconfirmed today :-(_

PS isc-dhcp-server.service & mongodb.service are not quite as serious, as they do not cause panic during startup FWIW.

holta:
@jvonau requests we resolve this issue now, and moving forward isc-dhcp-server.service progress will instead be tracked on #347.

Clarif: @jvonau fixed Sugarizer's red warning on boot per @kananigit's request in #359!

Issue #240, opened 2017-09-11
Sugarizer 0.8 to 0.9


holta:
Anyone know if this upgrade is as easy as changing [roles/sugarizer/defaults/main.yml](https://github.com/iiab/iiab/blob/master/roles/sugarizer/defaults/main.yml) from:
`sugarizer_version: 'sugarizer-0.8'`
to:
`sugarizer_version: 'sugarizer-0.9'`

...while dropping sugarizer-0.9.tar.gz into [download.iiab.io/packages](http://download.iiab.io/packages/) ?

holta:
Or: defer to [IIAB 6.5](https://github.com/iiab/iiab/milestone/2) if this upgrade is complex?

jvonau:
Placing sugarizer-0.9.tar.gz on d.iiab.io would allow a quick edit for local testing. I'd use local_vars to define sugarizer_version: during testing as a quick down & dirty way to see how 0.9 behaves.

holta:
@georgejhunt thank you for having posted [download.iiab.io/packages/sugarizer-0.9.zip](http://download.iiab.io/packages/sugarizer-0.9.zip) [since changed to [download.iiab.io/packages/sugarizer-0.9.tar.gz](http://download.iiab.io/packages/sugarizer-0.9.tar.gz)] from [github.com/llaske/sugarizer/archive/v0.9.zip](https://github.com/llaske/sugarizer/archive/v0.9.zip) [and similar places!]

Is there something I can smoke test, per @jvonau suggestion above or similar?

holta:
RELATED: during every boot Sugarizer (0.8 for now) says:

`[FAILED] Failed to start sugarizer.service.`

As reconfirmed by Curt Thompson, who finds this red error on boot to be jarring for new folk especially.

(Perhaps this boot-time ugliness can this be fixed with Sugarizer 0.9 now?)

jvonau:
from #211 warnings.txt 
> Sep 10 07:28:41 raspberrypi systemd[1]: [/etc/systemd/system/sugarizer.service:1] Unknown section 'unit'. Ignoring.
Sep 10 07:28:41 raspberrypi systemd[1]: [/etc/systemd/system/sugarizer.service:9] Failed to parse sec value, ignoring: 10                       # Restart service after 10 seconds if node service crashes
[...]

PR #359, opened 2017-10-02
disable systemd sugarizer.service

jvonau:


holta:
Should do 2 things:
1. Allow the following "permanent for now workaround" in all our defaults local_vars.yml files:
sugarizer_install: True
sugarizer_enabled: False
2. Solve #337 (hiding Sugarizer's red/alarmist warning on boot, similar to @georgejhunt's [suggestion from 18 days ago](https://github.com/iiab/iiab/issues/240#issuecomment-329554230) in #240)

PR #361, opened 2017-10-02
runtags - respect config_vars if present

jvonau:


holta:
Am resolving as I believe this is already done via #364.

Please re-open as necessary.

PR #363, opened 2017-10-02
make it always first run. re-examine issue for 6.5

georgejhunt:
blast always first_run


holta:
I chose compromise #364 instead of #363 and #365.

Please re-open if later necessary.

PR #364, opened 2017-10-02
use iiab.env for first_run test

jvonau:


jvonau:
Confirmed with jenkins this is the only change needed, log attached. Unsure why @holta 'clean' test failed by hand. Guess is some prep'ing done prior to runansible messed with the results and panic ensued. Suggested addition #361 
[run33.txt](https://github.com/iiab/iiab/files/1350678/run33.txt)

holta:
Smoke-tested & functionality-tested across [my usual 16 server app tests](http://wiki.laptop.org/go/IIAB/FAQ#What_can_I_do_with_E-books_and_Internet-in-a-Box.3F)  with our most common installer:
http://download.iiab.io/6.4/rpi/load-vpn.txt

Also with our 2nd most common installer:
http://download.iiab.io/6.4/rpi/load-min-vpn.txt

This seems like a great compromise (between #363 and #365) to get master working again, so I'll merge in the coming minutes if there are no objections.

PR #365, opened 2017-10-02
Revert "test revert"

jvonau:
Reverts iiab/iiab#362


holta:
I chose compromise #364 instead of #363 and #365.

Please re-open if later necessary.

PR #380, opened 2017-10-06
permissions - storaged as execuables in git - installed files need not be executable

jvonau:
Long standing laziness on my part that should be been addressed in XSCE. There are other examples throughout the git tree, but this PR deals with the network role only. Should this PR expanded to include other roles or should I do another PR for those instances?


holta:
Smoke-tested Ubuntu 16.04 LTS and earlier Raspbian.

Merging per community call (http://minutes.iiab.io)

PR #381, opened 2017-10-06
ensure network is completed before saying first_run is complete

jvonau:
untested.

### DIFF
--- a/roles/1-prep/tasks/prep.yml
+++ b/roles/1-prep/tasks/prep.yml
@@ -32,25 +32,6 @@
         mode=0755
         state=directory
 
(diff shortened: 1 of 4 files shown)


jvonau:
Now tested in jenkins

holta:
Confirmed Working!!

e.g. ./runansible broke twice on Raspbian Desktop due to "Temporary failure resolving 'mirrordirector.raspbian.org'" possibly caused by overly distant WiFi router, e.g. [2-common : Install common packages] and also after that.

Recovery / Continuing on from there now works, as a result of #381 being checked in.

PR #382, opened 2017-10-07
wondershaper doesn't work correctly out of the box and is unmaintained

jvonau:


holta:
Cosmetic changes to default_vars.yml & local_vars.yml are Very Important for new folk, to make onboarding very tangible!!

Hopefully we can integrate these (very important) readability changes together, sometime closer to IIAB 6.5 feature-freeze or code-freeze if possible?

holta:
default_vars.yml needs more work for sure, but has made huge progress 2 days ago with PRs #389 & #393 , so let's attack this in a more high-level way later.

PR #388, opened 2017-10-10
don't exclude ansible from updating for is_redhat

jvonau:


holta:
Looks extremely safe despite not having been smoke-tested, as we a need a PR to test Travis with now -- any objections to merging this in coming moments?

Issue #384, opened 2017-10-09
2 obsolete checkboxes in Admin Console?


holta:
These 2 checkboxes appear in most all installs:
```
[ ] Services for XO Laptops such as Registration, Backup, and the Activity Server
[ ] Collect statistics on the use of Sugar Activities.
```
Are they at all useful/functional at this time?

Should we hide them and/or document their limitations for [IIAB 6.5](https://github.com/iiab/iiab/milestone/2)?

Especially given 5-xo-services was commented out of the current [iiab.yml](https://github.com/iiab/iiab/blob/master/iiab.yml) as these painfully slows down RPi installs especially:

#370 "comment out & document 5-xo-services as it's slow as a dog"
PR #371 "comment out 5-xo-services (too damn slow)"

holta:
@tim-moody wrote:
> Make sure they are install false, not just missing 

Each machine's https://github.com/iiab/iiab/blob/master/vars/default_vars.yml contains:
```
sugar_stats_install: False
sugar_stats_enabled: False
```
Each machine's http://wiki.laptop.org/go/IIAB/local_vars.yml contains:
```
# sugar_stats_install: False
# sugar_stats_enabled: False
```
Where else should I look to explain these 2 "zombie" checkboxes?

PS Everything else under 5-XO-SERVICES appears likewise (False in default_vars, commented out in local_vars).

tim-moody:
xo-services is due to the name change to xo_services. sugar_stats was also changed from sugar-stats. admin console needs this change as well.

tim-moody:
actually adm cons already has xo_services. please confirm  that this is an issue since you added xo_services_install: False.

holta:
[default_vars.yml](https://github.com/iiab/iiab/blob/master/vars/default_vars.yml) contains the following: (on all these stock IIAB 6.4 machines)
```
xo_services_enabled: False
```
"xo_services_install:" (True or False) is missing above.  Is this in any way related to the "Services for XO Laptops" checkbox appearing on all these stock 6.4 machines?

[local_vars.yml](http://wiki.laptop.org/go/IIAB/local_vars.yml) contains the following: (on all these stock IIAB 6.4 machines)
```
# xo_services_install: False
# xo_services_enabled: False
```

holta:
Ref https://github.com/iiab/iiab-admin-console/issues/35 "name changes (- to _)"

tim-moody:
put xo_services_install: False in default_vars

[...]

PR #389, opened 2017-10-11
default_vars.yml: add "xo_services_install: False" & make file (more) readable

holta:


holta:
Added "xo_services_install: False" per @tim-moody's request in #384.

Smoke-tests on (existing) Ubuntu installation.

Unfortunately this doesn't seem to fix either of the obsolete (or certainly misbehaving) Admin Console checkboxes in #384

tim-moody:
if this is a fresh install or you have rebooted then I don't know what the issue is

holta:
> if this is a fresh install or you have rebooted then I don't know what the issue is

A clean install is not possible (kalite cannot install, due to #392) but I forced a clean install this morning by setting all 3 kalite variables to False in local_vars.yml
Context: clean install on Raspbian Lite 2019-09-07 attempted in the usual way (http://download.iiab.io/6.5/rpi/load.txt)

Great news is that both ghost checkboxes no longer appear!  As a result of "xo_services_install: False" that I added to default_vars.yml in #389 here.

Issue #398, opened 2017-10-12
IIAB 6.4+ fails to install on Raspbian Stretch with Desktop (Pixel)


holta:
IIAB 6.4 (and presumably IIAB 6.5/master) fails to install here:

>TASK [iiab-admin : Do the same if running on raspbian] ***************************
failed: [127.0.0.1]: FAILED! => {"failed": true, "msg": "The conditional check 'lx.stat.isdir is defined and lx.staat.isdir and is_rpi and is_debuntu' failed. The error was: error while evaluating conditional (lx.stat.isdir is defined and lx.staat.isdir and is_rpi and is_debuntu): 'dict object' has no attribute 'staat'\n\nThe error appears to have been in '/opt/iiab/iiab/roles/iiab-admin/tasks/main.yml': line 31, column 3, but may\nbe elsewhere in the file depending on the exact syntax problem.\n\nThe offending line appears to be:\n\n\n- name: Do the same if running on raspbian\n  ^ here\n"}

Bug report thanks to TK Kang.  See his screenshot here:

![20171012_111557s](https://user-images.githubusercontent.com/2458907/31495713-a992b324-af26-11e7-8d19-76f1c3db94af.jpg)

PR #399, opened 2017-10-12
fix roles/iiab-admin/tasks/main.yml typos for Raspbian With Desktop

holta:
Hopefully solves #398 !


holta:
Smoke-testing on Raspbian Stretch With Desktop succeeds past the point when [roles/iiab-admin/tasks/main.yml](https://github.com/iiab/iiab/blob/master/roles/iiab-admin/tasks/main.yml) failed for TK.

I will merge if there are no objections in the coming minutes, so TK, Mexicans & Haitians who need this can test it more fully.

PR #405, opened 2017-10-12
refresh-wiki-docs.sh was producing broken links so I overhauled it

holta:


holta:
@georgejhunt: prior regex's were overbroad creating many broken links.  I overhauled almost everything, with this central/lifesaver regex fixing many things:
```
sed -i -r "s|https://github.com/iiab/iiab/wiki/([-.0-9A-z]*)|\1.html|g" $DESTPATH/$f
```
[-.0-z] also would have worked [both work on Raspbian anyway, neither on Ubunty 16.04: see #407 for more universal regular expression!] but the above is more explanatory.  In short, the above line replaces these 2 older lines which were mangling many offline hyperlinks:
```
sed -i -r "/.*#.*/ s|$REPO/$REPONAME/wiki/(.*)(#.*)\">|./\1.html\2\">|" $OUTPUT/$FTRIMMED.html
sed -i -r "/.*#.*/! s|$REPO/$REPONAME/wiki/(.*)\">|./\1.html\">|" $OUTPUT/$FTRIMMED.html
```
Also the link-fixing logic was missing many of the downloaded files, so I re-organized putting that new for-loop at the end.

And many other small improvements hopefully making offlined docs (http://box/info) a lot more usable.

PS scraping/offlining is an art as much as a science, but at least the most important links now work offline.  We can improve upon this over time with "fieldback" informing what's truly needed!

holta:
@georgejhunt is the "exit 0" at the very end of refresh-wiki-docs.sh truly nec??

(i.e. does this permit Ansible to continue running with some kind of a "false positive" necessary evil when the scraping/offlining of IT docs happens to fail for connectivity-reasons-or-any-others-reasons ?)

holta:
Root Cause Clarif: links were being broken by sed's regular expressions greedily grabbing too much, across entire lines that happen to contain 2-or-more URL's, basically merging several URL's into 1, creating a bit of chaos.

In short: using `#` or `">` as a right-delimiter for URL's turned out not to be safe, when there are often many such URL's per line (the sed command seems to grab the longest regex-matching string it can find, hence creating a mess...)

Among many other small/stale patterns I hopefully cleaned up.

holta:
Smoke-tested and functionality-tested, so I will merge in coming minutes if there are no objections.

(Keep the improvements coming as this is not the final word in boxing/scraping for offline populations extremely eager to build their own digital libraries!)

PR #406, opened 2017-10-13
further basic refinements to refresh-wiki-docs.sh

holta:

PR #407, opened 2017-10-13
refresh-wiki-docs.sh regex [-.A-Za-z0-9] needed on Ubuntu etc where A-z fails

holta:


holta:
Smoke-tested on Ubuntu 16.04 LTS (4.2.2) & Raspbian Stretch Lite 2017-09-07 (sed 4.4) so if there are no objections I'll merge in coming moments to get Ubuntu back on track.

holta:
Fixes PR #405 -> PR #406 for Ubuntu, and possibly other OS's too.

PR  #408 will allow /usr/bin/iiab-refresh-wiki-docs to run on legacy IIAB systems too, after "git pull"

PR #408, opened 2017-10-13
refresh-wiki-docs.sh: permit "git pull" on earlier IIAB's, leaving legacy subdirs like info/html untouched

holta:


holta:
Smoke-tested & functional-tested on Raspbian Stretch Lite & Ubuntu 16.04 LTS.

Builds off PR's #405 #406 #407.

holta:
@georgejhunt should we follow the advice of https://stackoverflow.com/questions/20796200/how-to-iterate-over-files-in-a-directory-with-bash in case /library/www/html/info is empty due to connectivity glitches?

> whenever you iterate over files by globbing, it's good practice to avoid the corner case where the glob does not match (which makes the loop variable expand to the (un-matching) glob pattern string itself).
> 
> For example:
> 
> for file in Data/*.txt; do
>     [ -e "$file" ] || continue
>     # ... rest of the loop body
> done

Issue #419, opened 2017-10-17
Sugarizer blocks post-install Ansible runs @ "Actually get it installed on is_debuntu"


holta:
"./runtags sugarizer" and "./runansible" fail consistently immediately after a fresh install of http://download.iiab.io/6.5/rpi/load.txt on Raspbian Lite.  I've tried this on 2 different RPi3's.

Same as https://github.com/iiab/iiab/issues/193#issuecomment-335990225 from 6 days ago?  Which we had thought was fully fixed by #304, but apparently not:

> TASK [sugarizer : Actually get it installed on is_debuntu] *********************
> ok: [127.0.0.1] => (item=nodejs)
[...]

PR #420, opened 2017-10-17
Nodejs 6.X - npm is not available standalone

jvonau:
Think would be better to install /etc/apt/sources.list.d/nodesource.list and the needed key in 2-common or move the curl script to 2 common so any other NEW nodejs apps will used a common 6.X starting point.

### DIFF
--- a/roles/sugarizer/tasks/main.yml
+++ b/roles/sugarizer/tasks/main.yml
@@ -12,20 +12,24 @@
         src={{ sugarizer_location }}/{{ sugarizer_version }}
         state=link
 
+- name: Set up apt sources on is_debuntu
+  shell: curl -sL https://deb.nodesource.com/setup_6.x | bash -
+  when: internet_available and is_debuntu
+
 - name: Install sugarizer required packages - is_debuntu
   package: name={{ item }}
            state=present
-  with_items:
-    - node-gyp
   when: internet_available and is_debuntu
+  with_items:
+    - nodejs
 
 - name: Install npm non is_debuntu
   package: name={{ item }}
            state=present
+  when: internet_available and not is_debuntu
   with_items:
     - nodejs
     - npm
-  when: internet_available and not is_debuntu
 
 # attempting to reinstall npn is broken on raspbian 9
 - name: check for sugarizer already installed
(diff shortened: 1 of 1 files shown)


holta:
@georgejhunt: please read diagnosis @ #419 then review @jvonau's proposed solution here (that he worked very hard on with me today) towards putting an end to Sugarizer's many, many installation issues (#193), Thanks!

holta:
Smoke-tested & functionality-tested on an existing system, after removing these 2 &mdash; this solves #419:
```
/library/www/html/sugarizer
/library/www/html/sugarizer-0.9
```
For good measure, am also now running a from-scratch smoke test using http://download.iiab.io/6.5/rpi/load-sugarizer-vpn.txt

holta:
From scratch smoke-test and functionality test worked on Raspbian Lite with http://download.iiab.io/6.5/rpi/load-sugarizer-vpn.txt

(Even after Admin Console was installed, "./runtags sugarizer" continues to work...)

If there are no additions/objections to #420 here, I'd like to merge in coming hours ideally?  So master (./runansible) works far more reliably for all.

_Also let's further refine during Thursday's all (http://minutes.iiab.io) where further future-proofing is possible...to put Sugarizer's never-ending installation issues to rest :-)_

holta:
Obviously given the nature of this bug (conflicting Node.js packages enmeshed in your Sugarizer installation) there are no absolute guarantees for legacy installs among folks "riding master" on all different OS's.

You really do need to delete both Sugarizer directories in /library/www/html (as spelled out above) and/or better yet strongly consider a completely clean install on a clean OS, to make sure you end up with Node.js 6.x.x and a more future-proof Sugarizer installation.

See also @jvonau's comment about Debian from Item 3.a. in Monday's call http://minutes.iiab.io (Oct 16, 2017) below: 

> broke debian-9 with npm not found, might not be installed on the path. Side note ssh iiab-admin broken permission denied (publickey) in VM while accessing via openvpn

holta:
(Re)confirmation: when running Sugarizer's new playbook on a clean install on RPi, I observed the output of `apt -a list nodejs` change as these steps were scrolling by:
```
TASK [sugarizer : Set up apt sources on is_debuntu] ****************************
TASK [sugarizer : Install sugarizer required packages - is_debuntu] ************
```
The output of `apt -a list nodejs` originally was:
```
nodejs/stable 4.8.2~dfsg-1 armhf
```
...then became:
```
nodejs/unknown 6.11.4-1nodesource1 armhf
nodejs/stable 4.8.2~dfsg-1 armhf
```
...finally stabilizing as:
```
nodejs/unknown,now 6.11.4-1nodesource1 armhf [installed]
nodejs/stable 4.8.2~dfsg-1 armhf
```
_Most Important: running IIAB Admin Console's ./install no longer affects the state of Node.js's installed version thankfully &mdash; allowing subsequent runs of "./runtags sugarizer" to complete properly._

PR #378, opened 2017-10-05
Home page setting: http://box to WordPress etc?

jvonau:

PR #424, opened 2017-10-18
RFC on moving network last.

jvonau:
Not smoked tested as is and is for illustrative purposes and comments. Contains #378 #381, work towards removing /etc/sysconfig/iiab* files moving them into /etc/iiab/iiab.env, hostname handling, and moving network to be last.

### DIFF
--- a/roles/1-prep/tasks/computed_vars.yml
+++ b/roles/1-prep/tasks/computed_vars.yml
@@ -46,6 +46,15 @@
   file: path=/tmp/heart-beat.txt
         state=absent
 
+- name: no prior domain name
+  set_fact:
+    iiab_domain: "{{ iiab_domain }}"
+    iiab_hostname: "{{ iiab_hostname }}"
+
+# Now check FQDN
+- include: roles/2-common/tasks/hostname.yml
+  when: not first_run and "iiab_hostname.iiab_domain" != ansible_fqdn
+
 - name: add version section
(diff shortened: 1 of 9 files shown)


roles/httpd/tasks/main.yml
```diff
@@ -146,17 +146,11 @@
...
     - base
 
-- include: home-page.yml
-
 - name: place the script to generate home pages
   template: src=refresh-wiki-docs.sh
             dest=/usr/bin/iiab-refresh-wiki-docs
             mode=0755
 
-- name: generate the offline documents
```
holta: Minor quibble but is separating both [refresh-wiki-docs](https://github.com/iiab/iiab/blob/master/roles/httpd/templates/refresh-wiki-docs.sh) actions into 2 diff playbooks really worth it?

Makes it quite a bit harder for new contributors to understand, spread across these 2 playbooks:
* roles/httpd/tasks/main.yml
* roles/4-server-options/tasks/main.yml

Perhaps to pick 1 playbook or the other?

holta:
@georgejhunt @tim-moody @arky & All,

@jvonau is looking for feedback on the usefulness of moving /etc/sysconfig/iiab* files into /etc/iiab/iiab.env if you can review this prior to tomorrow's Thursday call (http://minutes.iiab.io Item 3.c.)

Thanks!

Issue #197, opened 2017-09-02
Permit installation of IIAB over Wi-Fi instead of Ethernet


holta:
Ref #194 (Pi wifi gw) outlines a future possibility that might perhaps allow people to build an IIAB system even when they lack access to an Ethernet cable providing Internet.

Instead of using Ethernet, they would download and install IIAB software over Wi-Fi (even though Wi-Fi constraints like connections/authentication/time-limits/bandwidth-limits mean this will never be quite as reliable as Ethernet, for so many reasons!)

jvonau:
addressed with #178 except for configuration of the actual connection and this a regression not an enhancement.

jvonau:
[headless wifi boot](https://www.raspberrypi.org/forums/viewtopic.php?t=191252)
Re: Raspbian Stretch Headless Setup Procedure
Thu Sep 07, 2017 9:06 pm

holta:
@tim-moody suggested "hotspot" functionality is possible at very the same time as you're downloading IIAB+ZIMs/etc to create an RPi3 server, so that non-guru humanitarians (e.g. Peace Corps Volunteers) can DIY their own Internet-in-a-Box &mdash; while beginning to "field test" (of sorts) immediately &mdash; without the need to toggle from "online provisioning/wifi mode" to "offline hotspot mode".

@jvonau suggested during our call we first need to find out if RPi's WiFi firmware supports AP mode while it's also a Wi-Fi client ?

While this would be amazing: I'm also nervous this could become a year-long project that might distract from the install-over-WiFi MVP (minimum viable product) we're seeking to prove out as part of [IIAB 6.5](https://github.com/iiab/iiab/milestone/2).  FWIW trying to enable simple AP mode on XO-1.5 became an almost year-long project earlier, that largely failed, when we valiantly tried that in 2015-2016 (also for very good reasons...and still today as then our R&D budget remains quite limited ;-)

Either way - schools/clinics in the developing world essentially all tell us they need a toggle to turn OFF upstream Internet - so we'll need something like this at the command-line (or Admin Console later) that rapidly (?) toggles back+forth between online-mode and offline-hotspot-mode.  Some of the Reasons Being...that developing world schools/clinics:
1. cannot possibly afford the costs of sharing their upstream connectivity to their IIAB users
[...]

Issue #430, opened 2017-10-19
Enable simultaneous Hotspot/AP & client-WiFi on RPi's internal WiFi?


holta:
If achievable, this would be particularly useful during IIAB installation/curation/testing.

And potentially also during field updates of Content Packs &mdash; though Wi-Fi access to the Internet can be quite rare within many rural/developing countries, where pricey 3G mobile data is usually far more common (though efforts like Google Station are seeking to change this...)

Simultaneous Hotspot/AP and WiFi-as-client could build upon #197 "Permit installation of IIAB over Wi-Fi instead of Ethernet".

tim-moody:
http://imti.co/post/145442415333/raspberry-pi-3-wifi-station-ap

holta:
> http://imti.co/post/145442415333/raspberry-pi-3-wifi-station-ap

@jvonau adds this interesting read from Oct 27 & onwards:

> How-to: AP+wifi client on Zero W with Stretch Lite
https://www.raspberrypi.org/forums/viewtopic.php?t=196263

jvonau:
iw list for .26 
> $ iw list
Wiphy phy0
	max # scan SSIDs: 10
	max scan IEs length: 2048 bytes
	max # sched scan SSIDs: 16
	max # match sets: 16
	max # scan plans: 1
	max scan plan interval: -1
	max scan plan iterations: 0
	Retry short limit: 7
	Retry long limit: 4
	Coverage class: 0 (up to 0m)
	Device supports T-DLS.
	Supported Ciphers:
		* WEP40 (00-0f-ac:1)
		* WEP104 (00-0f-ac:5)
		* TKIP (00-0f-ac:2)
		* CCMP-128 (00-0f-ac:4)
	Available Antennas: TX 0 RX 0
	Supported interface modes:
		 * IBSS
		 * managed
		 * AP
		 * P2P-client
		 * P2P-GO
		 * P2P-device
	Band 1:
		Capabilities: 0x1020
			HT20
			Static SM Power Save
			RX HT20 SGI
			No RX STBC
			Max AMSDU length: 3839 bytes
			DSSS/CCK HT40
		Maximum RX AMPDU length 65535 bytes (exponent: 0x003)
		Minimum RX AMPDU time spacing: 16 usec (0x07)
		HT TX/RX MCS rate indexes supported: 0-7
		Bitrates (non-HT):
			* 1.0 Mbps
			* 2.0 Mbps (short preamble supported)
[...]

PR #431, opened 2017-10-19
wifi as both upstream and downstream

georgejhunt:
I read #197 and realized that some of my earlier work had not been shared. As I remember I did see gateway function with wifi acting both as client and server. 
But this experimentation was using dnsmasq, and I used a script to blast config files via "here documents".
I'm not taking on ownership of this objective, but I'm pretty sure it is possible to integrate within our current codebase
I know rc.local is not what we want long term -- so there is still quite a bit of work
Not smoke tested on recent master (though rebased on it), and not really a candidate for inclusion into master.


holta:
Implements #430 experimentally, so this can be refined for [IIAB 6.6](https://github.com/iiab/iiab/milestone/3) &mdash; Thanks @georgejhunt

roles/8-mgmt-tools/meta/main.yml
```diff
@@ -8,3 +8,4 @@ dependencies:
    - { role: phpmyadmin, tags: ['services','phpmyadmin','tools'], when: phpmyadmin_install }
    - { role: awstats, tags: ['services','awstats','tools'], when: awstats_install }
    - { role: teamviewer, tags: ['services','teamviewer','tools'], when: teamviewer_install }
+    - { role: network, tags: ['services','base','network'], when: gui_desired_network_role is defined and network_late }
```
thotmx, line 11: Extra empty space breaks Travis CI

holta: @georgejhunt please advise, Thanks!

holta:
@georgejhunt should this be closed alongside #697 ?

(Or keep it open if there's good reason!)

Issue #433, opened 2017-10-19
Squid install fails @ Temporary failure resolving 'mirrordirector.raspbian.org'


holta:
This failure is new and extremely common over the past week or so.  Many different playbooks fail to download essential bits.  A long failure message is pasted in below, that happens to have occurred during [network : Install squid packages] but this download failure occurs in many other Ansible playbooks as well.

When this happens, "ping 8.8.8.8" works but "ping mit.edu" does not, confirming:
```
Temporary failure in name resolution
```
Confusingly /etc/resolv.conf appears largely empty (as follows) even on some RPi3's not experiencing this DNS failure:
```
# Generated by dhcpcd
# /etc/resolv.conf.head can replace this line
# /etc/resolv.conf.tail can replace this line
```
Though I also see the following on working IIAB-on-RPi3 machines:
```
# Generated by dhcpcd from eth0.dhcp
# /etc/resolv.conf.head can replace this line
nameserver 192.168.0.1
# /etc/resolv.conf.tail can replace this line
```
Possibly this flood of new download errors over the past few days is exacerbated by the "testing" version of dhcpcd5 that Raspbian has induced us to install in recent days, as a result of "apt upgrade" etc:
```
root@box:/opt/iiab/iiab# apt -a list dhcpcd5
Listing... Done
dhcpcd5/testing,now 1:6.11.5-1+rpt2 armhf [installed]
dhcpcd5/stable 6.10.1-1 armhf
```
```
root@box:/opt/iiab/iiab# apt -a list dhcpcd
Listing... Done
dhcpcd/stable 1:3.2.3-11 armhf
```
_Confusingly, installing IIAB over WiFi (experimentally, per #197) often yields the same error message despite different circumstances, i.e. when WiFi fails completely.  So when diagnosing, it's very important to test BOTH:_
```
ping 8.8.8.8
ping mit.edu
```
Finally, a typical error appears as follows:

> TASK [network : Install squid packages] ****************************************
[...]

PR #442, opened 2017-10-21
named / DNS fix for #433 Squid install fail

jvonau:
cherry-picked from an earlier branch and is part of #267 named - wrong place


holta:
Repeatedly tested as a fix for #433 over the past 4+ hours, using BIG-sized local_vars.yml

PR #467, opened 2017-10-27
cleaner Ansible output & README's

holta:
Editorial improvements to 25 files for more readable Ansible output.

A couple stale README's were also refreshed therein.

Smoke-testing just now beginning.


holta:
Smoke-tested on several Raspbian Lite and Desktop.

Builds on #466.

Issue #368, opened 2017-10-03
avoid initializing MySQL so many times during IIAB's Ansible install?


holta:
This takes gratuitously take 10 minutes during 5-xo-services &mdash; and possibly other places too?

(Certainly the vast majority of implementers don't use XO services in 2017/2018!)

As @jvonau [suggested](https://github.com/iiab/iiab/issues/330#issuecomment-333253611) earlier, this might be a relatively quick win for IIAB 6.5 or 6.6?

holta:
Is this also an issue with PostgreSQL, which appears to be faster to install, no matter how many times its own installation checks / initializations happen?

Is Jenkins-assisted Profiling possible?!

(MySQL smells like the real slowpoke during ./runansible but I could be wrong!)

holta:
Tangentially related: (not all having to do with commenting speedup by commenting out 5-xo-services in /opt/iiab/iiab.yml)

PR #117 lockout xo service on debian family
PR #367 6.4 needs the most-tested local_vars.yml (medium-sized) + restore 5-xo-services to iiab.yml
#370 comment out & document 5-xo-services as it's slow as a dog
PR #371 comment out 5-xo-services (too damn slow)

jvonau:
Turn on timestamps in Jenkins that should give the time required to repeat the execution of the duplicated plays. Now add the time up and compare that to the total time for the run should give a percentage spent doing nothing but repeating the role as percentage.

holta:
Added to Monday's call agenda @ http://minutes.iiab.io (Sept 9, 10AM NYC Time) even if MySQL is not the worst offender!

Great topic to deepen our understanding of Jenkins/CI/CD in general~

tim-moody:
or look at the install log. my impression is that the time for tasks other than the first install are in the milliseconds.

jvonau:
jenkin_ip:8080/job/IIAB-debian-jvonau/35/consoleFull using stock med_vars.

end runasible     12:55:46 ==> debian: 127.0.0.1   : ok=392  changed=292  unreachable=0    failed=0   
start runansible  12:34:11 ==> debian: + ./runansible
Total Time               21:35  = 1295 seconds

first install, is needed:
12:42:05 ==> debian: TASK [mysql : Install MySQL Debian]
12:43:39 ==> debian: changed: [127.0.0.1] => (item={u'option': u'enabled', u'value': True})
total 1min 34secs

now 'reinstall' for elgg that is enabled in the vars file
[...]

PR #472, opened 2017-10-27
Db - avoids MySQL & PostgreSQL Install-A-Thon

jvonau:
### Fixes Bug
#368 
jenkins/job/Mem-change-test/72/console

@holta

### DIFF
--- a/roles/3-base-server/tasks/main.yml
+++ b/roles/3-base-server/tasks/main.yml
@@ -18,6 +18,7 @@
 - name: MYSQL
   include_role:
     name: mysql
+  # has no "when: XXXXX_install" flag
   tags: base, mysql
 
 - name: Make sure there is a content directory
--- a/roles/3-base-server/tasks/main.yml
+++ b/roles/3-base-server/tasks/main.yml
@@ -15,12 +15,14 @@
   # has no "when: XXXXX_install" flag
   tags: base, iiab-admin
 
+- name: MYSQL
+  include_role:
+    name: mysql
+  tags: base, mysql
+
 - name: Make sure there is a content directory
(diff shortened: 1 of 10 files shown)


holta:
I'll merge experimentally in coming moments if there are no objections.

We can revert if there are problems.

holta:
QUESTION for @tim-moody & All: should both MySQL and PostgreSQL be installed in the same stage/section in future, to make our install process's modularity understandable?  If so (no obligation!) would that be Stage 3 or Stage 4?

BACKGROUND: I merged this as master was being very inefficient reinstalling MySQL so many times (and PostgreSQL a couple times) across single runs of ./runansible with [BIG-sized local_vars.yml](http://wiki.laptop.org/go/IIAB/local_vars_big.yml) especially, making Ansible's output impenetrable and slow.

PS this PR #472 completed smoke tests on existing Ubuntu and Raspbian systems as well as cleanly on Jenkins, and we needed a decision quickly, nye Analysis Paralysis.

holta:
Just a clarif that @jvonau's #472 here deleted these 7 files/dependencies, accomplishing such DB dependencies in another way:
```
roles/elgg/meta/main.yml       contained... { role: mysql }
roles/moodle/meta/main.yml     contained... { role: postgresql }
roles/nextcloud/meta/main.yml  contained... { role: mysql }
roles/owncloud/meta/main.yml   contained... { role: mysql }
roles/pathagar/meta/main.yml   contained... { role: postgresql }
roles/rachel/meta/main.yml     contained... { role: mysql, mysql_enabled: True }
roles/wordpress/meta/main.yml  contained... { role: mysql }
```

PR #474, opened 2017-10-28
Steps2 - avoid repetition of Stages 1-9 when ./runansible -> ./iiab-install is rerun

jvonau:
### Fixes Bug
interrupted installs starting at square one 
new iiab-steps.yml file for runansible
jenkins job/IIAB-debian-jvonau/62/console
@holta

### DIFF
--- a/iiab-network
+++ b/iiab-network
@@ -1,15 +1,8 @@
 #!/bin/bash
-
-if [ -f /etc/iiab/iiab.env ]
-then
-  . /etc/iiab/iiab.env
-  cd $IIAB_DIR
-else
-  echo /etc/iiab/iiab.env is missing. Playbook has not been run.
-  exit 1
-fi
-
+# running from a git repo
+# ansible files exist
 CWD=`pwd`
+export ANSIBLE_LOG_PATH="$CWD/iiab-network.log"
 
 if [ ! -f iiab-network.yml ]
 then
@@ -24,7 +17,6 @@ then
    touch /etc/iiab/config_vars.yml
 fi
 
-export ANSIBLE_LOG_PATH="$CWD/iiab-network.log"
 ansible-playbook -i ansible_hosts iiab-network.yml --connection=local
 echo "" >> iiab-network.log
 ls -la /etc/sys*/net*/ifcfg* >> iiab-network.log
--- a/iiab-steps.yml
+++ b/iiab-steps.yml
@@ -1,0 +1,29 @@
+---
+- hosts: all
+  become: yes
+
+  vars_files:
+  - vars/default_vars.yml
+  - vars/{{ ansible_local.local_facts.os_ver }}.yml
+  - vars/local_vars.yml
+
+  roles:
+      - { role: 0-init, tags: ['0-init'] }
+      - { role: 1-prep, tags: ['1-prep','platform','base'],
+          when: ansible_local.local_facts.stage|int < 1 }
+      - { role: 2-common, tags: ['2-common','base'],
+          when: ansible_local.local_facts.stage|int < 2 }
+      - { role: 3-base-server, tags: ['3-base-server','base'],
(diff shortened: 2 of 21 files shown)


jvonau:
> 15:25:20 ==> debian: TASK [3-base-server : Record STAGE] ********************************************
15:25:20 ==> debian: changed: [127.0.0.1]
15:25:20 ==> debian: 
15:25:20 ==> debian: TASK [3-base-server : ...HAS COMPLETED ====================================] ***
15:25:21 ==> debian: changed: [127.0.0.1]
15:25:21 ==> debian: 
15:25:21 ==> debian: TASK [4-server-options : ...IS BEGINNING ==================================] ***
15:25:21 ==> debian: changed: [127.0.0.1]
15:25:21 ==> debian: TASK [4-server-options : SSHD] *************************************************

Can be condensed to remove the extra echos, ie: 
TASK [3-base-server : Record STAGE 3 HAS COMPLETED ] 
and omit : ...IS BEGINNING

holta:
Naive Questions!
* Is Step 0 not part of this incremental recordkeeping, only Stages 1-9?
* Why does Line 93 of https://github.com/jvonau/iiab/blob/steps2/roles/1-prep/tasks/main.yml#L93 appear to refer to a non-existent file roles/1-prep/templates/iiab.env.j2 ?  How does this work?
* What would happen of STAGE 8 was recorded as successful in lines 50-54 of [roles/8-mgmt-tools/tasks/main.yml](https://github.com/jvonau/iiab/blob/steps2/roles/8-mgmt-tools/tasks/main.yml#L50-L54) but earlier stages had been forgotten?  e.g. if "./runtags 8-mgmt-tools" was run or some such, causing earlier steps to be accidentally missed?!
* Why is "IS BEGINNING" removed from Stages 4-9 and what is providing that functionality instead?  And/or why is it still part of Stages 0-3?

jvonau:
- Step 0 is the starting point and is more for detection of install state
- look again the file has always been in roles/1-prep/templates 
- e.g. if "./runtags 8-mgmt-tools, Think the 1-9* tags could go away as runtags is more specificity tied to individual or related roles than an arbitrary grouping.
- 4-9 are faster running and more clear visually than 0-3 where things are longer running.

holta:
> if "./runtags 8-mgmt-tools", Think the 1-9* tags could go away

Are you saying "./runtags" always clears the "what stages have been completed" [*] progress counter back to STAGE=1 ?

Forcing the next "./runansible" to run all Stages 0-9?  Am just trying to clarify the expected (or most desired) behavior of runtags...

[*] i.e. the [STAGE variable at Line 8](https://github.com/jvonau/iiab/blob/steps2/roles/1-prep/templates/iiab.env.j2#L8) of /etc/iiab/iiab.env's template.

> 4-9 are faster running and more clear visually than 0-3 where things are longer running.

Personally I'd much prefer all stages have consistent messaging, so the pattern is very clear to new implementers, as things like Stages 6 and 7 can take forever too!

jvonau:
On Sun, Oct 29, 2017 at 2:49 PM, A Holt <notifications@github.com> wrote:

> if "./runtags 8-mgmt-tools", Think the 1-9* tags could go away
>
> Are you saying "./runtags" always clears the "what stages have been
> completed" [*] progress counter back to STAGE=1 ?
>
 No, depending on ​which tags are passed ​would dictate what STAGE to be recorded or changed if at all​

> Forcing the next "./runansible" to run all Stages 0-9? Am just trying to
> clarify the expected (or most desired) behavior of runtags...
>
The alpha only tags would not alter the STAGE as their execution is not as
broad of a scope, where the numbered tags would record the STAGE of the
highest number tag when completing that stage.​

jvonau:
Ran into an issue with vnstats and an appliance mode network layout while testing this branch, not addressed in #477

PR #475, opened 2017-10-30
use ansible-2.4.1 on CentOS

jvonau:


holta:
If there are no objections / counter-proposals, I'll merge in coming moments.

PR #477, opened 2017-10-30
iiab-install: ./runansible -> ./iiab-install avoids repetition of Stages 1-9 when rerun

jvonau:
revised #474 as iiab-install


holta:
https://github.com/jvonau/iiab/blob/iiab-install/scripts/local_facts.fact#L44-L53 and https://github.com/jvonau/iiab/blob/merge-order4/scripts/local_facts.fact#L44-L53 currently appears as:
```
# upgrades return found, clean installs return 0
# interruptions return last stage number recorded (1-9)
if [ -f /etc/iiab/iiab.env ]
then
  STAGE=1
  source /etc/iiab/iiab.env
  STAGE=$STAGE
else
  STAGE=0
fi
```
No Rush To Decision (let's get a MVP/Minimum Viable Product first!)  But here's another option with stronger guarantees we might consider in a subsequent PR or whatever:
```
# Successful installs RETURN 9 (last non-zero stage completed/recorded)
# Interrupted installs RETURN 1-8 (last non-zero stage completed/recorded)
# RETURN 0 (STAGE=0) will never happen, on purpose, to avoid confusion (given Stage 0 is
#     ALWAYS run at the outset of every call to iiab-install...avoids misunderstandings from
#     new folk who fall in the trap believing that "STAGE=0 means that Stage 0 completed")
# Failed-To-Start installs RETURN -1 (universal failsafe/precondition)
STAGE=-1
if [ -f /etc/iiab/iiab.env ]; then
  source /etc/iiab/iiab.env
fi
```

holta:
Regarding jvonau/iiab's current testing branch [merge-order4](https://github.com/jvonau/iiab/tree/merge-order4):

1) Can Squid & Wondershaper's tags be fixed by moving "tags:" AFTER "when:" in both cases within [Lines 35-45 of roles/3-base-server/tasks/main.yml](https://github.com/jvonau/iiab/blob/merge-order4/roles/3-base-server/tasks/main.yml#L35-L45) as below?

2) Can each of the 4 "tags:" clauses within [Lines 23-45 of roles/3-base-server/tasks/main.yml](https://github.com/jvonau/iiab/blob/merge-order4/roles/3-base-server/tasks/main.yml#L23-L45) be tightened up onto 1 single line as below?
```
- include_tasks: roles/network/tasks/named.yml
  tags: named, network, domain

- include_tasks: roles/network/tasks/dhcpd.yml
  tags: dhcpd, network, domain

- include_tasks: roles/network/tasks/squid.yml
  when: squid_install
  tags: squid, network

- include_tasks: roles/network/tasks/wondershaper.yml
  when: wondershaper_install
  tags: wondershaper, network
```

holta:
Small suggestion @ https://github.com/jvonau/iiab/issues/16 "can we keep IS BEGINNING and HAS COMPLETED in same code block?" for code understandability~

Issue #484, opened 2017-11-04
full-text searching across all ZIMs breaks (& cached results are wiped) "15min into every lesson"


holta:
Is there any reason not to experiment with commenting out the cron job @ [roles/kiwix/tasks/kiwix_install.yml#L77-L88](https://github.com/iiab/iiab/blob/master/roles/kiwix/tasks/kiwix_install.yml#L77-L88) as a result of the much more reliable Kiwix 0.3.0 from 2017-10-25?  i.e. is commenting this out dangerous / premature for other known reasons?

Context: Kiwix's full-text searches work beautifully [across all ZIM files] @ http:// iiab . me : 3000 if you're willing to wait 2-to-7 minutes typically, which is fine at the very beginning of a school's lesson (if it works especially ;)  And thanks to [Kiwix 0.3.0](https://github.com/kiwix/kiwix-tools/commits/master) from [2017-10-25](http://download.kiwix.org/nightly/2017-10-25/) (e.g. https://github.com/iiab/iiab/pull/463) followup search results take about 30 seconds each typically (not fun, but tolerable!) when clicking on these buttons at the bottom of the page, e.g. for the 1st 250 search slow results: (i.e. 25 search results per page)

`<- 1 2 3 4 5 6 7 8 9 10 ->`

HOWEVER 15 minutes after every hour, you can no longer click on these 12 buttons, at the bottom of every search page &mdash; without triggering another 2-to-7 minutes search &mdash; as a result of the cron job installed in the above kiwix_install.yml :/

FYI these times were measured on a reasonably modern Debian 9.1 PC, running IIAB ~6.4 augmented with Kiwix 0.3.0

holta:
Somewhat Related:

https://github.com/kiwix/kiwix-tools/issues/92 "Why are Kiwix searches blocked when asking for more than 67 results?"

mgautierfr:
It would be better to not especially kill and restart the kiwix-serve processus.
Systemd can be configured to automatically restart a service if the process crash. And (it seems) it is what you want.

See https://jonarcher.info/2015/08/ensure-systemd-services-restart-on-failure/ for an example.

tim-moody:
@mgautierfr the problem was in the past that kiwix-serve went to 100% cpu and became unresponsive so people decided to restart it every 15 minutes. I think we should stop restarting it at all and continue to monitor it.

mgautierfr:
[...]

Issue #324, opened 2017-09-27
debian: KIWIX [WARNING]: Consider using file module with owner rather than running chown


arky:
Harmless warning in to file module in Kiwix task. Report from nightly Jenkins CI run. Please kindly close the bug if it is low priority.  

==> debian: TASK [kiwix : Set kiwix ownership] *********************************************
==> debian: changed: [127.0.0.1]
==> debian:  [WARNING]: Consider using file module with owner rather than running chown
==> debian: 
==> debian: TASK [kiwix : Make an entry in crontab to restart every hour] ******************
==> debian: changed: [127.0.0.1]
==> debian: 
==> debian: TASK [kiwix : Make an entry in crontab to restart every hour] ******************
==> debian: skipping: [127.0.0.1]
==> debian: 
==> debian: TASK [kiwix : Create kiwix-serve service] **************************************
==> debian: changed: [127.0.0.1] => (item={u'dest': u'/etc/systemd/system/kiwix-serve.service', u'src': u'kiwix-serve.service.j2', u'mode': u'0644'})
==> debian: changed: [127.0.0.1] => (item={u'dest': u'/usr/bin/iiab-make-kiwix-lib', u'src': u'iiab-make-kiwix-lib', u'mode': u'0755'})
==> debian: changed: [127.0.0.1] => (item={u'dest': u'/usr/bin/iiab-make-kiwix-lib.py', u'src': u'iiab-make-kiwix-lib.py', u'mode': u'0755'})
==> debian: changed: [127.0.0.1] => (item={u'dest': u'/usr/bin/iiab-make-apache-config.py', u'src': u'iiab-make-apache-config.py', u'mode': u'0755'})

holta:
@tim-moody can you explain and/or resolve?

jvonau:
code merged

PR #485, opened 2017-11-04
restart kiwix-serve 4AM nightly (not 15min after every hour) & restrict chown to /opt/iiab/kiwix

holta:
This is a proposed solution to #484 — full-text searching across all ZIMs breaks (& cached results are wiped) "15min into every lesson".

Not yet smoke-tested.


holta:
Ref #324 "debian: KIWIX [WARNING]: Consider using file module with owner rather than running chown"

holta:
Surfacing commit message from 20 min ago:

"Smoke-tested.  Please continue to test locally, so we can refine (or eliminate if nec?) this cron job's frequency as nec!

Closes #484"

PR #486, opened 2017-11-05
merge-order4: ./runansible -> ./iiab-install avoids repetition of Stages 1-9 when rerun

holta:
This is jvonau/iiab branch merge-order4 that builds off of #477's jvonau/iiab branch iiab-install.

Heavily tested over the past 5 days, on Raspbian Lite 2017-09-07 with and without updates &mdash; note that a reboot is REQUIRED prior to running ./runansible -> ./iiab-install if you want OpenVPN to work &mdash; as a result of #482 (or do not apply updates at all to 2017-09-07, but this is not a long-term solution!)

PR #487, opened 2017-11-05
iiab-order: ./runansible -> ./iiab-install avoids repetition of Stages 1-9 when rerun

jvonau:
replaces merge-order4 
tested on '.132'


holta:
Builds off of #486 and #477.

Issue #392, opened 2017-10-11
kalite blocks IIAB installs on RPi (python-pip's new /etc/pip.conf)


holta:
Certainly this blocked 1 install last night and 2 this morning, so this is not an isolated/intermittent download failure as I initially assumed?  (it's probably blocking all [IIAB 6.4](https://github.com/iiab/iiab/wiki/IIAB-6.4-Release-Notes) installs as well!)

Is this really as simple as a failure to download non-existent https://www.piwheels.hostedpi.com/simple/importlib/ etc?  (last line in error msg below)

(Or possibly a python issue?  Hypothetically relating to "apt dist-upgrade" which (as of a few hours ago) applies 29 packages above & beyond Raspbian Lite from 2017-09-07)

TASK [kalite : Install ka-lite dependencies with pip] **************************
[...]

PR #495, opened 2017-11-08
kalite static only

jvonau:
### Fixes Bug
Needing removal of /etc/pip.conf, needs further confirmation on clean test with /etc/pip.conf intact.
just run static libs
jenkins /job/IIAB-debian-jvonau/126/console
jenkins-kalite-static-test

jenkins job/iiab-CentOS-jv/75/console installed clean but install exited for other reasons

### DIFF
--- a/roles/kalite/tasks/install.yml
+++ b/roles/kalite/tasks/install.yml
@@ -10,11 +10,10 @@
   get_url: url={{ kalite_requirements }} dest={{ pip_packages_dir }}/kalite.txt
   when: internet_available
 
-#- name: Install KA Lite devel with pip - (debuntu)
+#- name: Install KA Lite non-static + reqs file with pip - (debuntu)
 #  pip: requirements={{ pip_packages_dir }}/kalite.txt
 #       virtualenv={{ kalite_venv }}
 #       virtualenv_site_packages=no
-#       extra_args="--no-cache-dir"
 #       extra_args="--no-cache-dir"
 #       extra_args="--disable-pip-version-check"
 #  when: internet_available and is_debuntu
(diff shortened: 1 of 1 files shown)


holta:
Per @benjaoming's observation (https://github.com/learningequality/ka-lite/issues/5533#issuecomment-342619392) that we should not be installing both static and non-static versions of KA Lite.

That emerged during the solving of #392 ("kalite blocks IIAB installs on RPi (python-pip's new /etc/pip.conf)") over the past week.

Hopefully this will speed up installation and reduce disk footprint too!

holta:
Should we comment out the reqs file entirely @ [Lines 9-11 of roles/kalite/tasks/install.yml](https://github.com/iiab/iiab/blob/master/roles/kalite/tasks/install.yml#L9-L11) &mdash; or is this still needed?
```
- name: Grab the requirements file
  get_url: url={{ kalite_requirements }} dest={{ pip_packages_dir }}/kalite.txt
  when: internet_available
```

holta:
> Should we comment out the reqs file entirely @ Lines 9-11 of roles/kalite/tasks/install.yml — or is this still needed?

Conclusion from @jvonau: lines 9-11 remain tiny/harmless/useful documentation so let's leave KA Lite's reqs file download in place, as is.

FYI /opt/iiab/pip-packages/kalite.txt (the current requirements file) as downloaded from https://raw.githubusercontent.com/learningequality/ka-lite/master/requirements.txt is only 31 lines and appears as follows:
```
# Django is NOT used but is actually bundled, we just put this requirement
# because later versions of django include a utf-8 named file that breaks
# on certain systems when packaging.
# This file:
# https://github.com/django/django/blob/master/tests/staticfiles_tests/apps/test
/static/test/%E2%8A%97.txt
# ...to explain: Some of the below packages depend on django, if we don't
# specify a version here, we'll get the latest.
# See also: https://github.com/learningequality/ka-lite/issues/5419
django==1.5.12

docopt>=0.6,<0.7
South==1.0.2
django-js-reverse==0.5.0
# used by django-js-reverse
ply==3.4
# used by django-js-reverse
slimit==0.8.1
six==1.10.0
django-appconf==1.0.1
peewee==2.6.4
python-dateutil==2.4.2
django-annoying==0.8.4
cherrypy==3.3.0
ifcfg==0.10.1
importlib==1.0.3
pbkdf2==1.3
rsa==3.4.2
youtube-dl==2015.11.24

django-tastypie-legacy
requests==2.11.1
```

benjaoming:
These are the requirements that `ka-lite-static` fetches and bundles (with a small gotcha, `django==1.5.12` is actually deleted again because there's a patched up version maintained alongside KA Lite in `kalite/packages/bundled`).

Issue #482, opened 2017-11-03
Raspbian kernel update breaks bridging


jvonau:
### Steps to reproduce the problem.

BROKEN
root@box:/opt/iiab/iiab# apt list -a raspberrypi-kernel
Listing... Done
raspberrypi-kernel/stable,now 1.20171029-1 armhf [installed,automatic]

root@box:/opt/iiab/iiab# uname -a
Linux box.lan 4.9.41-v7+ #1023 SMP Tue Aug 8 16:00:15 BST 2017 armv7l GNU/Linux

root@box:/opt/iiab/iiab# brctl addbr br1
add bridge failed: Package not installed

WORKING
root@box:/home/iiab-admin# apt list -a raspberrypi-kernel
Listing... Done
raspberrypi-kernel/testing 1.20171029-1 armhf [upgradable from: 1.20170811-1]
raspberrypi-kernel/now 1.20170811-1 armhf [installed,upgradable to: 1.20171029-1]

root@box:/home/iiab-admin# uname -a
Linux box.lan 4.9.41-v7+ #1023 SMP Tue Aug 8 16:00:15 BST 2017 armv7l GNU/Linux

root@box:/home/iiab-admin# brctl addbr br1
root@box:/home/iiab-admin# ip a
snip
8: br1: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc noqueue state UNKNOWN group default qlen 1000

jvonau:
secondary issue of why is 'testing' repo enabled by default as noted in #392 also.

jvonau:
Upstream issue https://github.com/raspberrypi/linux/issues/2256

holta:
Thanks much @jvonau for escalating this critical problem as we've received no response from irc channel #raspberrypi-kernel on irc.freenode.net over the past 72 hours...

Nor on Raspberry Pi's Forums just yet: https://www.raspberrypi.org/forums/viewtopic.php?t=196623

holta:
Clarif on how incredibly serious this is &mdash; the result is:
* Internet-in-a-Box WiFi hotspots no longer work (Access Points inside RPi3 or RPI Zero W, for the most basic functioning of Knowledge Hotspots)
* OpenVPN no longer works (for remote support)

jvonau:
TEBKAC - needs a reboot to use the new kernel. Please rethink install instructions, curl-me, scripts/ansible's apt-get -y update to handle updating kernel and where the needed reboot should be.

[...]

Issue #457, opened 2017-10-24
scripts/ansible does not upgrade or issue any message


tim-moody:
deb 9 vm with ansible 2.2.0.0
latest master - scripts/ansible
terminates quickly with no message
is this expected?

holta:
Can you use Ansible 2.4.0 per Step 5 of our [Do Everything from Scratch](https://github.com/iiab/iiab/wiki/IIAB-Installation#do-everything-from-scratch) installation instructions?  (That Josh Dennis helped us refine for Debian 9 over the past month.)

tim-moody:
root@debian9base:/opt/iiab/iiab# ansible --version
ansible 2.2.0.0
  config file = /opt/iiab/iiab/ansible.cfg
  configured module search path = ['/usr/share/ansible']
root@debian9base:/opt/iiab/iiab# apt purge ansible
Reading package lists... Done
Building dependency tree
Reading state information... Done
**Package 'ansible' is not installed, so not removed**
0 upgraded, 0 newly installed, 0 to remove and 0 not upgraded.

holta:
Is it possible 2.2.0 got installed onto Debian 9.1 (or whatever you have) thru pip or some other means than the OS's own "apt install ansible" etc?

(Any idea how 2.2.0 got there?)

tim-moody:
I think it got there from the previous version of scripts/ansible

holta:
Makes sense.  We have not focused on the installed base, as this has not been a concern to date.

But I do agree with your suggestion that scripts/ansible should clarify matters for "Experts" who become attached to obscure/buggy/older versions of Ansible for their own reasons...and as such we need a PSA right there at the command line...to make clear to them how to extricate...as we all know "overly experienced" sysadmins are people too :-) who need to be able to recover from their own secret sauce per https://github.com/iiab/iiab/wiki/IIAB-Installation#expert-mode (!)

tim-moody:
it's not a matter of attachment. there is no way to upgrade

holta:
There must be a way to remove Ansible 2.2.0

Unfortunately I'm not the expert )-:

Maybe @georgejhunt & @jvonau & @arky know?

tim-moody:
and, btw, this was the official means of installing ansible. it's just that iiab changed afterwards and the changed script is unable to undo what its predecessor did.

tim-moody:
see above

holta:
[...]

PR #498, opened 2017-11-08
experimental ./iiab-install blocks older/stale kernels, using version_gt()

holta:
Tested on Raspbian with both new & old kernels.

### DIFF
--- a/iiab-install
+++ b/iiab-install
@@ -14,7 +14,7 @@ export ANSIBLE_LOG_PATH="$CWD/iiab-install.log"
 
 if [ ! -f $PLAYBOOK ]; then
     echo "IIAB Playbook not found."
-    echo "Please run this command from the top level of the git repo."
+    echo "Please run this command from /opt/iiab/iiab (top level of the git repo)."
     echo "Exiting."
     exit 1
 fi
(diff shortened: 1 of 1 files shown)


holta:
Uses this bash function to compare version numbers:
```
function version_gt() { [ "$(printf '%s\n' "$@" | sort -V | head -1)" != "$1" ]; }
```
Caution: this function gives a False Positive when comparing 2.4.1.0 to 2.4.1 so structure the conditional carefully if using this for Ansible version checks too!

holta:
To protect installers/implementers from showstopper bugs like #482 with a PSA (explanation) instead.

holta:
(1) I've published `./iiab-install` (instead of `./runansible`) in the official install instructions @ https://github.com/iiab/iiab/wiki/IIAB-Installation#do-everything-from-scratch so we get increasingly more testing coverage.

[FAQ.IIAB.IO](http://FAQ.IIAB.IO) will be revised prior to [IIAB 6.5](https://github.com/iiab/iiab/milestone/2) release, with whatever newbie-nurturing strategies make sense.

@jvonau the kernel warning recommendation could be narrowed to "apt install raspberrypi-kernel" instead of the "apt dist-upgrade" possibly?  Anyway both work!

(2) I'm in favor of setting 2.4.0.0 or 2.4.1.0 as a baseline minimum within scripts/ansible going forward. Perhaps using the same "version greater than" comparison script:
```
function version_gt() { [ "$(printf '%s\n' "$@" | sort -V | head -1)" != "$1" ]; }
```
(i.e. using the full w.x.y.z makes comparisons completely safe with the above script....instead of colloquialisms like 2.4.1 etc.)

holta:
Ref #457 "scripts/ansible does not upgrade or issue any message"

PR #507, opened 2017-11-09
br0 - single interface mis-detection

jvonau:


holta:
Smoke-testing by both @jvonau & @tim-moody even if we need to understand better why casting seems to now be nec unlike in the past !!

PR #513, opened 2017-11-11
runtags enabled & documentation improvements

holta:
`echo "{}" > /etc/iiab/config_vars.yml` when /etc/iiab created in ./iiab-install (for now, so runtags works).

A better solution may be possible after discussing with @jvonau, @tim-moody and @georgejhunt etc during our Monday community/team call (http://minutes.iiab.io).


holta:
`echo "{}" > /etc/iiab/config_vars.yml` when /etc/iiab created in ./iiab-install (for now, so runtags works).

A better solution may be possible after discussing with @jvonau, @tim-moody and @georgejhunt etc during our Monday community/team call (http://minutes.iiab.io).

PR #514, opened 2017-11-12
./iiab-install pre-positions /etc/iiab/config_vars.yml for now (logic reinforced) + exFAT on debuntu

holta:
We'll discuss more during Monday's community/team call (http://minutes.iiab.io) how to get this on track!

One possibility is to remove config_vars.yml from ./iiab-install entirely, and have ./runtags & Admin Console check+establish their own prereqs ?

i.e. enforced preconditions + clear messaging of failures = failsafe SW

(But that's just one option...however we do this is fine with me...Monday's call will greatly help clarify/articulate.)

### DIFF
--- a/iiab-install
+++ b/iiab-install
@@ -72,7 +72,12 @@ echo "Placed /etc/ansible/facts.d/local_facts.fact into position."
 STAGE=0
 if [ ! -f /etc/iiab/iiab.env ]; then
     mkdir -p /etc/iiab
-    echo "{}" > /etc/iiab/config_vars.yml
+    echo "Ran 'mkdir -p /etc/iiab'"
+    # ./runtags and Admin Console currently require /etc/iiab/config_vars.yml (in future they may be self-sufficient?)
+    if [ ! -f /etc/iiab/config_vars.yml ]; then
+        echo "{}" > /etc/iiab/config_vars.yml
+        echo "Created stub /etc/iiab/config_vars.yml"
+    fi
 else
     if [[ `grep STAGE= /etc/iiab/iiab.env` ]]; then
         source /etc/iiab/iiab.env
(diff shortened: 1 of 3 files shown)


jvonau:
I'd rather see c8ba425af576cd95d4d739e4734e8096dcb87c26 than this hack in iiab-install

holta:
Monday if possible (http://minutes.iiab.io) let's try to inventory most all config files in /etc/iiab so everyone has a shared understanding of which should be more (or less) permanent for what specific purposes &mdash; starting when during installation &mdash; and which can/should be more transient: (in different ways!)

1. Some are/should be established rather permanently early on in all installations {iiab_env.py?, iiab_env.pyc?}
2. Some are more transient, on a spectrum from completely-optional...to deleted-routinely...to desired-for-certain-situations to... {iiab.env, iiab.ini, config_vars.yml, uuid, handle, iiab.ini, kiwix_catalog.json}

Possibly there are many more that the 8 above files.  Just recapitulating our short/evolving community spec for each would be great.  Their uses have evolved, are evolving, and will inevitably evolve more in new directions &mdash; so understanding/jotting down the intentions of each file (theory!) is just as important as understanding evolving/practical engineering realities (practice!)

holta:
This PR includes exfat-fuse & exfat-utils for debuntu in 2-common/tasks/packages.yml so teachers + operators can quickly read files off of modern disks, large USB sticks and large SD cards.

Separately: config_vars.yml + runtags dynamics will evolve in whatever direction consensus & on-the-ground-community needs drive these (during today's call http://minutes.iiab.io etc).

holta:
"exfat-fuse" and "exfat-utils" can be moved to "Install common packages" within 2-common/tasks/packages.yml if that's a truly better place to do this (instead of just on debuntu) and is safe on all OS's ?

holta:
Tested on Ubuntu and quickly on Raspbian.

jvonau:
The reason  exfat-fuse & exfat-utils were installable on demand was on redhat the packages are not part of the distro needing the installation additional repos, that is why the exFAT_enabled flag exists
originally only for the XOs.

jvonau:
Now has caused conflicts with #508 which YOU are free to fix.

PR #515, opened 2017-11-13
@mgautierfr recommends kiwix-serve 2017-11-12 better than 2017-10-25's 0.3.0

holta:
Images actually show up now thanks to https://github.com/kiwix/kiwix-tools/issues/102#issuecomment-343930567 :-)

### DIFF
--- a/roles/kiwix/tasks/main.yml
+++ b/roles/kiwix/tasks/main.yml
@@ -6,13 +6,13 @@
 
 - name: Set kiwix source file name x86_64
   set_fact:
-     kiwix_src_file: "kiwix-tools_linux64_2017-10-25.tar.gz"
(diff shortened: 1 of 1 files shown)


holta:
Tested on Debian 9.1

mgautierfr:
Just to be clear, I've not recommended you to use a nightly build I've just said that it was working with it :)

My recommendation (not always easy to follow) is to test with nightly and deploy release version.
If you are using a nightly, do it knowing all the facts (especially that we will no keep the file in the server for more than a month)

tim-moody:
in the interests of further clarity, where is the release you recommend? http://download.kiwix.org/bin/0.10/kiwix-0.10-linux-i686.tar.bz2?

PR #519, opened 2017-11-14
Node.js 6.11.5 & 6.12.0 were colliding on Raspbian: is this better?

holta:
Sugarizer playbook failing yet again, just when we thought we'd nailed every last Node.js bug etc :/

Ref: #193


holta:
Smoke-tested on Ubuntu & Raspbian.

Needs further testing on clean OS's.

Issue #502, opened 2017-11-08
Ubuntu 17.10 (Artful Aardvark) Stretch Goal for IIAB 6.5 - Calibre 3.7 + Nextcloud issues


holta:
Ubuntu 17.10 is NOT mainline (Ubuntu 16.04 LTS is our most popular OS today, after Raspbian!)  But still if this helps us all recruit new developers & implementers while preparing for April 2018's Ubuntu 18.04 LTS, that could be a very powerful thing.

As such the 3 people (@floydianslips, @deldesir, @jvonau) who are already making bits of progress installing IIAB 6.5 pre-releases onto Ubuntu 17.10 should (please!) keep notes together here, for the sake of all.  So far this is all I know about, from the Oct 26 call @ http://minutes.iiab.io:
* Create /opt/iiab/iiab/vars/ubuntu-17.yml by copying [vars/ubuntu-16.yml](https://github.com/iiab/iiab/blob/master/vars/ubuntu-16.yml)
* Insert the following line into [/opt/iiab/iiab/scripts/local_facts.fact](https://github.com/iiab/iiab/blob/master/scripts/local_facts.fact) :
`"ubuntu-17" | \`
* Comment out [Lines 49-52 from roles/postgresql/tasks/main.yml](https://github.com/iiab/iiab/blob/master/roles/postgresql/tasks/main.yml#L49-L52) (the hard-coded PostgresSQL version, that's followed by “when: is_ubuntu”)
* ?

_Thanks everyone for adding below as progress accelerates/drags/etc &mdash; regardless whether this slips to [IIAB 6.6](https://github.com/iiab/iiab/milestone/3) or [IIAB 7.0](https://github.com/iiab/iiab/milestone/4) or possibly comes together faster than planned!_

floydianslips:
Started a new install on a 17.10 VM and now I can't even git clone the repo without my github credentials. This is the first time I have had to put them in when just doing a clone. Any ideas?

holta:
@floydianslips:

@jvonau has a theory that you accidentally typed "git push" instead of "git pull" or some such.

If that doesn't solve it, can you paste in the exact input+output?

deldesir:
Useful change when ./iiab-install failed at task 3 due to php7.0 issue, just change php7.0 to the actual version of php installed on your distro (php7.1 in my ubuntu 17.10) in the ubuntu-17.yml file in vars.

deldesir:
[...]

PR #520, opened 2017-11-14
Fixes 1113 (Nov 13) for Ubuntu 17.10

jvonau:
### Fixes Bug
For #502


holta:
Is there concern this "ifupdown" might bloat some Debuntu-family distros?

Was it perhaps removed from Ubuntu 17.10 for a reason?  In any case, something to monitor...

PR #531, opened 2017-11-16
revision2: iiab_base_ver, iiab_revision & reserved_wifi

jvonau:

### DIFF
--- a/roles/0-init/defaults/main.yml
+++ b/roles/0-init/defaults/main.yml
@@ -1,6 +1,6 @@
 # use these as a tag a release at a point in time
 iiab_base_ver: 6.4
-gui_version: 2
+iiab_revision: 0
 
 # These entries should never be changed in this file.
 # These are defaults for boolean routines, 
--- a/roles/0-init/tasks/iiab_ini.yml
+++ b/roles/0-init/tasks/iiab_ini.yml
@@ -1,35 +1,39 @@
 # workaround for fact that auto create does not work on ini_file
 - name: Create iiab config file
-  file: dest='{{ iiab_config_file }}'
-        state=touch
+  file:
+      dest: '{{ iiab_config_file }}'
+      state: touch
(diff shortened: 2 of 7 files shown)


holta:
Example &mdash; put the following into local_vars.yml to ask IIAB not to use this wlan0 interface:
```
reserved_wifi: wlan0
```
(Our IIAB code defaults to the highest numbered, i.e. IIAB would normally use wlan1 if it found both {wlan0, wlan1} but the above ensures that wlan0 is indeed unused by IIAB...)

Then run...
```
cd /opt/iiab/iiab/
./iiab-network
```
_2017-11-20 ADDENDUM: "./iiab-network" now runs runs 0-init & 10-network (essentially the 1st and last Stage of an IIAB install) and going forward should be equivalent to "./runtags network"_

_Whereas in the past "./iiab-network" was more comprehensive than "./runtags network"._

_CLARIF from @jvonau: "the tags in network/main.yml were not being seen that deep into the role...
now network is a meta role of sorts and the tags are respected.."_

holta:
Merged after careful review with @tim-moody & @jvonau during community/team call @ http://minutes.iiab.io

@jvonau has tested this on Debian 9.1 VM on Jenkins.

holta:
[In order to change KA Lite's Admin password, Rick Castorani confirmed that typing "export KALITE_HOME=/library/ka-lite" ***was*** nec in IIAB 6.4, prior to "kalite manage changepassword Admin"]

[But in the end] I may not need to change documentation after all...

As [@jvonau] basically made "export KALITE_HOME=/library/ka-lite" happen automatically in roles/kalite/templates/kalite.sh.j2 -> /usr/bin/kalite within IIAB 6.5 going forward here:

https://github.com/iiab/iiab/pull/531/files#diff-1d4e54d2173b4cc1cb315529183c87b2

PR #532, opened 2017-11-17
Toughened phpMyAdmin playbook

holta:
Essentially phpMyAdmin playbook has been refactored to make it more resilient:

- Works when offline!  Either way:
  1. Rebuilds /opt/phpmyadmin or...
  2. Fails with clear messaging when prereq zip file is missing
- Better preserves state in /opt/iiab/downloads (better visibility of phpMyAdmin version number)
- Version Number now only needs to be set once in [roles/phpmyadmin/defaults/main.yml](https://github.com/iiab/iiab/blob/master/roles/phpmyadmin/defaults/main.yml) e.g.
  phpmyadmin_name: "phpMyAdmin-4.7.5-all-languages"
- Safer as link (/opt/phpmyadmin) & tree/content (e.g. /opt/phpMyAdmin-4.7.5-all-languages) are now all set to {{ apache_user }}:root (owner:group) (earlier the link was the only one whose owner was set, e.g. to www-data:root)
- Ansible output much improved
- Ansible syntax modernized, to its new multi-line format

Tested on Ubuntu 16.04 LTS.

Testing on other OS's would be great.

### DIFF
--- a/roles/mongodb/tasks/main.yml
+++ b/roles/mongodb/tasks/main.yml
@@ -60,10 +60,10 @@
   with_items: "{{ files_to_delete.files }}"
 
 - name: add mongodb to service list
-  ini_file: dest='{{ service_filelist }}'
+  ini_file: dest="{{ service_filelist }}"
             section=mongodb
-            option='{{ item.option }}'
-            value='"{{ item.value }}"'
+            option="{{ item.option }}"
+            value="{{ item.value }}"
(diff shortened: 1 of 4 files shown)


holta:
I'd like to merge this sometime Friday without delay if there are no objections?

Clarif: I meant /etc/iiab/iiab.ini in the commits above, not iiab.env

PR #537, opened 2017-11-19
had-1116-short: wifi install auto AP on reboot

jvonau:

### DIFF
--- a/iiab-from-console.yml
+++ b/iiab-from-console.yml
@@ -16,3 +16,4 @@
       - { role: 7-edu-apps, tags: ['7-edu-apps'] }
       - { role: 8-mgmt-tools, tags: ['8-mgmt-tools'] }
       - { role: 9-local-addons, tags: ['9-local-addons'] }
+      - { role: network, tags: ['network'] }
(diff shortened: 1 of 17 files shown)


holta:
Ambitious & impressive PR by @jvonau merged per quick review with @georgejhunt, @tim-moody, @jvonau, @holta during community/team call @ http://minutes.iiab.io

Going forward we've agreed to evaluate & revisit defaults etc &mdash; what works best for implementers in areas like:

1) Whether the default after reboot might be modified to remain a wifi-as-gateway until the implementer gives their final consent to run iiab-hotspot-on ?  @holta & @georgejhunt  are in favor.  This could greatly help dormroom-style and hotel-style installs where WiFi-to-Internet is the only option to install IIAB, and unplanned reboots can cause a remote IT helper person frustrations when the reboot converts the IIAB's internal WiFi to hotspot prematurely, for all sorts of reasons.
2) Whether "network" should be renamed 10-network (in iiab-install etc) if indeed this proves to be a solid idea to have networking installing during Stage 10 (instead of Stage 4, where it's been to date).

Personally I'm nervous that installs which happen to break in Stage 7-or-such would have been "basically functional" in the past...but are no longer given networking has been deferred from Stage 4 to Stage 10 (worry)

Ah well, we'll see how this plays out :)

PR #540, opened 2017-11-20
manual lint of sugarizer/tasks/main.yml per Ansible recs

holta:
Particularly shell: tar -> Ansible's own unarchive:


holta:
Tested on Ubuntu 16.04 LTS.

PR #546, opened 2017-11-20
revise "./iiab-install --reinstall"

jvonau:


holta:
DNS (name resolution) was breaking on all upgrades ("./iiab-install --reinstall") so this fix is urgent~

PR #557, opened 2017-11-24
proposed changes to default_vars.yml, per implementation norms

holta:


jvonau:
commenting out exFAT_enabled should be safe as it is defined to False in 0-init/defaults/main.yml

PR #271, opened 2017-09-19
openvpn - fix missing uuid on cloned image, keep together

jvonau:

Issue #229, opened 2017-09-08
/etc/iiab/uuid is never created if SD's 1st boot is in RPi Zero W


holta:
FYI this works in RPi3, but not in RPi Zero W.

Context: I (and others in Haiti etc) follow George's routine to "rm /etc/iiab/uuid /etc/iiab/handle" on the source image, and then use "dd" to clone IIAB to many other microSD's.

A /etc/iiab/uuid is then automatically created(*) during the first boot of each clone IF the 1st boot occurs in an RPi3.

However there appears to be a timing-or-similar bug when 1st boot occurs in an RPi Zero W machine, such that /etc/iiab/uuid is never created.

jvonau:
From what I can see a pass through 1-prep is required to recreate the uuid . I'd double check the behaviour on a freshly cloned rpi3 it should behave the same if runansible or the console has not been used yet. Fix for in the field  `./runtags prep`

holta:
@georgejhunt reminds me that xscenet@openvpn will most likely fail when the file /etc/iiab/uuid is absent.

jvonau:
#271

holta:
@georgejhunt, can you explain what auto-creates /etc/iiab/uuid during RPi3 boot, when none exists after dd / cloning?

Is [roles/openvpn/templates/announcer](https://github.com/iiab/iiab/blob/master/roles/openvpn/templates/announcer) somehow involved?

(As compared to the proposed approach @ PR #271 for example!)

georgejhunt:
At one earlier time, uuid was created when admin-console  gathered facts.
But Tim did not like using the fact gathering script for the purpose of
making sure that uuid existed. And at this point we do not have another
start up script that we can all agree upon. A simple place to put it would
be in rc.local -- but that has not been successful as a suggestion for this
sort of thing

I've written a PR which creates a "iiab-startup.sh" script in /usr/libexec
that is stimulated by systemd unit file. We can put the check for uuid
there.

holta:
TODO, now that PR #559 's been merged:
* Someone please report back if /etc/iiab/uuid is being created on RPi Zero W, Thanks!
* All 5 lines in [roles/2-common/tasks/iiab-startup.yml](https://github.com/iiab/iiab/blob/master/roles/2-common/tasks/iiab-startup.yml#L8-L26) that say:
```
when: startup_unit.stat.exists is defined and not startup_unit.stat.exists 
```
...can likely be abbreviated to:
```
[...]

PR #559, opened 2017-11-25
atstart: auto-create /etc/iiab/uuid on cloned images

georgejhunt:
### Fixes Bug
This is an alternative to #271. I do not like generating a uuid only if openvpn is installed.
### Description of changes proposed in this pull request. I 
This creates a bash script that is run at every boot, which in turn looks for uuid, and generates it if it does not exist`
I think a uuid is strategic for determining how many people are contributing to "field back".
rpi stretch (but not a fresh install)

### DIFF
--- a/roles/2-common/tasks/iiab-startup.yml
+++ b/roles/2-common/tasks/iiab-startup.yml
@@ -10,6 +10,7 @@
 - name: Copy startup script
   template: src=iiab-startup.sh
             dest=/usr/libexec/
+            mode=0755
   when: startup_unit.stat.exists is defined and not startup_unit.stat.exists
 
 - name: Ask systemd to recognize the changes
--- a/roles/2-common/tasks/main.yml
+++ b/roles/2-common/tasks/main.yml
@@ -44,6 +44,8 @@
 
 - include_tasks: udev.yml
 
+- include_tasks: iiab-startup.yml
+
 - name: Recording STAGE 2 HAS COMPLETED ==========================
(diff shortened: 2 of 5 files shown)


holta:
Ref #229

holta:
CLARIF!  During our Thur Nov 23 community/team call (http://minutes.iiab.io) we delayed uuid issues to [IIAB 6.6](https://github.com/iiab/iiab/milestone/2).

(As appropriate we can reverse this decision in coming days, if a consensus/compromise emerges and there's time to get this into IIAB 6.5?)

georgejhunt:
delay to 6.6 seems appropriate

holta:
> delay to 6.6 seems appropriate

Or maybe not.  If an agreed-upon & tested solution is available this week?!

Context: Tim has made enormous progress supporting Admin Console downloading of new ZIM files (that contain an internal index) into a local catalog that sorts out new & old &mdash; and also ZIM files that do not contain an index at all.

Also Calibre (etc!) have made massive improvements in recent days &mdash; _so it makes sense to delay IIAB 6.5's release a bit &mdash; such that Every Kid Worldwide gets one in their stocking ;)_

holta:
Monday during our call let's also talk over quickly:

* Community Analytics Purpose (moving in which directions, in consideration of kiwix/kiwix-tools#96 ?)
* Privacy-by-Design action plan ([GDPR](https://en.wikipedia.org/wiki/General_Data_Protection_Regulation) offers thoughtful principles!)

Issue #560, opened 2017-11-25
Runansible on rpi smoketest testing fails immediately with no error message


georgejhunt:
### Expected behavior and actual behavior.
Should at least output a message
### Steps to reproduce the problem.
On a machine loaded with an earlier version of IIAB, download current master, and ./runansible.
Was able to fix by removing the '-e' on the first line of runansible

### Specifications like the version of the IIAB, operating system version, or hardware details.
rpi3 with current master. There must be some obvious thing I'm overlooking. Others would have hit this.

holta:
Going forward `./runansible` should mention the recommended new alternative `./iiab-install` even when it does not fail.

_But we should also support the older approach `./runansible` with rudimentary error messaging if we can hopefully find out what the most common problem(s) are...is that clear in this situation?_

jvonau:
I can reproduce George's runansible error by altering IIAB_DIR= within /etc/iiab/iiab.env to some other value than "/opt/iiab/iiab" then the cd $IIAB_DIR  within runansible kicks in and returns to the prompt.

georgejhunt:
changing -e to -ex shows that runansible was erroring out at current line
34:(after the PR adding a message to use install-iiab

OLD=`grep XSCE /etc/iiab/iiab.env` This error can be eliminated by
bracketing the expected error with "+e" and "-e"

holta:
PR #565 looks to solve the root cause, whereas #563 papers it over?

Or do we want both possibly?

holta:
I'm concerned `grep XSCE /etc/iiab/iiab.env` will fail regardless in #565, just like it has to date.

However I still don't understand why it is failing &mdash; right after file /etc/iiab/iiab.env is confirmed to exist &mdash; does anybody understand this?

jvonau:
Think the -e is reacting to grep returning no value,  I changed to having the test always return a value with the use of wc -l and test for greater than zero.

holta:
> Think the -e to reacting to grep returning no value, I changed to having the test always return a value with the use of wc -l and test for greater than zero.

Confirmed grep indeed forces an error with "set -e"

(Which is counter-intuitive to ppl who run grep from the prompt, and do not see an error where the keyword/regexp is missing from the file...)

Thanks for explaining!

holta:
I'm merged PR #565 after testing it in isolation &mdash; if there's also a need for PR #563, let's discuss that Monday during our community/team call (http://minutes.iiab.io).

holta:
[...]

PR #563, opened 2017-11-25
toggle to "set +e" ignoring errors when working with /etc/iiab files (then back to "set -e")

georgejhunt:
### Fixes Bug
related to #560

yes on the failing rpi


holta:
Profound thanks to @georgejhunt who:
1) Saved legacy users of ./runansible with this testing
2) Clarified how valuable "set +e" then "set -e" are avoid known errors in bash scripts

PR #567, opened 2017-11-25
Kiwix playbook more readable; unused code commented out

holta:
Tested in many different ways on Ubuntu 16.04 LTS.

Including the official trigger to reinstall kiwix-tools, which is to remove /opt/iiab/kiwix/bin/kiwix-serve


holta:
Also tested on Raspbian Lite 2017-09-17 (that was fully updated with apt).

Warning: similar to other OS's the Kiwix playbook will fail if /library/zims  goes missing or is damaged in certain ways &mdash; the entire Kiwix playbook might be beefed up to be more resilient at a future date.

PR #582, opened 2017-11-30
rework scripts/ansible to handle upgrades for debian family

jvonau:


holta:
@tim-moody making progress with: https://github.com/iiab/iiab-admin-console/pull/48

holta:
Looking really good but @jvonau & all agreed during our community/team call this morning (http://minutes.iiab.io) to defer merging until this is tightened up just a bit more.

PR #585, opened 2017-12-02
apache proxy for kiwix

tim-moody:
rough and untested to get it on the table

### DIFF
--- a/roles/kiwix/defaults/main.yml
+++ b/roles/kiwix/defaults/main.yml
@@ -7,7 +7,7 @@ kiwix_src_file_i686: "kiwix-0.10-linux-i686.tar.bz2"   # Published Oct 2016 ("ex
 
 kiwix_port: 3000
 # Expected to be used soon for Kiwix proxy:
-kiwix_url: /kiwix
+kiwix_url: /kiwix/
 # Unused in Nov 2017, but should be:
 kiwix_path: "{{ iiab_base }}/kiwix"
 
--- a/roles/kiwix/tasks/kiwix_install.yml
+++ b/roles/kiwix/tasks/kiwix_install.yml
@@ -151,6 +151,9 @@
     dest: /etc/crontab
   when: kiwix_enabled and is_redhat
 
+- name: Restart apache, so it picks up kiwix.conf
+  service: name={{ apache_service }} state=restarted
+
 # 5. FINALIZE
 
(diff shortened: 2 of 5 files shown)


tim-moody:
pass smoke and function test on debian 9 vm

Please note it does not upgrade existing install as playbook only reinstalls the kiwix exec even on force

holta:
> Please note it does not upgrade existing install as playbook only reinstalls the kiwix exec even on force

Kiwix does upgrade itself reliably & completely with "./runtags kiwix" (certainly on Raspbian & Ubuntu 16.04 LTS) if /opt/iiab/kiwix/bin/kiwix-serve is removed per https://github.com/iiab/iiab/blob/master/roles/kiwix/tasks/kiwix_install.yml#L32-L40

This has been tested carefully on Raspbian and Ubuntu 16.04 LTS today and over the past week.

If anybody can reproduce different behavior across Debian or more broadly across multiple OS's please let us know if there's a genuine issue here.

PS Arguably IIAB's too heavy-handed, upgrading services (or at least many of their components) too often, whenever IIAB playbooks are rerun.  Example: Ansible's template module does a mandatory upgrade of all files @ https://github.com/tim-moody/iiab/blob/kiwix-proxy-12-2/roles/kiwix/tasks/kiwix_install.yml#L98-L112 for now.  In future we can scale this back if nec &mdash; e.g. with "force: no" [*] or "when: kiwix_force_install" etc &mdash; to prevent the following from being upgraded/overwritten every time:
* /usr/bin/iiab-make-kiwix-lib
* /usr/bin/iiab-make-kiwix-lib.py
* /etc/apache2/sites-available/kiwix.conf

[*] http://docs.ansible.com/ansible/latest/template_module.html

holta:
This PR #585 works on Ubuntu 16.04 LTS &mdash; in the sense that the Random button works when clicking on ZIM files @ http://box:3000 &mdash; any other things that need testing?

tim-moody:
the test is does box/kiwix/ bring up the kiwix home page

tim-moody:
iiab-menu has been changed to support this PR, so can now be merged

holta:
@tim-moody where should we document backward compatibility (presence or absence of kiwixUrl property in config.json determines proxy or no proxy) ?

FYI For All: test this using Test http://box/kiwix/ (trailing slash important!) &mdash; noting that all visits to main page http://box menuing system now use this, &mdash; in turn uses port 3000 to access ZIM files under the covers.

PR #588, opened 2017-12-03
force br0 to always be available on rpi

jvonau:


tim-moody:
where does wlan0 get added to the bridge?

holta:
> where does wlan0 get added to the bridge?

@jvonau answered "when hostapd starts..." in live chat.

holta:
Am merging this in coming moments to bring about wider testing, unless there are objections.

Can easily be reversed if nec.

Issue #412, opened 2017-10-14
Captive Portal design choices: onboarding non-Latin communities / URL literacy


holta:
New users (rich and poor, literate and non-literate) often struggle to type in http://box and http://box.lan into their phones, even when they do know a Latin-based alphabet.

And increasingly many do not know a Latin-based alphabet, as new countries and communities get involved, outside of major cities etc!

It is possible to force client machines' browsers to open a page automatically in support of new users — however each OS has its own quirks.

@tim-moody, @m-anish and @georgejhunt have done extensive research here since 2016: _please do contact them if you can help._

jvonau:
How about a look at what code that is hidden from view that exists now contains? Smells like a hidden fork for profit of the few otherwise.

holta:
@jvonau my understanding (unfortunately extremely limited!) is that these diverse code efforts, while all over the map, are all public in place likes https://coova.github.io

Some of the original "DNS jail" work is part of other Kiwix-connected and RACHEL-related projects if I recall correctly?

But @tim-moody can share much more during our upcoming community calls (http://minutes.iiab.io) as he's gathered the most comprehensive overall view of all these diverse open source efforts last I checked; and possibly @georgejhunt too.

Solving this browser/human literacy issue (which is much more complex of a UX challenge than it initially appears appears on the surface, hence the reason this struggle has gone on+off repeatedly for about 2 years now) involves browser tricks+popups used by Chrome, Microsoft, Molliza, iOS, Android itself (etc) that are not only very subtle, often undocumented, but also ever-evolving...

@darkenvy & @floydianslips have also done some basic R&D here too, along these lines?
https://github.com/nodogsplash/nodogsplash (and the related NoCatSplash)

Also Ivan Savov @ Learning Equality (KA Lite) had this "Captive DNS" suggestion for us after our August http://OFF.NETWORK hackathon too:
https://github.com/learningequality/kolibri-sandisk-connect/blob/master/ansible/roles/system/files/dnsd
https://github.com/learningequality/kolibri-sandisk-connect/blob/master/ansible/roles/system/files/resolve_all_dns_to_host_ip.py

_PS there's also an opposing view that this entire "Captive Portal" effort might be way more trouble than it's worth (?)  When a giant sticker (http://BOX.LAN) pasted on these handheld Internet-in-a-Box devices solves ~99% of the problem rather efficiently, and in a more personable way, empowering local cultures arguably..._

slaporte:
[...]

PR #597, opened 2017-12-05
initial check in

tim-moody:
This is smoke tested on rpi
Not connected to any top level role
Needs more refinement on splash page
ToDo add auto submit on splash page

### DIFF
--- a/roles/nodogsplash/defaults/main.yml
+++ b/roles/nodogsplash/defaults/main.yml
@@ -1,0 +1,3 @@
+nodogsplash_install : False
+nodogsplash_enabled : False
+nodogsplash_arm_deb : nodogsplash_2.0.0-1_armhf.deb
--- a/roles/nodogsplash/tasks/main.yml
+++ b/roles/nodogsplash/tasks/main.yml
@@ -1,0 +1,3 @@
+- name: Install nodogsplash (Raspbian only)
+  include_tasks: rpi.yml
+  when: is_rpi
--- a/roles/nodogsplash/tasks/rpi.yml
+++ b/roles/nodogsplash/tasks/rpi.yml
@@ -1,0 +1,68 @@
+- name: nodogsplash dependencies
+  package:
+    name: libmicrohttpd12
+    state: present
+
+- name: Get the nodogsplash software
+  get_url:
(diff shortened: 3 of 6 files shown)


tim-moody:
Not sure where this belongs in the stepwise install. It needs to own iptables.

holta:
Ref #412 "Captive Portal design choices: onboarding non-Latin communities / URL literacy"

tim-moody:
not sure if it restores previous iptables rules on stop - it does

tim-moody:
anecdotally enabling nds and rebooting yields the correct iptables rules

tim-moody:
Need to check status of pass through which seems enabled

holta:
@deldesir is interested in making this work on Ubuntu 16.04 and 18.04 &mdash; thanks @deldesir if you get a chance to work with @tim-moody here!

tim-moody:
Client Platform | Test Results
-- | --
Win 10 Laptop | IIAB home page shown on connect to SSID
Android 7.0 | Got redirected to home page, but still says Sign In at top   Went to wifi connection page and took ‘keep this connection; looks like I   have access to internet sites’   Now the splash page no longer appears   Restarted nodogsplash   Rebooted Android   Forgot connection   Still don’t get splash page   Now checked and nodogsplash no running   Started it and phone went to splash page; I clicked continue and now   don’t see signin (?but is it because it remembers the box page from previous   access)   I still think I am getting passthrough   12/8 changed config to redirect to home page and put variable back in   form field. Even when clicking continue I still see sign in to network in   bold letters at top (and fonts are wrong)
Android 4.3 | Connected without popup; has message sign into network   I clicked that and the splash screen came up and auto redirected to   the home page (which does not say signin)
iPhone 4, iOS 6.1.3 | Connected; then popped up and auto redirected to home page   BUT home page says login   I set it down and then it went back to AirX   Clicked iiab ssid and got iiab.me?!   Entered box and got to home page, BUT maybe remembered it.   BTW, Appendicitis video can not be started (nor pancreatitis)
Mac Air | Connected and showed splash then auto redirected.   Menu does not show sign in message (actuall in top bar it says “join   Internet in a Box”   BTW, Appendicitis video can not be started

PR #605, opened 2017-12-07
always restart dns server

jvonau:


holta:
A WiFi-based install of MIN-sized RPi3 completed and could ping 8.8.8.8, but DNS was not working.

Let's hope this fixed things!

Issue #594, opened 2017-12-05
"get_url:" needs "timeout: 1800" (or such) for low-bandwidth/WiFi installs


holta:
This worked, allowing WiFi-based install of [big-sized](http://wiki.laptop.org/go/IIAB/local_vars_big.yml) IIAB 6.5/master to complete on a RPi3 with a ~1mbit/sec connection ("./iiab-install" was constantly failing prior to increasing Ansible's default timeouts below...)

Basic Idea...insert "timeout: 1800" (or similar) after "get_url:" in these places:

* [calibre/tasks/create-db.yml](https://github.com/iiab/iiab/blob/master/roles/calibre/tasks/create-db.yml#L13) downloads 513KB Metamorphosis-jackson.epub
* [kiwix/tasks/main.yml](https://github.com/iiab/iiab/blob/master/roles/kiwix/tasks/main.yml#L27) downloads 98MB kiwix-tools_armhf_2017-12-01.tar.gz
* [nextcloud/tasks/main.yml](https://github.com/iiab/iiab/blob/master/roles/nextcloud/tasks/main.yml#L17) downloads 42MB nextcloud_latest-12.tar.bz2
* [phpmyadmin/tasks/main.yml](https://github.com/iiab/iiab/blob/master/roles/phpmyadmin/tasks/main.yml#L2) downloads 11MB phpMyAdmin-4.7.6-all-languages.zip
* [sugarizer/tasks/main.yml](https://github.com/iiab/iiab/blob/master/roles/sugarizer/tasks/main.yml#L2) downloads 217MB sugarizer-0.9.tar.gz
* [wordpress/tasks/install.yml](https://github.com/iiab/iiab/blob/master/roles/wordpress/tasks/install.yml#L15) downloads 9.2MB wordpress-4.9.1.tar.gz

_Caveat: KA Lite's installation can still fail as its own internal downloader can be similarly impatient, see [#589](https://github.com/iiab/iiab/issues/589)._

jvonau:
I don't like the one size fits all, big stick approach. If you are going to add a timeout do something like 
>  get_url:
    url: "{{ some_path }}"
    dest: "{{ downloads_dir }}/"
    timeout: "{{ max_timeout }}"

Next define max_timeout in roles/0-init/defaults/main.yml, keep this set to a sane/safe value for wired devices. This allows for an override that could be adjusted in local_vars.yml to suit the end-user's environment without waiting for hours for a failure to show up when a wired device is used.

[...]

PR #613, opened 2017-12-08
Resilient/Recoverable pkg downloads during IIAB installation, even if WiFi < 1mbit/sec

holta:
`timeout: "{{ download_timeout }}"` was added to all the `get_url:` calls in common usage, per design discussion @ #594.

Variable `download_timeout` is initially set to 100 seconds within [default_vars.yml](https://github.com/iiab/iiab/blob/master/vars/default_vars.yml) and [local_vars.yml](https://github.com/iiab/iiab/blob/master/vars/medium.localvars) as Ansible's default timeout for "get_url:" downloads (10 seconds) often fails.

_Resumable downloads with multiple "retries:" have not been implemented&mdash; but might possibly be considered in future?_

### DIFF
--- a/roles/0-init/tasks/main.yml
+++ b/roles/0-init/tasks/main.yml
@@ -1,7 +1,8 @@
 # Initialize
 
 - name: ...IS BEGINNING ============================================
-  stat: path=/etc/iiab/iiab.env
+  stat:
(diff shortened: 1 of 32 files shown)


holta:
Tested on Ubuntu 16.04 LTS and Raspbian Lite.

PR #615, opened 2017-12-08
kiwix-serve -> kiwix in /etc/iiab/iiab.ini for iiab-make-kiwix-lib.py used by ./cp-menus

holta:
1) The name "kiwix" seems more broadly appropriate than "kiwix-serve" and is now part of the kiwix playbook, as placed in /etc/iiab/iiab.ini automatically during install.  Similar to the "./runtags kiwix" name etc.

2) Does anyone know if this kiwix section (formerly kiwix-serve section) within /etc/iiab/iiab.ini is referenced in any other places, beyond just iiab-make-kiwix-lib.py ?  (as used by https://github.com/iiab/iiab-menu/blob/master/cp-menus)

### DIFF
--- a/roles/kiwix/templates/iiab-make-kiwix-lib.py
+++ b/roles/kiwix/templates/iiab-make-kiwix-lib.py
@@ -108,8 +108,8 @@ def init():
     config = ConfigParser.SafeConfigParser()
     config.read(iiab_config_file)
     iiab_base_path = config.get('location','iiab_base')
-    iiab_zim_path = config.get('kiwix-serve','iiab_zim_path')
-    kiwix_library_xml = config.get('kiwix-serve','kiwix_library_xml')
+    iiab_zim_path = config.get('kiwix','iiab_zim_path')
+    kiwix_library_xml = config.get('kiwix','kiwix_library_xml')
     kiwix_manage = iiab_base_path + "/kiwix/bin/kiwix-manage"
 
 # Now start the application


holta:
I will merge this later this afternoon if there are no objections.

Most immediately so ./cp-menus stops throwing errors &mdash; but also because Emmanuel & All generally prefer to call this functionality "kiwix" rather than "kiwix-serve", when referring to all 5 executables within /opt/iiab/kiwix/bin.

tim-moody:
Please stop renaming things for the sake of renaming as this tends to break things and have unforeseen consequences. The name of the application is kiwix-serve. Kiwix is the organization that produces it.

holta:
I humbly disagree.  /etc/iiab/iiab.ini uses the popular human names like [calibre] unlike its technocratic internals like calibre-serve.service etc.  Quickly understandable human-readable files like /etc/iiab/iiab.ini can greatly help new people understand the ropes, even if yes every now and then arbitrary traditions-for-tradition's-sake (that I've caused, or whoever has caused) need to pass into history in favor of new more vivid documentation.

e.g. when the culture moves forward in one direction or another (I tried calling this "kiwix-serve" over the past year, but this refers to something larger including kiwix-manage etc too, and regardless folks strongly prefer to call it "kiwix" in this case, as a more memorable mnemonic, the verdict is clear after a year, when talking with Emmanuel and many others.)

If there are consequences to this name change, I will help fix them.  So far I've looked carefully and have not found any beyond iiab-make-kiwix-lib.py.  But if surprises arise, I will help.

PR #622, opened 2017-12-13
iiab-install: [[ `grep` ]] -> grep -q; ansible version logic cleaner

holta:


holta:
Smoke-tested on Ubuntu 16.04 LTS & Raspbian Lite.

PR #635, opened 2017-12-27
Experiment trying to get Ethernet working during WiFi installs

holta:
Highly-experimental attempt at fixing #616 "WiFi-installed IIAB does not connect using an Ethernet cable"

_Hopefully @jvonau and others can help refine this!_

Context: too many WiFi installs are failing, and implementers get profoundly stuck, needing Ethernet-to-Internet in order to recover. 

Once they finally get through `./iiab-install` (using Ethernet-to-Internet or WiFi-to-Internet) they of course need to run `iiab-hotspot-on` if they began the process over WiFi &mdash; as documented @ https://github.com/iiab/iiab/wiki/IIAB-Installation#do-everything-from-scratch (Item 3.)

Also Related: https://github.com/iiab/iiab/issues/558#issuecomment-350842090, PR #617 _(still nec or not?),_ PR #634 _merged._

PR #647, opened 2018-01-05
Experiment with promiscuous mode on wlan0 to work around #638?

holta:
Possible interim/temporary fix to mitigate the worst of #638 WiFi failures:
- Should this be conditioned on flag RPi/Raspbian within roles/2-common/templates/iiab-startup.sh and roles/network/templates/network/iiab-hotspot-off ?
- Might iiab-hotspot-off possibly disable promiscuous mode, if our RPi3 AP is already proven sufficiently reliable streaming video to 32 clients etc?  (If so, perhaps experiment with including `ip link set dev wlan0 promisc off` inside iiab-hotspot-off?)


tim-moody:
shouldn't this only be done if wifi is a gateway?

tim-moody:
when: startup_unit.stat.exists is defined and not startup_unit.stat.exists
means startup service can not be upgraded, BTW

holta:
> shouldn't this only be done if wifi is a gateway?

I tend to agree, but James Cameron has [concerns](https://github.com/iiab/iiab/issues/638#issuecomment-355467217) and also note that PR #635 [functionality](https://github.com/holta/iiab/blob/ethernet/roles/network/tasks/main.yml#L7-L16) hasn't been merged yet (hopefully soon!) so yeah this needs to keep evolving for sure.

> when: startup_unit.stat.exists is defined and not startup_unit.stat.exists
means startup service can not be upgraded, BTW

@georgejhunt i think that's a question from @tim-moody to you?

georgejhunt:
We got into the ping/promiscuous experimentation, trying to understand an
unreliable upstream connection.  I don't think we have any direct
experience with upstream connection metrics (outside of continuous ping),
or their relation to promiscuous mode. Is downloading packages more
reliable? Does the upstream connection stay up for days?

Seems like this PR is necessary, but not sufficient, without further
testing.

holta:
> Is downloading packages more reliable?

No more, no less.  Let's keep an eye it, to be sure, but the real goal is below.

> Does the upstream connection stay up for days?

Yes, so far after almost a week's testing it appears far more robust.

PS this PR #647 is included in the larger PR #635.

holta:
This functionality is included within PR #635 which has been merged.

Can re-open later as nec.

PR #652, opened 2018-01-11
Turn on wlan0 promiscuous mode within network playbook...or much earlier?! (RPi)

holta:
wlan0 promiscuous mode should be turned on much earlier during ./iiab-install, so that WiFi-to-Internet is far more stable on RPi when installing over WiFi!

Unfortunately [/usr/libexec/iiab-startup.sh at the end of Stage 2](https://github.com/iiab/iiab/blob/master/roles/2-common/templates/iiab-startup.sh) (2-common) does not do this &mdash; long before detected_network.yml is run.

So for now enabling this within the "network" playbook/stage (after Stage 9, equivalent to ./iiab-network) will have to do ?

### DIFF
--- a/roles/network/tasks/main.yml
+++ b/roles/network/tasks/main.yml
@@ -18,6 +18,9 @@
 # Separate Idea, Not Without Risks: should WiFi-as-gateway detection logic
 # be encapsulated into roles/network/tasks/hostapd.yml in future?  Whereas
 # today "./runtags hostapd" doesn't exist & "./runtags AP" is UNSUPPORTED!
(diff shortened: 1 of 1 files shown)


holta:
@jvonau please review today if poss!  Either way:
1) We can go with this PR that enables wlan0 promiscuous mode near the end of ./iiab-install (not ideal, but far better than nothing!)
2) Or we can find a more clever way to enable wlan0 promiscuous mode far earlier on in IIAB's install process, to make these install-over-WiFi sagas more reliable & achievable??

(If 2. is possible, RPi install-over-WiFi will certainly have far fewer "10SEC" glitches per #638, just as I faced myself trying to install this afternoon!)

PR #659, opened 2018-01-24
need sqlite3 for rachel maps

georgejhunt:
### Fixes Bug
rachel maps require sqlite3 module
not smoke tested

### DIFF
--- a/roles/httpd/tasks/main.yml
+++ b/roles/httpd/tasks/main.yml
@@ -14,9 +14,9 @@
   package: name=php{{ php_version }}-sqlite
   when: is_debian and ansible_distribution_major_version == "8"
 
-#- name: Debian changed sqlite name
-#  package: name=php{{ php_version }}-sqlite3
-#  when: ansible_local.local_facts.os_ver == "debian-9"
+- name: Debian changed sqlite name
+  package: name=php{{ php_version }}-sqlite3
+  when: is_debian and ansible_distribution_major_version == "9"
 
 -  name: Install httpd required packages
    package: name={{ item }}


holta:
@georgejhunt this works for cities without accents, Thanks!

Do you know why the final click (after autocomplete indeed works!) fails to show those cities, for all cities that contain accents in their names?

tim-moody:
sqlite install was a by-product of installing admin console, which is now optional.

tim-moody:
maybe I missed the point if this is due to a debian version change. could put in os.yml

holta:
@tim-moody & @jvonau please clarify if this PR should be merged, i.e. if it is the best way to accomplish "apt install sql7.0-sqlite3" as Raspbian requires to make city search work, per http://dev.worldpossible.org/mods/en-worldmap-10/map.html

Thanks again to @georgejhunt who solved this making a HUGE different in many Mexican schools 2 weeks ago!

tim-moody:
it is a way to install php7.0-sqlite3, though I don't understand why  when: ansible_local.local_facts.os_ver == "debian-9" didn't work. I don't object to merging after a smoke test.

holta:
> it is a way to install php7.0-sqlite3, though I don't understand why when: ansible_local.local_facts.os_ver == "debian-9" didn't work.

Because it was commented out?

> I don't object to merging after a smoke test.

Agreed: let's merge after a smoke test on Raspbian &mdash; and ideally a functionality test of city searching too!

holta:
Smoke-tested on Raspbian Lite on RPi3.

PR #665, opened 2018-02-07
httpd indentation/syntax per new Ansible documentation

holta:


holta:
Smoke-tested on Ubuntu 16.04 LTS, so I will now merge.

PR #668, opened 2018-02-10
do not ignore gui request for fixed ip

georgejhunt:
### Fixes Bug
admin-console does not set ip address
yes

### DIFF
--- a/roles/network/tasks/computed_network.yml
+++ b/roles/network/tasks/computed_network.yml
@@ -46,7 +46,7 @@
     wan_netmask: "{{ gui_static_wan_netmask }}"
     wan_gateway: "{{ gui_static_wan_gateway }}"
     wan_nameserver: "{{ gui_static_wan_nameserver }}"
-  when: gui_static_wan
+  when: gui_static_wan or user_wan_iface != "auto"
 
 # we need to have an interface name for ifcfg-WAN to be able to change gateway
 # the DEVICE from the gui. Thanks to George for proving my point about knowing
--- a/roles/network/tasks/computed_network.yml
+++ b/roles/network/tasks/computed_network.yml
@@ -46,7 +46,7 @@
     wan_netmask: "{{ gui_static_wan_netmask }}"
     wan_gateway: "{{ gui_static_wan_gateway }}"
     wan_nameserver: "{{ gui_static_wan_nameserver }}"
-  when: gui_static_wan and user_wan_iface != "auto"
+  when: gui_static_wan
 
 # we need to have an interface name for ifcfg-WAN to be able to change gateway
 # the DEVICE from the gui. Thanks to George for proving my point about knowing


georgejhunt:
Requires reboot after running ./iiab-network.

holta:
@jvonau can you review this removal of `and user_wan_iface != "auto"` (just in case you notice any unintended consequences?)

Hence allowing fixed IP addresses to be set in Admin Console's web GUI, if indeed this is safe to merge?

(@georgejhunt mentions this code should only affect RPi, mitigating risks on other HW platforms)

holta:
Look great as this functionality had not been working for the past 6 months or more!

Thanks @tim-moody who did basic functionality testing on RPi3, allowing the setting of a fixed IP address:

1) http://box/admin -> Configure -> Network Parameters -> Override IP Addresses -> apply checkbox "Check to use a static WAN IP Address instead of DHCP"
2) enter chosen IP address in "WAN IP Address on RPi3"
3) not sure this step is required, but likely: click "Save Configuration" then "Install Configured Options"
4) confirm above I.C.O. is complete within Utilities menu -> Display Job Status -> Refresh Status
5) reboot

_Bad news is the above process cannot be reversed by the GUI for now, according to @tim-moody's initial tests &mdash; the static IP address remains even when you try to change back to gateway, save, ICO, reboot._ [[CORRECTION BELOW]](https://github.com/iiab/iiab/pull/668#issuecomment-364672235)

tim-moody:
Tests in console on rpi3 with medium local vars:
1) checked static ip address save and install conf options - no change to ip addr
2) changed to appliance mode and ran - no change
3) reboot - has the static address
4) set mode back to gw, save and ico - still static
5) reboot - still static
6) uncheck static ip,  save and ico:
 2: eth0: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc pfifo_fast state UP group default qlen 1000
    link/ether b8:27:eb:6f:70:ac brd ff:ff:ff:ff:ff:ff
    inet 192.168.0.194/24 brd 192.168.0.255 scope global eth0 [**static**]
       valid_lft forever preferred_lft forever
    inet 192.168.0.36/24 brd 192.168.0.255 scope global secondary eth0 [**dhcp**]
       valid_lft forever preferred_lft forever
7) reboot - only dhcpd address on eth0

holta:
Seems I misunderstood @tim-moody earlier.  Using http://box/admin to revert/remove a Fixed IP (and then rebooting) did indeed work for him.  Sorry for the false alarm.

georgejhunt:
It occurred to me after the call this morning that maybe a better solution would be to change the "and" to "or" in the original line 49 of computed_network.yml.  This would permit local_vars to specify static ip  (via user_wan_iface and associated variables) without the GUI (Jerry's original intention I think)

holta:
> maybe a better solution would be to change the "and" to "or" in the original line 49 of computed_network.yml. This would permit local_vars to specify static ip (via user_wan_iface and associated variables) without the GUI (Jerry's original intention I think)

@jvonau @tim-moody should we go ahead with the above?

holta:
Let's merge George's new suggestion ("or" rather than "and") to increase testing coverage.

Reverting or enhancements can considered anytime as new info comes to light.

Thanks all who've been testing extensively in recent weeks!

Issue #669, opened 2018-02-10
is Ansible 2.4.3 running out of memory when installing IIAB 6.5/master on RPi3?


holta:
If #666 is the devil's number, we can revert it :)

Quick summary (of "cannot allocate memory" when running the ["network" role](https://github.com/iiab/iiab/tree/master/roles/network)) @ https://github.com/iiab/iiab/pull/666#issuecomment-364667294

The workaround is to reboot and re-run Ansible:
- cd /opt/iiab/iiab
- ./iiab-install

The suspicion (needs confirmation) is that this problem results from Ansible 2.4.3 released on 2018-01-31.

If confirmed/serious, reverting to Ansible 2.4.2 may be the pragmatic answer until the upstream issue can be addressed.

holta:
@tim-moody @georgejhunt can you paste in the error message onto this ticket if poss?

tim-moody:
An exception occurred during task execution. To see the full traceback, use -vvv. The error was: OSError: [Errno 12] Cannot allocate memory

holta:
> OSError: [Errno 12] Cannot allocate memory

Unrelated to IIAB, others experienced this with Ansible 2.4.1:
https://github.com/ansible/ansible/issues/32854

And yet this is (alleged) to be improved in Ansible 2.4.3:
https://github.com/ansible/ansible/issues/31673

Thanks all for more reports helping us isolate the issue & how to reproduce it.

holta:
@jvonau : Ansible 2.5 was released in beta form yesterday @ https://releases.ansible.com/ansible/ so let's look into how we might pin [/opt/iiab/iiab/scripts/ansible](https://github.com/iiab/iiab/blob/master/scripts/ansible) to 2.4.x possibly?

holta:
A workaround for RPi &mdash; and Ubuntu 16.04 LTS too if nec &mdash; until Ansible 2.4.3 is fixed:

> wget http://download.iiab.io/packages/ansible_2.4.2.0-1ppa~xenial_all.deb
> apt install ./ansible_2.4.2.0-1ppa~xenial_all.deb

I've cached Ansible 2.4.2's PPA above, as it's expected to disappear from https://launchpad.net/~ansible/+archive/ubuntu/ansible/+files/ansible_2.4.2.0-1ppa~xenial_all.deb shortly.

holta:
[...]

PR #666, opened 2018-02-07
Ansible 2.4.2 -> 2.4.3 in scripts/ansible

holta:
Ansible 2.4.3 was released 2018-01-31.

Ansible 2.4.2 was released 2017-11-29.

Changelog:
https://github.com/ansible/ansible/blob/stable-2.4/CHANGELOG.md

PR #671, opened 2018-02-11
install Ansible 2.4.2 on RPi/Deb/Ubuntu from download.iiab.io (until 2.4.3 fixed!)

holta:
Mitigates Ansible 2.4.3 memory issue/bug summarized @ #669

(Does more than just revert PR #666 ...so as to ensure Ansible 2.4.2-rather-than-the-latest is installed on Raspbian ...and also on Ubuntu/Debian too FWIW, even if those will presumably run out of memory much more rarely?)

### DIFF
--- a/scripts/ansible
+++ b/scripts/ansible
@@ -3,7 +3,7 @@
 # Installs or upgrades to the best possible Ansible release, so iiab-install
 # can proceed.  Ensure you're online before running this script!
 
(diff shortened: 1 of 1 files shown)


holta:
Smoke-tested on Ubuntu 16.04 and Raspbian.

Seems like an important-if-temporary workaround that should be merged very shortly...

Issue #654, opened 2018-01-16
usbmount to http://box/usb doesn't work w/ modern exFAT sticks


holta:
Ref #329 "http://box/usb "teacher content" does not appear from USB sticks" _[after cold boot]_

holta:
Moving to IIAB 6.5 for discussion.  This is starting to affect teachers in Mexico who can no longer use IIAB to display/share their content, as exFAT-formatted 32+ GB memory sticks are becoming mainstream.

holta:
> [affects] teachers in Mexico who can no longer use IIAB to display/share their [USB stick] content, as exFAT-formatted 32+ GB memory sticks are becoming mainstream.

@georgejhunt are you able to confirm this unfortunate behavior?

And/or suggest a path forward if so?

holta:
Thursday Community/team call excerpt from http://minutes.iiab.io :
- @tim-moody wonders if security of exFAT executables is related to these new store-bought 32+ GB sticks failing to display teacher content?
- @georgejhunt / @floydianslips / others suspect we need to tweak automounting of exfat filesystems &mdash; i.e. http://box/usb functionality might/likely start working as soon as we get exfat-formatted sticks to automount (manually mounting them is not a problem today, since IIAB introduced exFAT support late in 2017)
- @jvonau suggests we might need to add "exfat" to usbmount's own list of filesystems it's supposed to automount
- @georgejhunt / others ask that we verify this with **store-bought exFAT sticks** to be sure (not just our own USB sticks, that we've reformatted to exFAT)

holta:
Is it possible we just need to add "exfat" to the following line in /etc/usbmount/usbmount.conf ?
```
FILESYSTEMS="vfat ext2 ext3 ext4 hfsplus"
```
_How is this best achieved within [roles/usb-lib](https://github.com/iiab/iiab/tree/master/roles/usb-lib) presumably?_

Note To Self: documentation [roles/usb-lib/README.rst](https://github.com/iiab/iiab/blob/master/roles/usb-lib/README.rst) is very stale and needs to be replaced with http://wiki.laptop.org/go/IIAB/FAQ#Can_teachers_display_their_own_content.3F or similar after we finalize/merge this fix.

holta:
Worrisome blog post from someone facing this very same "how to automount exfat" obstacle:
https://unix.stackexchange.com/questions/403591/using-usbmount-with-exfat-drives

This other blog post might possibly be more reassuring...contradicting the above :)
[...]

PR #682, opened 2018-02-15
automount USB exFAT sticks (& possibly portable NTFS disks too)

georgejhunt:
### Fixes Bug
exfat stick to not mount
let systemd manage the usbmount. Udev does not wait long enough
only on top of a previous install via ./runtags


holta:
@georgejhunt might this automount NTFS-formatted hard disks too?

Ref: #654 ("usbmount to http://box/usb doesn't work w/ modern exFAT sticks")

holta:
George's Summary during our community/live call (http://minutes.iiab.io): This PR creates a unit file, returning quickly so udev routine can return quickly.

PR #543, opened 2017-11-20
gh-rpi-wifi-gw2: WiFi as both upstream & downstream

jvonau:
### Fixes Bug
replacement for #431

PR #690, opened 2018-02-26
Wifi2 - simultaneous AP & WiFi-as-client on RPi (using dnsmasq)

georgejhunt:

- Substitute dnsmasq for dhcpd and bind9/named. I could not get dhcpd to play nice with hostapd.
- Put a bash snippet into iiab-startup.sh to split the wifi into two adapters and assign separate MAC addresses.
- Rewrite iiab-hotspot-on/off to conform to the new wifi config. Rewrite iptables to open WAN during iiab-hotspot-off.
- Let dhcpcd manage wlan0 during iiab-hotspot-off boot process (otherwise disable dhcpcd for wlan0).
- Write a new script "iiab-internet-on" with optional "toall" *[renamed to `-f` or `--force` [below](https://github.com/iiab/iiab/pull/690#issuecomment-368557871)]* parameter (which temporarily ignores vars/local_vars.yml's iiab_gateway_enabled). 
- iiab-internet-on lasts until the next reboot. Preliminary testing suggests that placing "iiab-internet-on" in /etc/rc.local works as expected if prefixed with a 20 second delay.
-  Created an image, uploaded to http://download.iiab.io/6.5/rpi/old/iiab-6.5-180223-updown_wifi-g07bc1ef.img.zip, downloaded the image, copied to SD, verified function in rpi3 and rpiW.

1. Raspbian.
2. Yet to test in Ubuntu, Debian, Fedora, and CentOS.


tim-moody (commented):
Current master has two scripts, ansible, which installs 2.4.2, and ansible-2.2.0. This change seems to convert the second into the first.

holta:
Community/team call (http://minutes.iiab.io) converged on this syntax: (a bit more clear than "toall" as earlier suggested/above)
```
iiab-internet-on -f
iiab-internet-on --force
```
Both above flags being equivalent: the effect is to enable passthrough / masquerading, ignoring the value of local_vars.yml's "iiab_gateway_enabled".  **WARNING: the `iiab-internet-on` command (regardless of flag or not) would Always Be Temporary, persisting only until the next boot, as George could not make it work across reboots[*].  This is unlike all other commands like `iiab-hotspot-on|off` and `iiab-remote-on|off`.**

[*] unless it's added into rc.local with a 20sec delay.  Which might cause issues with remote support, affecting OpenVPN's initial start possibly?

Fringe benefit of this "only until next reboot" behavior, is that it offers some additional security: whereas *intentionality* is very clear when plugging Ethernet into an upstream Internet/DHCP server &mdash; the requirement to type "iiab-internet-on -f" (e.g. into an RPi Zero W every time, from a hotel room, after entering hotel's WiFi credential into /etc/wpa_supplicant/wpa_supplicant.conf) reaffirms that the operator truly wants its WiFi online each time (i.e. acting as an Internet-connected WiFi client _and_ acting as an AP!)

holta:
@tim-moody recommends “chmod 600” for networking files like the following, that contain WiFi credentials (and similar):
- /etc/wpa_supplicant/wpa_supplicant.conf
- /etc/hostapd/hostapd.conf
- /etc/hostapd/hostapd.conf.iiab
- any others?

holta:
Most of us are confident that http://169.254.x.x (Link-Local) IP’s (instead of http://box) will continue to work as-good-as-permanently upon inserting an Ethernet cable, as saved the day in Nigeria earlier this month (indeed!)  Thanks to [RFC3927 Section 2.1](https://tools.ietf.org/html/rfc3927#section-2.1) &mdash; among facilities that choose not to have a DHCP server &mdash; but this needs to be verified of course.

Aside: http://box.local did not work in Nigeria, which might have been nice, but folks are more than happy to bookmark that specific computer lab's http://169.254.x.x (instead of the usual http://box or http://box.lan or http://172.18.96.1) within every client PC's browser. 

Caveat: link-local URL's for IIAB (which work great on Windows & Mac) might not work on some Linux distros/PC's.

roles/network/tasks/rpi_debian.yml
```diff
@@ -20,11 +20,16 @@
     dest: /etc/dhcpcd.conf
     src: network/dhcpcd.conf.j2
 
-- name: Copy the bridge script for RPi
+- name: Copy the neworking script for RPi
```
holta: @georgejhunt typo..plz change "neworking" to "networking" on Line 23 of roles/network/tasks/rpi_debian.yml

holta:
Background to help explain this PR #690 &mdash; enabling WiFi-as-AP and WiFi-as-client simultaneously:
- this builds off #430, PR #431, PR #543 (etc!) and relates in part to #197 from 2017-09-02 ("Permit installation of IIAB over Wi-Fi instead of Ethernet", the original/narrow goal of IIAB 6.5)
- upstream/downstream WiFi (AP mode & WiFi-as-client) MUST use the same radio channel!

georgejhunt:
There is a candidate page for the IIAB wiki at https://github.com/georgejhunt/iiab/wiki/Using-Wifi-during-IIAB-Installation-on-the-Pi-3-and-W

tim-moody:
I find the text nicely written and to be commended for including the reference to install instructions rather than repeating them. (A cross reference in the opposite direction might be good.) One question: is associate the right word to refer to making an ssh connection? Isn't association the act of making a wifi connection to an ssid prior to logging in?

tim-moody:
If the user issues iiab-internet-on while connected to the rpi internal hotspot will that connection be lost when hostapd is restarted in order to use the upstream channel?

georgejhunt:
Yes Tim is correct that iiab-internet-on, as it is currently written, would
need to be reissued after iiab-hotspot-on.  There is also the problem that
iiab-hotspot-on will not chose the correct channel, if the credentials are
not correct for wpa_supplicant.

(in hotspot-on, there is a function which first uses wpa_supplicant to
connect, and fetch the channel in use currently, in order to rewrite
hostapd.conf with that channel).

holta:
The old link works but FYI I've renamed https://github.com/georgejhunt/iiab/wiki/Using-Wifi-during-IIAB-Installation-on-the-Pi-3-and-W to https://github.com/georgejhunt/iiab/wiki/Using-WiFi-during-IIAB-Installation-on-the-Pi-3-and-W, to benefit a far wider demographic.

(GitHub does not allow the official spelling "Wi-Fi" within article title itself...but more official spellings are now used in the text of the article.)

PS unfortunately the article's edit history seems to be lost as a result the above GitHub dash-in-title problem; in general GitHub's support for wiki-editing documentation is extremely minimal :/

holta:
> unfortunately the article's edit history seems to be lost

Thankfully the above is not true.

The article's complete edit history is still available even after title renaming &mdash; by clicking on "revisions" (just under the article's title) and then click on "History" &mdash; which shows all its ~50 edits to date.

holta:
http://FAQ.IIAB.IO Question 16 (["Any other networking tips?"](http://wiki.laptop.org/go/IIAB/FAQ#Any_other_networking_tips.3F)) now links to proposed implementer summary doc: https://github.com/georgejhunt/iiab/wiki/Using-WiFi-during-IIAB-Installation-on-the-Pi-3-and-W

roles/network/templates/network/iiab-internet-on
```diff
@@ -19,7 +19,7 @@ sleep 1
 dhclient wlan0
 
 # may need to change the iptables
-GATEWAY=`route -n | awk '/UG/ { print $8 }'
+GATEWAY=`route -n | awk '/UG/ { print $8 }`
```
holta: @georgejhunt do you mean...
```
GATEWAY=`route -n | awk '/UG/ { print $8 }'`
```

georgejhunt:
Replaced by #697

PR #692, opened 2018-02-28
AWStats icons broken: some confusion about alias name, add the other one

georgejhunt:
### Fixes Bug
icons in awstats broken
rpi3
@tim-moody  said icons were messed up. This change cleaned up the rendered page for me. Still not sure it this was the requested help. -- but it is a direct means of communicating.


tim-moody:
looks good to me. fixes the problem. I would merge

PR #702, opened 2018-03-21
MediaWiki - introduce new role as part of IIAB

tim-moody:
add mediawiki role


tim-moody:
smoke and functional tests passed on deb 9 vm

holta:
Merged as @tim-moody had also tested this on RPi / Raspbian.

PR #708, opened 2018-03-22
/mediawiki corrected/clarified in local_vars.yml

holta:


tim-moody:
I am not really expecting mediawiki to be the home page, though I guess there's nothing to prevent it.

Issue #698, opened 2018-03-14
Ansible fails to enable iiab-cmdsrv on ubuntu-18 -- maybe other os's


georgejhunt:
(I do not know why this is surfacing now)
This is a bug in ansible -- line 251 fails to create a link in /etc/systemd/system/multi-user.target.wants:

[line 251 roles/cmdsrv/tasks/main.yml](https://github.com/iiab/iiab-admin-console/blob/master/roles/cmdsrv/tasks/main.yml#L252-L255):
- name: Enable iiab-cmdsrv service
  service: name=iiab-cmdsrv
           enabled=yes
           state=restarted

root@box:/etc/systemd/system/multi-user.target.wants# ls
apache2.service		  netfilter-persistent.service
atd.service		  NetworkManager.service
avahi-daemon.service	  ondemand.service
bind9.service		  open-vm-tools.service
chrony.service		  pollinate.service
console-setup.service	  pppd-dns.service
cron.service		  remote-fs.target
dhcpd.service		  rsync.service
ebtables.service	  rsyslog.service
iiab-startup.service	  snapd.autoimport.service
irqbalance.service	  snapd.core-fixup.service
isc-dhcp-server6.service  snapd.service
kalite-serve.service	  ssh.service
kiwix-serve.service	  systemd-resolved.service
lxcfs.service		  udev-reload.service
lxd-containers.service	  ufw.service
mariadb.service		  unattended-upgrades.service
ModemManager.service	  vnstat.service
munin-node.service	  wpa_supplicant.service

root@box:/etc/systemd/system/multi-user.target.wants# systemctl status iiab-cmdsrv
● iiab-cmdsrv.service - Provides the IIAB Command Server
   Loaded: loaded (/etc/systemd/system/iiab-cmdsrv.service; enabled-runtime; ven
   Active: inactive (dead)

I broke line 251 ff into two separate plays -- one for enable and one for start -- still failed

Then I  called systemctl directly, and the link was generated:
- name: Enable iiab-cmdsrv service
  shell: systemctl enable iiab-cmdsrv

After re-running /opt/iiab/iiab-admin-console/install (with the above direct call to systemctl)
root@box:/opt/iiab/iiab-admin-console# ls /etc/systemd/system/multi-user.target.wants/
apache2.service		  netfilter-persistent.service
atd.service		  NetworkManager.service
avahi-daemon.service	  ondemand.service
bind9.service		  open-vm-tools.service
chrony.service		  pollinate.service
[...]

Issue #300, opened 2017-09-21
"Install Configured Options" button fails; works from CLI (Ansible 2.2.1 issue?)


holta:
Error from http://box/admin -> Configure tab -> "Install Configured Options" button (when viewed in Utilities tab -> Display Job Status) is:
```
TASK [setup] *******************************************************************
fatal: [127.0.0.1]: FAILED! => {"failed": true, "msg": "failed to transfer file to ~/.ansible/tmp/ansible-
tmp-1506017011.8-22480284034983/setup.py: [Errno 2] No such file or directory: '~/.ansible
/tmp/ansible-tmp-1506017011.8-22480284034983/setup.py'"}
```

Possible Explanations:
https://github.com/ansible/ansible/issues/21562
https://github.com/ansible/ansible/issues/20332

@tim-moody & @jvonau are working very hard on this as we speak :)

holta:
Miracle... @tim-moody & @jvonau seem to have solved this already.

PR forthcoming likely @ https://github.com/iiab/iiab-admin-console

holta:
Ref PR https://github.com/iiab/iiab-admin-console/pull/29 ("use the iiab ansible.cfg rather than default")

holta:
Ref PR #301 ("move ansible tmp file to eliminate ~")

holta:
Ref PR #302 (typo)

holta:
Appears fully fixed during basic functional testing -- and smoked tested by several of us earlier today.

PR #62, opened 2017-06-26
Update ansible.cfg to match upstream

jvonau:
This fixes ansible versions 2.2.1+ & 2.3.0  See: https://github.com/ansible/ansible/pull/1855

PR #711, opened 2018-03-23
3 scripts for IIAB 6.5 & beyond: ansible-2.4.x + ansible-2.5.x + scripts/ansible

holta:
Ansible 2.5.0 smoke-tested on Raspbian Lite 2018-03-13 (medium-sized local_vars.yml) and Ubuntu 16.04 (big-sized local_vars.yml).

Ref's
#669 "is Ansible 2.4.3 running out of memory when installing IIAB 6.5/master on RPi3?"
#698 "Ansible fails to enable iiab-cmdsrv on ubuntu-18 -- maybe other os's"


holta:
Warning: our /opt/iiab/iiab/ansible.cfg is diverging more and more from the official /etc/ansible/ansible.cfg and needs a sanity check in such areas as:
```
system_warnings = False   #let's restore the default (True)
deprecation_warnings = False   #let's restore the default (True)
```
Thanks @jvonau for the warning here:
https://docs.ansible.com/ansible/devel/reference_appendices/config.html

holta:
> Warning: our /opt/iiab/iiab/ansible.cfg is diverging more and more from the official /etc/ansible/ansible.cfg

After thinking about it, @jvonau recommends removing /opt/iiab/iiab/ansible.cfg entirely.  Let's trial this approach (hewing closer to the Ansible community's actual defaults, to keep IIAB config files from bitrotting too far away from mainline usage).

If so: where should IIAB-specific customizations be stored?  Who knows how many of these are critical??  e.g. are things like [Lines 16-17 of /opt/iiab/iiab/ansible.cfg](https://github.com/iiab/iiab/blob/master/ansible.cfg#L16-L17) (pasted in below) still important?
```
remote_tmp     = /opt/iiab/iiab/.ansible/tmp
local_tmp = /opt/iiab/iiab/.ansible/tmp
```

jvonau:
Yes, I agree, that is why in chat I suggested the removal of the file from the git repo and replaced with a readme on ansible's over-rides or maybe auto creation of ansible.cfg from scripts/ansible should iiab even be need to use an override in the future. Less is better IMHO. Leaving the override in place cleans up the feedback but hides bit-rot.

holta:
Should we start fresh, with a new /opt/iiab/iiab/ansible.cfg containing just these 2 lines?
```
remote_tmp = /opt/iiab/iiab/.ansible/tmp
local_tmp = /opt/iiab/iiab/.ansible/tmp
```
(Or is it possible the above are no longer needed?)

jvonau:
That change was for a [bug](https://github.com/ansible/ansible/pull/18551) in ansible that has since been fixed, so I believe no longer needed.

holta:
/opt/iiab/iiab/ansible.cfg → /opt/iiab/iiab/ansible.cfg.deprecated

(Experimental renaming is now part of this PR #711 &mdash; plz all review the contents of https://github.com/iiab/iiab/blob/master/ansible.cfg for any IIAB customizations that we don't want to lose!)

holta:
Ref from July 4, 2017:
PR #62 "Update ansible.cfg to match upstream" (fix for Ansible 2.2.1 & 2.3.0)

jvonau:
https://github.com/iiab/iiab/issues/300 has a reference to https://github.com/iiab/iiab-admin-console/pull/29 that would affect iiab-admin-console, the symlink in {{ cmdsrv_dir }} should be removed also to evaluate its behaviour with this change in place.

holta:
./[runtags](https://github.com/iiab/iiab/blob/master/runtags) does not work with Ansible 2.5.0, likely due to stale syntax in /opt/iiab/iiab/[iiab.yml](https://github.com/iiab/iiab/blob/master/iiab.yml)

We should try to use the newer syntax similar to /opt/iiab/iiab/[iiab-stages.yml](https://github.com/iiab/iiab/blob/master/iiab-stages.yml) ?

holta:
Thanks to @jvonau and @tim-moody who proposed an empty stub /opt/iiab/iiab/ansible.cfg as follows:
```
# Future overrides of /etc/ansible/ansible.cfg belong in this file.
# Used by iiab-admin-console
```
Which is now part of this PR #711 &mdash; while the original /opt/iiab/iiab/ansible.cfg is safeguarded at /opt/iiab/iiab/ansible.cfg.deprecated

holta:
Plz see @jvonau's summary of Ansible 2.4.x & 2.5.x packages/sites @ https://github.com/iiab/iiab/issues/715#issuecomment-377086655

holta:
All 3 scripts tested.

Makes it easier for [lIAB load scripts](http://download.iiab.io/6.5/rpi/) to toggle between Ansible 2.4, 2.5 and 2.6

REMINDER: downgrading (e.g. from Ansible 2.5.0 to 2.4.x pinning) requires that you remove repo lines from:
- /etc/apt/sources.list
- /etc/apt/sources.list.d/ansible*.list

Specifically, lines like the following would need to be removed: (e.g. to downgraded and pin to Ansible 2.4.x)
- deb http://ppa.launchpad.net/ansible/ansible/ubuntu xenial main
- deb http://ppa.launchpad.net/ansible/ansible-2.5/ubuntu xenial main

PR #712, opened 2018-03-25
Ansible 2.5.x breaks ./runtags: include_role to import_role in Stages 3-9?

holta:
This PR restores `./runtags` functionality on Ansible 2.5.0 for PR #711.  (It's one way &mdash; read below for other possible solutions.)

Thanks to @jvonau, who found that attributes (incl tags) are no longer inherited (to tasks within `include_role:`) in Ansible 2.5:

https://docs.ansible.com/ansible/devel/porting_guides/porting_guide_2.5.html#dynamic-includes-and-attribute-inheritance

This PR attempts to use `import_role:` instead of `include_role:` to keep code tight.

Another way to solve this would have been to preserve `include_role:` but encapsulate the actual contents of every role within [`block: ... tags: ACTUALTAG`](http://docs.ansible.com/ansible/latest/playbooks_blocks.html) in each tasks/main.yml [blocks are logical grouping of tasks, introduced by Ansible 2.0].

### DIFF
--- a/roles/3-base-server/tasks/main.yml
+++ b/roles/3-base-server/tasks/main.yml
@@ -4,19 +4,19 @@
   command: echo
 
 - name: HTTPD
-  include_role:
+  import_role:
     name: httpd
   # has no "when: XXXXX_install" flag
   tags: base, httpd
 
 - name: IIAB-ADMIN
-  include_role:
+  import_role:
     name: iiab-admin
   # has no "when: XXXXX_install" flag
   tags: base, iiab-admin
 
 - name: MYSQL
-  include_role:
+  import_role:
     name: mysql
   # has no "when: XXXXX_install" flag
   tags: base, mysql
--- a/roles/3-base-server/tasks/main.yml
+++ b/roles/3-base-server/tasks/main.yml
@@ -1,23 +1,29 @@
 # Base Server
 
 - name: ...IS BEGINNING =====================================
-  command: echo
+  shell: echo
 
-- name: HTTPD
-  import_role:
-    name: httpd
+- block:
+    - name: HTTPD
+      shell: echo
+    - import_role:
+        name: httpd
   # has no "when: XXXXX_install" flag
   tags: base, httpd
 
-- name: IIAB-ADMIN
-  import_role:
-    name: iiab-admin
+- block:
+    - name: IIAB-ADMIN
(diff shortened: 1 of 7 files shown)


holta:
./runtags confirmed fixed (and "./iiab-install --reinstall" smoke-tested) on Ansible 2.5.0 on Ubuntu 16.04.

Both successfully tested on Ansible 2.4.2 on Ubuntu 16.04 too, for backwards compatibility!

PS smoke-testing on Ansible 2.4.2 & 2.5.0 on Raspbian 2018-03-13 looks good (ongoing...)

holta:
Oops swapping `include_roles:` -> `import_roles:` causes Ansible's output to lose explanatory messages like `[9-local-addons : CALIBRE]`.

The best solution I've found (so far, to retain explanatory output) is snippets like the following:
```
- block:
    - name: CALIBRE
      shell: echo
    - import_role:
        name: calibre
  when: calibre_install
  tags: calibre
```
Instead of the original snippets like:
```
- name: CALIBRE
  include_role:
    name: calibre
  when: calibre_install
  tags: calibre
```
e.g. within /opt/iiab/iiab/roles/9-local-addons/tasks/main.yml and similar.

holta:
Am disappointed that pages of "skipping" clutter/slowness have returned to Ansible's output.

Possibly we should consider this proposal instead:

> Another way to solve this would have been to preserve `include_role:` but encapsulate the actual contents of every role within [`block: ... tags: ACTUALTAG`](http://docs.ansible.com/ansible/latest/playbooks_blocks.html) in each tasks/main.yml [blocks are logical grouping of tasks, introduced by Ansible 2.0].

holta:
A possible future Python-based approach (to restructuring ./runtags and other things) that @tim-moody came up with during our live community/team call today @ http://minutes.iiab.io :

#714 “Enhancement - Generate Dynamic Playbook”

holta:

holta:
FYI the newly released [Ansible 2.5.1](https://github.com/iiab/iiab/pull/740#issuecomment-382581867) fixes upstream bug https://github.com/ansible/ansible/pull/36372 &mdash; allowing looping in `include_role:` (should we choose that approach!)

holta:
No longer seems a priority 1.5 years later.

Issue #715, opened 2018-03-26
Pin IIAB 6.5 to Ansible 2.4.x in scripts/ansible ?


holta:
@jvonau do you know how to do this in [scripts/ansible](https://github.com/iiab/iiab/blob/master/scripts/ansible) ?

PS pip within scripts/ansible-2.2.0 -> ansible-2.4.2 -> ansible-2.4.4 (soon) is one option.  But there's prob a better way, in preparation for 2.4.5 backported bug fixes etc.

holta:
Related:
#669 "is Ansible 2.4.3 running out of memory when installing IIAB 6.5/master on RPi3?"
PR #711 "restore scripts/ansible now that Ansible 2.5.0 is released & smoke-tested"
PR #712 "support Ansible 2.5.x: include_role to import_role in Stages 3-9"
#714 "Enhancement - Generate Dynamic Playbook"
PR #716 "install.yml in every core IIAB role - also fixes ./runtags for Ansible 2.5"

jvonau:
Looks like upstream ansible as reworked the repo urls again from
http://ppa.launchpad.net/ansible/ansible/ubuntu
to:
http://ppa.launchpad.net/ansible/ansible-2.5/ubuntu 
http://ppa.launchpad.net/ansible/ansible-2.4/ubuntu 

See and expand the 'Technical details about this PPA' link
https://launchpad.net/~ansible/+archive/ubuntu/ansible-2.5
https://launchpad.net/~ansible/+archive/ubuntu/ansible-2.4

Modifying the respective lines in scripts/ansible should pin you to 2.4 or 2.5 

Preventing ansible from updating from 2.4.2 that is being installed to 2.4.3 until 2.4.5 is released is the issue. I'd ensure that /etc/apt/sources.list does not contain ansible defination and /etc/apt/sources.list.d/ansible*.list does not exist to prevent apt from seeing the ansible repo and thus any available update available from upstream until 2.4.5 is released. 

Personally I feel the out of memory issue uncovered in #669 can be mitigated with a more adequate amount of swap space and 2.5 has shown the same behaviour now.

holta:
Excellent summary above by @jvonau.

If Ansible 2.4.4 is released shortly, this might be best for IIAB 6.5 legacy systems, given the serious [maximum recursion depth](https://github.com/iiab/iiab/pull/716#issuecomment-376406165) issues recently uncovered in 2.4.2

[...]

Issue #714, opened 2018-03-26
Enhancement - Generate Dynamic Playbook - ./runrole instead of ./runtags


tim-moody:
Give a list of desired tags to run as currently done by runtags, generate a dynamic top level playbook to include the vars section, aggregate 0 [AKA [Stage 0](https://github.com/iiab/iiab/tree/master/roles/0-init)] and the role of each passed tag, assuming that tag name and role name are the same.

holta:
Given that Ansible 2.5 has broken ./runtags (neither #712 nor #716 fully resolve this just yet) @tim-moody is also exploring a possibly much simplified replacement command (`./runrole`, that works quite separately from Ansible's tag system).  Perhaps with /opt/iiab/iiab/iiab-run-one-role.yml along these lines:
```
---
- hosts: all
  become: yes

  vars_files:
  - roles/0-init/defaults/main.yml
  - vars/default_vars.yml
  - vars/{{ ansible_local.local_facts.os_ver }}.yml
  - vars/local_vars.yml

  tasks:

  - name: 0-init
    include_role:
      name: 0-init
    tags: 0-init

  - name: Runtag
    include_role:
      name: {{ run_role }}

ansible-playbook -i ansible_hosts iiab-tmp-run.yml --connection=local -e "run_role=$1"
```
If running multiple roles is important, perhaps something like...
```
ansible-playbook -i ansible_hosts run_roles.yml  --connection=local -e "role_list=kiwix,kalite"
```
List's syntax needs fixing, perhaps...
```
role_list=['kiwix','kalite']
```
If multiple roles are even necessary.  Regardless, our 11 aggregate roles (Stage 0-9 and network) can be invoked, using `./runrole 7-edu-apps` etc.

tim-moody:
the multi-role syntax does not work

holta:
@tim-moody reports that his draft `./runrole` currently has a problem getting access to critical vars in [roles/0-init/defaults/main.yml](https://github.com/iiab/iiab/blob/master/roles/0-init/defaults/main.yml)

Reason: [default_vars.yml](https://github.com/iiab/iiab/blob/master/vars/default_vars.yml) are not shared (as if private: True) ?

So apparently `./runrole kiwix` works but `./runrole kalite` does not, lacking access to `is_F18`.

One workaround might be putting all the variables from roles/0-init/defaults/main.yml into default_vars.yml ?

holta:
Let's confirm that the following...
```
  vars_files:
[...]

PR #716, opened 2018-03-27
install.yml in every core IIAB role - also fixes ./runtags for Ansible 2.5

holta:
_This PR #716 is intended to replace PR #712, preserving (1) ./runtags functionality, (2) ./iiab-install efficiency (only runs the playbooks that are needed), and (3) explanatory output ...all with the latest versions of Ansible 2.4.x _and_ 2.5.x &mdash; here's the PR summary in 5 points:_

1. tasks/main.yml was moved to tasks/install.yml across all 37 (of 43) core IIAB roles/playbooks where tasks/install.yml did not already exist.

EXCEPTION 1: a 44th role was also not touched: that's the deprecated [moodle-1.9/moodle](https://github.com/iiab/iiab/tree/master/roles/moodle-1.9/moodle).

EXCEPTION 2: of 55 grand total roles, the 11 aggregate roles (Stage 0 to 9, as well as network) were also/intentionally left untouched &mdash; as ./runtags already works to invoke these: {0-init, 1-prep, 2-common, 3-base-server, 4-server-options, 5-xo-services, 6-generic-apps, 7-edu-apps, 8-mgmt-tools, 9-local-addons, network}.

2. New files tasks/main.yml were created in the above 37 (of 43) core IIAB roles &mdash; intentionally minial 2-line wrappers containing:
```
- import_tasks: install.yml
  tags: xovis   # example! (also: _ was changed to - for tags debian-schooltool and ejabberd-xs)
```
install.yml is not only good practice, but these 37 tags cleanly fix ./runtags on Ansible 2.5, while preserving ./iiab-install's conditional execution across Stage 3-9.  These aggregate roles (.yml for all stages) remain untouched.  They continue to use `include_role` to dynamically/conditionally/efficiently invoke underlying roles, whose [attributes/tags are no longer inherited by include_* as of Ansible 2.5](https://docs.ansible.com/ansible/devel/porting_guides/porting_guide_2.5.html#dynamic-includes-and-attribute-inheritance)).

CLARIF 1: A conditional `when: SOME_ROLE_install` clause was NOT included within the 37 roles' tasks/main.yml files to preserve exact functionality of each of these roles/playbooks (for now).  Such `when:` clauses accomplish nothing (in this case) with Ansible's (static not dynamic) `import_tasks:` especially.  Furthermore, conditional execution of underlying/core roles based on `when:` clauses is already taken care of in Stage 0-9 and network (the 11 aggregate roles) when invoked by ./iiab-install

CLARIF 2: `./runtags` continues to use `iiab.yml` not `iiab-stages.yml` for now.

3. These 6 (of 43) core IIAB roles had pre-existing tasks/install.yml files:
- awstats
- dokuwiki
- kalite
- mediawiki
- teamviewer
- wordpress

So in the above 6 cases, each contains a new file main.yml which calls main2.yml (the original main.yml) &mdash; which in turn calls install.yml &mdash; preserving the exact functionality of the original role/playbook.  These 6 small files (main2.yml) can later be simplified (or possibly eliminated) by merging them into install.yml (or any other) as best suits each individual playbook.

4. The `runtags` bash script was tightened up for readability.

5. Grammar was fixed in roles/1-prep/templates/iiab.env.j2


holta:
Smoke-tested on Raspbian & Ubuntu 16.04 with Ansible 2.5.0

However there's a bug in Ansible 2.4.2 that prevents this valid code from working with big-sized local_vars.yml:
```
TASK [4-server-options : SAMBA] ************************************************
ERROR! Unexpected Exception, this is probably a bug: maximum recursion depth exceeded in cmp
to see the full traceback, use -vvv
```
Ansible 2.4.4 is a likely workaround (when it's released shortly) for those who prefer to stick with Ansible 2.4.x, as suggested by these associated Ansible Issues:

https://github.com/ansible/ansible/issues/23609 "include/include_role: maximum recursion depth exceeded"
https://github.com/ansible/ansible/issues/33528 "maximum recursion depth exceeded in cmp"
https://github.com/ansible/ansible/issues/36053 "Maximum recursion depth exceeded while using include_task"

holta:
Somewhat Related:

#669 "is Ansible 2.4.3 running out of memory when installing IIAB 6.5/master on RPi3?"
#698 "Ansible fails to enable iiab-cmdsrv on ubuntu-18 -- maybe other os's"
PR #711 "restore scripts/ansible now that Ansible 2.5.0 is released & smoke-tested"
PR #712 "Ansible 2.5.x breaks ./runtags: include_role to import_role in Stages 3-9?"
#714 "Enhancement - Generate Dynamic Playbook"
#715 "Pin IIAB 6.5 to Ansible 2.4.x in scripts/ansible ?"

holta:
Bad News: ./runtags does not invoke `include_tasks:` commands within playbooks themselves.

An example is `./runtags kiwix` which runs roles/kiwix/tasks/main.yml which in turn DOES run import_tasks: install.yml (on Ansible 2.5.0, with this PR) _but nested subroutine include_tasks: kiwix_install.yml does not run :-(_

Once again this is because tags (and other attributes) no longer inherit across dynamic include_* as of Ansible 2.5, unless you use static import_*, given Ansible's reinforced security model.

So these low-level `include_tasks:` might all have to be changed over to `import_tasks:` (which means playbooks that are actually run...would run a little slower...following the old model of showing many "skipped" lines within their subroutines).

holta:
@tim-moody wrote:
> so you have reversed the meaning of the flag

Which flag?  (Are you referring to kiwix_src_bin_only -> kiwix_src_contains_bin in PR #717 ?)

holta:

holta:
No longer seems a priority 1.5 years later.

Issue #710, opened 2018-03-23
"Failed to start usbmount@dev-sda.service." during every IIAB boot on RPi


holta:
Error occurs on all boots, after installing IIAB 6.5/master on Raspbian Lite 2018-03-13 on a (regular!) RPi 3, using 1-line install script http://download.iiab.io/6.5/rpi/load.txt

FYI "systemctl start usbmount@dev-sda.service" also fails (after boot).  Inspired by /var/log/boot.log which recommended "systemctl status usbmount@dev-sda.service"

This seems to be a very new IIAB bug arising in recent days (or weeks?)

Somewhat Related:
#682 "automount USB exFAT sticks (& possibly portable NTFS disks too)"
#695 "Option to show entire USB stick/drive content (at http://box/usb)"

holta:
Notes:
- error/warning appears during bootup when a USB stick/drive is inserted
- error/warning does NOT appear when rebooting without a stick inserted
- error/warning does not prevent actual usage of http://box/usb (i.e. USB stick's content is still visible, despite loud red message on boot)

georgejhunt:
Note:
systemctl disable usbmount\@.service makes the error message dissappear

holta:
> `systemctl disable usbmount@dev-sda.service` makes the error message disappear

Is [roles/usb-lib/tasks/main.yml](https://github.com/iiab/iiab/blob/master/roles/usb-lib/tasks/main.yml) the place to make this happen?

holta:
@jvonau experiment shows another method to make red error go away...adding `After=rc-local.service` to the end of [Unit] section in /etc/systemd/system/usbmount@.service so that it appears as follows:
```
[Unit]
BindTo=%i.device
After=%i.device
After=rc-local.service

...
```
Where should this be implemented / in which playbook?

holta:
(Resolved by @jvonau's PR #722 "delay usbmount's actions during boot to be as late as possible")

tkantor81:
Error is still visible.

holta:
TK Kang mentions this is an issue when installing IIAB 7.1 pre-releases on on Debian 10?

PR #720, opened 2018-03-29
disable systemd usbmount

georgejhunt:
### Fixes Bug
Not really a bug -- just annoying boot message

### DIFF
--- a/roles/network/tasks/rpi_debian.yml
+++ b/roles/network/tasks/rpi_debian.yml
@@ -29,6 +29,10 @@
 - name: Stopping services
   include_tasks: down-debian.yml
 
+- name: Disable systemd usbmount which throws error messages
+  shell: systemctl disable usbmount\@.service
+  ignore_errors: True
+
 - name: Reload systemd
   shell: systemctl daemon-reload
 


holta:
@georgejhunt if you have 4 USB sticks inserted, do you know if `systemctl disable usbmount\@.service` acts as a regexp, as if running all 4 commands below?

1. `systemctl disable usbmount@dev-sda.service`
2. `systemctl disable usbmount@dev-sdb.service`
3. `systemctl disable usbmount@dev-sdc.service`
4. `systemctl disable usbmount@dev-sdd.service`

PS this PR mitigates #710

jvonau:
Any reason for this undo function of usb-lib's changes is placed the network section? Seems it would be better to not install the systemd service file in roles/usb-lib/tasks/main.yml's for the rpi's to begin with.

jvonau:
The provided systemd service file might benefit from having an After= present in the unit section to delay the running of this service as not to mount the filesystem too early resulting in read-only. I'll suggest to try After=rc-local.service to delay the mount as late as possible in the boot sequence.

holta:
@jvonau experiment shows another method to make red error go away...adding `After=rc-local.service` to the end of [Unit] section in /etc/systemd/system/usbmount@.service so that it appears as follows:
```
[Unit]
BindTo=%i.device
After=%i.device
After=rc-local.service

...
```

georgejhunt:
Jerry had a better fix

holta:
> Jerry had a better fix

Where should this be implemented / in which playbook?

holta:
Resolved by #722

PR #725, opened 2018-03-29
script to run one role

tim-moody:
also moved global 0-init defaults to vars/default_vars.yml


tim-moody:
./runrole kiwix and ./runrole kalite work on deb 9 vm

holta:
Implements the basic version of #714 ("Enhancement - Generate Dynamic Playbook - ./runrole instead of ./runtags").

holta:
Consider renaming `run-one-role.yml` to `iiab-runrole.yml` to be consistent with the 5 other .yml's in this directory?

jvonau:
Perhaps using import_role in place of include_role for 0-init would be enough to always access 0-init/defaults/main.yml. I really don't like having the starting points for detection routines that are recorded in iiab.ini being exposed in defaults_vars.yml where the impression is that those are available to altered like the other options that are exposed. Those is_* definitions at the bottom of default_vars.yml should really be in 0-init/defaults/main.yml and should really never be altered. Having those variables exposed in what could/does end up being a user editable file is just asking for the variables to be altered with unexpected results. Just my 2 cents worth.

holta:
https://github.com/ansible/ansible/pull/36372 "Fix name parameter templating in include_role module"

tim-moody:
import role doesn't work either see #727

tim-moody:
I prefer this approach

jvonau:
Well I left those variables cause nobody else attempted to clean up the dumping ground in 'defaults' as a reminder for someone to review the related code before moving 1-prep, 2-common related variables as those should be in roles/1|2*/defaults/main.yml as required. I was hoping for a cleanup/removal of the related code (92-rtc-i2c.rules done twice) and the variables would be going away, but that has not happened for some reason. I'd be fine with global detection items being stored apart from the user editable default_vars items like how the os_ver.yml variables are handled, this new file would just be the defaults for the os-ver.yml variables that are now in use. Again I dislike exposing what should not be user editable in default_vars.yml, we have users who copy default_vars to local_vars before starting, now the 'defaults' set to False are overriding what is in the detected os_ver.yml files.

holta:
Separating "untouchable" **constant_vars.yml** from "please don't touch" [default_vars.yml](https://github.com/iiab/iiab/blob/master/vars/default_vars.yml) sounds like a healthy approach.

If this is a good idea, can somebody please suggest exactly what constant_vars.yml should contain!?

tim-moody:
constant variable is an oxymoron I considered and thought better of. internal, working, global would be better names, but I think you will find the division between that and default not to be black and white, so there will be much more discussion. default vars is not supposed to be edited and I opposed copying everything there into local vars.

no matter where a variable is defined a user can override it in local vars. but if you want internal or some such and it is in vars, not a role, I'm ready to put this to rest.

holta:

PR #727, opened 2018-03-31
Rerunrole

tim-moody:
this doesn't work


jvonau:
Failure on 2.4 at kalite where is_F18 is used. That is the reason for the use of the explicit use of default/main.yml for iiab-stages.yml.
 
> TASK [kalite : Calc KA Lite db file name (Fedora 18)] **************************
task path: /opt/schoolserver/iiab/roles/kalite/tasks/main.yml:5
fatal: [127.0.0.1]: FAILED! => {"failed": true, "msg": "The conditional check 'is_F18' failed. The error was: error while evaluating conditional (is_F18): 'is_F18' is undefined\n\nThe error appears to have been in '/opt/schoolserver/iiab/roles/kalite/tasks/main.yml': line 5, column 3, but may\nbe elsewhere in the file depending on the exact syntax problem.\n\nThe offending line appears to be:\n\n\n- name: Calc KA Lite db file name (Fedora 18)\n  ^ here\n"}

While the attached ran through the kalite role on 2.4, I would like to know if this runs on 2.5
[run-one-role2.yml.txt](https://github.com/iiab/iiab/files/1865824/run-one-role2.yml.txt)

Think the finer point of using tasks: vs roles: in the top level play-book comes into the mix. Looks to me of being a case of should 'roles:' be used the defaults are utilized versus being imported/included under 'tasks:' where it appears the values are ignored or just used within the role and are not global.

tim-moody:
So basically defaults must be declared under vars: I don't see having two defaults.

holta:

PR #730, opened 2018-04-03
Prune bitrot from /opt/iiab/iiab/ansible.cfg

holta:
@jvonau & @tim-moody recommend a minimal/clean slate at /opt/iiab/iiab/ansible.cfg &mdash; that does not keep diverging (too far!) from the broader Ansible community's norms (at /etc/ansible/ansible.cfg, that evolves quasi-annually-or-more-often, upon major new releases of Ansible).


holta:
CLARIF- it's intentional that we;re not turning on both of the following, so we hew closer to Ansible's best practices:
```
system_warnings = True
deprecation_warnings = True
```
Rather than red errors of times past, these yellow/orange warnings in Ansible's output serve to keep us honest and informed about problematic features/syntax that are deprecated from upcoming versions of Ansible.

holta:
Smoke-tested on Ubuntu 16.04

PR #732, opened 2018-04-03
Prune bitrot from /opt/iiab/iiab/ansible.cfg

holta:
@jvonau & @tim-moody recommend a minimal/clean slate at /opt/iiab/iiab/ansible.cfg — i.e. a much shorter & more understandable file than we have today &mdash; that does not keep diverging (too far!) from the broader Ansible community's norms (at /etc/ansible/ansible.cfg, that evolves quasi-annually-or-more-often, upon major new releases of Ansible).

CLARIF 1 - It's intentional that we're turning on both of the following, so we hew closer to Ansible's best practices:
```
system_warnings = True
deprecation_warnings = True
```
Rather than red errors of times past, these yellow/orange warnings in Ansible's output serve to keep us honest and informed about problematic features/syntax that will be deprecated from upcoming versions of Ansible.

CLARIF 2 - @jvonau asked me to split this off from PR #711 

CLARIF 3 - This PR is a cleanup of #730 & #731

CLARIF 4 - Smoke-tested on Ubuntu 16.04 and Raspbian Lite.


holta:
Interesting that our original /opt/iiab/iiab/ansible.cfg was actually the *cause* of these 4 warnings/problems:
```
[DEPRECATION WARNING]: [defaults]hostfile option, The key is misleading as it
can also be a list of hosts, a directory or a list of paths , use [defaults]
inventory=/path/to/file|dir instead. This feature will be removed in version
2.8. Deprecation warnings can be disabled by setting deprecation_warnings=False
 in ansible.cfg.
[DEPRECATION WARNING]: DEFAULT_SUDO_EXE option, In favor of Ansible Become,
which is a generic framework. See become_exe. , use become instead. This
feature will be removed in version 2.8. Deprecation warnings can be disabled by
 setting deprecation_warnings=False in ansible.cfg.
[DEPRECATION WARNING]: DEFAULT_MODULE_LANG option, Modules are coded to set
their own locale if needed for screenscraping . This feature will be removed in
 version 2.9. Deprecation warnings can be disabled by setting
deprecation_warnings=False in ansible.cfg.
[DEPRECATION WARNING]: DEFAULT_SUDO_USER option, In favor of Ansible Become,
which is a generic framework. See become_user. , use become instead. This
feature will be removed in version 2.8. Deprecation warnings can be disabled by
 setting deprecation_warnings=False in ansible.cfg.
```
Great news the above purple warnings are now gone from ./iiab-install's Ansible output.

These 4 lines below (from our older /opt/iiab/iiab/ansible.cfg, now put aside as /opt/iiab/iiab/[ansible.cfg.deprecated](https://github.com/iiab/iiab/blob/master/ansible.cfg.deprecated)) were the underlying cause:
```
hostfile       = /etc/ansible/hosts
sudo_user      = root
module_lang    = C
sudo_exe = sudo
```
If there's a need to add values to our newly cleaned out /opt/iiab/iiab/ansible.cfg, remember that headers (such as the following 10) are required:
```
[defaults]
[inventory]
[privilege_escalation]
[paramiko_connection]
[ssh_connection]
[persistent_connection]
[accelerate]
[selinux]
[colors]
[diff]
```

PR #739, opened 2018-04-07
cal-rpi2: refine Calibre install approaches for each OS

jvonau:
reworked calibre

### DIFF
--- a/roles/calibre/defaults/main.yml
+++ b/roles/calibre/defaults/main.yml
@@ -12,5 +12,7 @@ calibre_sample_book: "Metamorphosis-jackson.epub"
 
 calibre_src_url: "https://raw.githubusercontent.com/kovidgoyal/calibre/master/setup/linux-installer.py"
 
-calibre_debs_on_debian: true
 # USE TO TEST debs.yml (RASPBIAN APPROACH!) ON DEBIAN 9.X
+calibre_debs_on_debian: True
+# enable calibre unstable
+calibre_unstable: False
--- a/roles/calibre/tasks/debs.yml
+++ b/roles/calibre/tasks/debs.yml
@@ -1,6 +1,6 @@
 - name: Start by installing OS's Calibre package
   package:
-    name: calibre calibre-bin
+    name: calibre
     state: present
 
 # April 5 2018 raspbian is now tracking the latest Calibre with a small delay.
--- a/roles/calibre/tasks/debs.yml
+++ b/roles/calibre/tasks/debs.yml
@@ -5,13 +5,16 @@
 
 # April 5 2018 raspbian is now tracking the latest Calibre with a small delay.
 # Should you really want the latest just run the below script standalone, it is
-# strongly suggested that waiting for the lastest deb from Rasbian is the best.
+# strongly suggested that waiting for the lastest deb from Raspbian is the best.
 
-- name: Upgrade latest Calibre
+- name: Upgrade to testing Calibre - RPi
+  command: scripts/calibre-install-latest-rpi.sh
+  when: is_rpi
+
+- name: Upgrade to tesing Calibre - non-RPi
   command: scripts/calibre-install-latest.sh
-  #when: not is_rpi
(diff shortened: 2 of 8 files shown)


holta:
If poss let's permit upgrades of Calibre 3.20 on Ubuntu 18.04 (with a flag if necessary, along the lines of http://FAQ.IIAB.IO Item 22?)

OR: we could simply mandate that Calibre's playbook always upgrades to the latest version of Calibre possible (when run online).

FYI this PR replaces PR #738.

holta:
> I'm always against using 'latest' in any of the 4-9 stages as that always requires internet access while the admin-console reuses those plays for simple things like enabling/disabling services.

Internet can often be available in the early months of a deployment, allowing tuning/upgrading of critical apps.  Hence the reason many playbooks have an "upgrade if you can" toggle (documented in http://FAQ.IIAB.IO Item 22).

These 2 opposing use cases are each surprisingly common, roughly summarized as:
- always upgrade this app when there's Internet
- never upgrade this app when there's Internet

jvonau:
I might be coming around, the scripts contain 'apt update' that would force a cache refresh, those should have the when: internet_available conditional applied to them. I'm working out the correct format required for more than one conditional with the newer syntax. Think it would be ok to omit internet_available from the unstable step as that would be a manual edit, and the user should know if the internet is available at the time of editing local_vars. Thoughts?

holta:
> Think it would be ok to omit internet_available from the unstable step as that would be a manual edit, and the user should know if the internet is available at the time of editing local_vars. Thoughts?

My experience (over the past year) is that Ansible is only really used (only really safe to use!) when IIAB's online.

That could certainly change over time (investments in auditing/refactoring?)

One Possible Ideal to shoot for...

"Predictable+Deliberate patterns of behavior (offline reinstalls vs. online upgrades, use cases?) across All mainstream IIAB playbooks."

jvonau:
FYI it looks like Raspbian has removed the Debian keys from apt so the only way to install the latest Calibre on Raspbian is to use their repo, or import Debian the key.
>apt-key list
/etc/apt/trusted.gpg
--------------------
pub   rsa2048 2012-04-01 [SC]
      A0DA 38D0 D76E 8B5D 6388  7281 9165 938D 90FD DD2E
uid           [ unknown] Mike Thompson (Raspberry Pi Debian armhf ARMv6+VFP) <mpthompson@gmail.com>
sub   rsa2048 2012-04-01 [E]

pub   rsa2048 2012-06-17 [SC]
      CF8A 1AF5 02A2 AA2D 763B  AE7E 82B1 2992 7FA3 303E
uid           [ unknown] Raspberry Pi Archive Signing Key
sub   rsa2048 2012-06-17 [E]

pub   rsa4096 2014-06-10 [SC]
      6125 E2A8 C77F 2818 FB7B  D15B 93C4 A3FD 7BB9 C367
uid           [ unknown] Launchpad PPA for Ansible, Inc.

pub   rsa4096 2014-06-13 [SC]
      9FD3 B784 BC1C 6FC3 1A8A  0A1C 1655 A0AB 6857 6280
uid           [ unknown] NodeSource <gpg@nodesource.com>
sub   rsa4096 2014-06-13 [E]

holta:
> FYI it looks like Raspbian has removed the Debian keys from apt so the only way to install the latest
> Calibre on Raspbian is to use their repo, or import Debian the key.
>> apt-key list
>> /etc/apt/trusted.gpg

Should this (or similar) be automated?

jvonau:
Think splitting the routine between debian and raspbian would be the easiest at this point, lets not do what is not needed unless absolutely required. The down side is one has to wait for raspbian to make the package available. The unstable repo would have the package available sooner for those who do not wish to wait, that is where I feel it would be useful to have raspbian import the needed key to enable the unstable repo(debain & ubuntu I believe has the needed key for testing and unstable pre-installed ), that could be added to unstable if really required but is not part of this PR.

holta:
@jvonau tested on RPi.

PR #726, opened 2018-03-30
fix6.5 - auto-enable hostapd (during Ethernet install of IIAB)

georgejhunt:
### Description of changes proposed in this pull request.
Enable hostapd if wan is wired

### Smoke-tested in operating system.
2018-3-13 raspbian with new 4.14 kernel. Hostapd functions without a reboot, and survives a reboot. Also wifi functions headless when SD card is transferred to rpiW.

PR #741, opened 2018-04-07
fix enabling hostapd

jvonau:
on all OSs

### DIFF
--- a/iiab-network
+++ b/iiab-network
@@ -19,11 +19,11 @@ fi
 
 OS="unknown"   # will be overridden below, if /etc/iiab/iiab.env is legit
 if [ -f /etc/iiab/iiab.env ]; then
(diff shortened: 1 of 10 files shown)


jvonau:
Enabling using service: was broken upstream as noted in #726 with 2ba4737abaf8eb2391e89fcf528fc06e89a29c5e adding to the confusion as it is a systemd unit file.

jvonau:
related https://github.com/ansible/ansible-modules-core/issues/3764 where deleting /etc/init.d/foo was listed as a workaround but using systemd: in place of service: should be looked at going forward. I'll test spin using systemd in place of shell: in this chunk of code, as this was a quick look.

holta:
@jvonau tested on Raspbian.  Should work on Ubuntu 16.04 and 18.04.

PR #742, opened 2018-04-09
scripts/ansible* cleanup - see github.com/jvonau/iiab/issues/18

jvonau:
Issue from the skype is reproduced see https://github.com/jvonau/iiab/issues/18


jvonau:
fixes
>Installing --- Please Wait
grep: /etc/lsb-release: No such file or directory

Issue #759, opened 2018-04-28
Moodle 3.1.11 LTS impossible on Ubuntu 18.04? (due to php7.2, even w/ encrypted password)


holta:
BACKGROUND: Moodle 3.1.11 LTS failed to install on Ubuntu 18.04 LTS even after PR #758 fixed "postgresql_version: 10" (was "10.3") in [vars/ubuntu-18.yml](https://github.com/iiab/iiab/blob/master/vars/ubuntu-18.yml)

Clarif: above fixed "systemctl restart postgresql-iiab" for Ubuntu 18.04, after I also deleted /opt/iiab/moodle and data directory /library/pgsql-iiab (per http://FAQ.IIAB.IO "22 Can I upgrade or reinstall server apps?")

SOLVABLE: However Moodle still won't install &mdash; the error is pasted in below...

Our [Moodle playbook](https://github.com/iiab/iiab/tree/master/roles/moodle) may be straightforward, but see also @martignoni's suggestions at https://github.com/iiab/iiab/issues/627#issuecomment-384779601

> TASK [moodle : Create db user] *****************************************************************************************************************************************
An exception occurred during task execution. To see the full traceback, use -vvv. The error was: HINT:  Remove UNENCRYPTED to store the password in encrypted form instead.
fatal: [127.0.0.1]: FAILED! => {"changed": false, "module_stderr": "Traceback (most recent call last):\n  File \"/tmp/ansible_Mc6pNR/ansible_module_postgresql_user.py\", line 855, in <module>\n    main()\n  File \"/tmp/ansible_Mc6pNR/ansible_module_postgresql_user.py\", line 816, in main\n    role_attr_flags, encrypted, expires, conn_limit)\n  File \"/tmp/ansible_Mc6pNR/ansible_module_postgresql_user.py\", line 270, in user_add\n    cursor.execute(query, query_password_data)\n  File \"/usr/lib/python2.7/dist-packages/psycopg2/extras.py\", line 144, in execute\n    return super(DictCursor, self).execute(query, vars)\npsycopg2.NotSupportedError: UNENCRYPTED PASSWORD is no longer supported\nLINE 1: CREATE USER \"Admin\" WITH UNENCRYPTED PASSWORD E'changeme' NO...\n                                 ^\nHINT:  Remove UNENCRYPTED to store the password in encrypted form instead.\n\n", "module_stdout": "", "msg": "MODULE FAILURE", "rc": 1}

@jvonau summarized:
[...]

PR #760, opened 2018-04-28
Ubuntu 18.04's PostgreSQL 10.3 requires Moodle encrypt (hash) its password

holta:
Smoke-tested on Ubuntu 18.04 LTS.

Of course Moodle 3.11.1 still doesn't work for the reasons explained in #759 (php7.2 is more strict, not permitting Moodle versions prior to 3.4)

### DIFF
--- a/roles/moodle/tasks/main.yml
+++ b/roles/moodle/tasks/main.yml
@@ -106,6 +106,7 @@
   postgresql_user:
     name: Admin
     password: changeme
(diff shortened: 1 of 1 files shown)


holta:
Should also be tested on Raspbian clean installs (etc) to be sure this works OK with prior versions of PostgreSQL 9.x

holta:
> Should also be tested on Raspbian clean installs (etc) to be sure this works OK with prior versions of PostgreSQL 9.x

Also smoke-tested successful on the latest Raspbian Lite.

PR #763, opened 2018-04-29
Minor refactor of {iiab-vpn, iiab-remote-off, iiab-remote-on} for readability

holta:
Also iiab-vpn was changed from `#!/bin/sh` to `#!/bin/bash`


holta:
Smoke-tested on Ubuntu 16.04

PR #772, opened 2018-05-02
Remove br0 in Appliance Mode for NetworkManager (e.g. if no internal WiFi)

jvonau:

### DIFF
--- a/roles/network/defaults/main.yml
+++ b/roles/network/defaults/main.yml
@@ -19,8 +19,8 @@ discovered_wired_iface: "none"
 discovered_wireless_iface: "none"
 
 #Redhat
-iiab_wired_lan_iface: "none"
-iiab_wireless_lan_iface: "none"
+#iiab_wired_lan_iface: "none"
+#iiab_wireless_lan_iface: "none"
 has_WAN: False
 has_ifcfg_gw: "none"
 has_wifi_gw: "none"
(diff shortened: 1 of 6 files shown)


holta:
General Clarif: Ubuntu 18.04/Server uses a completely diff networking system (networkd) from 18.04/Desktop (NetworkManager).

FYI Ubuntu 18.04 (as previewed in 17.10) also introduces http://Netplan.io instead of the earlier network configuration utility/package (ifupdown).

holta:
Smoke-tested on a clean install of Ubuntu 18.04/Server.

Plz confirm if this is mergeable, all who can take a look!

* Is a separate PR needed to cure #773's 2MIN hang on every boot?  (@jvonau has a solution here.)
* And/or a PR to disable cloud-init on 18.04/Server?  (If not on 18.04/Desktop too?)  #774

jvonau:
I've added the fix for #773 to this PR, https://github.com/iiab/iiab/pull/772/commits/76e84bec8ba6433b715aa3394f52cdb4300b22f8 has been tested om RPi with no ill results

holta:
Food for thought for the future &mdash; @jvonau & @georgejhunt agree we can't ignore IPv6 forever, but for now: "there is a failure for isc-dhcp-server6.service which is the ipv6 version of dhcpd that should be disabled but does not really affect the operation of our setup..."

PR #780, opened 2018-05-05
correct false positive, numbers need quotes (debuntu wan_in_interfaces)

jvonau:


holta:
What tests are most urgent here?
(If this should be merged into 6.5)

PR #781, opened 2018-05-05
ejabberd-iiab.service 755->644; ejabberd-iiab.init 644->755

holta:

PR #782, opened 2018-05-05
revert to stock ejabberd config

jvonau:
reverts to a stock ejabberd configuration while still providing the "online" group script that would need to be run should the need arise when trying to support really old olpc clients

### DIFF
--- a/roles/ejabberd/tasks/main.yml
+++ b/roles/ejabberd/tasks/main.yml
@@ -6,21 +6,21 @@
   tags:
     - download
 
-- name: Configure ejabberd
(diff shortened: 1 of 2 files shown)


jvonau:
Yes but ejabberd-iiab.service and all related files go away, while ejabberd-iiab.service would need to be manually disabled if already installed previously.

holta:
> Yes but ejabberd-iiab.service and all related files go away, while ejabberd-iiab.service would need to be manually disabled if already installed previously.

I'll try to rebuild #782 on top of #781

Issue #786, opened 2018-05-11
http://box/awstats/awstats.pl fails on 1st use (RPi)


holta:
### Expected behavior and actual behavior.

The error is:

> Error: Command for pipe '/usr/share/awstats/tools/logresolvemerge.pl /var/log/apache2/access.log* |' failed
>
> Setup ('/etc/awstats/awstats.conf' file, web server or permissions) may be wrong.
> Check config file, permissions and AWStats documentation (in 'docs' directory). 

### Steps to reproduce the problem.

- install IIAB 6.5/master onto RPi (medium or big local_vars.yml)
- visit http://box/awstats/awstats.pl
- the top of the page show this is red: `Last Update: | Never updated (See 'Build/Update' on awstats_setup.html page)       Update now`
- click `Update now` (at the very top/right)

Aside: above-mentioned doc /usr/share/doc/awstats/html/awstats_setup.html is most easily viewable at https://awstats.sourceforge.io/docs/awstats_setup.html#BUILD_UPDATE

### Specifications like the version of the IIAB, operating system version, or hardware details.

- RPi3 running latest Raspbian Lite, with ["big" local_vars.yml](http://wiki.laptop.org/go/IIAB/local_vars_big.yml)

holta:
FYI I'm hoping @georgejhunt can investigate these (permissions etc?) on the live RPi3 currently at 10.8.0.18:
```
root@box:~# ls -l /etc/awstats/awstats.conf
lrwxrwxrwx 1 root root 38 May 11 06:36 /etc/awstats/awstats.conf -> /etc/awstats/awstats.schoolserver.conf
root@box:~# ls -l /etc/awstats/awstats.schoolserver.conf
-rw-r--r-- 1 root root 61986 May 11 06:36 /etc/awstats/awstats.schoolserver.conf
```

holta:
Here's hoping that PR #788 (just merged) fixed this issue #786.

(But if not, please re-open as nec!)

PR #788, opened 2018-05-11
recurse into apache log directory with www-data

georgejhunt:
### Fixes Bug
/var/log/apache2 was owned by www-data, but the actual log files were not, because files module does not recurse by default

tested


holta:
Thanks @georgejhunt !

Hopefully this resolves #786?

tim-moody:
If the problem is permissions, why does it work subsequently? If you wait a day, first use also works. Prior to that there is just no data.

PR #790, opened 2018-05-12
RFC netplan systemd-networkd static support (fixed IP address on Ubuntu 18.04/Server)

jvonau:

### DIFF
--- a/roles/2-common/tasks/packages.yml
+++ b/roles/2-common/tasks/packages.yml
@@ -61,6 +61,7 @@
     - bzip2
     - i2c-tools
     - bridge-utils
+    - netmask
     - usbutils
     - hostapd
     - wget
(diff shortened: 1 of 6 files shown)


georgejhunt:
Does it make sense to put an if clause in the template so that the admin console can toggle fixed ip?

Right now, the template is only written if not dhcp. I think it would be better to put the conditional in the template, and then it would work in both directions

jvonau:
Yea was thinking about that, then the hockey game started 'go jets' and I paused for a bit.

holta:
Thank you to @georgejhunt for making this all possible, with the Ubuntu 18.04/Server test machine and ongoing review.

georgejhunt:
I wonder if there is a reason for debian-auto -> patch-auto. I have been tripped up many times when file names or variable names change without reason (worst case is into the ini file).

jvonau:
I created the template as debian-auto with the intent to deploy it like that and Tim suggested patch-auto as the final name.

holta:
PR #791 (that we hope to merge very shortly!) supercedes this, per discussion during our community/team call this morning @ http://minutes.iiab.io

PR #800, opened 2018-05-17
really call hostname.yml (U18 hostname was "ubuntu18base" even after reboot)

jvonau:
not sure where https://github.com/iiab/iiab/commit/db38c9a1d0d392facd32cccc5fb5463b13464094 went

### DIFF
--- a/roles/0-init/tasks/hostname.yml
+++ b/roles/0-init/tasks/hostname.yml
@@ -1,7 +1,7 @@
 - name: Is ubuntu-18 server
   stat:
     path: /etc/cloud/cloud.cfg
-    register: U18_server
+  register: U18_server
 
 - name: edit cloud.cfg yaml
   lineinfile:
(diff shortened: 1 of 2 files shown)


holta:
@tim-moody can you test this in a VM?

(Hostname is already set correctly on my Ubuntu 18.04/Server for whatever reason; whereas hostname remained incorrectly set as "ubuntu18base" on Tim's VMs, even after installation of IIAB and reboot...)

holta:
Context- Ubuntu 18.04 networking is even more complicated than we thought, due to cloud-init etc, as @jvonau confirmed in these articles:

https://websiteforstudents.com/updating-changing-the-hostname-on-ubuntu-18-04-beta-server/

https://askubuntu.com/questions/1028633/host-name-reverts-to-old-name-after-reboot-in-18-04-lts

holta:
Merging so others can test more broadly — @jvonau confirms this was well smoke-tested on sev platforms:

> fresh installs on a RPi won't try to change /etc/cloud/cloud.cfg because of the stat test... I tested the code block on my RPi and your U18-server in /0-init/main.yml to ensure the the edit was skipped/enforced when needed without having to fiddle around with hostname. Tim reported that the PR worked on his problem install

PR #805, opened 2018-05-20
handle systemd-networkd having enabled-runtime

jvonau:

### DIFF
--- a/roles/network/tasks/detected_network.yml
+++ b/roles/network/tasks/detected_network.yml
@@ -28,6 +28,11 @@
     systemd_networkd_active: True
   when: 'ansible_local.local_facts.systemd_networkd == "enabled"'
 
+- name: Setting systemd_networkd-2 results
+  set_fact:
+    systemd_networkd_active: True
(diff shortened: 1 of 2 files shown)


holta:
Deals with issue @tim-moody ran into with non-live .iso installer of Ubuntu 18.04 (Server image not Desktop image) need by our main IIAB demo server.

CONTEXT: @tim-moody could not use the standard live installer b/c it does not support [[1](https://bugs.launchpad.net/ubuntu/+source/subiquity/+bug/1751656), [2](https://bugs.launchpad.net/ubuntu/+source/subiquity/+bug/1750645)] pre-existing partitions:
http://releases.ubuntu.com/18.04/ubuntu-18.04-live-server-amd64.iso

So he is using this one instead:
http://cdimage.ubuntu.com/releases/18.04/release/ubuntu-18.04-server-arm64.iso

Other non-live installers are available here: (as we've used on VM's, and on bare metal in the case of virgin HDD's)
http://cdimage.ubuntu.com/releases/18.04/release/

PR #810, opened 2018-05-21
call /bin/ls to guard against an alias being set

jvonau:


georgejhunt:
when I tried this, it did not fix the problem. This is issue #811

Issue #817, opened 2018-05-25
ZIM full-text search fails on all IIAB 6.5 machines "unless upgraded from Ubuntu 16.04" ?


holta:
I don't understand this serious failure (the inability to use Kiwix's full-text [FT] search feature, even when the ZIM file contains a proven-valid [FT] search index).

This failure appears very widespread on essentially all IIAB 6.5 machines (RPi and Ubuntu 18.04) &mdash; except for one Ubuntu 18.04/Desktop machine that was upgraded from Ubuntu 16.04/Desktop &mdash; this is the only one found to work:

http://10.8.0.22/kiwix/search?content=wiktionary_es_all_novid_2017-08&pattern=canada+

Whereas this same full-text search query fails on all other machines, even when they have the *identical* [kiwix-tools 0.5.0](http://download.kiwix.org/release/kiwix-tools/) and the *identical* ZIM http://download.kiwix.org/zim/wiktionary/wiktionary_es_all_novid_2017-08.zim &mdash; all verified to be bit-for-bit identical using sha256sum &mdash; and yet this seach query still fails: 

http:// iiab.me /kiwix/search?content=wiktionary_es_all_novid_2017-08&pattern=canada+

Those results being:

> Not Found
>
> There is no article with the title "canada " and the fulltext search engine is not available for this content.

holta:
CORRECTION: the above bug report described the problem erroneously, BUT kiwix-tools 0.5.0 DOES FAIL to support full-text search with ZIM's that contain an internal index.

EVIDENCE shows (so far) that updating it to http://download.kiwix.org/nightly/2018-05-24/ fixes the problem...testing is ongoing and will likely require a PR/fix to IIAB 6.5 in coming hours...

PR #819, opened 2018-05-25
kiwix-tools 0.5.0 (2018-04-23) -> 2018-05-24 fixes FT search w/ internal index

holta:
SEE #817 -> PR #818 (emergency patch applied to release-6.5) to fix full-text searching with ZIM files that contain an internal index.


holta:
Tested on 5 OS's as per https://github.com/iiab/iiab/pull/818#issue-190482626

Issue #824, opened 2018-05-31
http://box[.lan] fails on iPhones "if Fixed IP set on RPi3's eth0"


holta:
All 3 experienced IIAB/iPhone 7 and iPhone SE users were completely stumped today, because the WiFi connection was no longer appearing on the top-left of their iPhone screens (also "No Internet Connection" in yellow, and an exclamation point on the right, See Screenshot Below!)

The usual tricks all failed:

- http://box/ and http://box.lan/

- Forcing the iPhones to "Forget" the WiFi SSID

- Rebooting the iPhones

The only workaround was to force each iPhone to http://172.18.96.1 and then make a "Favorite" of that in the iPhone's Safari browser.  Naturally this is Not a satisfactory solution for new users — and it's even more frustrating for experienced iPhone users who never had trouble with http://box/ in the past.

We'll try to investigate more...not quite sure how...but getting more urgent as all iPhone users are now blaming IIAB......until we can explain better why iPhones are no longer working with IIAB (at today's school anyway!)

FYI, whether this is related or not: today IIAB had a Fixed IP address set by Admin Console -> Save Configuration -> hand fix /etc/dhcpcd.conf to include "interface eth0" instead of "interface none" -> cd /opt/iiab/iiab -> ./iiab-network

![img-20180530-wa0014](https://user-images.githubusercontent.com/2458907/40760900-914acc46-645e-11e8-9388-379c4fc2abff.jpg)

Theory Corroborated:

When the same iPhone is used with another IIAB (where a Fixed IP address has NOT been set) http://box/ suddenly comes back to life — see above screenshot — showing an iPhone's normal WiFi settings when attached to IIAB (compare to earlier screenshot at top!)

Conclusion: IIAB's internal WiFi hotspot appears to behave very differently with iPhones, after a Fixed IP address is set within IIAB on eth0.

(Any ideas why?)

![img-20180530-wa0016](https://user-images.githubusercontent.com/2458907/40760912-a0f0206a-645e-11e8-971f-7c49e07e55b3.jpg)

holta:
[...]

PR #825, opened 2018-05-31
Fallback to old gateway info when WAN is not detected 

jvonau:
Fallback to old gateway info when WAN is not detected when configuring static ip. Otherwise WAN device becomes 'none'

### Fixes Bug
Adam's static setup while offline

### DIFF
--- a/roles/4-server-options/tasks/main.yml
+++ b/roles/4-server-options/tasks/main.yml
@@ -16,10 +16,10 @@
   when: openvpn_install
   tags: openvpn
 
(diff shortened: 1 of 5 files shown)


holta:
Partially Related to #824 (http://box[.lan] fails on iPhones "if Fixed IP set on RPi3's eth0")

holta:
@georgejhunt & @tim-moody plz review this so we can merge during-or-prior-to Thursday's community call (http://minutes.iiab.io), Thanks!

PR #837, opened 2018-06-14
Fix PR #825 (dnsmasq / fallback to old gateway info when WAN is not detected)

holta:


holta:
Fixes PR #825, we hope!  (Smoke-tested on Raspbian Lite on an original RPi 3 B...)

PR #836, opened 2018-06-14
MediaWiki 1.30.0 -> 1.31.0 LTS

holta:

PR #844, opened 2018-06-20
Fix LocalSettings.php [and auto-generate it during MediaWiki installs]

tim-moody:
generated local settings has server param due to --server supplied to maintenance/install.php
this led to quick fix of supplying hardcoded LocalSettings which is not compatible with version 1.31
this tries to fix all that


tim-moody:
passes smoke and functional tests

holta:
@tim-moody: Tim, should custom settings in LocalSettings.php be auto-saved in a generic/nearby place (like /library/mediawiki-1.31.0/LocalSettings.php.old etc) instead of obliterated?  In case MediaWiki is being reinstalled, as is bound to happen eventually?

Aside: this PR #844 was tested on Ubuntu 18.04/Server.  More/general details on /library/mediawiki-1.31.0/LocalSettings.php at #836 ("MediaWiki 1.30.0 -> 1.31.0 LTS").

tim-moody:
rerun and see if it installs with local settings still there

holta:
> rerun and see if it installs with local settings still there

"./runrole mediawiki" preserves /library/mediawiki-1.31.0/LocalSettings.php if it already exists &mdash; and generally fails to complete MediaWiki's (re)install.

So I've changed http://FAQ.IIAB.IO Item 22 ("Can I upgrade or reinstall server apps?") asking at-your-own-risk upgraders / reinstallers to first delete /library/mediawiki-ACTUAL.VERSION.NUMBER/LocalSettings.php

PR #845, opened 2018-06-20
kiwix-tools 2018-05-24 -> 0.6.0 (2018-06-20)

holta:
Please test on all 3 platforms, those who can!

- armhf
- x86_64
- i586

Context: all 3 installers are now cached at http://download.iiab.io/packages/ (and also available at https://download.kiwix.org/release/kiwix-tools/), including kiwix-serve which is the piece we care about most :)

### DIFF
--- a/roles/kiwix/defaults/main.yml
+++ b/roles/kiwix/defaults/main.yml
@@ -1,9 +1,9 @@
 # Which kiwix-tools to download from http://download.iiab.io/packages/
 # As obtained from http://download.kiwix.org/release/kiwix-tools/ or http://download.kiwix.org/nightly/
 
-kiwix_version_armhf: "kiwix-tools_linux-armhf-2018-05-24"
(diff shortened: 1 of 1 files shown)


holta:
Smoke-tested on 2 different Ubuntu 18.04/Server machines...

While not fatal (http://box/kiwix/ and http://box:3000/kiwix/ do work) it's unfortunate that the classic/original URL http://box:3000 suddenly now fails with message:
```
The requested URL "/" was not found on this server.
```
_(This worked with Kiwix 2018-05-24...possibly this might be restored in a future version of Kiwix?)_

holta:
> unfortunate that the classic/original URL http://box:3000 suddenly now fails

@tim-moody does http://box:3000/kiwix/ avoid the proxy, for those who want to research delicate topics without leaving a data trail as http://box/kiwix/ does?

tim-moody:
looks like it

PR #847, opened 2018-06-20
scripts/ansible needs dirmngr only; prereqs not needed: python-pip python-setuptools python-wheel patch

holta:
Extensively tested over past 6 months.  Time to set it loose even further, so it gets tested in yet more scenarios as part of master.


holta:
Ref: PR #843

holta:
Testing on Debian 9.x especially would be appreciated!

jvonau:
The python packages are actually ansible requirements that may not be installed along with ansible due to apt's --recommended flag making these dependencies optional but we need then later anyway. However  those are installed in packages.yml so six of one a half dozen the other. Better to keep the dependencies together IMHO.

PR #848, opened 2018-06-21
Calibre 3.26.0 permits revert of PR833 & PR839

holta:

PR #839, opened 2018-06-18
Latest Calibre(s) insertable-tho-not-runnable in RPi Zero W: use pkgs from Calibre 3.23

holta:

Issue #835, opened 2018-06-12
Pi Zero [W] not working with IIAB [if Calibre 3.24+ installed via Debian-flavored .deb's]


Piflyer:
### Expected behavior and actual behavior.
To have it boot up to Raspbian with iiab in it
### Steps to reproduce the problem.
Download on a Raspberry Pi 3 B and transfer it to a Pi zero. It was in ‘panic mode’ and I lot would kill init.

tim-moody:
I do this regularly without any problems, but generally with small installs. Also, I have not yet tried 3 B+.

holta:
@kananigit can you please comment on your own experience today building up IIAB 6.5 within RPi 3 (or 3 B+?) to create microSD's that you then insert into RPi Zero W?

PS this has worked reliably for me in recent months.  I would like @Piflyer to provide photos of the error(s) in question, excerpt from systemd logs (using the journalctl command) and context as to what exact local_vars.yml he/she is using from http://download.iiab.io/6.5

Piflyer:
![79c3877c-ce42-40c2-87ee-28a70c21a017](https://user-images.githubusercontent.com/37424588/41322608-ebcb102c-6e77-11e8-8177-5ae5d359ea84.jpeg)

Piflyer:
It said, `[ 1.187737] Kernel panic - not syncing: Attempted to kill init! exit code 0x0000000b`

holta:
@jvonau : somehow @Piflyer 's screen photo from yesterday isn't posted above, so am re-posting — any ideas?

![41322608-ebcb102c-6e77-11e8-8177-5ae5d359ea84](https://user-images.githubusercontent.com/2458907/41355393-66e15dd0-6eef-11e8-9aa7-2e09b116889f.jpeg)

tim-moody:
I guess we need local_vars.yml to see what might be installed to cause this. Is this during fsck?

holta:
>  we need local_vars.yml to see what might be installed to cause this

Yes: @Piflyer please post your /opt/iiab/iiab/vars/local_vars.yml

Clarif: this is usually derived from _{min, medium, or big}_ local_vars.yml here...

http://wiki.laptop.org/go/IIAB/local_vars.yml

holta:
@Piflyer please RE-confirm whether you've prepared the microSD in a:
- 3 B (RPi released 2016-02-29)
or a...
- 3 B+ (RPi released 2018-03-14)

Thanks!

[...]

PR #834, opened 2018-06-10
install Calibre via Debian .debs insteads of Raspbian .debs (fixes 831)

holta:
This was tested on Raspbian Lite to fix #831

This PR834 (identical to PR #833) had been a stopgap solution for MEDIUM-sized and BIG-sized IIAB 6.5 (originally released on 2018-05-24) that allowed the installation Calibre 3.24 and 3.25 despite its broken Raspbian-flavor .debs

POSTSCRIPT: this was NOT a great solution, as it prevented the resulting IIAB microSD cards from booting inside RPi Zero W's.

PR #856, opened 2018-06-27
Avoiding OS's ancient version(s) of Calibre (blocks install on Ubuntu 16.04, slows others)

holta:
Tested to work on Raspbian Lite on RPi 3 B+, whose microSD was later moved to Zero W.

All other OS's use the [Python approach](https://github.com/iiab/iiab/blob/master/roles/calibre/tasks/py-installer.yml) (not the [.deb approach](https://github.com/iiab/iiab/blob/master/roles/calibre/tasks/debs.yml) used on RPi) to installing Calibre.

### DIFF
--- a/roles/calibre/tasks/main.yml
+++ b/roles/calibre/tasks/main.yml
@@ -5,14 +5,15 @@
     path: "/usr/bin/calibre"
   register: calib_executable
 
-- name: Install Calibre via OS's package installer (IF /usr/bin/calibre MISSING)
-  package:
-    name: "{{ item }}"
-    state: latest
-  with_items:
-    - calibre
-    - calibre-bin
-  when: internet_available and (not calib_executable.stat.exists)
+# The 8 lines below appear unecessary as of 2018-06-26 -- as tested on RPi.
+#- name: Install Calibre via OS's package installer (IF /usr/bin/calibre MISSING)
+#  package:
+#    name: "{{ item }}"
+#    state: latest
+#  with_items:
+#    - calibre
+#    - calibre-bin
+#  when: internet_available and (not calib_executable.stat.exists)
 
 - name: Install Calibre experimental .debs IF calibre_via_debs (AND /usr/bin/calibre WAS MISSING)
   include_tasks: debs.yml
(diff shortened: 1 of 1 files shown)


holta:
Builds On:

PR #848, PR #846, PR #839, #835, PR #834, PR #833, #831, #830

holta:
Clarif from this morning's community/team call (http://minutes.iiab.io):

Even though this is tested to work on RPi, @jvonau is concerned that older OS's especially (e.g. Ubuntu 16.04) likely still require the OS's own very stale version of Calibre as a prerequisite, in order to bring in all graphical libraries/dependencies.

Even if the evidence (as of June 2018) shows that Raspbian Stretch and Ubuntu 18.04 don't currently require the OS's own version of Calibre.

While I'm not too concerned about Ubuntu 16.04 now that most all are migrating to mainline Ubuntu 18.04, still it sounds like this PR is more appropriate for later (possibly even as late as mid-2019, when even more OS's, especially Debian 10/derivatives including an overhauled Raspbian around mid-2019) will presumably have solved this even more cleanly with a very modern version of Calibre OOTB.

holta:
Deferred this PR from IIAB 6.6 to IIAB 6.7 as we continue to monitor Calibre's Install Resilience going into 2018/2019 &mdash; quite likely this PR will be very safe to merge around the time of IIAB 7.0 but we'll keep an eye on it then...

holta:
It's possible this PR #856 should be reconsidered in the short term, as Ubuntu 16.04 now fails to install "its very own" Calibre 2.55 (from 2.5 years ago) due to recent library/dependency/security apt updates.

holta:
Fixes Ubuntu 16.04 install, where master is currently broken due to #883

Smoke-tested "./runrole calibre" on Raspbian Lite and Ubuntu 18.04 &mdash; all tests successful &mdash; both with and without the existence of trigger file /usr/bin/calibre

holta:
CLARIF: this is surely not the final word, but this allows IIAB 6.6/master to once again work on Ubuntu 16.04, and will allow us to refine any crucial dependencies/logic for mainline OS's going forward.

PR #862, opened 2018-06-29
Remove Travis tag

m-anish:


holta:
@arky please let us know if there's a better way to mitigate false alarms from Travis, Thanks!

Issue #830, opened 2018-06-08
Enable Calibre adds/removes/metadata edits using /library/calibre/users.sqlite


holta:
In addition to Admin/changeme, a generic read-only username/password (e.g. box/box) is also necessary...unless we (@darkenvy?) also get calibre-web front-end to work (#816) !

1. Create the users/passwords in /library/calibre/users.sqlite by running:
```
systemctl stop calibre-serve
calibre-server --manage-users --userdb /library/calibre/users.sqlite
```
2. Remember to remove read/write permissions from student accounts like box/box or mx/mx etc, using:
> 2) Edit an existing user
> 3) Change read/write permission for [ACTUAL ACCOUNT NAME]

3. Unit file /etc/systemd/system/calibre-serve.service must contain:

> ExecStart=/usr/bin/calibre-server --daemonize --log=/var/log/calibre.log --pidfile=/var/run/calibre.pid --port=8080 --enable-auth --userdb=/library/calibre/users.sqlite /library/calibre

4. Should be obvious, but before testing http://box:8080 you need to have run:
```
systemctl daemon-reload
systemctl start calibre-serve
```
5. Also remember that most browsers REMEMBER Calibre's username/password.  So you need to fully close your browser AND forcefully flush cached password(s) from your browser, if changing to/from teacher credential (typically Admin/changeme) from/to student credential (typically box/box or mx/mx or whatever...)

holta:
Background Details @ #747 ("Automate installing Calibre username/password for students & teachers")

holta:
Parametrizing Calibre's port (8080 by default, but conflicts will inevitably arise) summarized under: https://github.com/iiab/iiab/issues/660#issuecomment-361425259

holta:
SEE #529 'Calibre 2.x images & Calibre 3.x books don't appear @ http://box/books (http://box:8080 works!)'

holta:
> Parametrizing Calibre's port (8080 by default, but conflicts will inevitably arise) summarized under: https://github.com/iiab/iiab/issues/660#issuecomment-361425259

Workaround fix port in /library/www/html/iiab-menu/config.json summarized in more detail here:

https://github.com/iiab/iiab-menu/issues/4#issuecomment-397791038

Best case we're hoping https://github.com/iiab/iiab-menu can be made to read directly from default_vars.yml and local_vars.yml

holta:
Initial stub /library/calibre/users.sqlite will contain...

[...]

PR #879, opened 2018-07-07
Calibre with logins: implements #830 so teachers can add/remove/convert books & edit their metadata

holta:
Implements #830


holta:
"./runrole calibre" successfully tested on Ubuntu 18.04 and Raspbian Lite, with and without the existence of trigger file /usr/bin/calibre (its absence forces installation).

Remember that /library/calibre/users.sqlite is just a stub on installation, that implements these 1 teacher account + 7 student usernames (for Haiti, Mexico, Peru) as listed @ https://github.com/iiab/iiab/issues/830#issuecomment-403236981

Issue #867, opened 2018-07-02
[Debian 9] calibre-serve Service Fails to start: Missing arguments "--log" "TimeoutStartSec'


arky:
The 'Calibre-Serve' services fails to start with multiple warnings. 

$ calibre-server --version
calibre-server (calibre 2.75.1)

TASK [calibre : Remove calibre.conf link if disabled (debuntu)] ****************
skipping: [127.0.0.1]

TASK [calibre : Enable Calibre service -- runs calibre-server by Kovid Goyal] ***
fatal: [127.0.0.1]: FAILED! => {"changed": false, "msg": "Unable to start service calibre-serve: Job for calibre-serve.service failed because of unavailable resources or another system error.\nSee \"systemctl status calibre-serve.service\" and \"journalctl -xe\" for details.\n"}
	to retry, use: --limit @/opt/iiab/iiab/run-one-role.retry

PLAY RECAP *********************************************************************
127.0.0.1                  : ok=26   changed=6    unreachable=0    failed=1   

$ sudo service calibre-serve status
● calibre-serve.service - Provides the Calibre Book Server
   Loaded: loaded (/etc/systemd/system/calibre-serve.service; enabled; vendor preset: enabled)
   Active: failed (Result: exit-code) since Mon 2018-07-02 12:26:42 EDT; 2min 25s ago
  Process: 4672 ExecStart=/usr/bin/calibre-server --daemonize --log=/var/log/calibre.log --pidfile=/var/run/calibre.pid --port=8080 /library/calibre (

Jul 02 12:26:41 box.lan calibre-server[4672]: allows you to browse you calibre library by categories. You can also
Jul 02 12:26:41 box.lan calibre-server[4672]: access an interface optimized for mobile browsers at /mobile and an
Jul 02 12:26:41 box.lan calibre-server[4672]: OPDS based interface for use with reading applications at /opds.
Jul 02 12:26:41 box.lan calibre-server[4672]: The OPDS interface is advertised via BonJour automatically.
Jul 02 12:26:41 box.lan calibre-server[4672]: Whenever you pass arguments to calibre-server that have spaces in them, enclose the arguments in quotati
Jul 02 12:26:41 box.lan calibre-server[4672]: calibre-server: error: no such option: --log
Jul 02 12:26:42 box.lan systemd[1]: calibre-serve.service: Control process exited, code=exited status=2
Jul 02 12:26:42 box.lan systemd[1]: Failed to start Provides the Calibre Book Server.
[...]

PR #881, opened 2018-07-09
/library/calibre must be created prior to placing users.sqlite in there

holta:
Fixes: #879

Related: #867


holta:
Tested on Debian 9.4

Issue #869, opened 2018-07-02
[Debian 9] Sugarizer.service: Failed to parse output specifier, ignoring: syslog


arky:
Seems to be harmless parsing error. Filing for documentation purposes only. 

https://github.com/iiab/iiab/blob/b04a6147e960f7e4ec6e7e9454cd46389146ff87/roles/sugarizer/templates/sugarizer.service.j2#L12
https://github.com/iiab/iiab/blob/b04a6147e960f7e4ec6e7e9454cd46389146ff87/roles/sugarizer/templates/sugarizer.service.j2#L13

Errors
Jul 02 13:10:37 box.lan systemd[1]: [/etc/systemd/system/sugarizer.service:1] Unknown section 'unit'. Ignoring.
Jul 02 13:10:37 box.lan systemd[1]: [/etc/systemd/system/sugarizer.service:10] Failed to parse output specifier, ignoring: syslog               # Output to syslog
Jul 02 13:10:37 box.lan systemd[1]: [/etc/systemd/system/sugarizer.service:11] Failed to parse output specifier, ignoring: syslog                # Outut to syslog
Jul 02 13:10:37 box.lan systemd[1]: apt-daily-upgrade.timer: Adding 20min 11.919953s random time.
Jul 02 13:10:37 box.lan systemd[1]: apt-daily.timer: Adding 10h 56min 48.251582s random time.
Jul 02 13:18:36 box.lan systemd[1]: Reloading.
Jul 02 13:18:36 box.lan systemd[1]: [/etc/systemd/system/sugarizer.service:1] Unknown section 'unit'. Ignoring.
Jul 02 13:18:36 box.lan systemd[1]: [/etc/systemd/system/sugarizer.service:10] Failed to parse output specifier, ignoring: syslog               # Output to syslog
Jul 02 13:18:36 box.lan systemd[1]: [/etc/systemd/system/sugarizer.service:11] Failed to parse output specifier, ignoring: syslog                # Output to syslog
Jul 02 13:18:36 box.lan systemd[1]: apt-daily.timer: Adding 6h 30min 40.149135s random time.
Jul 02 13:18:36 box.lan systemd[1]: apt-daily-upgrade.timer: Adding 8min 50.496711s random time.
Jul 02 13:22:05 box.lan systemd[1]: Reloading.
Jul 02 13:22:05 box.lan systemd[1]: [/etc/systemd/system/sugarizer.service:10] Failed to parse output specifier, ignoring: syslog               # Output to syslog
[...]

PR #851, opened 2018-06-21
http://box/sugarizer fix on RPi & All OS's?  (Sugarizer 1.0+)

georgejhunt:
1. opens port 8089 on ipv4 as well as ipv6. -- added spec for ipv4 to server.listen() in sugarizer.js
2. Our earlier testing should have involved "iiab-network" in order to change the iptables rules.
3. These changes pass smoke test "iiab-install --reinstall"
3.5 http://box.lan/sugarizer should work if associated with the wifi, but http://192.168.0.1/sugarizer will not-- because apache proxies to http://box.lan/ which will not resolve.
4. I noticed that sugarizer is installed, even when local_vars.yml says install=False. But I don't understand it because https://github.com/iiab/iiab/blob/master/roles/7-edu-apps/tasks/main.yml lines 36 and following: 
```<quote>
- name: SUGARIZER
  include_role:
    name: sugarizer
  when: sugarizer_install
  tags: sugarizer
<unquote>
```
I did not look into this problem.  It may be related to the reorganization of states that has occurred.

PR #814, opened 2018-05-23
Sugarizer v1.0 migration

georgejhunt:
### Description of changes proposed in this pull request.
This PR upgrades to the most recent sugarizer [implementing #798]

### Smoke-tested in operating system.
2018-04-18 stretch lite. 

Known issues: the proxy from apache /sugarizer to localhost:8089 does not work. But then the kiwix, and kalite proxies do not work either.  I remember that I had them working once. But I've lost the magic sauce.

Tim's menu system goes directly to the desired port, so maybe it does not matter.

PR #882, opened 2018-07-09
Sugarizer: fixes part of #869 (comments must be on LEFT, in systemd unit files)

holta:
See #814, #851, and the immediate bug / circumstances: https://github.com/iiab/iiab/issues/869#issuecomment-403401145


holta:
Tested to fix https://github.com/iiab/iiab/issues/869#issuecomment-403401145 on Debian 9.4

PR #897, opened 2018-07-13
Correct URI for IIAB license file

arky:
Correct the URI  http://github.com/XSCE/iiab/blob/master/LICENSE to http://github.com/XSCE/iiab/blob/master/LICENSE.md

### DIFF
--- a/roles/httpd/files/html/credits.html
+++ b/roles/httpd/files/html/credits.html
@@ -38,7 +38,7 @@
 
   It is licensed under the terms of the GNU Library General Public License as published by the Free Software Foundation; either version 2 of the License, or (at your option) any later version.<br><br>
 
-  Licensing information may be found at  <a href="http://github.com/XSCE/iiab/blob/master/LICENSE">github.com/XSCE/iiab/blob/master/LICENSE</a>.<br>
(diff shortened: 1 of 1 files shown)


tim-moody:
why not https://github.com/iiab/iiab/blob/master/LICENSE.md

tim-moody:
also, I don't think that credits.html is used anywhere.

rather https://github.com/iiab/iiab-menu/blob/master/menu-files/html/credits.html

holta:
> also, I don't think that credits.html is used anywhere.
>
> rather https://github.com/iiab/iiab-menu/blob/master/menu-files/html/credits.html

Somehow I missed Tim's above comment.

In any case, this credits.html that @arky found is now more up-to-date:
https://github.com/iiab/iiab/blob/master/roles/httpd/files/html/credits.html

@tim-moody should this new one be moved to https://github.com/iiab/iiab-menu/blob/master/menu-files/html/credits.html

(And deleted from roles/httpd/files/html if so?)

tim-moody:
they look identical to me

tim-moody:
one needs to be removed

holta:
> they look identical to me

https://github.com/iiab/iiab/blob/master/roles/httpd/files/html/credits.html has a dozen improvements and corrections I've added over the last hour.

If poss @tim-moody can you copy/move this to https://github.com/iiab/iiab-menu/blob/master/menu-files/html/credits.html ?

Then we should delete roles/httpd/files/html/credits.html

PR #918, opened 2018-07-17
Fixes #917 [Kolibri systemd service says “active (exited)” instead of “active (running)”]

### LINKED ISSUE #917: Kolibri systemd service says “active (exited)” instead of “active (running)”
This message appears on all mainline IIAB Linux OS's (Raspbian, Ubuntu 18.04, Debian 9.5) &mdash; upon running `systemctl status kolibri`, which responds with `active (exited)` in green &mdash; rather than `active (running)` in green, as expected of systemd services in general.

@arky, @jvonau & @georgejhunt would you know how to fix Kolibri's systemd unit file ([/etc/systemd/system/kolibri.service](https://github.com/iiab/iiab/blob/master/roles/kolibri/templates/kolibri.service.j2)) to make its status appear as `active (running)` ?  This should not only help it play well with others (other systemd services) but also help with logging/restarting correctly etc, per systemd standards.

holta:
@arky this changes "Type=oneshot" to "Type=forking" in systemd unit file &mdash; thanks much to @jvonau.

Tested on Raspbian Lite.

Please test on Ubuntu 18.04 and Debian 9.5, Thanks All!

### DIFF
--- a/roles/kolibri/templates/kolibri.service.j2
+++ b/roles/kolibri/templates/kolibri.service.j2
@@ -2,7 +2,7 @@
 Description=Kolibri
 
 [Service]
-Type=oneshot
+Type=forking
 RemainAfterExit=yes
 Environment=KOLIBRI_USER={{ kolibri_user }}
 Environment=KOLIBRI_HOME={{ kolibri_home }}


arky:
@holta Kudos!

Issue #948, opened 2018-07-22
Calibre 3.27.1 (Raspbian .deb's released today) fails to install


holta:
Very unfortunate that our mid/late June workaround (https://github.com/iiab/iiab/blob/master/scripts/calibre-install-packages.sh from PR #839) does not appear to work this time.

More Details @ https://www.mobileread.com/forums/showthread.php?p=3729117#post3729117

Thanks @jvonau / all for any ideas towards installing `python-pyqt5 (>= 5.11.2+dfsg-1)` in any other way?

Even despite the following:
```
root@box:~# echo "deb http://raspbian.raspberrypi.org/raspbian/ testing main" > /etc/apt/sources.list.d/rpi-testing.list
root@box:~# apt update
...
768 packages can be upgraded. Run 'apt list --upgradable' to see them.
root@box:~# apt -a list python-pyqt5
Listing... Done
python-pyqt5/testing 5.10.1+dfsg-2+rpi1 armhf [upgradable from: 5.7+dfsg-5]
python-pyqt5/stable,now 5.7+dfsg-5 armhf [installed,upgradable to: 5.10.1+dfsg-2+rpi1]
```

PR #949, opened 2018-07-22
Calibre 3.27.1 glitch so remove from MEDIUM-sized local_vars.yml

holta:
See #948


holta:
Ideally this can be reverted within days...if the almighty (or we) find a Raspbian workaround/fix for Calibre 3.27 or 3.28! (#948)

PR #966, opened 2018-07-24
Label unused files as .deprecated in /opt/iiab/iiab

holta:
Make /opt/iiab/iiab more readable and understandable to newcomers.


holta:
There are likely others that should be deprecated to make IIAB's code base more approachable, but this is a start.

PR #968, opened 2018-07-25
kiwix proxypassreverse

georgejhunt:
Smoke tested. At first http://box.lan/kiwix did not work. But then it did. May have been a startup, or cache issue. 

Always worked at http://10.10.123.206/kiwix or  http://10.10.123.206/kiwix/


tim-moody:
what is the purpose of this PR, that is what problem does it intend to solve?

georgejhunt:
Sorry, I had not read enough into the issues. I don't think this solves anything

PR #1000, opened 2018-08-15
OpenVPN resiliency improvements

holta:
Implements & refines #995 &mdash; enabling variable `openvpn_handle` in /etc/iiab/local_vars.yml to customize /etc/iiab/openvpn_handle &mdash; by building on PRs #996, #997, #998.

[Testing confirmed below.]

PR #998, opened 2018-08-14
OpenVPN announcer fixes/cleanup

holta:
Builds on PRs #996 & #997

PR #996, opened 2018-08-13
Use openvpn_handle variable to set /etc/iiab/handle

holta:
Testing now beginning.

Fixes #995

Issue #995, opened 2018-08-13
Enable local_vars.yml to set /etc/iiab/openvpn_handle during IIAB installation


holta:
Implementers want uninterrupted installs (or at least as streamlined as possible).

So prescribing /etc/iiab/handle in advance within /etc/iiab/local_vars.yml would allow this to happen, regardless whether local_vars.yml is constructed interactively by a sexy install, or more traditionally using a text editor like nano!

Related:
- [x] #317 Install OpenVPN/TeamViewer/support tool(s) BEFORE installs begin [or near the very beginning!] 
- [ ] #872 Reboot + Auto-Continue...IIAB Install Scripts (for OS updates etc)
- [ ] #958 Non-ASCII Characters in install.txt Bash Script
- [ ] #963, https://github.com/iiab/iiab-factory/pull/33 Host install.txt on Github
- [ ] #994 Should IIAB support NOOBS to broaden participation?
- [x] PRs #996, #997, #998, #1000, #1001, #1003, #1004 making the [openvpn playbook/role](https://github.com/iiab/iiab/tree/master/roles/openvpn) more readable, while implementing variable openvpn_handle

_**[ CHECKLIST ABOVE HAS NOW MOVED TO IIAB 6.7 TICKET #1028 ]**_

PR #1001, opened 2018-08-16
OpenVPN code readability

holta:
Explanations/formatting cleaned up, so that OpenVPN playbook is more maintainable in future.

Builds on #995 and PRs #996 #997 #998 #1000

### DIFF
--- a/roles/openvpn/tasks/main.yml
+++ b/roles/openvpn/tasks/main.yml
@@ -70,8 +70,8 @@
 #    backup: yes
 #  when: openvpn_handle is defined
 
-# Comment out in future?  up_wan was being installed twice (also above) and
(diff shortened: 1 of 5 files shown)


tim-moody:
# openvpn_handle variable is mandatory as of August 2018 (if set to "", /etc/iiab/uuid will be used instead)
a variable that can have no value  should not be said to be mandatory.

holta:
CLARIF:

The openvpn_handle variable value "" is allowed (empty string is still a string).

What is mandatory is that variable openvpn_handle not be left completely undefined.

Hence the reason it is set to "" in https://github.com/iiab/iiab/blob/master/vars/default_vars.yml and https://github.com/iiab/iiab/blob/master/roles/openvpn/defaults/main.yml and http://wiki.laptop.org/go/IIAB/local_vars.yml

PR #1013, opened 2018-08-20
Keys

georgejhunt:
As requested at last week IIAB meeting


georgejhunt:
you're right, Thanks

georgejhunt:
tested on a missing file, and with openvpn_install=true. Fails to remove the existing lines when openvpn_install changed to false. 
Tried putting "^" at beginning of regexp.  Don't know what else to try. Maybe an ansible bug.

holta:
Looks good @georgejhunt (but shouldn't this code be encapsulated within openvpn's own playbook, so that [./install-support](https://github.com/iiab/iiab/blob/master/install-support) and "./runrole openvpn" work as expected?)

georgejhunt:
True

holta:
> tested on a missing file, and with openvpn_install=true. Fails to remove the existing lines when openvpn_install changed to false.
Tried putting "^" at beginning of regexp. Don't know what else to try. Maybe an ansible bug.

One proven way to fix this is by inserting a backslash in front of every plus sign (+) in each public key (there are lots!)

(Hopefully there's an even more elegant way to fix this?!)

In any case: plus signs (+) aren't literals but are interpreted as part of [Python regular expressions](https://docs.python.org/2/library/re.html), per the regexp parameter as implemented by Ansible's [lineinfile module](https://docs.ansible.com/ansible/2.6/modules/lineinfile_module.html).

holta:
Tested on Ubuntu 18.04/Desktop &mdash; please test on other OS's after merging.

PR #1016, opened 2018-08-21
httpd playbook output clarified

holta:


holta:
Tested on Ubuntu 18.04

PR #1021, opened 2018-08-21
More fully deprecate TeamViewer playbook (part 2)

holta:
Fix PR #1020


holta:
Tested on Ubuntu 18.04 and Raspbian.

Issue #1041, opened 2018-08-28
IIAB Apps need: Upgrades, Reinstalls, Backups, Accounts, Content, Audits (provisioning norms needed!)


arky:
@holta  Please kindly allow users to do provisioning with 'calibreweb_provision' variable. This is pretty standard practice in configuration management deployments. 

Just checking if metadata.db /  app.db exits seems like a good idea but I would recommend using variables to allow more fine grained control of deployments.  

https://github.com/iiab/iiab/commit/2e6ab33874774205f16a0de81eef2180eb25eb1e

holta:
As written the default `calibreweb_provision` was too dangerous, in that it would have caused deletion of e-books' metadata among implementers who often run "iiab-install --reinstall" as they're adding content, racing like crazy in the days prior to delivering Internet-in-a-Box units to schools.

This is the reason that the majority of our IIAB services/apps use a trigger file, that allows sysadmins to force an upgrade or reinstall of an individual IIAB service/app &mdash; without harming others and their content/configurations &mdash; even if that sysadmin/implementer has zero understanding of Ansible: http://wiki.laptop.org/go/IIAB/FAQ#Can_I_upgrade_or_reinstall_server_apps.3F

But if an additional `calibreweb_provision` flag is truly also needed, in addition to a trigger file (in this case `/library/calibre-web/metadata.db`) then let's just make sure that is designed carefully to be failsafe, so both use cases don't collide with each other~

Refs: #816 -> PR #999 -> PR #1038 -> PR #1039

holta:
Here are the 2 most common upgrade/clean install scenarios requested by IIAB implementers:

1) UPGRADE by OVERLAYING (some IIAB service/app) to the latest version WITHOUT losing content or settings/accounts/passwords!

2) UPGRADE by WIPING (some IIAB service/app) to the latest version WHILE forcing all settings/accounts/passwords to the default &mdash; _even if that means losing existing content!_  HUGE BONUS: if the implementer is _forced_ to create a complete backup prior to proceeding, e.g. by moving the entire directory of content + settings (if such exists) then informed consent is 100% confirmed.

Separately here are a few other scenarios that arise on rare occasion, but are not nearly as common:

3) REINSTALL by OVERLAYING the existing/old version (of some IIAB service/app) WITHOUT losing content or settings/accounts/passwords!

[...]

PR #1011, opened 2018-08-20
New Transmission playbook: eases KA Lite downloads using BitTorrent

arky:
### Fixes Bug #884
https://github.com/iiab/iiab/issues/884

### Description of changes proposed in this pull request.
This adds a new playbook for transmission-daemon. The provisioning system adds ka-lite torrents for download.

### Smoke-tested in operating system.

Smoke tested on Debian 10 i586 machine. 

The ports 9091 51413 needs to be opened for WAN and local lan. 

### Mention a team member for further information or comment using @ name
@holta

Issue #884, opened 2018-07-10
Automate torrenting+installing of KA Lite videos in 7 languages


holta:
@arky & All: The overwhelming majority of IIAB implementers and schools want the download/installation of videos+thumbnails to be fully automated from http://pantry.learningequality.org/downloads/ka-lite/0.17/content/ (currently holds 7 languages, English is the largest at ~44GB) into KA Lite.  For their main language(s).

As microSD cards are plenty big enough (and affordable enough) to hold every compressed KA video in 2018.

So I've started a conversation with Learning Equality to see if they have suggestions as to how best to proceed and collaborate here, e.g. if for example this selection among the 7 currently torrentable KA languages (to be downloaded using above URL torrents) can be made increasingly foolproof AND future-proof ?

Applying to both KA Kite & Kolibri ?  e.g. see #841 "Kolibri playbook for IIAB"

And ideally applicable to other popular/bulk content that is yet to appear ?

Stayed tuned!  Keep in touch all interested in working towards this very-high-leverage install automation..as first discussed with @tim-moody earlier in 2018.

tim-moody:
adm cons should probably do this instead of just opening the kalite gui.

holta:
> adm cons should probably do this

Also [/etc/iiab/local_vars.yml](http://wiki.laptop.org/go/IIAB/local_vars.yml) [formerly /opt/iiab/iiab/vars/local_vars.yml] can fully specify the primary (if not secondary) language, so that "overnight" installs become fully automated.

> instead of just opening the kalite gui

Currently KA Lite's GUI is useless to download/torrent compressed vids.  KA Lite's GUI and its taxonomy/hierarchy of videos (and checkboxes) are however very useful for those schools/implementers/curriculum designers that want to REMOVE certain videos, e.g. from the ~44GB set of compressed/torrented English vids.

arky:
I think this is easily done with systemd bittorrent daemon. The daemon could download and copy the finished download to the KA-Lite library.
 
The daemon could be triggered by ansible install commands or triggered by config file or adding .torrent files to a specific download folder.

jamalex:
Heads up that in Kolibri, all KA videos are already pre-compressed, so the entire English channel is 36.5 GB, when downloaded through the regular methods (UI or management commands). It's also more frequently updated than KA Lite content, and hence torrents would get stale quickly (and hence dilute the seeder pool), making torrents less than optimal.

[...]

PR #1044, opened 2018-08-29
BitTorrent tool "Transmission" eases KA Lite downloads

holta:
Builds off @arky's hard work at #884 -> PR #1011 adding:
- instructions at http://FAQ.IIAB.IO#KA_Lite_Administration:_What_tips_.26_tricks_exist.3F, in the [playbook itself](https://github.com/iiab/iiab/tree/master/roles/transmission) and in [local_vars.yml](http://wiki.iiab.io/local_vars.yml)
- login to http://box:9091 to monitor KA Lite torrent downloads, and initiate others (enhances access from all IP addresses, instead of the restrictive rpc-whitelist and rpc-host-whitelist that had been in /etc/transmission-daemon/settings.json, both of which are now turned off).  Username/Password (Admin/changeme) can be turned off using variable rpc-authentication-required if some individuals prefer that.
- integration with [Stage 8](https://github.com/iiab/iiab/blob/master/roles/8-mgmt-tools/tasks/main.yml) of IIAB's install, [default_vars.yml](https://github.com/iiab/iiab/blob/master/vars/default_vars.yml#L364-L401) and [all 3 local_vars files](https://github.com/iiab/iiab/tree/master/vars) (all greatly cleaned up!)
- deletion of all 3 local_vars_*_vpn.yml files as OpenVPN is now installed in different ways
- IMHO commenting out all 7 languages under transmission_kalite_languages within /etc/iiab/local_vars.yml is more tangible/visceral than creating yet another variable like `transmission_provision` (which has the same effect).  Nevertheless I've left the variable in place, because @arky may have other uses for it (#1041).  If not, we should remove this variable.
- successful testing on Raspbian (Transmission 2.92), Ubuntu 18.04 (Transmission 2.92) and Ubuntu 16.04 (Transmission 2.84)

Future:
- we might try to upgrade to Transmission 2.94+ in future, as 2.92 has bugs whereby http://box:9091 sometimes displays very erroneous info (wrongly labelling torrents/sizes) until you do a browser reload.  If so, this will also reduce IIAB's exposure to vulnerabilities from 2016/2017 etc.
- we might implement http://box/transmission at some point?  (Though that's longer than http://box:9091 so perhaps not wise!)

Question: I don't know why @arky chose the following owner/group in roles/transmission/defaults/main.yml:
```
transmission_user: debian-transmission
transmission_group: root
```
(But I can confirm that it works!  As such, should we delete the [user/group Known Issue](https://github.com/iiab/iiab/tree/master/roles/transmission#known-issues) from roles/transmission/README.md ?)


holta:
Let's push this into IIAB 6.6 RC 1 now and see how people use it~

tim-moody:
re:
IMHO commenting out all 7 languages under transmission_kalite_languages within /etc/iiab/local_vars.yml is more tangible/visceral than creating yet another variable like transmission_provision (which has the same effect). Nevertheless I've left the variable in place, because @arky may have other uses for it (#1041). If not, we should remove this variable.

one is a boolean and the other is a list of languages. e.g.

with_items: "{{ transmission_kalite_languages }}"
45 | ignore_errors: yes
46 | when: transmission_provision
47

if transmission_provision were kalite_use_transmission or some such it could go in local vars

holta:
> if transmission_provision were kalite_use_transmission or some such it could go in local vars

Excellent question as to how Transmission should be wired into other playbooks & sundry content!

For now, local_vars.yml takes an initial stab at solving this for KA Lite with these 2 explanations:
- https://github.com/iiab/iiab/blob/master/vars/local_vars_medium.yml#L181
- https://github.com/iiab/iiab/blob/master/vars/local_vars_medium.yml#L212-L232

...hoping that the most prevalent such social norms develop into programmatic solutions for all, sometime in the coming months (in support of 1st Time Implementers, who really do need tight community workflows **not** endless corner cases that cause them to Give Up ;)

PR #1079, opened 2018-09-02
Experimental fix for serious OpenVPN flaw

holta:
This is _a_ possible fix to a structural flaw that's existed in our OpenVPN code for years, and manifested itself as OpenVPN accidentally turning on, even when @tim-moody & @kananigit had set their [/etc/iiab/local_vars.yml](http://wiki.laptop.org/go/IIAB/local_vars.yml) to: 
```
openvpn_install: True
openvpn_enabled: False
```
Testing ongoing...

PR #1087, opened 2018-09-03
Vpn

jvonau:
This is a bit of a can of worms between the OS's methods to launch openvpn, some are a very legacy approach. Lets stay with the systemd @ way for maximum compatibility across the various distros.

### DIFF
--- a/roles/openvpn/tasks/main.yml
+++ b/roles/openvpn/tasks/main.yml
@@ -72,7 +72,7 @@
     - { src: 'announce', dest: '/etc/openvpn/scripts/announce', mode: '0755' }
     - { src: 'announcer.j2', dest: '/etc/openvpn/scripts/announcer', mode: '0755' }
     - { src: 'silence', dest: '/etc/openvpn/scripts/silence', mode: '0755' }
-    - { src: 'xscenet.conf.j2', dest: '/etc/openvpn/xscenet.conf', mode: '0644' }
+    - { src: 'xscenet.conf.j2', dest: '/etc/openvpn/client/xscenet.conf', mode: '0644' }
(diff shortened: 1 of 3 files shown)


jvonau:
#1079

holta:
_Disabling_ OpenVPN when the implementer/operator just specified `openvpn_enabled: True` seems very wrong (it's precisely the opposite of what the implementer/operator just asked for, and might even be depending on for other OpenVPN connections.)

"Legibility" i.e. abiding by OpenVPN industry/global norms is very important for long-term code maintainability, wherever we can, as outlined @ https://github.com/iiab/iiab/pull/1079#issuecomment-418175439

In short: this current PR (feels to me!?) like a step backwards not forwards &mdash; when IIAB should (try hard, where we can) to work within industry/global norms as much as possible.

Let's talk more during Thursday's 10AM NYC Time call (http://minutes.iiab.io) as this approach feels wrong to me, but perhaps there is more to the story...

jvonau:
I'm going to have to take exception with your position in respect to what the openvpn developers intended method for starting tunnels is.. In the (openvpn git repo)[ https://github.com/OpenVPN/openvpn/tree/master/distro/systemd] there are no shipped init.d scripts so the origin of the startup routine used by debian is the package maintainer, and I'll go with it's a shitty way of dealing with systemd  What I see is upstream from debian seriously wants the use of the client|server unit files to be used to workaround the debian's flawed method of starting, upstream does not ship a openvpn unit file and then lying to system with the bin true trick... What I think should be the plan is to use the open-client@ unit file as upstream wants and leave the shitty use of openvpn disabled because it's really flawed on so many levels Next would be to circle back and fix the other support files, might need to move the announcer stuff

jvonau:
https://github.com/OpenVPN/openvpn/tree/master/distro/systemd

PR #1081, opened 2018-09-03
Cleanup of /opt/iiab/iiab/install-support, iiab-support.yml etc

holta:
For code readability / Ansible norms.

PR #1090, opened 2018-09-05
Fix iiab-remote-on|off which had longstanding (OpenVPN) issues

holta:
Also clarifies stale & unused code.

Builds on #995 PRs #996 #997 #998 #1000 #1001 #1003 #1004 #1079 #1080 #1081 #1083 #1085 #1086 #1088

Aside: #1087 is still being debated.  From what I can tell it defies the wishes of the OpenVPN community (which installs out-of-the-box with AUTOSTART="ALL" in places like /etc/default/openvpn and/or /etc/init.d/openvpn, and runs a parent-like service "openvpn").  Hence #1087 seems to encourage long-term maintenance problems or at least risks, the more we override OpenVPN's default behaviors.  But @jvonau is investigating further, and we'll try to talk tonight.


tim-moody:
please rename to iiab-vpn-on and off. these do nothing to anything 'remote'

tim-moody:
on should only start, not enable

holta:
> on should only start, not enable

Why?  I don't agree with this change in behavior.  We (and OpenVPN global norms) have always enabled parent service "openvpn" as the OpenVPN community refines that to start as early as possible during boot, to keep the connection as active as possible, as often as possible (after reboots especially!)

holta:
> please rename to iiab-vpn-on and off. these do nothing to anything 'remote'

CLARIF: George & I chose these established commands 3-4 years ago (xs-remote-on|off, now iiab-remote-on|off) as human-not-techie terms, to allow operators to turn on/off "all" remote access for extended support over multiple reboots / multiple days.  (Which suggests we should maybe include "systemctl disable ssh" etc too, as we intended to do long ago!)

tim-moody:
will it make my tv work?

please then add iiab-vpn-on and off so the user knows what they're getting.

holta:
> please then add iiab-vpn-on

I've added iiab-vpn-on|off for those who dislike the larger scope of security/privacy protections from network intrusion, that iiab-remote-off is in fact intended for, even if it is not yet fully implemented.

holta:
Tested & re-tested on Raspbian & Ubuntu 18.04

PR #1091, opened 2018-09-05
Fix for Ansible install on Ubuntu, thx to @kananigit

holta:
Fixes this bug that @kananigit experienced on Ubuntu 18.04/Server:
```
scripts/ansible will now try to install Ansible --- Please Wait
Reading package lists... Done
Building dependency tree
Reading state information... Done
Package python-wheel is not available, but is referred to by another package.
This may mean that the package is missing, has been obsoleted, or
is only available from another source

E: Unable to locate package python-pip
E: Package 'python-wheel' has no installation candidate
```
Thereupon commenting out the old Line 43 of /opt/iiab/iiab/scripts/ansible ("apt -y install python-pip python-setuptools python-wheel patch") had this effect:
```
The following packages have unmet dependencies:
 ansible : Depends: sshpass but it is not installable
E: Unable to correct problems, you have held broken packages.
```
The solution was to scrap the old Line 43 entirely, and use the following instead:
```
apt -y install software-properties-common
```


holta:
More complete details on @kananigit's nightmare Ubuntu 18.04.1/Server install documented here:

https://github.com/iiab/iiab/wiki/IIAB-6.6-Release-Notes#known-issues

(In short, some 18.04.1/Server installers released on 2018-07-25 fail to populate /etc/apt/sources.list !)

PR #1113, opened 2018-09-08
sync dhcpd so comes up at boot on rpi w

tim-moody:
### Fixes Bug
dhcpd fails to start at boot on rpi w

code tested on w
ansible run not tested


holta:
Am pushing this to encourage wider testing on @jvonau's recommendation; @tim-moody believes it's low-risk.

Many thanks to @tim-moody & @jvonau for battling this out today!

(In support of Raspberry Zero W's, where dhcpd was not starting...)

PR #1138, opened 2018-09-18
Wrong variable name [in iiab-make-kiwix-lib]

tim-moody:


tim-moody:
currently testing on rpi

holta:
Should this be merged today?

tim-moody:
it's clearly bug, but since it only happens in an exception I have not been able to trigger it. risk of merging is very low

holta:
> it's clearly bug, but since it only happens in an exception I have not been able to trigger it. risk of merging is very low

Thanks...and thanks to all for testing `iiab-make-kiwix-lib` in coming days!

Issue #1141, opened 2018-09-19
OpenVPN starts too late in IIAB's install--remote support cannot wait 30-45 min


holta:
I'd like @jvonau & others to help refine where OpenVPN can be installed+enabled _far earlier_ than [Stage 4](https://github.com/iiab/iiab/blob/master/roles/4-server-options/tasks/main.yml#L13-L17) (as close as possible to the very beginning of IIAB's installation process, to make remote support realistic!)

jvonau:
There is an interactive question and early activation in https://github.com/iiab/iiab-factory/pull/33/files

holta:
I generally lean against having an interactive question for OpenVPN support, as I think this opens the door to too many people using a collegial service (whether their intentions sincere or not, and whether their security practices are secure or not ;)

Folks who wants service/support should generally talk to us and explain themselves first in my opinion.  Eventually we can send them to a paid support vendor of one kind or another if we cannot handle the volume (not an unlikely scenario!)

Anyway this is the reason I removed OpenVPN support for 1-line installers for the general public from IIAB 6.6, as I'm concerned the risks are too high, with so many important contributors homes being put at potential risk as we now begin to grow.

holta:
Closing as PR #1148 merged after successful testing.

(Feel free to re-open this issue if a larger discussion is useful!)

PR #1148, opened 2018-09-19
OpenVPN rapid support - implements #1141 for realistic remote support

holta:
Resolves longstanding pain point #1141 

Needs testing on all mainline OS's

### DIFF
--- a/roles/1-prep/tasks/main.yml
+++ b/roles/1-prep/tasks/main.yml
@@ -37,6 +37,12 @@
   set_fact:
     uuid: "{{ stored_uuid.stdout_lines[0] }}"
 
+- name: OPENVPN
+  include_role:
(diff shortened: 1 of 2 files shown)


holta:
Tested successfully on Ubuntu 18.04 and Raspbian Lite.

Further testing including fresh install(s) on Debian 9.5 etc will also be useful after merge.

jvonau:
Being able to ssh in as iiab-admin won't occur till a bit later until ssh is enabled in 4-server-options/tasks/main.yml and the iiab-admin user is created in 3-base-server/tasks/main.yml, perhaps those related routines may need to occur earlier to have the full desired effect of coming available earlier. This change also may also prohibit the admin-console from disabling the vpn service by way of iiab-from-console.yml

PR #1151, opened 2018-09-20
Fix & Revision to PR #1148 (OpenVPN rapid start during installs)

holta:
Builds on #1141 PR #1148


holta:
Tested on fresh install of Debian 9.5 too.

Issue #1146, opened 2018-09-19
Is IIAB setting Apache/systemd/CMDSRV/etc permissions in a non-standard way?


holta:
When installing software on Raspbian after IIAB is installed (e.g. emacs, and others) I commonly get this message:
```
 14 files changed, 2995 insertions(+), 1 deletion(-)
 create mode 120000 apache2/mods-enabled/socache_shmcb.load
 create mode 120000 apache2/mods-enabled/ssl.conf
 create mode 120000 apache2/mods-enabled/ssl.load
 create mode 100644 apache2/sites-available/admin-console.conf
 create mode 100644 apache2/sites-available/cmd-service.conf
 create mode 120000 apache2/sites-enabled/admin-console.conf
 create mode 120000 apache2/sites-enabled/cmd-service.conf
 create mode 100644 iiab/config_vars.yml
 create mode 100644 iiab/kiwix_catalog.json
 create mode 100644 iiab/oer2go_catalog.json
 create mode 100644 systemd/system/iiab-cmdsrv.service
 create mode 120000 systemd/system/multi-user.target.wants/iiab-cmdsrv.service
```
@jvonau & others: should we modify chmod permission of any of the above, to better follow standards?

jvonau:
With etckeeper creating a git repo of /etc and records the changes when apt runs... Think nobody is really using etckeeper and it's bloating the filesystem every time apt runs.

holta:
> With etckeeper creating a git repo of /etc and records the changes when apt runs... Think nobody is really using etckeeper and it's bloating the filesystem every time apt runs.

Ok.  Should etckeeper be commented out from Line 51 of [roles/2-common/tasks/packages.yml](https://github.com/iiab/iiab/blob/master/roles/2-common/tasks/packages.yml#L51) now?

holta:
> 
> 
> > With etckeeper creating a git repo of /etc and records the changes when apt runs... Think nobody is really using etckeeper and it's bloating the filesystem every time apt runs.
> 
> Ok. Should etckeeper be commented out from Line 51 of [roles/2-common/tasks/packages.yml](https://github.com/iiab/iiab/blob/master/roles/2-common/tasks/packages.yml#L51) now?

Idea added to http://minutes.iiab.io call agenda for Thur 2018-09-20 Item 2. a. so it's not lost or forgotten.

PR #1152, opened 2018-09-20
Remove etckeeper & clean roles/2-common/tasks/packages.yml

holta:
Per @jvonau's suggestion at https://github.com/iiab/iiab/issues/1146#issuecomment-422981195

Is this a good idea to ditch etckeeper?  Very easy change if so.


holta:
Smoke-tested on all [3 mainline OS's](https://github.com/iiab/iiab/wiki/IIAB-Platforms)

PR #870, opened 2018-07-03
Tweaking iptables for Captive Portal (Anish's Python approach, thx to Nikos Fotiou)

m-anish:
### Description of changes proposed in this pull request.

Based on @m-anish 's https://github.com/m-anish/named_redirect which builds off Nikos Fotiou's https://github.com/nikosft/captive-portal/blob/master/captive_portal.py

Issue #826, opened 2018-06-05
Nodogsplash Captive Portal requires internet connection [try dnsmasq approach?]


tim-moody:
see nodogsplash/nodogsplash#154

holta:
Refs #412 #575 #597 #608

holta:
@m-anish is currently reviewing https://github.com/nikosft/captive-portal/blob/master/captive_portal.py with @tim-moody, as a possible alternative Captive Portal &mdash; for new ideas at least &mdash; to break this log jam?

As usual testing of Android-and-similar against RPi servers is the key, and they look to be making progress here...

holta:
Chat Excerpt:

Anish: https://github.com/nikosft/captive-portal/blob/master/captive_portal.py
Tim: have you tried it?
Anish: so it is fairly simple script.  it works when there is internet on WAN IF
Tim: so does nodogsplash, but that is not our case
Anish: (or basically when DNS resolves).  yeah...  but i think we can use the iiab flag to do proper iptables redirecting.  we can even make is dynamic.  as it is fairly simple python code.  iiab internet_accessible flag
Tim: I think we need to emulate the traffic to test for internet by different devices
Anish: yea, agree.  this is not wispr or anything like that compliant (if there is such a thing)
Tim: I did a wireshark capture of nodogsplash with WAN internet (works) and without (doesn't work)
Anish: but offers an easy solution to a complex-ish problem...  tim, i think the difference here might be that it is just a few lines of python code, that we can expand and build on without consulting the remote corners of the internet for scarce documentation :)
Tim: I have been through nearly 10 approaches that are 'supposed to work', none works without an internet connection. I agree that the simplicity is a plus. BTW, there's another one in python used by kiwix's pi-box. it doesn't work for android either.
Anish: this is essentially the redirect it does...

-A PREROUTING -i br0 -p tcp -m tcp --dport 80 -j DNAT --to-destination 172.18.96.1:9090

Tim: Anish: do you have the wherewithal to analyze my wireshark captures...  The problem happens before the redirect. You only get to that stage if the device decides it is a captive port situation and so far for me it never does unless dns works
Anish: yeah.  so.. maybe we can have a dynamic bind rule based on whether internet works.  we could do (1) dns blackhole if no internet  or (2) this if there is internet.  because dns blackhole works everytime ... just that it will never redirect to anything on the internet
[...]

Issue #608, opened 2017-12-07
Captive Portal experimental support summary in FAQ.IIAB.IO ?


holta:
I'm in favor, even if just a couple sentences that we work on together @ http://FAQ.IIAB.IO on how to try out @tim-moody's nodogsplash PR #597 prototype for IIAB 6.5's release, implementing #412.

holta:
Experimental Support Documented in http://minutes.iiab.io under Dec 28, 2017 Item 2. h.

Hoping to migrate this to http://FAQ.IIAB.IO with @deldesir's help in January...or in coming days!

holta:
@tim-moody has also suggested @georgejhunt research an iptables approach.

holta:
As nodogsplash Captive Portal support might be changing to an iptables approach instead, I won't push these DRAFT configuration materials (transcribed from http://minutes.iiab.io under Dec 28, 2017 Item 2. h.) into FAQ.IIAB.IO until the outcome is more clear...

Captive Portal Experimentation (#597) Summary:

1. Set these both to True in your /opt/iiab/iiab/vars/local_vars.yml :
nodogsplash_install: True
nodogsplash_enabled: True
2. Arrange for Ansible to run the nodogsplash playbook AFTER ./iiab-install (incl network playbook!) have completed, similar to the 5-line “CALIBRE” section in  roles/9-local-addons/tasks/main.yml but remember the “network” playbook runs AFTER Stage 9. 
   1. _Explanation: IIAB’s networking playbook will revert iptables.  So you need the nodogsplash playbook To Run At The Very End!_
3. Works very well on Windows/Mac...but Android sign-in page never quite goes away?  Also a time-out issue might hit Android folk etc.  So encourage them to bookmark it after they get thru to http://box or whatever.  Android is the worst, old or new Android :-(  (Possibly Firefox on Android isn’t so bad?)
   1. Tim pasted in his testing across 4 platforms, so others can confirm problem clients: https://github.com/iiab/iiab/pull/597#issuecomment-354317071 
4. Modify its configuration until you get your favorite browsers/clients working better:
   1. **/etc/nodogsplash/nodogsplash.conf ← modify this one especially!**
   2. /etc/nodogsplash/htdocs/splash.html
   3. Of course run “systemctl restart nodogsplash” each time to test!
   4. Possibly change the timeout (which uses client devices’ MAC addresses) reducing it to 10min or so? (George's suggestion)
   5. Later do the same with Ansible?
[...]

PR #1179, opened 2018-09-29
Captive Portal -- Android 5.1,6.1,7.1 Mac 10.13.1, Windows 10.0, iOS 12.0

georgejhunt:
### Test Procedure:
1. Start with 6.7 master
2. git remote add ghunt https://github.com/georgejhunt/iiab, git fetch --all, git checkout -b squash ghunt/squash
3. Update /etc/iiab/local_vars.yml, enable py_captive_portal, dnsmasq, change ssid
4. ./iiab-install, and do iiab-admin-console, and iiab-menus
5. reboot
### Test results (first with iiab_gateway_enabled, and ethernet wire)
1. On Mac, brings up default browser with a splash page link. Click on link brings up home tab, and an extra tab with "success" on the tab , and previous splash
2. On iOS, brings up a splash page, click on link brings a "open in Safari" button, which leads to fully functional Safari browser (turning 90 degrees browses landscape mode).
3. Android 5.1, select ssid, brings up "sign in to wifi", which leads to splash, which clicked leads to crippled browser (sign in to network remains at the top, and 90 degree pivot fails)
4. Android 7.0, fails -- displaying a "sign in to network" for less than a second
5. Windows 10 -- displays splash, which requires click on link, which displays "Success" and home in two tabs
### Test results with ethernet wire removed
1. Same as above, except that #4 (Android 7) succeeds. That is a "sign in to network" appears, pressed, that leads to a splash page, which displays a message saying "Please ignore the SECURITY warning which appears after clicking the first button". Press the first button displays "The network you're trying to join has security issues. For examble, the login page may not belong to the organization shown. Continue anyway via browser". This leads to the splash page again. Clicking on the second link leads to a fully functional (rotate 90 degrees ) browser
#### known issues
1. cannot escape dumbed down browser on Android 5
2. The online failure of Android 7 should be easy to fix (but I've already lost the test log once by clicking the wrong button). So I want to generate this PR now, and then update it when I fix the bug


tim-moody:
I now see that we have the same experience with respect to 'dumbed down browser'; we just interpret it differently. I also see the sign in to wifi message, even on latest android, but did not interpret that to be dumbed down, because I have seen it over the home page with all javascript working. So you may be right, but I agree that it is pretty much impossible to get rid of the SIGN IN TO WIFI message.

tim-moody:
BTW, many are prepared to live with the cosmetic problem of the signin message not disappearing if everything else works.

holta:
Refs: #412 #608 #826 PR #870

holta:
@jvonau & @m-anish can you review this captive portal PR from @georgejhunt, ideally today or tmrw if you have time, so it can be merged (or revised) as appropriate?

holta:
@georgejhunt can you look over this PR feedback from @jvonau ?

- looks like some/the templates and support scripts are not provided
- there is a duplicate when clause
- not sure why hosts.yml is being modified
- why is box.lan being hardcoded in the templates?
- IIAB_GATEWAY_ENABLED LAN_IP in iiab.env should detect the current state not be a [one-off] when the the file is first installed

holta:
> * there is a duplicate when clause

Lines 47 and 49 of roles/network/tasks/captive_portal.yml :
https://github.com/iiab/iiab/pull/1179/files#diff-5fb8b8548104837f4087a06b933430beR49

@jvonau also mentions:
- from the commit messages looks like sqlite is used but not requested as a package along with python-dateutil that could go in packages
- [ Ansible's `service:` module should be [systemd:](https://docs.ansible.com/ansible/latest/modules/systemd_module.html) with `daemon-reload: yes` ]

georgejhunt:
Thanks for the extra set of eyes.
* looks like some/the templates and support scripts are not provided
-- I think templates and support scripts are all in place -- did a clean install
* there is a duplicate when clause
fixed
* not sure why hosts.yml is being modified
-- I'm not sure either,reverted -- perhaps when I was playing with virtual hosts in apache
* why is box.lan being hardcoded in the templates?
-- I use jinja2 to display the splash, -- and ansible also uses jinja2 -- cannot both interpret {% %} as I want
* IIAB_GATEWAY_ENABLED LAN_IP in iiab.env should detect the current state not be a [one-off] when the the file is first installed
-- Don't understand. Template should pick up the current state of variable during ansible install

holta:
Superseded by PRs #1300, #1327, #1333

Issue #1200, opened 2018-10-09
Roles missing READMEs


veeara282:
The following roles do not have `README.*` files:
```
0-init
awstats
calibre
cups
ejabberd
homepage
httpd
mediawiki
mongodb
monit
moodle
munin
mysql
network
nextcloud
openvpn
osm
phpmyadmin
postgresql
sshd
sugarizer
teamviewer
vnstat
wordpress
```

I used the script in PR #1201 to generate this list.

holta:
Thanks much @aidan-fitz for generating the above list of 35 IIAB roles, needing more of an explanation!

FYI the ~12~ 11 roles below should probably be removed from your list, as they're rarely if ever used: (as of Oct 2018 anyway!)

1. authserver
2. debian_schooltool
3. ejabberd_xs
4. idmgr
5. ajenti
6. moodle-1.9
7. pathagar
8. schooltool
9. nodogsplash
10. docker
11. sugar-stats

holta:
My mistake...now corrected above: **sshd** is in fact an active role :)

holta:
> Those in favor should mention how many of the existing readmes they have read.

Some of these README's are absolutely top-notch, but they have been overlooked as they were too hard to find.

So I'm changing that now, with direct links to the best README's from http://FAQ.IIAB.IO which has a very high Google page ranking.

This is a big job, and I'm not done yet (to make our best documentation accessible!)  But please for now see ["What can I do with E-books and Internet-in-a-Box?"](http://wiki.laptop.org/go/IIAB/FAQ#What_can_I_do_with_E-books_and_Internet-in-a-Box.3F) which is helping to make IIAB roles (playbooks) much more understandable to new implementers and go-getter teachers/students around the world.  Folks who truly want to understand (and where possible help us all make more/better digital libraries in coming years!)

Ref: PR #1285

PR #1201, opened 2018-10-09
scripts/roles_needing_docs.py to show missing documentation

veeara282:
### Fixes Bug
N/A

Added a Python script that checks every role in the project and prints its name to `stdout` if the role directory does not contain a `README` file. For ease of use, you can pipe the output of this script to a file or to a clipboard utility (e.g. `pbcopy` on macOS, `xclip` on Linux).

Ubuntu 18.04.1

@holta


veeara282:
Per your comment on #1200, I can set this up to ignore a given list of roles.

holta:
> Per your comment on #1200, I can set this up to ignore a given list of roles.

Run the output thru /usr/bin/sort or similar?

PS then we should document this in http://FAQ.IIAB.IO or somewhere in https://github.com/iiab/iiab/wiki

holta:
Can we perhaps rename this to (something like) roles-without-docs.py or roles-needing-docs.py ?

veeara282:
I like that—it's in more of a functional style of naming programs.

veeara282:
I should note that the script can be tricked by putting an empty `README` in a role directory.

PR #1204, opened 2018-10-09
Tweaks to roles-needing-docs.py & unmaintained-roles.txt

holta:
- builds on @aidan-fitz's #1200 & PR #1201 so it can be run from any directory
- moves scripts/docs_ignore to /opt/iiab/iiab/unmaintained-roles.txt
- uses dashes not underscores per IIAB norm (similar to /usr/bin/iiab-* commands)


veeara282:
This can only be run if the repo is in the `/opt/iiab/iiab` directory. I have it in the same place as my other project repos on my development machine.

PR #1202, opened 2018-10-09
@jvonau's new fix for dhcp_service2, as community transitions from named/BIND towards dnsmasq

jvonau:

PR #1207, opened 2018-10-09
Refining @jvonau's new fix for dhcp_service2

holta:
Building on @jvonau's PR #1202.  Please see his original 2 commits and their descriptions:

https://github.com/iiab/iiab/commit/bb6b3c2ab252791b25ea0ad0ba2cd3f85bd1b440
> Explicitly disable dnsmasq when in Appliance mode
> otherwise dnsmasq will become active upon reboot because enable_services.yml
> would still be acting upon local_vars.yml where dnsmasq_enabled is set to True
> while there is not LAN interface available to bind the service to.

https://github.com/iiab/iiab/commit/9d616cf3018bf7ec2ea7891fca8fdc04ca5907dc
> explicitly identify which primary dhcp server is being disabled
> dhcp_service2 is used in restart.yml when not in Appliance mode

(Needs further testing...)


holta:
Basic smoke testing comlpeted using `./iiab-network` on Ubuntu 18.04, Debian 9.5 & Raspian Lite.

jvonau:
squid and friends' state are already recorded in enable_services.yml under the respective services.

tim-moody:
Don't see the need for dhcp_service and dhcp_service2 and prefer not to overload variables. dhcp_service2 is sometimes the name of the service and sometimes a boolean. If it is a working variable lets keep it out of the whole default/local vars system.

Issue #1193, opened 2018-10-09
Transmission (BitTorrent) error when no languages are picked


holta:
@arky if you have a moment, can you consider cleaner error handling for IIAB 6.7's release, so that red errors are not part of Ansible output unless necessary?

> TASK [transmission : Add PAUSED KA Lite torrent(s) to transmission-daemon's queue] ***
failed: [127.0.0.1] (item=None) => {"changed": true, "cmd": "/usr/bin/transmission-remote --start-paused -n Admin:changeme -a http://pantry.learningequality.org/downloads/ka-lite/0.17/content/ka-lite-0.17-resized-videos-.torrent", "delta": "0:00:00.378235", "end": "2018-10-08 23:17:18.668073", "item": null, "msg": "non-zero return code", "rc": 1, "start": "2018-10-08 23:17:18.289838", "stderr": "", "stderr_lines": [], "stdout": "localhost:9091/transmission/rpc/ responded: \"success\"\nError: gotMetadataFromURL: http error 404: Not Found", "stdout_lines": ["localhost:9091/transmission/rpc/ responded: \"success\"", "Error: gotMetadataFromURL: http error 404: Not Found"]}
...ignoring

holta:
Appears fixed by PR #1216

Issue #1078, opened 2018-09-02
BitTorrent and/or IPFS for education-centric grassroots-driven Sneakernets-of-Alexandria


holta:
Building off #884, PRs #1011, #1044, @tim-moody's suggestions at PR https://github.com/iiab/iiab/pull/1061#issuecomment-417859271

holta:
RELATED: #977 (Kolibri 0.10.2+ Release Polish...as it evolves to 0.11 and beyond!)

arky:
It seems to me that there is no consensus on when/if IPFS  will be adopted to address this issue in upstream projects. 

Perhaps we could only focus on streamlining existing transmission code in this bug until at certain point in future when an alternative solution becomes available.

holta:
There is progress here — a basic version of Internet Archive's Decentralized Web services should be integrated as an option within IIAB 7.0 in coming weeks — and hopefully much more from @m-anish as part of IIAB 7.1:

Refs: #1546 #1576 PR #1617

mitra42:
But note ... we disable IPFS in the Offline Archive code because of the CPU/bandwidth overhead. It also won't work in offline mode, and (at least for our content) is still single-point-of-failure, i.e. has to be able to reach the archive.org servers. They've got a lot on their plate at IPFS, and making it work in low CPU / low bandwidth / single-point-of failure is AFAIK not on anyone's agenda.

BitTorrent / WebTorrent has some potential for these environments, especially for rebroadcast of video between end-user devices, I hope to get a tracker build into the offline archive at some point, but don't have the time to do so at the moment.

PR #1216, opened 2018-10-11
Check transmission_kalite_languages is undefined

arky:
Check if transmission_kalite_languages is defined.

https://github.com/iiab/iiab/issues/1193

### Fixes Bug

Proposed fix for bug #1193.

Check If there any languages are defined in transmission_kalite_languages. 

Debian 10

@holta

### DIFF
--- a/roles/transmission/tasks/main.yml
+++ b/roles/transmission/tasks/main.yml
@@ -42,7 +42,8 @@
     --start-paused
     -n {{ transmission_username }}:{{ transmission_password }}
     -a http://pantry.learningequality.org/downloads/ka-lite/{{ transmission_kalite_version }}/content/ka-lite-0.17-resized-videos-{{ item }}.torrent
-  with_items: "{{ transmission_kalite_languages | default('english') }}"
+  with_items: "{{ transmission_kalite_languages }}"
+  when: transmission_kalite_languages is defined and transmission_provision
   ignore_errors: yes
 
 - name: Disable transmission-daemon service
--- a/roles/transmission/tasks/main.yml
+++ b/roles/transmission/tasks/main.yml
@@ -18,7 +18,7 @@
   systemd:
     name: transmission-daemon
     state: stopped
-  ignore_errors: yes  
+  ignore_errors: yes
 
 - name: Create transmission-daemon settings
   template:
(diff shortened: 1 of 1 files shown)


holta:
Quite Popular Scenario:

A BitTorrent service (e.g. Transmission) is desired/needed to be turned on for _future_ usages, rather than for immediate downloads, particularly in areas where Internet costs are expensive.

(Am I perhaps misunderstanding this PR when I say that it would not support this common need to defer content torrenting ?)

holta:
> @holta This PR is a simple check for null variable. We are using jinja2 filters to catch the edge cases where the variable is undefined.

I'm concerned implementers will be stuck downloading (and paying for the download of) ~43GB of English content against their will.

Wouldn't it be better to test if {{ transmission_kalite_languages }} is null, and (if indeed null) then launch /usr/bin/transmission-remote WITHOUT the `-a <URL>` flag?

holta:
Interestingly http://box:9091 works immediately during IIAB 6.7/master installations, even when a language is not set in /etc/iiab/local_vars.yml

In other words: functionality is complete (from my perspective) however the red error as shown in #1193 really should be mitigated to avoid confusing new implementers &mdash; who will very reasonably perceive (incorrectly!) that http://box:9091 failed to start :/

arky:
> I'm concerned implementers will be stuck downloading (and paying for the download of) ~43GB of English content against their will.
> 
Set always transmission_provision: False to prevent adding any torrents. 

> Wouldn't it be better to test if {{ transmission_kalite_languages }} is null, and (if indeed null) then launch /usr/bin/transmission-remote WITHOUT the `-a <URL>` flag?

Tried that already. The Ansible looping with_items is limited. This is very simple solution for an edge case.

arky:
> Interestingly http://box:9091 works immediately during IIAB 6.7/master installations, even when a language is not set in /etc/iiab/local_vars.yml
> 
> In other words: functionality is complete (from my perspective) however the red error as shown in #1193 really should be mitigated to avoid confusing new implementers — who will very reasonably perceive (incorrectly!) that http://box:9091 failed to start :/

The failure or success of the following task has no impact on the transmission-daemon process. Seems to me like totally different problem.

arky:
@holta Updated the pull request and changed the PR title.

tim-moody:
when: transmission_kalite_languages is **not** defined and transmission_provision

seems backwards

arky:
@tim-moody Typo fixed

holta:
Thanks all for continuing to push this forward so we can get passionate educators far closer to the Sneakernet-of-Alexandria vision (#1078 and similar) of efficiently exchanging the best local/peer-proven OER's [Open Education Resources].

Fixes #1193

Issue #1189, opened 2018-10-09
Install Script: prompt to set iiab-admin password


veeara282:
### Expected behavior and actual behavior.

Currently, the install script prompts the user to modify `local_vars.yml` and then uses the default value for `iiab_admin_passw_hash`. The script should prompt the user to set the password and then add the hash to the config file.

When logging in via SSH, the VM sends over this warning:
```
SSH is enabled and the default password for the 'iiab-admin' user is unchanged.
This is a security risk - please login as the 'iiab-admin' user and type 'passwd' to change password.
```

Is there a default password? If so, there should be a user-friendly way to override it.

### Steps to reproduce the problem.

1. Run `curl download.iiab.io/6.7/install.txt | sudo bash`
2. When prompted for `local_vars.yml` choice, select 1
3. When prompted `Edit /etc/iiab/local_vars.yml to customize your Internet-in-a-Box? [Y/n]`, select yes
4. Edit and save `local_vars.yml`, then continue installation
5. After installation, try to login or `su` as `iiab-admin`

### Specifications like the version of the IIAB, operating system version, or hardware details.

IIAB 6.7 pre-release, Ubuntu 18.04 Azure VM

holta:
Thanks @aidan-fitz for the feedback.  I'll look into this for the upcoming IIAB 6.7 release or its immediate successor.

FYI passwords are published here: http://FAQ.IIAB.IO under Question 11 ("What are the default passwords?")

PR #1226, opened 2018-10-15
(1) iiab-admin role variables (2) alerts of pi & iiab-admin published passwords (3) introduce iiab_env_file (4) service_filelist | iiab_config_file -> iiab_ini_file

holta:
In support of and building on #1189 "Install Script: prompt to set iiab-admin password"


holta:
Heavily tested on Ubuntu 18.04; testing ongoing on Raspbian Lite.

tim-moody:
I've said it before, but I think it is a mistake to put every variable in each variant of local vars

https://www.defit.org/data-redundancy/

PR #1236, opened 2018-10-17
Add new fonts and icons

tim-moody:
for use by menuing and anything else

font awesome expects files in webfonts, not fonts


tim-moody:
passed smoke test on ub18.04 vm

holta:
> font awesome expects files in webfonts, not fonts

In the big picture, how will IIAB be using https://fontawesome.com = https://en.wikipedia.org/wiki/Font_Awesome for what purposes?

PR #1239, opened 2018-10-17
[TravisCI] Test all maintained roles. tests/test.yml refactor

arky:
### Fixes Bug
None 

Update the list of all IIAB roles. And check their syntax using tests/test.yml 
Add roles_path to point to the correct '../roles/' directory. 

TravisCI

@holta

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,7 +17,7 @@ install:
   - pip install ansible
 
   # Create ansible.cfg with correct roles_path and local_tmp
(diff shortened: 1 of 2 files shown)


holta:
Slightly revised to keep these 2 synchronized:
https://github.com/iiab/iiab/blob/master/tests/test.yml

If there's a clever way to automate this synchronization (or better yet to maintain only 1 such file, in conjunction with @aidan-fitz's [scripts/roles-needing-docs.py](https://github.com/iiab/iiab/blob/master/scripts/roles-needing-docs.py)) please by all means go for it, anybody that can make this happen!

PR #1244, opened 2018-10-19
Clarify ProxyPassReverse in Apache's kiwix.conf.j2 & calibre-web.conf.j2

holta:
Hopefully these 2 files can become models for developers wanting cleaner playbooks, that work reliably at http://box/[SOME-COMMUNITY-SERVICE] from day one:

https://github.com/iiab/iiab/blob/master/roles/calibre-web/templates/calibre-web.conf.j2
https://github.com/iiab/iiab/blob/master/roles/kiwix/templates/kiwix.conf.j2

Thanks to @tim-moody & @arky


arky:
@holta  Thanks for documenting this. Let me give you bit more further information. 

Reverse proxy in web applications is used for not only for human friendly URLs (URI rewriting) but also for performance and security benefits. 

"A reverse proxy taking requests from the Internet and forwarding them to servers in an internal network. Those making requests to the proxy may not be aware of the internal network."

https://en.wikipedia.org/wiki/Reverse_proxy

PR #1247, opened 2018-10-23
Offer Calibre 3.33.1+ (or whatever is latest) if implementer opts in within local_vars.yml

holta:
This PR also pins scripts/calibre-install-pinned-rpi.sh to Calibre 3.33.1 if some folks still want that.

HOWEVER/NOTE: retesting every release of Calibre (every 2-3 weeks typically) is a bad use of my time, so the default install is now reverting from scripts/calibre-install-pinned-rpi.sh to scripts/calibre-install-latest-rpi.sh _(each of which occasionally fail due to upstream dependency churn, so might as well let implementers experiment with the very latest Calibre at their own discretion...)_


holta:
Tested on Raspbian Lite 2018-10-09

Issue #1130, opened 2018-09-13
Remove Ansible's pink/red warnings. Prep for Ansible 2.8/2.9 and 2.11 deprecations.


holta:
Not all of Ansible's pink/red warnings are removable &mdash; indicating too many callouts to bash/sudo etc &mdash; but most will be!

Ref: #1100 (Migrate to Ansible 2.7 on "2018-10-04")

holta:
See PR #746 "Speedup by squashing package installs in ansible.cfg (condenses loop)" which is related &mdash; that PR was a real lifesaver (saving ~20 min on each RPi install, by avoiding dozens of unnec calls to "apt update") as contributed by @jvonau in April 2018.

arky:
It seems this is a trivial fix. You shouldn't use 'with_items' looping in package module anymore.

Use 'name:' instead of 'with_items' looping: 

```
package:
    name: "{{ item }}"
    state: present
  with_items:
    - openvpn
    - nmap
```

```
package:
     name:
       - openvpn
       - nmap
      state: present
 
```

holta:
More than 90% of our Ansible 2.7.1 warnings are now gone!

But this 1 FAIL and 2 WARNINGS remain: (can others help solve these 3?)
```
TASK [postgresql : Initialize the postgres db (debuntu)] ***********************
 [WARNING]: Consider using 'become', 'become_method', and 'become_user' rather
than running su

TASK [kolibri : Create kolibri default facility name, admin account and language] ***
[...]

PR #746, opened 2018-04-19
Speedup by squashing package installs in ansible.cfg (condenses loop)

holta:
### Description of changes proposed in this pull request.

Recommended by @jvonau per https://github.com/ansible/ansible/issues/24581.

### Smoke-tested in operating system.

Tested by @jvonau in Raspbian Lite on RPi 3.

PR #1259, opened 2018-10-28
Modernize package installations by removing squash_actions from ansible.cfg

holta:
Reverts PR #746 to use the more modern approach (just as fast, with lists that are more readable than loops) as required by Ansible 2.11+

The change in Ansible is summarized here: https://docs.ansible.com/ansible/devel/porting_guides/porting_guide_2.7.html#using-a-loop-on-a-package-module-via-squash-actions

This fixes (most of) #1130 !


holta:
Smoke-tested on Ubuntu 18.04

Needs further testing on Raspbian 2018-10-09 and Debian 9.5

PR #1276, opened 2018-11-02
Kiwix 0.6.1-1 -> 0.7.0

holta:
Please test the new kiwix-tools 0.7.0 with diverse content!


tim-moody:
may break make-kiwix-lib

holta:
> may break make-kiwix-lib

Can you confirm either way?

holta:
Excerpt from https://github.com/kiwix/kiwix-tools/blob/master/Changelog :
```
kiwix-tools 0.7.0
=================

 * Adapt to kiwix-lib new API

kiwix-serve
-----------

 * Dumps only valid books in the opdsfeed.
 * Allow the opds feed to be filtered by lang and paginated.

kiwix-manage
------------

 * Add a download command to download a remote book locally
 * Book are referenced by bookId not index.
 * No more indexType option as it is always XAPIAN.

PR #1289, opened 2018-11-08
Kiwix 0.7.0 -> 0.7.0-1

holta:
The 3 needed Kiwix packages are uploaded to: http://download.iiab.io/packages/


kelson42:
@holta This is only a new compilation, maybe because an underlying library has changed. I would ignore it, it is just a measure to avoid to overwrite the older 0.7.0 files.

Issue #1293, opened 2018-11-10
Lokole (1) needs essential documentation (2) reply buttons should pre-populate "To" field


holta:
Tracking ticket to keep us on track for Lokole as part of IIAB 6.7's release in December &mdash; many thanks to @aidan-fitz & @c-w for their ongoing help:

- [ ] PR #1305: Teachers need documentation (https://github.com/ascoderu/opwen-webapp/issues/81) building off @c-w's https://github.com/iiab/iiab/pull/1282#issuecomment-435704091 explaining how to:
  - [ ] Suspend & unsuspend accounts
  - [ ] Establish a superuser account to change others' passwords
  - [ ] Delegate such superuser privs to other users
  - [x] #1294 Change the right side of email address (currently appearing as USERNAME@iiab-box.lokole.ca)
    - [ ] https://github.com/iiab/iiab/pull/1294#issuecomment-439044880 How to wipe Lokole's database(s) clean, to freshly install a new version?  (With clean out-of-box settings if possible!)  e.g. `./runrole lokole` doesn't fix the right side of email addresses after PR #1294 merge.
- [x] Replying to email should pre-populate "To" field (https://github.com/ascoderu/opwen-webapp/issues/79)

Refs: https://github.com/iiab/iiab/pull/894 https://github.com/iiab/iiab/pull/1282#issuecomment-435622419

veeara282:
@holta, think we're done with this issue now?

holta:
> @holta, think we're done with this issue now?

If you think there's sufficient testing for now, yes.  The checklist above and the checklist at #1305 may be useful to spot-check future/evolving versions of Lokole, so perhaps bookmark 'em :)

PR #1294, opened 2018-11-12
Update right side of Lokole email address

c-w:
Previously, the email address appeared as `username@iiab-box.lokole.ca`.
With this change, the email address appears as `username@box.lan`

See https://github.com/iiab/iiab/issues/1293

cc @aidan-fitz @holta

### DIFF
--- a/roles/lokole/templates/webapp_secrets.sh.j2
+++ b/roles/lokole/templates/webapp_secrets.sh.j2
@@ -1,4 +1,5 @@
 export OPWEN_STATE_DIRECTORY='{{lokole_run_directory}}'
 export OPWEN_SESSION_KEY='{{ lookup('password', '/dev/null chars=ascii_letters,digits,_ length=32') }}'
 export OPWEN_PASSWORD_SALT='{{ lookup('password', '/dev/null chars=ascii_letters,digits,_ length=16') }}'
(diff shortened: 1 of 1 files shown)


holta:
This seems very low-risk so I'll go ahead and merge it.

Thanks all for continued testing!

holta:
Is there a way to reinstall a fresh version of Lokole on Internet-in-a-Box ?

FYI most IIAB apps allow you do the following: (to freshly install that IIAB app/service)
```
cd /opt/iiab/iiab
./runrole lokole
```
But when I do the above, email addresses remain as USERNAME@box.lokole.ca (instead of USERNAME@box.lan as I'd hoped!)

_Is there a way to wipe Lokole's database(s) clean?
(Ideally with Lokole's out-of-the-box settings wiped completely clean, or some such / similar!)_

c-w:
@holta The configuration values are persisted in a file at `{{ lokole_run_directory }}/webapp_secrets.sh` during the `configure` step, so to update the values I'd assume that this step has to be re-run (or the file changed manually).

This is the code that creates the config file: https://github.com/ascoderu/iiab/blob/382fa66570f57be84c47bb7fdde7166d213f4bcd/roles/lokole/tasks/install.yml#L41-L46

PR #1299, opened 2018-11-16
Add more funky fonts

tim-moody:
Bubble Gum
Finger Paint
Flavors
Freckle Face


holta:
This is smoke-tested / safe to merge right?

tim-moody:
not smoke, but functionally tested, but there is no ansible change, so smoke test highly likely to work.

PR #1301, opened 2018-11-17
handle some old non-canonical file names, like gutenberg

tim-moody:
old names don't parse because of _MM_YYYY pattern instead of YYYY-MM


tim-moody:
functionally tested. no change to ansible so smoke test should work.

PR #1303, opened 2018-11-18
bind9 and dnsmasq fighting over port 53

georgejhunt:
@jvonau -- This seems like an obvious and required fix, but check me. Seems that installing a package starts it up. Would you rather install named, and then disable it?

### DIFF
--- a/roles/4-server-options/tasks/main.yml
+++ b/roles/4-server-options/tasks/main.yml
@@ -10,7 +10,7 @@
 
 - name: Install named / BIND
   include_tasks: roles/network/tasks/named.yml
-  when: named_install
+  when: named_install and not dnsmasq_enabled
   tags: base, named, network, domain
 
 - name: Installing captive portal


holta:
@jvonau can you review @georgejhunt 's suggestion here?

holta:
@georgejhunt have you tested this as solving the problem?

@mrdavidhaag seems to indicate the port 53 collision runs even deeper as he's described within #1306 ?

georgejhunt:
try "systemctl disable systemd-resolvd", and reboot, before running our
installer in the VM

holta:
@georgejhunt is this PR still necessary for #1306 etc?

Or has it been supplanted by other approaches e.g. @jvonau's PR #1311 or similar?

holta:
This is superseded by #1318 (choosing a smaller hammer, that allows folks to install _both_ named/BIND9 _and_ dnsmasq, and choose which to turn on later...)

PR #1300, opened 2018-11-16
Cap2

georgejhunt:
retested in new role location, with py_captive_portal_enabled changed to captive_portal_enabled, online and offline, for android 5,7, windows 7,8,10, macbook OSX 10.13.6, ios11 on a clean build of raspbian 2018-10-9.

PR #1311, opened 2018-11-29
Staging of captive portal

jvonau:


georgejhunt:
I feel the captive portal should be it's own role, and not part of the networking role. Though I recognize that it does add some configuration to dnsmasq, and requires dnsmasq rather than bind9

holta:
@jvonau wrote:
> If CP is to be stand-alone please clarify the use of code in iiab-gen-iptables that is part of the original PR [#1300]

( Let's try to get to the bottom of this during our Wedn call: http://minutes.iiab.io )

holta:
@georgejhunt can you please help @jvonau understand https://github.com/iiab/iiab/blob/master/roles/network/templates/gateway/iiab-gen-iptables#L126-L128 ?

Thanks!

holta:
Discussed with @georgejhunt during today's http://minutes.iiab.io call &mdash; on hold pending @georgejhunt's review of iptables &mdash; possible this iptables rule is causing wider problems ?

PR #1327, opened 2018-12-09
Cap3 [captive-portal under Apache; disconnected from wsgi & iptables]

georgejhunt:
Removes systemd startup of debug wsgi service, and puts the captive portal under apache.
Removes any reference to captive-portal in iptables
Tested, and works, on mac, windows 7,10, iphone, android 5,7
I did a fresh install, found errors, and corrected them. But did not again confirm that a fresh install was flawless. I ran out of time and patience

PR #1334, opened 2018-12-12
captive-portal tweak (code maintainability, not functional changes)

holta:
Builds on PRs #1300, #1327, #1333


tim-moody:
pity so much time is spent on tweaking code and docs cosmetics but no one wants to verify that it works.

PR #1336, opened 2018-12-13
Netplan - WIP

jvonau:
### Fixes Bug
aimed at #1320 
WIP

PR #1349, opened 2019-01-02
Update iiab-hotspot-on & -off commands from dhcpd to dnsmasq

holta:
As explained on PR #1336, thanks to @jvonau


holta:
@jvonau recommends pushing this now, so I will.

PR #1352, opened 2019-01-03
Redefine default_vars.yml from MEDIUM-sized to MIN-sized

holta:
As discussed at length during this morning's call (http://minutes.iiab.io)

CLARIF that the word "default" has 2 meanings going forward:
- New implementers will still "default" to [MEDIUM-sized /opt/iiab/local_vars.yml](http://wiki.laptop.org/go/IIAB/local_vars.yml) when running IIAB's 1-line installer (http://d.iiab.io) allowing them to get to know IIAB, by exploring about a dozen of IIAB's most popular apps/services.  These [highlight] IIAB's best apps/services, from the longer/full showcase at http://FAQ.IIAB.IO (Item 3 [and Item 9]).
- Veteran implementers (typically upgrading IIAB often) often do not want all those "dozen" IIAB apps/services, are typically wedded to their own custom local_vars.yml, and do no want default_vars.yml to change often (imposing infra/services changes).  Sync'ing `default_vars.yml` functionality to the [MIN-sized local_vars.yml](http://wiki.laptop.org/go/IIAB/local_vars_min.yml) going forward (instead of MEDIUM-sized, as has been true to date) is the best way to mitigate this tension between newbie/veteran installers in my opinion.

BACKGROUND: the 3 local_vars.yml files (MIN, MEDIUM, BIG) contain carefully-written explanations for new implementers, to help educate newcomers around their primary IIAB install-time choices as they edit their {MIN, MEDIUM or BIG} /opt/iiab/local_vars.yml, right before they launch their (often multi-hour!) IIAB installation by running "sudo iiab" (which calls iiab-install etc).  Whereas default_vars.yml is more of a technical reference for advanced gurus, containing a laundry-list of most all IIAB variables ever created since the beginning of time.  Recap: /opt/iiab/iiab/vars/default_vars.yml is very important, but its values are overridden by those in /opt/iiab/local_vars.yml


holta:
CLARIF that this PR also:
- ends our universal practice of installing named and dhcpd (while giving you the option in local_vars.yml to change that, if dnsmasq really isn't sufficient)
- ends the practice of installing+enabling Captive Portal in MIN-sized local_vars.yml & default_vars.yml
- `iiab_gateway_enabled: False` is now set in default_vars.yml, per established norms in recent years (set long ago in all 3 local_vars files)
- changes CUPS in MEDIUM-sized local_vars.yml from:
```
cups_install: True
cups_enabled: False
```
...to:
```
cups_install: False
cups_enabled: False
```
...just the same as MIN-sized local_vars.yml.  Whereas CUPS remains as follows in BIG-sized local_vars.yml:
```
cups_install: True
cups_enabled: True
```

holta:
FYI This systemic change (towards more minimal set[s] of IIAB apps/services, without installing the kitchen sink as we used to do!) arises in part because of XSCE's (flawed) assumption from years ago, that most/all services should be installed across the board at XSCE/IIAB install-time &mdash; the assumption had been that these services would later be toggled on/off in the field.

(While in fact such field reconfigurations rarely if ever happened after the install/implementation phase, nor were reconfigs even requested [rarely if ever] as we discovered over more than half a decade.)

holta:
Testing is ongoing, and will continue after this is merged very shortly now.

Thanks to all helping test all 3 sizes of local_vars.yml

PR #1354, opened 2019-01-03
Revert "Lokole: change admin to Admin per IIAB app norms"

holta:
Reverts iiab/iiab#1353 as this managed to break http://box/lokole logins attempted with Admin/changeme

(Apologies @aidan-fitz & @c-w the original ` lokole_admin_user: admin` was indeed appropriate!)

Ref #1293

### DIFF
--- a/roles/lokole/defaults/main.yml
+++ b/roles/lokole/defaults/main.yml
@@ -1,6 +1,6 @@
 # Info needed to install Lokole
 lokole_version: "0.1.26"
-lokole_admin_user: Admin
+lokole_admin_user: admin
 lokole_admin_password: changeme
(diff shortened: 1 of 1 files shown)


veeara282:
Okay, thanks for letting us know!

veeara282:
It does get used in an email address, I think, so case wouldn't make a difference.

holta:
> It does get used in an email address, I think, so case wouldn't make a difference.

While that's true when entering your username (admin or Admin) into http://box/lokole &mdash; I did confirm that uppercase `lokole_admin_user: Admin` within /opt/iiab/iiab/roles/lokole/defaults/main.yml definitely broke the ability to login (with admin or Admin!)

PR #1377, opened 2019-01-10
0.3 patch1

georgejhunt:
Some additional changes that were not included in the first try


tim-moody:
I suspect changing the size value to human readable with break adm console, which as I recall expects to do that on the front end.

holta:
Just FYI this PR includes duplicate copies of commits/patches from months ago making it very hard to read.

Rebasing isn't easy for sure!  But "11 Files changed" (near the top-right of this ticket) is simply not accurate.  It appears the true number is "2 Files changed" here?

In any case, please let's avoid duplicate commits/patches (already merged months ago) in future.  Thereby making our work genuinely visible/understandable to all.

holta:
> Otherwise when we merge 0.3-devel into master, all
Adam's commits into master will be reversed.

I don't believe that's true.  My understanding is that only new commits are layered on.

> Maybe I still do not understand how git works.

Agreed rebasing is annoyingly hard.

(I don't see how already committed patches from months ago are relevant here, but then again I'm not the expert on rebasing...)

holta:
> > Otherwise when we merge 0.3-devel into master, all
> > Adam's commits into master will be reversed.
> 
> I don't believe that's true. My understanding is that [] new commits are [merely] layered on.

(Without affecting the hundreds of other commits that have been merged into master over recent months!)

georgejhunt:
The other issue is that for testing, we'd like to be as close to what
will be the case when we land in master.
We will not have that, if master has diverged since we started testing
in 0.3-devel

holta:
Refs:
https://github.com/iiab/iiab/pull/1393
https://github.com/iiab/iiab-admin-console/pull/121

Issue #935, opened 2018-07-20
[Debian 10] kalite-serve.service fails to start 


arky:
### Expected behavior and actual behavior.
Kalite server fails to start on Debian 10. Python virtual environment bug in Debian / Ubuntu ?  

● kalite-serve.service - Provides the kalite Server
   Loaded: loaded (/etc/systemd/system/kalite-serve.service; enabled; vendor preset: enabled)
   Active: failed (Result: exit-code) since Fri 2018-07-20 09:58:37 EDT; 28s ago
  Process: 2337 ExecStart=/usr/local/kalite/venv/bin/kalite start --port=8008 (code=exited, status=1/FAILURE)

Jul 20 09:58:37 box.lan kalite[2337]:   File "/usr/local/kalite/venv/local/lib/python2.7/site-packages/kalite/packages/dist/cherrypy/lib/cptools.py", 
Jul 20 09:58:37 box.lan kalite[2337]:     import logging
Jul 20 09:58:37 box.lan kalite[2337]:   File "/usr/lib/python2.7/logging/__init__.py", line 26, in <module>
Jul 20 09:58:37 box.lan kalite[2337]:     import sys, os, time, cStringIO, traceback, warnings, weakref, collections
Jul 20 09:58:37 box.lan kalite[2337]:   File "/usr/lib/python2.7/weakref.py", line 14, in <module>
Jul 20 09:58:37 box.lan kalite[2337]:     from _weakref import (
Jul 20 09:58:37 box.lan kalite[2337]: ImportError: cannot import name _remove_dead_weakref
Jul 20 09:58:37 box.lan systemd[1]: kalite-serve.service: Control process exited, code=exited status=1
Jul 20 09:58:37 box.lan systemd[1]: kalite-serve.service: Failed with result 'exit-code'.
Jul 20 09:58:37 box.lan systemd[1]: Failed to start Provides the kalite Server.
~

### Steps to reproduce the problem.

(venv) root@box:/opt# cd /usr/local/kalite/venv/
(venv) root@box:/usr/local/kalite/venv# ls
bin  include  lib  local  pip-selfcheck.json  share
(venv) root@box:/usr/local/kalite/venv# virtualenv . --system-site-packages
Running virtualenv with interpreter /usr/local/kalite/venv/bin/python2
Traceback (most recent call last):
  File "/usr/lib/python3/dist-packages/virtualenv.py", line 21, in <module>
    import logging
[...]

Issue #928, opened 2018-07-19
Update "php_version: 7.1" to 7.2 in debian-10.yml if mcrypt solved?


holta:
See @arky's #902 & PR #904 that reverted [debian-10.yml Line 19](https://github.com/iiab/iiab/blob/master/vars/debian-10.yml#L19) from `php_version: 7.2` to 7.1

holta:
Appears solved by PR #1386 bringing bars/debian-10.yml up to PHP 7.3

PR #899, opened 2018-07-14
Enable Debian 9 rsyslog service

arky:
Ensure that rsyslog service is installed and enabled when sugarizer is
enabled.

Fixes #869 https://github.com/iiab/iiab/issues/869

### Description of changes proposed in this pull request.
When sugarizer is enabled this change installs rsyslog package and starts the service. 

### Smoke-tested in operating system.

Smoked-tested on Debian 9

TASK [2-common : include_tasks] ************************************************
included: /opt/iiab/iiab/roles/2-common/tasks/debian.yml for 127.0.0.1

TASK [2-common : Install Debian 9 rsyslog server when sugarizer is enabled.] ***
changed: [127.0.0.1]

TASK [2-common : Enable Debian 9 rsyslog service when sugarizer is enabled.] ***
ok: [127.0.0.1]

TASK [2-common : include_tasks] ************************************************
skipping: [127.0.0.1]

TASK [2-common : include_tasks] ************************************************
skipping: [127.0.0.1]

$ sudo systemctl list-units |grep rsys
rsyslog.service                                                                          loaded active running   System Logging Service

Issue #878, opened 2018-07-07
Recruit someone to test IIAB on Debian 9 & 10


holta:
[IIAB 6.6 pre-releases](http://download.iiab.io/6.6) are well tested on Raspbian & Ubuntu 18.04, but this is not the case with [Debian 9](https://www.debian.org/distrib/).

@arky & All: can you help find someone to install/test Internet-in-a-Box on Debian 9.4+ on x86_64 virtual machine or actual hardware?

Any body who can help test [[Debian 9.5](https://www.debian.org/News/2018/20180714)] — please run http://download.iiab.io/6.6 to check if the 12+ issues below are resolved or close:

- [x] #864 [i586] Invalid Kiwix Download Link '.bz2'
  - [x] #865 [i586] Kiwix-tools-src doesn't contain bin/ directory
  - [x] #866 Kiwix for i586 and i686 - towards resolving #864 & #865
  - [x] #895 Kiwix-tools code for i586
- [x] #867 calibre-serve Service Fails to start: Missing arguments "--log" "TimeoutStartSec'
- [x] #869 Sugarizer.service: Failed to parse output specifier, ignoring: syslog
  - [x] #899 Enable Debian 9 rsyslog service
- [x] #880 Install script [sometimes] fails to create /opt/iiab/downloads folder
- [x] 2018-07-15 http://download.iiab.io/6.6/load-big.txt tested on [Debian 9.5](https://www.debian.org/News/2018/20180714)
- [x] #902 [Debian 10] Ansible reports os_ver as "OS_not_supported"
  - [x] #904 [Debian 10] [Fixes #902] Generate correct os_ver for Debian 10/Buster Release
- [x] #928 & PR #1386 Update "php_version: 7.1" to 7.3 (& PostgreSQL 10 to 11) in debian-10.yml if mcrypt solved?
- [x] #935 [Debian 10] kalite-serve.service fails to start
- [x] #1386 Fixes for 2019's Debian 10 "Buster" + 3 outstanding issues incl workarounds
  - [ ] #1387 Debian 10 Buster: 3 outstanding issues w/ workarounds + 2 bugs (Kolibri & MongoDB)

CONVERSELY: if a volunteer if not forthcoming, we can move Debian to "secondarily" supported @ https://github.com/iiab/iiab/wiki/IIAB-Platforms (similar to Ubuntu 16.04 LTS, which is now beginning to fade in popularity).

holta:
[...]

PR #1386, opened 2019-01-13
Fixes for 2019's Debian 10 "Buster" + 2 outstanding issues incl workarounds

holta:
BIG-sized http://d.iiab.io/6.7 installed to completion on the [testing branch of Debian 10 "Buster", daily build 2019-01-13](https://www.debian.org/devel/debian-installer/) after forcing the following line into `/etc/os-release` :
```
VERSION_ID="10"
```
(The above allows [iiab-install](https://github.com/iiab/iiab/blob/master/iiab-install) to run [/opt/iiab/iiab/scripts/local_facts.fact](https://github.com/iiab/iiab/blob/master/scripts/local_facts.fact) -> `/etc/ansible/facts.d/local_facts.fact` properly, setting OS_VER to "debian-10" so we're ready for Debian 10's beta releases in coming months and final release by/around mid-2019.)

THIS PR FIXES:
- #928 PHP 7.3 & PostgreSQL 11 go into [vars/debian-10.yml ](https://github.com/iiab/iiab/blob/master/vars/debian-10.yml)
- Nextcloud install is fixed by ending the installation of `php{{ php_version }}-mcrypt` on all OS's released after 2017.  This is accomplished by better self-identifying all OS's within [/opt/iiab/iiab/vars](https://github.com/iiab/iiab/tree/master/vars).  @jvonau can hopefully confirm that mcrypt is completely unnec going forward?

OUTSTANDING ISSUES &mdash; INCL WORKAROUNDS:
- The [3 lines](https://github.com/iiab/iiab/blob/master/roles/httpd/templates/refresh-wiki-docs.sh#L36-L38) in /usr/bin/iiab-refresh-wiki-docs that use lynx to d/l from github.com need to be commented out for now (mysteriously GitHub is blocking all connections from lynx on Debian 10, even when lynx issues exactly the same headers as from Ubuntu/Raspbian or Debian 9!)  Possibly this is a subtle SSL/TLS problem, even when lynx claims to be using the same SSL as on the other distros?!
- ["Stop ejabberd service"](https://github.com/iiab/iiab/blob/master/roles/ejabberd/tasks/main.yml#L62-L68) froze during ejabberd's install, so `systemctl stop ejabberd` had to be run manually.  Installation then continued on its own: relaunching IIAB's install with `sudo iiab` was not even nec this time!

https://github.com/iiab/iiab-admin-console/pull/116 FIXED:

- [Admin Console will now install](https://github.com/iiab/iiab-factory/blob/master/iiab#L203-L205) as obsolete `pip install --download <dir> SomePackage` syntax was removed and replaced with `pip download --dest <dir> SomePackage` in 2 sections of [roles/cmdsrv/tasks/main.yml](https://github.com/iiab/iiab-admin-console/blob/master/roles/cmdsrv/tasks/main.yml).  Specifically `--download` was deprecated after pip 9.x as documented at https://het.as.utexas.edu/HET/Software/Pip/reference/pip_download.html &mdash; and currently Debian 10 uses pip 18.1.  _Errors had been: pip has "no such option: --download"_ in both "TASK [cmdsrv : Download zeromq and python binding]" and "TASK [cmdsrv : Download speedtest-cli]".  PR116 fixed pip syntax and was tested across 3 distros.

Others Refs: #878 #899 #928 #935

### DIFF
--- a/roles/nextcloud/tasks/main.yml
+++ b/roles/nextcloud/tasks/main.yml
@@ -65,7 +65,7 @@
     state: present
   when: is_debuntu and (is_debian_8 or is_debian_9 or is_ubuntu_16 or is_ubuntu_17)
   # NOT NEC TO TEST FOR is_raspbian_8 OR is_raspbian_9 AS /opt/iiab/iiab/vars/<OS>.yml
-  # DEFINES THESE AS SUBSETS OF is_debian_8 OR is_debian_9
+  # DEFINES THESE AS SUBSETS OF is_debian_8 OR is_debian_9 (FOR NOW!)
 
 # we need to install the rpm in order to get the dependencies
 # but we only need to do this the first time
--- a/roles/nextcloud/tasks/main.yml
+++ b/roles/nextcloud/tasks/main.yml
@@ -59,7 +59,7 @@
     state: present
   when: is_debuntu
 
-- name: 'Install php{{ php_version }}-mcrypt -- on "debuntu" distros released prior to 2018. NOTE: php7.1 deprecated mcrypt 1-Dec-2016 and php7.2 dropped it completely 30-Nov-2017, as it should no longer be nec.'
+- name: 'Install php{{ php_version }}-mcrypt IF this is a "pre-2018" distro in the debuntu family. NOTE: PHP 7.1 deprecated mcrypt 1-Dec-2016 and PHP 7.2 dropped it completely 30-Nov-2017, as it should no longer be nec.'
   package:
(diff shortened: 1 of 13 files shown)


holta:
Smoke-tested on Raspbian Lite, Ubuntu 18.04, Debian 10 daily (testing branch).

jvonau:
It's not that mcrypt is not needed but the library was dropped/removed with the 7.2 release of php, whatever code used that library should of been ported to use openssl as a replacement. It would be wise to check with with upstream NextCloud on what is really required.

Issue #1240, opened 2018-10-17
"./iiab-install --reinstall" should check/create essential directories


holta:
Currently https://github.com/iiab/iiab/blob/master/roles/0-init/tasks/fl.yml is only run on the very 1st run of /opt/iiab/iiab/iiab-install

Three or four implementers so far have become confused by the current arrangement where by the existence (or non-existence!) of /etc/iiab/iiab.env has so many consequences:
https://github.com/iiab/iiab/blob/master/roles/0-init/tasks/main.yml#L3-L17
https://github.com/iiab/iiab/blob/master/roles/0-init/tasks/first_run.yml
https://github.com/iiab/iiab/blob/master/roles/0-init/tasks/iiab_ini.yml

@tim-moody suggests 0-init could be improved by checking/creating the essential directories whenever `./iiab-install --reinstall` is run (e.g. during his testing of #1236).

Taking this further...

Every single run of /opt/iiab/iiab/iiab-install could (and eventually should) verify its essential requirements are in place (not just directories, and not just during the very first run) as any well-spec'd program (should!) do.

[We need to wait for a GSoC volunteer to do this, but for sure let's organize the best way(s) to do this, to put on GSoC project list at #1055 or https://github.com/iiab/iiab/projects/1 regardless who architects this!]

holta:
Good enough!

PR #1279, opened 2018-11-03
Enhance "iiab-install --reinstall" to create/verify essential directories (for #1240)

holta:
This is a manual rebase of @jvonau's PR #1246.

Needs review & testing: some of these core IIAB directories might be needed in Stage 1, in which case this might not work?

PR #1246, opened 2018-10-23
Enable "iiab-install --reinstall" to verify essential directories (for #1240)

jvonau:
### Fixes Bug
#1240
### Description of changes proposed in this pull request.
allow iiab-install --reinstall to verify directory structure 
### Smoke-tested in operating system.
has not been tested
### Mention a team member for further information or comment using @ name
@holta

Issue #963, opened 2018-07-24
Move logic from web curl script to iiab-install or factory


jvonau:
Move the bulk of [web curl script](http://download.iiab.io/6.6/install.txt) to iiab-install to avoid duplication and provide easier maintenance or move install.txt to factory and point the url to github.

holta:
Related:
PRs #931 #964 #965

tim-moody:
I'd like to see a separate script for setup: clone repos, create initial dir, etc. before the actual install starts.

jvonau:
[Move to github](https://github.com/iiab/iiab-factory/pull/33/files)

holta:
Refs:
- https://github.com/iiab/iiab-factory/pull/33
- #1160 "Proactively discourage pi/raspberry password compromises"
- http://download.iiab.io/6.7/install.txt (major UX improvements over the past 2 months, so newbie implementers don't forget to reboot into new kernel, ETC)

PR #1397, opened 2019-01-15
move fl.yml

jvonau:
### Fixes Bug
#1240
slimmed down version of #1246 #1279


holta:
Moving fl.yml is a good idea but creation of /etc/iiab should not happen in 2 places.

Rather it should remain a strict precondition to running ./iiab-install (that gracefully warns & educates naive implementers) as outlined in https://github.com/iiab/iiab/pull/1279#issuecomment-450946096

jvonau:
Can you point out where /etc/iiab is being created please?

holta:
Oops I now understand this PR better: it looks ready to go after smoke-testing.

> Can you point out where /etc/iiab is being created please?

It's created automatically by https://github.com/iiab/iiab-factory/blob/master/iiab#L86 or manually in Step 6. of https://github.com/iiab/iiab/wiki/IIAB-Installation#do-everything-from-scratch

In either case, naive implementers are gracefully warned that existence of /etc/iiab and local_vars.yml therein is a (strict) precondition to running ./iiab-install: https://github.com/iiab/iiab/blob/master/iiab-install#L27-L35

Given the summer 2018 modularity decision steers us away from the very sloppy practice of duplicate code / duplicate functionality spread all around: https://github.com/iiab/iiab/issues/963#issuecomment-407559871

PR #1402, opened 2019-01-16
Don't auto reinstall named, dhcpd when install is False

jvonau:
Trigger is change in hostname away from what is set in local_vars.

### DIFF
--- a/roles/network/tasks/main.yml
+++ b/roles/network/tasks/main.yml
@@ -55,22 +55,22 @@
     - hostname
     - domain
 
-- name: Configuring wondershaper
+- name: Configure wondershaper
   include_tasks: wondershaper.yml
   when: wondershaper_install
   tags:
     - network
     - wondershaper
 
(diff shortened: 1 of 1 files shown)


holta:
Thanks @m-anish who uncovered this.

@jvonau writes: to test change the hostname and run iiab-network without this PR bind and named get installed..

@m-anish writes: then, after that manually uninstall named, change hostname [again], and rerun iiab-network, and it should not install.. right

m-anish:
Okay, this PR seems to work.. I did the following test..

1. Ran iiab-install, big without this PR in Appliance mode
2. Change hostname 
3. Ran iiab-network
4. bind9, bind9utils **got installed**
5. apt remove bind9 bind9utils
6. Incorporate this PR's changes
7. Run iiab-install --reinstall
8. Change hostname
9. Ran iiab-network
10. bind9, bind9utils did **not** get installed

PR #1403, opened 2019-01-16
Create nodejs as a shared role for Sugarizer and Node-RED

jvonau:
Removes duplicate code, for easier future changes. Have to confirm current ansible behavior but might result in the nodejs install only being called once when both parent roles are being installed but would install with either parent role.


m-anish:
Okay, I tested this PR in the following manner on two platforms. Ubuntu 18.04 and Debian stretch (which covers both the routes to nodejs install)

1. Do a minimal install + nodered + mosquitto
2. Check <ip>:1880
3. Nodered is accessible
4. Also, from the log, both the installs went smoothly, and I can see node installed. 

I think this should be merged!

PR #1398, opened 2019-01-15
REVISED: Add Node-RED and Mosquitto (MQTT) IIAB playbooks for IoT

m-anish:
This is the updated and fixed PR for #854

PR #1408, opened 2019-01-17
Changes for Node-RED menu to install automatically

georgejhunt:
permits menu for node-red to autopopulate


holta:
Builds off @m-anish's PR #1398

PR #1411, opened 2019-01-18
Simplify node-js installation from nodesource and a variable for version

m-anish:
For 7.0, we should think about simplifying the way nodejs is installed. This PR does that:

1. Introduces a variable to set nodejs version
2. Installs from nodesource regardless of platform

### DIFF
--- a/roles/nodejs/defaults/main.yml
+++ b/roles/nodejs/defaults/main.yml
@@ -1,0 +1,1 @@
+nodejs_version: 8.x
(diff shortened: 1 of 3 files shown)


holta:
@jvonau what's the risk of going with https://github.com/iiab/iiab/pull/1411 here?  (It might even be safer to keep Sugarizer running reliably on Node.js 8.x across all distros...)

holta:
Clarif: PR #1430 extends this to work with Red Hat-family OS's like CentOS 7 too, i.e. our legacy distros that IIAB has always tried to support informally over the years (but to a lesser degree than Raspbian/Ubuntu/Debian).

m-anish:
> Clarif: PR #1430 extends this to work with Red Hat-family OS's like CentOS 7 too, i.e. our legacy distros that IIAB has always tried to support informally over the years (but to a lesser degree than Raspbian/Ubuntu/Debian).

+1, closing this

PR #1414, opened 2019-01-18
Alt

jvonau:
Based on feedback in chat, an alternate implementation of #1411

### DIFF
--- a/roles/nodejs/tasks/main.yml
+++ b/roles/nodejs/tasks/main.yml
@@ -10,11 +10,17 @@
 # This May Change: thanks all for running "apt -a list nodejs" on Buster's
 # daily builds @ www.debian.org/devel/debian-installer/ to keep us informed!
 
+- name: Set up Node.js 10.x apt sources (all debuntu distros)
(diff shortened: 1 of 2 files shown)


m-anish:
I feel the approach in #1411 keeps things simpler. I would prefer that, following discussion with @holta on skype.

holta:
Let's try for PR #1430 instead &mdash; to greatly simplify Ansible playbooks {pbx (Asterix/FreePBX), nodered (Node-RED), sugarizer (Sugarizer)} across OS's/distros &mdash; with the same /usr/bin/nodejs and /usr/bin/npm in every single case.

This avoids having to dance back & forth between /usr/local/bin and /usr/bin across different distros, and streamlining a much tighter/uniform approach.

The Catch: removing nodejs is more confusing as we standardize all Node.js installation to use the nodesource.com method:

- https://github.com/nodesource/distributions#debinstall
- https://github.com/nodesource/distributions#rpminstall

i.e. after installing using nodesource.com, `apt -a list nodejs` gives implementers the false impression that Node.js is not installed.  _While in fact `apt remove nodejs` and `apt purge nodejs` still DO work!_

PR #1415, opened 2019-01-19
Sugarizer should fail the install if nodejs version is incorrect

m-anish:
As the title says, and as we discussed on skype (between me and @holta ) sugarizer should fail the install process if the nodejs version is incorrect. 

Merge after #1411


holta:
The crux of this PR is now part of PR #1430, so it can likely be closed in coming days.

m-anish:
Yep. Lets go with #1430 instead. Closing.

PR #1416, opened 2019-01-20
ejabberd flaky so don't install it by default, even in BIG-sized local_vars.yml

holta:
Why?  The installation of ejabberd has been intermittently very flaky in recent weeks, as reconfirmed by @m-anish as well.


m-anish:
I do believe I owe you a coffee!

Issue #1431, opened 2019-01-29
Node-RED fails on Ubuntu Server 18.04: node-red.service.j2 needs /usr/bin/node-red-pi NOT /usr/local/bin


holta:
@m-anish can you please help investigate?  Node-RED's service is failing a fresh install of IIAB 6.7/master (with PR #1430 for Node 10.x & the new Sugarizer) on a fresh Ubuntu Server 18.04 VM:
```
root@box:~# systemctl restart node-red
root@box:~# systemctl status node-red
● node-red.service - Node-RED
   Loaded: loaded (/etc/systemd/system/node-red.service; enabled; vendor preset: enabled)
   Active: failed (Result: exit-code) since Tue 2019-01-29 16:12:34 EST; 374ms ago
  Process: 2379 ExecStart=/usr/local/bin/node-red-pi --max-old-space-size=128 -v (code=exited, status=203/EXEC)
 Main PID: 2379 (code=exited, status=203/EXEC)

Jan 29 16:12:34 box.lan systemd[1]: node-red.service: Service hold-off time over, scheduling restart.
Jan 29 16:12:34 box.lan systemd[1]: node-red.service: Scheduled restart job, restart counter is at 5.
Jan 29 16:12:34 box.lan systemd[1]: Stopped Node-RED.
Jan 29 16:12:34 box.lan systemd[1]: node-red.service: Start request repeated too quickly.
Jan 29 16:12:34 box.lan systemd[1]: node-red.service: Failed with result 'exit-code'.
Jan 29 16:12:34 box.lan systemd[1]: Failed to start Node-RED.
```

holta:
Conversely/Strangely: Node-RED (http://box:1880 and its systemd service node-red) both continue to work on RPi after I upgraded Node.js 8.x to 10.x (and rebooted).

I do not understand why this same upgrade (Node.js 8.x to 10.x with the new Sugarizer, at PR #1430) somehow caused Node-RED to fail on Ubuntu 18.04 :/

holta:
On Ubuntu Server 18.04, `journalctl -u node-red` shows:
```
Jan 29 16:55:13 box.lan systemd[1]: node-red.service: Service hold-off time over, scheduling restart.
Jan 29 16:55:13 box.lan systemd[1]: node-red.service: Scheduled restart job, restart counter is at 4.
Jan 29 16:55:13 box.lan systemd[1]: Stopped Node-RED.
[...]

PR #1430, opened 2019-01-29
sugarizer v1.1.0 + sugarizer-server dev branch on Node.js 10.x [FAILURE to reach MongoDB]

holta:
To incorporate @llaske's latest fixes (for Node.js 10.x LTS especially) from:
https://github.com/llaske/sugarizer
https://github.com/llaske/sugarizer-server

Refs:
#974 "Sugarizer 1.0.1+ Release Polish"
#1411 "Simplify node-js installation from nodesource and a variable for version"
#1415 "Sugarizer should fail the install if nodejs version is incorrect"
https://github.com/iiab/iiab/pull/1424#issuecomment-458518728 "Sugarizer 1.0.1 -> 1.1.0"
#1429 "Validate Sugarizer Server 1.1.0 on release in/around March 2019"

PR #1432, opened 2019-01-29
Node-RED service fix for Ubuntu Server 18.04 & other OS's too?

holta:
@m-anish please review this as a possible fix for #1431 ?

This approach would deprecate /usr/**local**/bin/node-red-pi (in favor of /usr/bin/node-red-pi) across all OS's.

(Or...do you know of any OS's/distros that still require /usr/**local**/bin/node-red-pi ?)

### DIFF
--- a/roles/nodered/templates/node-red.service.j2
+++ b/roles/nodered/templates/node-red.service.j2
@@ -3,11 +3,12 @@ Description=Node-RED
 After=syslog.target network.target
 
 [Service]
-{% if is_debian_8 or is_debian_9 or is_ubuntu_16 or is_ubuntu_17 %}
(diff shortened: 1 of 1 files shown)


holta:
CLARIF, thanks to @m-anish:

Th underlying problem (#1431) is arising because PR #1430 (like PR #1411) installs node-red-pi to /usr/bin (hopefully something we can standardize around) instead of the apt (and other package managers?) which traditionally install it to /usr/local/bin

So this PR (#1432) should be merged with #1430 if that happens in coming hours/days...to simplify/standardize around /usr/bin/node-red-pi

holta:
This PR (#1432) is effectively now included within PR #1430.

m-anish:
die 8.x die! :-)

PR #1440, opened 2019-01-31
Resolving issue #1050: fix Appliance Mode & expose Calibre-Web port 8083 consistently with others

### LINKED ISSUE #1050: Open Calibre-Web port 8083
Devops generally restrict access to ports running behind a reverse proxy for security reasons. Is there a specific use case need to open port 8083 to public?

>http://192.168.0.x:8083 does not work, as a result of iptables, even when services_externally_visible: true. This is fixable, but perhaps it's not a priority, as URL's like {http://192.168.0.x/books, http://10.8.0.x/books, http://127.0.0.1/books and http://box/books} all work.

https://github.com/iiab/iiab/blob/master/roles/calibre-web/README.rst

holta:
Fixes #1050, even if not everyone agrees understandably, on the plus side local testing of services is now consistent/possible :/

### DIFF
--- a/roles/network/tasks/computed_services.yml
+++ b/roles/network/tasks/computed_services.yml
@@ -62,7 +62,7 @@
     value: "{{ squid_enabled }}"
   - option: wondershaper_enabled
     value: "{{ wondershaper_enabled }}"
-  - option: iiab_network_mode(applied)
+  - option: iiab_network_mode_applied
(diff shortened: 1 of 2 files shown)


holta:
Merged after testing ./iiab-network on Raspbian & Ubuntu 18.04 -- please test more broadly in Appliance Mode etc.

PR #1447, opened 2019-02-03
Enforce Node.js precondition (that no other version is installed)

### LINKED ISSUE #1445: roles/nodejs should test for any pre-installed version(s) of Node.js
@m-anish likely IIAB should refuse to proceed during installation if [roles/nodejs](https://github.com/iiab/iiab/blob/master/roles/nodejs/tasks/main.yml) detects that a different version of Node.js is already installed?

(While issuing an explanatory warning of course, explaining which version of Node.js is already installed, and which is needed!)

Option: Ansible playbook roles/nodejs need not reinstall Node.js 10.x if a good one is already installed ??

Please anyone who can, feel free to fill in this stub/comment here...designing or coding this up:
https://github.com/iiab/iiab/blob/master/roles/nodejs/tasks/main.yml#L6-L8

holta:
Fixes #1445 and cleans up [roles/nodejs/tasks/main.yml](https://github.com/iiab/iiab/blob/master/roles/nodejs/tasks/main.yml)


holta:
Tested on Ubuntu 18.04 with and without (various) pre-installed versions of nodejs.

PR #1453, opened 2019-02-05
Change iiab.j2 to rpi.j2 to fix iiab-network dup route failings?

### LINKED ISSUE #1452: iiab-network was failing at "Restart the networking service if appropriate" due to duplicate routes
@georgejhunt on an RPi 3 machine running Raspbian Lite ("192.168.0.131") ./iiab-network was failing each time it was run &mdash; even after reboot:

> TASK [network : Restart the networking service if appropriate] *****************
fatal: [127.0.0.1]: FAILED! => {"changed": false, "msg": "Unable to start service networking: Job for networking.service failed because the control process exited with error code.\nSee \"systemctl status networking.service\" and \"journalctl -xe\" for details.\n"}

@jvonau asked me to try out testing using rpi.j2 instead of iiab.j2 within the `Copy the bridge script for RPi` stanza here:
https://github.com/iiab/iiab/blob/master/roles/network/tasks/rpi_debian.yml#L42

i.e. trying this one...
https://github.com/iiab/iiab/blob/master/roles/network/templates/network/iiab.j2

instead of this one...
https://github.com/iiab/iiab/blob/master/roles/network/templates/network/rpi.j2

Quick testing on this particular RPi 3 (192.168.0.131) shows this fixes the problem &mdash; after I made the above change, and then ran ./iiab-network &mdash; dnsmasq now stays running even after reboot, yay!

Can we consider a PR doing exactly this?  (Whether this happens in IIAB 6.7 or more likely afterwards!)

holta:
Fixes #1452 it appears?

Thanks @georgejhunt & @jvonau: let's take as much time as we need to review & test this, to ensure it's appropriate and safe, whether this (or something like this fix!) are part of IIAB 6.7 or later.


holta:
Pushing this experimentally towards solving #1519 WiFi DNS/dnsmasq issues for br0.

Accompanies @jvonau's PR #1525 to fix WiFi installs of IIAB on RPi especially.

Issue #1465, opened 2019-02-08
Include Minetest client software for Windows/Android for offline communities? [Luanti, Minecraft]


holta:
IIAB could possibly offer http://box/software as a synonym for the pre-existing http://box/local_content == http://box/usb

Reason: what client software to offer each community tends to be highly-dependent on local needs / local language / curriculum / IT particulars / etc...as outlined @ https://github.com/iiab/iiab/pull/1454#issuecomment-461515928

Refs: PR #1454, PR #1461, #1462, #1464

holta:
These 3 (from https://www.minetest.net/downloads/) are probably enough for starters:
- https://github.com/minetest/minetest/releases/download/0.4.17.1/minetest-0.4.17.1-win64.zip 14.0 MB
- https://github.com/minetest/minetest/releases/download/0.4.17.1/minetest-0.4.17.1-win32.zip 13.7 MB
- https://github.com/minetest/minetest/releases/download/0.4.17.1/Minetest-0.4.17.1-2.apk 11.3 MB

(Less than 40MB total.)

holta:
Something to think about in 2025?

Or, conversely, we can close this ticket — if it's just not the best way forward in the end, in this new era where parents/educators/students in almost all countries can now install the very latest Luanti (formerly Minetest) client software more directly via Android Play Store etc — as discussed here:

- PR #3985 
- https://github.com/iiab/iiab/blob/master/roles/luanti/README.rst

tim-moody:
Side loading android is problematic, but the windows and/or Linux client might be worthwhile. Taking it a step at a time makes sense.

Issue #1462, opened 2019-02-08
Minetest (Minecraft clone) for Ubuntu etc?


holta:
So far PRs #1454 #1461 have only implemented the Minetest server on RPi.

It would be great if this ran on Ubuntu 18.04+ & Debian 9.7+ if not others!

holta:
http://www.nathansalapat.com/downloads/0.4.17.1.tar.gz (14MB) is now backed up to http://download.iiab.io/packages/minetest-0.4.17.1.tar.gz in case we need it later, using `iiab_download_url: http://download.iiab.io/packages` as defined in [default_vars.yml](https://github.com/iiab/iiab/blob/master/vars/default_vars.yml#L17)

holta:
@tim-moody is working on it.  There's a good chance the package will/might work on Debian too.

holta:
Appears completed by PR #1471 

@tim-moody is investigating further Minetest additions like Carbone NG (carbone-ng) and flat world for TK Kang.

PR #1454, opened 2019-02-06
minetest for rpi initial checkin [Minecraft clone]

tim-moody:
minetest server

PR #1471, opened 2019-02-09
Minetest

tim-moody:
tested on ubuntu 18.04.1 vm and rpi
probably should pull some constants out of default vars and back to defaults in play


holta:
> tested on ubuntu 18.04.1 vm and rpi

Great, what was the trick to finally get it working on Ubuntu?

> probably should pull some constants out of [default_vars.yml] and back to defaults in play

Got time to do that before Monday / RC3?

PS let's recruit others to test on Debian 9.7+ if poss...

Refs: #1454 #1461 #1462 #1464 #1465 #1472

holta:
Should Minetest become part of local_vars_big.yml, particularly if Debian support is confirmed?

And/or should we first ship its offline client software? (#1465)

PR #1473, opened 2019-02-10
Minetest README.rst, default_vars.yml, defaults/main.yml, local_vars_big.yml

holta:
Please [edit directly](https://github.com/holta/iiab/blob/minetest-doc/roles/minetest/README.rst) so this can be merged prior to Monday's IIAB 6.7 RC3!


holta:
Further testing on Debian 9.7+ would be appreciated!

Issue #1456, opened 2019-02-06
Add http://box/nodered (and possibly also http://box/sugar ?)


holta:
@m-anish if you can add http://nodered (if not others) that'd be great!

Context: teachers love these rapid-access short URL's so kids don't get distracted by http://box and go immediately to the lesson material.

Ref: #923 "Failing/Short URL's—parent tkt to consolidate diverse issues"

tim-moody:
minetest is not a web app, so there is no point in having box/minetest. a browser visiting it will get nothing

holta:
@m-anish & all decided not to bother with http://box/sugar during our community/team call today (http://minutes.iiab.io) as that might get complicated, and http://box/sugarizer is enough.

Issue #923, opened 2018-07-18
Failing/Short URL's—parent tkt to consolidate diverse issues


holta:
Internet-in-a-Box URL's are failing in many annoying ways.

So this ticket serves to consolidate all these glitches that drive teachers and implementers crazy.

Such that the failure patterns can be attacked together—more coherently building up a knowledge base of proven techniques—for developers to implement the clean/short URL's that teachers and students wants on Day 1:

- [x] #111 -> PR #967 Nextcloud not accessible from VPN
- [ ] #201 Automate re-homing of http://HOSTNAME/wordpress? (instead of RELOCATE flag)
- [ ] #529 Calibre 2.x images & Calibre 3.x books don't appear @ http://box/books (http://box:8080 works, and http://box/calibre-web worked with PR #999 prototype!)
- [ ] #551 How to mitigate campuswide failures for Nextcloud, WordPress, Moodle, Elgg etc?
- [x] #851 -> #898 Fix http://box/sugarizer (worked with Sugarizer 0.7, 0.8, 0.9)
- [ ] #852 Fix http://box:3000 for those who want privacy & original/clean/short Kiwix URL
- [x] #913 PR #1412 learningequality/kolibri#4127 http://box/kolibri should work (not just http://box:8009)
- [ ] #10 #941 http://10.0.0.x/moodle must be customized on installation
- [x] PRs #980 #985 #1060 Enable http://box/kiwix w/o trailing slash
- [x] PR #992 http://10.0.0.x/wordpress/ not displaying correctly & http://10.0.0.x/wordpress/wp-admin/ not working ("UPSTREAM ISSUE, INHERENT TO WORDPRESS")
- [x] #1109 Clean up RedirectMatch rules (for Kolibri & others?)
- [x] #1196 -> PR #1213 Calibre-Web's http://box/books restored to work with Apache, as if http://localhost:8083
- [x] #1262 Shorten http://box/awstats/awstats.pl to http://box/awstats
- [x] #1282 http://box/lokole and http://box/lokole/static are currently both failing (certainly on Ubuntu 18.04; I've not tested other OS's)
- [x] #1456 Add box/nodered (and possibly also box/sugar ?)
  - [x] PR #1475 Add /sugar RedirectMatch to /sugarizer — for http://box/sugar
  - [x] PR #1476 Node-RED reverse proxy — for http://box/nodered
[...]

PR #1475, opened 2019-02-10
Add /sugar RedirectMatch to /sugarizer — for http://box/sugar

m-anish:
Creates a simple redirect from sugar to sugarizer.


holta:
For #1456

Ref: #923

PR #1476, opened 2019-02-10
Node-RED reverse proxy — for http://box/nodered

m-anish:
This PR creates a reverse proxy for the node red service on /nodered


holta:
Fixes #1456

Ref: #923

m-anish:
Think this is good to merge. 

Feel free to do a smoke test though.

holta:
> Think this is good to merge.
> 
> Feel free to do a smoke test though.

Strange: http://box/nodered doesn't work for me on Ubuntu Server 18.04.1 and Raspbian Lite after installing this PR. No matter what URL I try (http://WHATEVER/nodered), in every case local & remote browsers show 404 error:
```
Cannot GET /nodered
```

PR #1490, opened 2019-02-11
Minetest: rename service template

tim-moody:


holta:
This completes the fix begun with PR #1488

Note that minetest-server.service is installed automatically on Ubuntu & Debian.

Hence this rename from minetest-serve to minetest-server on all OS's incl Raspbian.

tim-moody:
there was no reason to rename except for rpi as the package determines the name elsewhere

PR #1491, opened 2019-02-11
Docs: pbx/README.rst, tasks/asterisk.yml, tasks/freepbx.yml, minetest/README.rst

m-anish:
Made the description clearer.


holta:
I'm not sure PR #1491 is correct, as is.

Even when installing IIAB as a gateway, several of us do its web-based administration from the WAN side where pbx.lan does not resolve.

Warning is important to a larger set of situations (no less people who have no idea what an Appliance/LAN Controller/Gateway are, these people are increasingly important to us!)

m-anish:
I guess best to leave it as is.

Issue #1387, opened 2019-01-13
Debian 10 Buster: remaining issues/workarounds e.g. with MongoDB/Sugarizer & Monit


holta:
Installing IIAB 6.7/master (http://d.iiab.io/6.7) on [Debian 10 "Buster"](https://www.debian.org/devel/debian-installer/) pre-releases can work!  With these 3+2 known caveats:

- [x] Ensure the line `VERSION_ID="10"` (or `VERSION_ID="10.0"`) exists in `/etc/os-release`.  This allows [iiab-install](https://github.com/iiab/iiab/blob/master/iiab-install) to run [/opt/iiab/iiab/scripts/local_facts.fact](https://github.com/iiab/iiab/blob/master/scripts/local_facts.fact) -> `/etc/ansible/facts.d/local_facts.fact` properly, setting OS_VER to "debian-10" so we're ready for Debian 10's pre-releases in coming months.  (Final releases will contain a similar such line by/around mid-2019.)
- [x] The [3 lines](https://github.com/iiab/iiab/blob/master/roles/httpd/templates/refresh-wiki-docs.sh#L36-L38) in /usr/bin/iiab-refresh-wiki-docs that use lynx to download from github.com (for http://box/info documentation) need to be commented out for now &mdash; mysteriously GitHub is blocking all connections from lynx on Debian 10, even when lynx issues exactly the same headers as from Ubuntu/Raspbian or Debian 9!  Possibly this is a subtle SSL/TLS problem, even when lynx claims to be using the same SSL as on the other distros?!
- [x] ["Stop ejabberd service"](https://github.com/iiab/iiab/blob/master/roles/ejabberd/tasks/main.yml#L62-L68) froze during ejabberd's install (when installing BIG-sized IIAB) so `systemctl stop ejabberd` had to be run manually.  Installation then continued on its own: relaunching IIAB's install with `sudo iiab` was not even nec this time!

This/these issue(s) below appear more serious?

- [x] Kolibri 0.11.1 -> 0.12.1 (http://box:8009 -> http://box/kolibri) doesn't seem to work on Debian 10, despite its IIAB install completing without obvious errors &mdash; please run `cd /opt/iiab/iiab` then `./runrole kolibri` to help us look more close if you can!
- [ ] MongoDB (for Sugarizer) not installable due to missing packages: #1437
  - [x] Workaround implemented, bypassing MongoDB & Sugarizer on Debian 10+ for now: PRs #1850 #1852
[...]

PR #1594, opened 2019-04-11
Debian 10 needs auto just like 9

georgejhunt:
### Fixes Bug
Stalled twice on network restart on real debian 10 hardware. Interface not up -- appliance mode


holta:
Ref: #1387

PR #1636, opened 2019-05-07
Increase hostapd_wait from 1 to 5 sec (?) at end of roles/network (restart.yml) for dnsmasq

holta:
Thanks to @jvonau and Tony Anderson, this fix is essential for IIAB's initial install when the network role (after Stage 0-9) is running on certain hardware.

Note the problem ocurred during network role near the very end &mdash; as you can see if you're running IIAB's roles/network manually:
```
cd /opt/iiab/iiab
./iiab-network
```
Ref: #1306

### DIFF
--- a/roles/network/defaults/main.yml
+++ b/roles/network/defaults/main.yml
@@ -40,7 +40,7 @@ wondershaper_upspeed: "1024"
 
 # Wi-Fi
 host_ssid: IIAB
-hostapd_wait: 1
+hostapd_wait: 5
 host_wifi_mode: g
 host_channel: 6
 host_wireless_n: False


holta:
Tony Anderson confirms this helps on his Acer XC-885 desktop machine, so @jvonau agrees we should merge this now, so help many others.

_PS please all feel free to revise to 3 seconds or 10 seconds as appropriate?!_

holta:
@jvonau: I've beefed up the comment @ [roles/network/tasks/restart.yml#L56](https://github.com/iiab/iiab/blob/master/roles/network/tasks/restart.yml#L56) to explain why this is necessary &mdash; is the following accurate / good enough?
```
- name: Waiting {{ hostapd_wait }} seconds for network to stabilize (iiap-gen-iptables will fail if dnsmasq isn't ready!)

PR #1596, opened 2019-04-15
iiab-footprint [iiab-diagnostics to speed up remote support of networks especially?]

georgejhunt:
Responds to #1575 
Creates a more automated means of gathering together network settings, and communicating them back to developers.

Issue #1575, opened 2019-03-19
More automated reporting of networking/similar issues


holta:
I'm thinking we should automate reporting of networking [and other] bugs so we can be 10X more efficient during the initial phases of debugging especially.  Instead of spending days-upon-days and weeks-upon-weeks gathering context each time (whereupon we lose people often, permanently in many cases...)

Networking issues will never entirely go away, as we've discovered over 7 years now &mdash; nor will other bugs of course &mdash; as there will always Yet Another wacky networking topology thrown at us, Etc.  No matter what we tell implementers we simply cannot control them.

So it would seem wise that IIAB include a script to auto-publish the output of 10-or-so key log files & network/configuration files, so initial reporting becomes an order of magnitude more efficient going forward.  Non-technical implementers (and there are many) will happily partake if they're leading themselves and all of us forward.  e.g. the best would be 1 short command, that tests for Internet and then [with consent] uploads all these files to pastebin/GitHub or similar.

And also somehow emails us at the same time?

A more mature version of this "iiab-diagnostics" command (or whatever we call it) could also dump everything into /opt/iiab/diagnostics (or wherever) with crisp instructions for implementers to (1) move these files to USB stick (2) publish these to the web from Windows [or Android] typically.

Refs: #1306 #1519 #1520 #1527 #1536 #1569 #1572

holta:
If folks can add to this ticket their recommendations of the 10-or-so-most-useful IIAB config & networking & log files that would be Extremely Useful!

Also: which such files might will be the most privacy-sensitive, and as such perhaps should not be published to GitHub etc?

holta:
What IIAB files are most useful for remote support / remote diagnostics?  _DRAFT:_
- /etc/iiab/local_vars.yml
- /etc/iiab/iiab.env
- /etc/iiab/iiab.ini
- /opt/iiab/iiab/iiab-install.log
- /opt/iiab/iiab/iiab-network.log (if `./iiab-network` has been run)
- /opt/iiab/iiab/iiab-debug.log (if `./runrole <ROLE>` has been run)
- /library/www/html/home/menu.json
- output of `uname -a`
- output of `ip a`
- output of `brctl show`
- /etc/resolv.conf & other DNS files?
- hostapd &/or wpa_supplicant config files ?
- is it possible to auto-verify whether or not an X Windows-like graphical desktop is installed?
- _output of other networking files & commands?_

mikkokotila:
[...]

PR #1645, opened 2019-05-09
./install-support prompts for openvpn_handle & confirms resulting 10.8.0.* IP address

holta:
Also toughens up the deprecated command /usr/bin/iiab-handle &mdash; for those who insist on a temporary /etc/iiab/openvpn_handle (that will be wiped during the next Ansible run!)

Tangentially related to: #1575, PR #1596

### DIFF
--- a/iiab-support.yml
+++ b/iiab-support.yml
@@ -1,4 +1,3 @@
----
 - hosts: all
   become: yes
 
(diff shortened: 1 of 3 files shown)


holta:
`./install-support` also now runs faster, bypassing most Ansible stuff in Stage 1, only running the core playbook ([roles/openvpn](https://github.com/iiab/iiab/tree/master/roles/openvpn)).

Yes this assumes Stage 1 ([1-prep](https://github.com/iiab/iiab/tree/master/roles/1-prep)) was already run at an earlier date, but I think mandating that is a very fair compromise for a much faster runtime, particularly given Stage 1 = 1-prep is run from the get-go on most all IIAB systems.

Heavily tested on Ubuntu 18.04 and Raspbian.

Will merge shortly, to amp up "fieldback" / grassroots community testing...if further improvements are later appropriate we'll make those happen.

holta:
Later on `./install-support` should be further improved, to advertise the `/usr/bin/iiab-diagnostics` command, e.g. after PR #1596 converges into something the public (or at least newbie implementers) can easily grasp.

PR #1667, opened 2019-05-16
Nextcloud 15.0.x on Debian 9 & Raspbian 9 (PHP 7.1+ is missing) -- but Nextcloud 16.0.y on all others distros

holta:
Needs testing to be sure!

Refs:  #1324, PR #1614, #1615, #1649

Issue #1632, opened 2019-05-02
Ansible 2.8+ ADVISORY: avoid warnings by using "when: var | bool" for top-level BARE vars (in case they're strings, instead of boolean)


holta:
`apt` now installs Ansible 2.8.0rc2 automatically as part of IIAB installs, as a result of https://launchpad.net/~ansible/+archive/ubuntu/ansible

This is a surprise given Release Candidates were not pushed by apt in the past, and I had not expected we'd face Ansible 2.8.x until 2.8.0's official release date of 2019-05-16.

Ansible 2.8 Porting Guide
https://docs.ansible.com/ansible/devel/porting_guides/porting_guide_2.8.html

In any case, this new Ansible 2.8.x issues many warnings we need to address in due course, for example:

> [DEPRECATION WARNING]: Distribution Ubuntu 18.04 on host 127.0.0.1 should use
/usr/bin/python3, but is using /usr/bin/python for backward compatibility with
prior Ansible releases. A future Ansible release will default to using the
discovered platform python for this host. See https://docs.ansible.com/ansible/
2.8/reference_appendices/interpreter_discovery.html for more information. This
feature will be removed in version 2.12. Deprecation warnings can be disabled
by setting deprecation_warnings=False in ansible.cfg.

Whereas on Raspbian, Ansible 2.8.x shows this instead:

> [WARNING]: Platform linux on host 127.0.0.1 is using the discovered Python interpreter at /usr/bin/python, but future installation of another Python interpreter could change this. See https://docs.ansible.com/ansible/2.8/reference_appendices/interpreter_discovery.html for more information.

Running a BIG-sized install of IIAB yielded 222 of these warnings:

> TASK [0-init : Set port 443 for Admin Console if adm_cons_force_ssl] *********** [DEPRECATION WARNING]: evaluating adm_cons_force_ssl as a bare variable, this behaviour will go away and you might need to add |bool to the expression in the future. Also see CONDITIONAL_BARE_VARS configuration toggle.. This feature will be removed in version 2.12. Deprecation warnings can be disabled by setting deprecation_warnings=False in ansible.cfg.

As well as these 7+ others:

> 2019-05-02 15:27:49,566 p=9952 u=root |  TASK [0-init : Add 'runtime' variable values to /etc/iiab/iiab.ini] ************
...
[...]

PR #1696, opened 2019-05-25
Revise boolean syntax for Ansible 2.8 e.g. 'when: X' -> 'when: X | bool' (eliminates most warnings)

holta:
The crux was converting 248 instances of 'when: X' -> 'when: X | bool' (i.e. Ansible 2.8's annoying new/safer syntax) by running:
```
cd /opt/iiab/iiab
find . -type f -exec sed -i 's/[#[:blank:]]when:[[:blank:]][0-Z_]*$/& | bool/' {} \;
```
Thereby resolving most of #1632

ALSO:
- I avoided 12 instances of `changed_when: False` by using `[#[:blank:]]` above
- I manually fixed 2 instances of `grep -r 'when:[[:blank:]][0-Z_]*[[:blank:]]*#'` (in roles/sugarizer/tasks/main.yml and roles/nextcloud/tasks/main.yml)
- I manually fixed 3 instances of `grep -r 'when:[[:blank:]][0-Z_]*[[:blank:]]*}'` (in roles/sugarizer/meta/main.yml and roles/nodered/meta/main.yml )
- I cleaned up 5 instances of `: false` to `: False` and 6 instances of `: true` (to either `: True` or `: yes`) after verifying each
- Duplicate when: clauses were fixed in nextcloud/tasks/main.yml, arising from PR #1667
- I ran `grep -r 'when:[0-Z_]'` and `grep -r 'when:[[:blank:]][[:blank:]]' | wc` to make sure none of those instances existed
- /opt/iiab/iiab-admin-console has about 17+ similar instances that trigger Ansible 2.8's bool warning, but it might be easier to fix those manually

PS this was smoke-tested with a BIG-sized IIAB on Ubuntu 18.04 using "./iiab-install --reinstall" but it should also be tested with a fresh/clean install on another OS ideally.

### DIFF
--- a/roles/0-init/tasks/main.yml
+++ b/roles/0-init/tasks/main.yml
@@ -109,7 +109,7 @@
 #  set_fact:
 #    mongodb_install: True
 #    mongodb_enabled: True
-#  when: sugarizer_enabled
+#  when: sugarizer_enabled | bool
 
 # There might be other db's
 - name: Turn on both vars for PostgreSQL if moodle_enabled or pathagar_enabled
--- a/roles/1-prep/tasks/main.yml
+++ b/roles/1-prep/tasks/main.yml
@@ -9,7 +9,7 @@
       - uuid-runtime
       - sudo
     state: present
-  when: is_debuntu
+  when: is_debuntu | bool
 
 - name: Does /etc/iiab/uuid file exist?
   stat:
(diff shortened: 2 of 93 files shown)


holta:
Also smoke-tested on BIG-sized Raspian and Ubuntu 16.04, using `date; time ./iiab-install --reinstall`

_Thanks to everyone who can further test fresh installs of http://download.iiab.io on any/similar OS!_

PR #1697, opened 2019-05-25
Not just is_ubuntu_18: make U19/U20 viable [Debian 10 already OK?] [extends use of php{{ php_version }}-sqlite3 to future OS's for now...somewhat controversially]

holta:
Building off PR #1696

Should be evaluated/tested in context with:

#1387 Debian 10 Buster: 3 outstanding issues w/ workarounds + 2 bugs (Kolibri & MongoDB)

### DIFF
--- a/roles/calibre/tasks/debs.yml
+++ b/roles/calibre/tasks/debs.yml
@@ -46,7 +46,7 @@
 
 - name: Install/Upgrade to Calibre testing .deb's - target Ubuntu 16.04 (not rpi and not ubuntu_18)
   command: scripts/calibre-install-latest.sh
-  when: not is_rpi and (is_ubuntu_16 or is_debian_9) and internet_available
+  when: (not is_rpi) and (is_ubuntu_16 or is_debian_9) and internet_available
   #when: not is_rpi and not is_ubuntu_18 and internet_available
 
 - name: Install/Upgrade to Calibre unstable .deb's IF calibre_unstable_debs
--- a/roles/calibre/tasks/debs.yml
+++ b/roles/calibre/tasks/debs.yml
@@ -46,7 +46,8 @@
 
 - name: Install/Upgrade to Calibre testing .deb's - target Ubuntu 16.04 (not rpi and not ubuntu_18)
   command: scripts/calibre-install-latest.sh
-  when: not is_rpi and not is_ubuntu_18 and internet_available
+  when: not is_rpi and (is_ubuntu_16 or is_debian_9) and internet_available
+  #when: not is_rpi and not is_ubuntu_18 and internet_available
 
 - name: Install/Upgrade to Calibre unstable .deb's IF calibre_unstable_debs
   command: scripts/calibre-install-unstable.sh
--- a/roles/calibre/tasks/debs.yml
+++ b/roles/calibre/tasks/debs.yml
@@ -5,6 +5,8 @@
 # /opt/iiab/iiab/vars/raspbian-9.yml
 # /opt/iiab/iiab/vars/raspbian-10.yml
 # /opt/iiab/iiab/vars/debian-10.yml
+#
+# As viewable live @ https://github.com/iiab/iiab/tree/master/vars
 
 # If you want the latest Calibre, run the appropriate script below, standalone.
 # HOWEVER: it's strongly suggested you wait for apt (blessed by your OS!) to
--- a/roles/calibre/tasks/debs.yml
+++ b/roles/calibre/tasks/debs.yml
@@ -1,8 +1,9 @@
-# roles/calibre/tasks/main.yml REQUIRES calibre_via_debs (TO BE True) BEFORE
-# CALLING THIS SCRIPT.  As of 2018-10-23 this is set in only 3 places:
+# calibre_via_debs MUST BE TRUE IN ORDER FOR roles/calibre/tasks/main.yml TO
+# INVOKE THIS SCRIPT. As of 2019-05-30 calibre_via_debs is set by these 4 OS's:
 #
-# vars/raspbian-9.yml
 # vars/raspbian-8.yml
+# vars/raspbian-9.yml
+# vars/raspbian-10.yml
 # vars/debian-10.yml
 
 # If you want the latest Calibre, run the appropriate script below, standalone.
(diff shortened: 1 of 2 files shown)


holta:
@tim-moody and @georgejhunt is it possible that `php{{ php_version }}-sqlite` and `php{{ php_version }}-sqlite3` are no longer used?

This came up b/c http://box.lan/admin would seem to work fine on Ubuntu 16.04 even without these (!) as you can see from the absence of these packages here:
```
root@box:~# apt -a list | grep php | grep sqlite

php-sqlite3/xenial-updates,xenial-updates 1:7.0+35ubuntu6.1 all
php-sqlite3/xenial,xenial 1:7.0+35ubuntu6 all
php7.0-sqlite3/xenial-updates,xenial-security 7.0.33-0ubuntu0.16.04.4 amd64
php7.0-sqlite3/xenial 7.0.4-7ubuntu2 amd64
```
So @jvonau and I are asking if you can help us shed light on why IIAB occasionally installs these.  It does not (seem?) that CMDSRV is using either based on https://paste.ubuntu.com/p/Wq2xNbDtWP/ ?

(In any case, https://github.com/holta/iiab/blob/ubuntu-18-19-20/roles/httpd/tasks/main.yml#L26-L37 should be commented out entirely if indeed phpXYZ-sqlite[3] are no long in use?!)

holta:
@tim-moody writes "I think we decided that because sqlite3 and php are part of the base install the connector should be too."

holta:
_Just for the record in case this comes up again..._

Here's the current output with IIAB 7.0 pre-release on Raspbian 9:
```
root@box:~# apt -a list | grep php | grep sqlite

php-sqlite3/stable 1:7.0+49 all
php7.0-sqlite3/stable,now 7.0.33-0+deb9u3 armhf [installed]
```
Here's the current output with IIAB 7.0 pre-release on Ubuntu 18.04:
```
root@box:~# apt -a list | grep php | grep sqlite

php-sqlite3/bionic,bionic 1:7.2+60ubuntu1 all
php7.2-sqlite3/bionic-updates,bionic-security,now 7.2.17-0ubuntu0.18.04.1 amd64 [installed]
php7.2-sqlite3/bionic 7.2.3-1ubuntu1 amd64
```

tim-moody:
what's the actual issue? does anything break?

holta:
> what's the actual issue? does anything break?

Nothing breaks.

This arose as I'm auditing/reviewing each use of `is_ubuntu_18` to see what was really intended, what was actually needed, and what might make sense going forward.

Aside: Likewise each use of `is_raspbian_9` and `is_debian_9` should be audited/reviewed too at some point in future, as we transition to `is_raspbian_10` and `is_debian_10` in coming months.  A variable `is_debuntu_10` could later be created (if absolutely necessary).

tim-moody:
I think the is_xxx strategy is running out of steam. for each operation we will get an increasingly long list of when clauses which will get slower and slower. we should move to per os yml files that can be included and only do one os.

tim-moody:
as an example, imagine that each os vars file had a common_profile variable that is either 'generic' for legacy or something that describes the broadest set of packages for that os. 2-common main task would then include {{ common_profile }}, so going forward the 'when's would start to disappear while old oses still worked.

holta:
> we should move to per os yml files that can be included and only do one os.

One concern is that (in most cases, but not all cases!) an inflationary number of yml files can create its own form of chaos.

One example is Calibre.  Creating a different yml for each OS created a lot of code duplication which became much more expensive to maintain / harder to understand as time went on.  i.e. as a consequence of having many more lines of code to maintain, the intent of the code became harder the grasp/glean, and much harder to maintain as time went on, causing bugs / code debt to accumulate.

tim-moody:
no idea what you're talking about. neither calibre nor calibre-web has per os yml file. as long as ansible only parses files used, which it seems to, individual files will be faster. and they are way easier to read as they contain far fewer conditionals. it's true that there may be some redundancy, but it falls away over time as a given os becomes stable and then no longer used. we currently carry the burden of every os ever contemplated in every file and the complexity goes up exponentially with each new os or release.

holta:
I don't want to criticize Calibre, nor the many people who tried very hard to make it stable on RPi!

But just in the case of that example, these 6 files showcase an accumulating number of code paths attempted across different OS's month-by-month (for what is essentially a single job, whether yml files, or bash files, or whatever) creating a growing danger of bitrotting / expensive-to-maintain code:

- calibre-install-latest-rpi-plus.sh
- calibre-install-latest-rpi.sh
- calibre-install-latest.sh
- calibre-install-packages.sh
- calibre-install-pinned-rpi.sh
- calibre-install-unstable.sh

(Above are currently in https://github.com/iiab/iiab/tree/master/scripts ...as intended to juggle the diverse requirements on Debian 9, Debian 9, Debian 10, Ubuntu 18...)

These very real 'pain points' will of course never quite be sane, as a result of the fact these 3-or-more OS's and at least 2 HW architectures keep "interbreeding", such that an ever-evolving number of code paths are needed in some cases...it gets ugly fast in such cases...so in that case I chose to back off as the more-often-than-monthly churn became too expensive to test &mdash; with absent/broken dependencies requiring a code change almost every month :/

holta:
@georgejhunt do you use php{{ php_version }}-sqlite3 anywhere in IIAB?

At the moment it would seem to be completely unused.

(Thanks for clarifying!)

holta:
> All the maps use php-sqlite for the serving up of tiles

Thanks @georgejhunt for clarifying above.  Should https://github.com/holta/iiab/blob/ubuntu-18-19-20/roles/httpd/tasks/main.yml#L26-L48 perhaps be moved to https://github.com/iiab/iiab/blob/master/roles/osm-vector-maps/tasks/main.yml ?

georgejhunt:
I remember having to squash a bug where some of the distro's had
stopped installing sqlite-php by default. I'm not in favor of making
php-sqlite optional

holta:
In the end this PR is 100% about documenting our existing code and its intentions, with few material changes for now.

If at some later date it's confirmed that `php{{ php_version }}-sqlite3` is no longer used anywhere except for `roles/osm-vector-maps` then its installation might be moved into that Ansible role.

Issue #1705, opened 2019-05-30
Validate Calibre-Web 0.6.3


holta:
Details at #1624

holta:
Implemented by PR #1709 which seems ok after (very quick) testing on Ubuntu 18.04 & Raspbian.

Issue #1624, opened 2019-04-29
Calibre-Web install fails during pip on Raspbian [& Calibre-Web versioning questions]


holta:
@tim-moody, @jvonau & @OzzieIsaacs do you understand why Calibre-Web is suddenly no longer installing on Raspbian?

Specifically, Internet-in-a-Box's standard Calibre-Web installer playbook appears to now fail every time at https://github.com/iiab/iiab/blob/master/roles/calibre-web/tasks/main.yml#L49-L54 with the following error:

> TASK [calibre-web : Download Calibre-Web dependencies (using pip) into virtual environment] ***
[...]

PR #1709, opened 2019-05-31
IIAB should use Calibre-Web 0.6.3 instead of master branch

holta:
Per #1624 #1705


holta:
Tested on Ubuntu 18.04

Testing on Raspbian next...

PR #1716, opened 2019-06-02
Bluetooth PAN and Terminal servers

tim-moody:


tim-moody:
passed smoke test on non-rpi
bt-term not working on rpi

tim-moody:
passed smoke test on rpi
win 10 can connect to pan

jvonau:
To ensure updates don't undue the change please copy the Bluetooth service file to /etc/systemd/system/ and edit that copy.

tim-moody:
good point. thanks for catching it.

georgejhunt:
What a pain to have to add "| bool" to all our flags. Some are missing in this PR

tim-moody:
It's my impression that when there is a Boolean operation the cast is not required. so when: is_rpi and bluetooth_install but when: is_rpi |bool

georgejhunt:
good to know, thanks

jvonau (approved):
Let's go with it..

PR #1732, opened 2019-06-13
Ncat [for OpenVPN on Debian 10 Buster]

georgejhunt:
Maybe this "need_ncat" is better than having to add is_debian-10 ... as a new release comes out-- but only a little better.


georgejhunt:
not tested

georgejhunt:
successful announcement of openvpn handle on debian buster(10)

Issue #1675, opened 2019-05-18
Validate Kolibri 0.12.4 after its .pex pip installer is published (0.12.4 was released 2019-06-12)


holta:
Building on PR #1608 (Kolibri 0.12.3)

Kolibri 0.12.4 Milestone:
https://github.com/learningequality/kolibri/milestone/36

holta:
Kolibri 0.12.4 was released earlier today:
https://github.com/learningequality/kolibri/releases/tag/v0.12.4

It will likely take a few days for them to publish their .pex (pip installer) as usual: https://learningequality.org/download/

holta:
> Kolibri 0.12.4 was released earlier today:
> https://github.com/learningequality/kolibri/releases/tag/v0.12.4
> 
> It will likely take a few days for them to publish their .pex (pip installer) as usual: https://learningequality.org/download/

@benjaoming Kolibri 0.12.4's .pex (pip install) is now available, and installs on Ubuntu 18.04, but fails completely to start its systemd service:

> TASK [kolibri : Enable & (Re)Start kolibri service] ****************************
fatal: [127.0.0.1]: FAILED! => {"changed": false, "msg": "Unable to restart service kolibri: Job for kolibri.service failed because the control process exited with error code.\nSee \"systemctl status kolibri.service\" and \"journalctl -xe\" for details.\n"}

A bit more detail here:
```
root@box:~# systemctl start kolibri
Job for kolibri.service failed because the control process exited with error code.
See "systemctl status kolibri.service" and "journalctl -xe" for details.

root@box:~# systemctl restart kolibri
Job for kolibri.service failed because the control process exited with error code.
See "systemctl status kolibri.service" and "journalctl -xe" for details.

root@box:~# systemctl status kolibri
● kolibri.service - Kolibri
   Loaded: loaded (/etc/systemd/system/kolibri.service; enabled; vendor preset: enabled)
   Active: failed (Result: exit-code) since Thu 2019-06-13 15:44:55 EDT; 1s ago
  Process: 18855 ExecStart=/usr/local/kolibri/bin/kolibri start (code=exited, status=1/FAILURE)
 Main PID: 1585 (code=killed, signal=TERM)

Jun 13 15:44:55 box.lan kolibri[18855]:   File "/usr/local/kolibri/lib/python2.7/site-packages/kolibri/dist/django/apps/registry.py", line 116, in populate
Jun 13 15:44:55 box.lan kolibri[18855]:     app_config.ready()
[...]

PR #1735, opened 2019-06-13
Revert to Kolibri 0.12.3, due to Kolibri 0.12.4 bug

holta:
Tested to support both options in future:

1) A specific Kolibri version number hard-coded into [roles/kolibri/defaults/main.yml](https://github.com/iiab/iiab/blob/master/roles/kolibri/defaults/main.yml#L9-L10) (as variable kolibri_version)

or

2) Comment out that [line that defines kolibri_version](https://github.com/iiab/iiab/blob/master/roles/kolibri/defaults/main.yml#L9-L10), if you instead want to attempt the installation of the very latest Kolibri version

Refs: #1675 and https://github.com/learningequality/kolibri/issues/5664


holta:
Tested on Ubuntu 18.04

PR #1733, opened 2019-06-13
Initial checkin of AzuraCast role [for streaming-based radio stations]

m-anish:
Doesn't have checks yet to disable install on rpi

Issue #1546, opened 2019-03-05
Podcasting platform for IIAB [offline YouTube? offline-not-brainless social media?]


m-anish:
I am interested to explore whether it is possible to have a nicely done open source podcasting platform on the IIAB. It should ideally:
1. Come with some sort of app/web-frontend that lets users search, discover, subscribe, listen
2. Allow podcasters to create and publish the podcast on the app/web-frontend itself

m-anish:
[This](http://www.podcastgenerator.net/) looks interesting.

holta:
@mikkokotila should we align IIAB with any of the emerging "Offline YouTube" platforms like [PeerTube](https://en.wikipedia.org/wiki/PeerTube), and if so which one?

Some of these appear to be gutters for viral hate speech:
https://www.polygon.com/2018/3/7/17087668/steemit-dtube-bitchute-youtube-purge

Likewise Mastodon is used for "legalized child pornography" in Japan...which is (in fact!) the very last thing IIAB schools want to be involved with:
http://www.ethanzuckerman.com/blog/2017/08/18/mastodon-is-big-in-japan-the-reason-why-is-uncomfortable/

Refs: #1530 #1538 #1540 #1541 #1542

m-anish:
Another open source solution - [Podcast Server](http://davinkevin.github.io/Podcast-Server/)

holta:
@ericnitschke recommends YouPHPTube which is mostly just to "offline" i.e. rehost/stream YouTube videos — but that's a start:

https://github.com/DanielnetoDotCom/YouPHPTube
https://www.youphptube.com
https://demo.youphptube.com
https://youphptube.readthedocs.io

ericnitschke:
> @ericnitschke recommends YouPHPTube which is mostly just to "offline" i.e. rehost/stream YouTube videos — but that's a start:

Users can upload videos/podcasts, and record their own right within the service.  We use it during workshops because participants can DL others' videos immediately to their phones.

Not a perfect solution but it's working well so far. Will keep you all updated.

m-anish:
YouPHPtube looks interesting. As a test, I was able to broadcast a livestream on OBS, and watch it at the same time in a youphptube browser window (albiet with some lags, and the buffering dialog popping up in between sometimes). But promising that it "just worked"

Initial thoughts:
1. It might be slightly bandwidth heavy, but it remains to be seen
[...]

PR #1743, opened 2019-06-19
Add 'Cham' basic online live Videostream capability

m-anish:
This PR adds basic live video streaming capability to the IIAB. It installs nginx from source alongside the mod-rtmp module. After this playbook is installed and running, a client can stream video to IIAB (through any rtmp streaming software like OBS - Open Broadcaster Studio), and users will be able to watch the stream at http://box.lan:8081 (which is the port used by nginx). 

This playbook should run on all IIAB supported platforms (rpi, amd64-debian/ubuntu). Potential issues may be the names of the packages which are installed as dependencies which may break the compile of nginx. So please test. I have done so on ubuntu 18.04.2


holta:
@m-anish wants us to consider this for inclusion within IIAB 7.0 during our "feature freeze" call starting in just over 12 h from now: http://minutes.iiab.io

FYI the web site for 'Cham' itself is emerging here: https://github.com/eka-foundation/cham

Refs: #1546, PR #1733

roles/cham/defaults/main.yml
```diff
@@ -0,0 +1,25 @@
...
+nginx_rtmp_module_src_file: master.zip
+nginx_rtmp_module_src_dir: nginx-rtmp-module-master
+cham_url: https://github.com/eka-foundation/cham/archive/
+cham_src_file: v0.1.tar.gz
+streaming_root: /library/streaming
+hls_dir: hls
+flv_dir: flv
+recordings_dir: recordings
+
+# If true, recordings will also be made available at box.lan/content/recordings
```
holta:
```suggestion
# If true, recordings will also be made available at box.lan/local_content/recordings
```

georgejhunt:
I think we had a feature freeze a few weeks ago. I do not think our target
users will have internet or mesh network. It seems to me that the risks of
breaking something at this late in the release cycle are greater than the
benefits.  Once we have branched off 7.0, we can merge Cham, (perhaps after
we know it works on rpi)

I'd like to find out more about why nginx was installed from source, rather
than using a package where someone else has ironed out the
incompatibilities.

I'm sorry to be a kill-joy. And I'm willing to be convinced otherwise at
our meeting in less than 12 hours.

holta:
As to whenever this is ready to merge, @m-anish has a strong sentence he's provided here (for http://FAQ.IIAB.IO Item 9, IIAB's 
showcase table of IIAB Apps) to illustrate Cham's purpose & possibilities:

> Cham is a lightweight live video streaming platform with adaptive bitrates for Internet-in-a-Box.

holta:
@m-anish

1) Can you explain this PR #1743's design for @georgejhunt, so he understands why you're compiling NGINX from source?  This approach (can/often) make for incredibly-hard-to-maintain software & forward-going support paths.

   Also/separately: if NGINX were to become a mainline part of IIAB in the distant future, how might that be navigated/integrated?  (In resource-tight environments like RPi especially!)

2) What does 'live video streaming' mean here?

   Are you saying that video files cannot be streamed?  (i.e. that only live events can be streamed, with a video camera or similar?)

   See the bottom of http://FAQ.IIAB.IO Item 9.  So that the explanation (of what Cham is, and what Cham does, with a 'Getting Started' for less technical operators if at all possible?) is crystal clear to as many people as possible, going a bit beyond https://github.com/m-anish/iiab/blob/videostream/roles/cham/README.rst, Thanks!

m-anish:
To be honest, the system packages for nginx should also probably work, but they bring extra clutter (read system bureaucracy) which starts to make sense when we use nginx more widely. Right now with this PR, nginx just serves this one web app. 

And it's pretty impressive in what it does. Give it an rtmp stream, and it will give you an adaptive bitrate output hls stream.

Is it suitable for an rpi. It _should_ work, but I would probably limit the transcoding defined in the conf file to maximum one stream, and I would also explore if the rpi offers any hardware encoding of videos. So, this can work on a pi, with a different nginx.conf file (which doesn't place a lot of load on system resources). 

I don't mind bumping this to 7.1 at all! Lets make it more feature ready. Until then, I would love more testing, on more platforms. 

Update: I also added support for VAAPI (hardware accelerated encoding) when this runs on platforms where that option is there (like intel kaby-lake based NUC's). This greatly reduces system load while streaming is happening, and now we start to get into territory of asking (okay how many parallel live streams can we do?)

holta:
Thank you @m-anish 

I agree this great work should be bumped to [IIAB 7.1](https://github.com/iiab/iiab/milestone/6)

But we can refine that decision during Thursday's community/team call (http://minutes.iiab.io)

roles/cham/README.rst
```diff
@@ -0,0 +1,39 @@
...
+  * Icecast password: g0adm1n
+  * Streaming port: 10005
+  * Default bitrate: 64k
+  * Default mount point: live
+* Make sure that autodj in AzuraCast is properly setup with the above settings.
+
+Attribution
+-----------
+
+This 'cham' playbook is a contribution of the `EKA foundation <https://github.com/eka-foundaiton>`.
```
holta:
```suggestion
This 'cham' playbook is a contribution of the `EKA foundation <https://github.com/eka-foundation>`.
```

m-anish:
Tagging @pdey and @soale, as we would eventually need to work to upstream this. 

Backstory: We got a working solution for Zanskar!

m-anish:
Updated PR and tested on 19.10 clean install. Please test. 

Things I haven't tested:
VAAPI hardware encoding - I don't have the NUC anymore that I can test hardware encoding with. Can do this after I get the NUC back. 

Things I tested:
Clean, fresh install. Add this role. Test live streaming.

m-anish:
This PR is now much more compatible than before. Earlier, we didn't have nginx as a standard part of the IIAB. Now that we have it, this PR does not compile nginx or rtmp-module from source, rather, installs the rtmp as a standard system package. Config files have been tuned to be compatible with how other web services are being handled within nginx in IIAB. 

I did however had to add a `include` file directive to `nginx.conf` to add a `rtmp` context under the `master` context, since the conf.d directory limits the context to `http>server`

m-anish:
Pending steps: 
Comprehensive testing on different platforms. 

Future feature:
Use the rpi's hardware encoding features as well. This is a small improvement, unlike a big improvement in the case of new intel hardware, but nevertheless, it will save some system resources.

m-anish (commented):
Looks good on a eyeball-review :)

roles/cham/tasks/main.yml
```diff
@@ -0,0 +1,7 @@
+- name: Install Cham
+  include_tasks: install.yml
+  when: internet_available and cham_install
+
+- name: Uninstall Cham
+  include_tasks: uninstall.yml
+  when: internet_available and (not cham_install)
```
jvonau: I'd use internet_available and (not cham_enabled)

roles/cham/tasks/install.yml
```diff
@@ -0,0 +1,131 @@
...
+    src: rtmp.conf.j2
+    dest: "{{ nginx_config_dir }}/../rtmp.conf"
+    owner: root
+    group: root
+    mode: 0755
+
+- name: Cham - nginx config - Enable nginx rtmp webserver configuration file
+  lineinfile: 
+    path: "{{ nginx_config_dir }}/../nginx.conf"
+    line: "include {{ nginx_config_dir }}/../rtmp.conf;"
```
jvonau: could define {{ nginx_dir }} in default_vars and use that in place of {{ nginx_config_dir }}../

roles/cham/tasks/install.yml
```diff
@@ -0,0 +1,131 @@
...
+    state: present
+    insertafter: EOF 
+
```
jvonau: Might be wise to have the include as conditional in the nginx.conf template also, so 'runrole nginx' does the right thing. The above code would act as a failsafe for 'runrole cham' where nginx.conf would not be installed during runrole.

jvonau (commented):
Good pretty good

roles/cham/tasks/install.yml
m-anish: agree with that. can you please do that :)

roles/cham/tasks/install.yml
m-anish: I handle that process in the uninstall.yml part. line 19

jvonau: Lets work up the Tim's approach with include servers.d

jvonau:
@tim-moody

tim-moody:
{{ nginx_config_dir }}/../rtmp.conf;" seriously?

roles/cham/tasks/install.yml
m-anish: could you explain a bit more. happy to make any changes, but would need to understand :)

m-anish:
> {{ nginx_config_dir }}/../rtmp.conf;" seriously?

I don't know of a variable that points to /etc/nginx, and didn't want to use absolute paths. 

Suggestion?

tim-moody:
? put it in {{ nginx_etc_dir }}/protocols.d - it's just to declare rtmp, right?

tim-moody:
I think lineinfile is a last resort when we have no other control

m-anish:
> ? put it in {{ nginx_etc_dir }}/protocols.d - it's just to declare rtmp, right?

Don't see any directory with that name.

m-anish:
> I think lineinfile is a last resort when we have no other control

could you suggest another approach

tim-moody:
declare {{ nginx_etc_dir }}
cause  {{ nginx_etc_dir }}/protocols.d to be created
put it in {{ nginx_etc_dir }}/protocols.d and include {{ nginx_etc_dir }}/protocols.d/*

m-anish:
> {{ nginx_etc_dir }}

where should I make that declaration?

tim-moody:
default vars:

nginx_etc_dir: /etc/nginx
nginx_config_dir: {{ nginx_etc_dir }}/conf.d

m-anish:
> default vars:
> 
> nginx_etc_dir: /etc/nginx
> nginx_config_dir: {{ nginx_etc_dir }}/conf.d

cool. I'll make the changes. Thanks for the tip :)

m-anish:
> default vars:
> 
> nginx_etc_dir: /etc/nginx
> nginx_config_dir: {{ nginx_etc_dir }}/conf.d

cool. I'll make the changes. Thanks for the tip :) 

> declare {{ nginx_etc_dir }}
> put it in {{ nginx_etc_dir }}/protocols.d and include {{ nginx_etc_dir }}/protocols.d/*

this should probably happen in the role that installs nginx?

m-anish:
> default vars:
> 
> nginx_etc_dir: /etc/nginx
> nginx_config_dir: {{ nginx_etc_dir }}/conf.d

cool. I'll make the changes. Thanks for the tip :) 

> declare {{ nginx_etc_dir }}
> put it in {{ nginx_etc_dir }}/protocols.d and include {{ nginx_etc_dir }}/protocols.d/*

this should probably happen in the role that installs nginx?

tim-moody:
I would use iiab/vars/default_vars.yml for the variables
I would create /etc/nginx/protocols.d in nginx role (or wherever other nginx dirs are created)

m-anish:
@tim-moody done. please review :)

m-anish:
these last two reverts were per @jvonau recommendation, to leave protocols.d dir creation in install.yml

tim-moody:
protocol.d belongs with nginx. what if another app needs it too?

m-anish:
> protocol.d belongs with nginx. what if another app needs it too?

its part of the nginx role. it will get installed as part of a regular iiab install. me and jerry were just discussing where to best put it in the nginx playbook

holta:
Cham (and this PR) were extensively discussed during today's Internet-in-a-Box (IIAB) call &mdash; please see the notes/minutes here: http://minutes.iiab.io

Refs: #2209 #2157 https://github.com/eka-foundation/home/issues/31

Issue #1744, opened 2019-06-19
nameserver setting [with static IP address]


mikerjensen:
IIAB 7.0's /etc/resolv.conf sets nameserver to 127.0.0.1 - this doesn't work for me, and I can't change the settings on the DHCP server to which the RPI is connected via Ethernet, so after an RPI reboot i always need to edit /etc/resolv.conf and change the nameserver setting to an external DNS, e.g 8.8.8.8 

Is there a way to set this permanently?

tim-moody:
@mikerjensen we need to know the network topology

mikerjensen:
RPI connected via Ethernet cable to broadband/wireless router.

holta:
@mikerjensen please run:
```
systemctl status dnsmasq | pastebinit
```
And reply with the URL, Thanks!

mikerjensen:
Seems there is an unacceptable character for pastebinit...
    
    
    $ systemctl status dnsmasq | pastebinit
      Traceback (most recent call last):
        File "/usr/bin/pastebinit", line 424, in <module>
          params = urlencode(params)
        File "/usr/lib/python3.5/urllib/parse.py", line 801, in
      urlencode
          v = quote_via(str(v), safe, encoding, errors)
        File "/usr/lib/python3.5/urllib/parse.py", line 729, in
      quote_plus
          string = quote(string, safe + space, encoding, errors)
        File "/usr/lib/python3.5/urllib/parse.py", line 707, in quote
          string = string.encode(encoding, errors)
      UnicodeEncodeError: 'utf-8' codec can't encode character '\udce2'
      in position 0: surrogates not allowed
    pi@box:~ $ systemctl status dnsmasq 
      ● dnsmasq.service - dnsmasq - A lightweight DHCP and caching DNS
      server
         Loaded: loaded (/lib/systemd/system/dnsmasq.service; disabled;
      vendor preset:
         Active: inactive (dead)
      
    
    On 19/06/19 13:39, A Holt wrote:
    
    
      
      @mikerjensen
        please run:
      systemctl status dnsmasq | pastebinit
      And reply with the URL, Thanks!
      —
        You are receiving this because you were mentioned.
        Reply to this email directly, view it on GitHub, or mute the thread.
      [
{
"@context": "http://schema.org",
"@type": "EmailMessage",
"potentialAction": {
"@type": "ViewAction",
[...]

PR #1745, opened 2019-06-19
Update computed_services.yml [dnsmasq should/generally run, by default]

jvonau:
Fix


holta:
Towards resolving #1744 in a long-term way.

PR #1750, opened 2019-06-20
network role should restart dnsmasq in appliance mode too!

holta:
As recommended by @jvonau


holta:
Testing is ongoing...

holta:
Tested perfectly on Raspbian in appliance mode: many thanks to @jvonau for catching this!

PR #1753, opened 2019-06-20
Update detected_network.yml [for Docker]

jvonau:
Docker

### DIFF
--- a/roles/network/tasks/computed_network.yml
+++ b/roles/network/tasks/computed_network.yml
@@ -157,7 +157,7 @@
 
 # so this works
 - name: Interface count
-  shell: ls /sys/class/net | grep -v -e lo | wc | awk '{print $1}'
+  shell: ls /sys/class/net | grep -v -e lo -e bridge0 -e veth | wc | awk '{print $1}'
   register: adapter_count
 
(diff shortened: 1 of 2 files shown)


holta:
@m-anish & all can you review this Docker-oriented change?

Thanks!

(Not directly related to yesterday's networking tweaks @ PR #1745, #1750)

jvonau:
Still need to deal with iptables

holta:
> Still need to deal with iptables

Ok.  Is this likely for IIAB 7.0?

Or is this more suitable for IIAB 7.1?

m-anish (changes requested):
Okay. From the code it seems we're trying to exclude interfaces here which we're not interested in. 
While we're at it, can we exclude:

* "br-*" because these are bridges azuracast/docker creates
* docker0 because this is the default docker bridge

m-anish:
Try this patch on.

[0001-Add-br-and-docker-to-network-interface-list-to-exclu.patch.zip](https://github.com/iiab/iiab/files/3317436/0001-Add-br-and-docker-to-network-interface-list-to-exclu.patch.zip)

m-anish:
Superseded by #1765

PR #1760, opened 2019-06-22
Fix bug in KA Lite parsing of regex for ifconfig output

m-anish:


holta:
@m-anish has tested this many times on Ubuntu 18.04

PR #1686, opened 2019-05-22
Workaround for upstream FreePBX breakage

m-anish:
### Fixes Bug
Freepbx upstream is broken. This PR gets around that temporarily. 

### Description of changes proposed in this pull request.
Applies a patch - https://github.com/FreePBX/framework/pull/71

### Smoke-tested in operating system.
 - [x] VM-18.04
 - [ ] NUC-18.04

Fixes #1685

Issue #1685, opened 2019-05-21
Freepbx-15 broken upstream


m-anish:
Opening a ticket here to keep track of this...

https://github.com/FreePBX/framework/pull/71#issuecomment-494390926

m-anish:
PR #1686

holta:
@m-anish is FreePBX-15 now fixed upstream?

Why was PR #1686 closed?

Can this tkt be closed too?

holta:
No response from @m-anish as to why he closed PR #1686.

@m-anish please share the status of FreePBX as part of your IIAB playbook `pbx` when you can!

PR #1761, opened 2019-06-22
FreePBX fixes and changes

m-anish:
* Patches upstream freepbx bug which hasn't been fixed yet
* Moves pbx.lan -> box.lan:8888
* Fixes documentation accordingly

### DIFF
--- a/roles/network/tasks/hosts.yml
+++ b/roles/network/tasks/hosts.yml
@@ -10,7 +10,7 @@
   lineinfile:
     path: /etc/hosts
     regexp: '^172\.18\.96\.1'
-    line: '172.18.96.1     {{ iiab_hostname }}.{{ iiab_domain }} {{ iiab_hostname }} box box.lan pbx pbx.lan'
+    line: '172.18.96.1     {{ iiab_hostname }}.{{ iiab_domain }} {{ iiab_hostname }} box box.lan'
     state: present
   when: iiab_lan_iface != "none" and not installing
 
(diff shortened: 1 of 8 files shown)


holta:
Replaces PR #1686 towards solving #1685

tim-moody:
can't break the 8000 habit

holta:
@web-anish

1) plz verify "5060, 5160, 5161, 8888, 10000-20000" are correct and that all are TCP in https://github.com/iiab/iiab/wiki/IIAB-Networking#list-of-ports--services ?  _**[Done]**_

2) Supposedly 8888 is already used by "Default for IPython / Jupyter notebook dashboards" according to https://en.wikipedia.org/wiki/List_of_TCP_and_UDP_port_numbers#Registered_ports

(Do we / @mikkokotila care?)

holta:
Emailed @mikkokotila to try to quickly clear this up, so @m-anish can finalize a port number.

@tim-moody suggests possibly considering port 10080.

holta:
> @tim-moody suggests possibly considering port 10080.

Already taken by AzuraCast (PR #1733) per https://github.com/iiab/iiab/wiki/IIAB-Networking#list-of-ports--services

mikkokotila:
8888 is the default for Jupyter, but I think it automatically goes to the next available if 8888 is already in use. I'm not 100% on how this plays out in different systems, but I guess it would be like that across the board. Still to be safe, maybe use other than 8888.

holta:
> 8888 is the default for Jupyter, but I think it automatically goes to the next available if 8888 is already in use. I'm not 100% on how this plays out in different systems, but I guess it would be like that across the board. Still to be safe, maybe use other than 8888.

Agreed!  Let's go with 83 as suggested by @tim-moody, that Anish has documented here: https://github.com/iiab/iiab/wiki/IIAB-Networking#list-of-ports--services

holta:
I've updated the following from 8888 to 83:

http://FAQ.IIAB.IO Item 9 "What services (IIAB apps) are suggested during installation?"

Issue #1759, opened 2019-06-22
Assess MediaWiki/Discourse/CRM/FB + Trello/Asana/Todoist + Slack/Twist/Teams needs to scale up IIAB Community Support


holta:
IIAB is transitioning to a new, more secure http://download.iiab.io server as we speak.

And a lot of good work has gone into reinforcing OpenVPN to become substantially more robust over the past year.

But thought needs to be put into larger critical/core community platforms & tooling — those essential to our scaling impactfully as a contributor community in coming years.

Might this include...?

- MediaWiki
- Discourse
- CRM like CiviCRM or otherwise
- "Facebook" more appropriate than Twitter for educators ?
- ???

veeara282:
For task management, do we plan to continue using the GitHub issue tracker for everything or use a separate app like Trello and Asana for general task management? (Disclosure: I currently work at @asana.)

Trello has [public boards](https://help.trello.com/article/789-changing-the-visibility-of-a-board-to-public-private-or-team), which anyone on the Internet can view but only members of the board can edit. I think this is a good feature for an open source project. In Asana, only members of the workspace or org can see anything in that workspace/org.

Both [Trello](https://blog.trello.com/github-and-trello-integrate-your-commits) and [Asana](https://asana.com/apps/github) have GitHub integrations, so tasks can be synchronized with issues and PRs.

veeara282:
I think that [Slack](https://slack.com/), [Twist](https://twist.com/?lang=en), [Microsoft Teams](https://products.office.com/en-us/microsoft-teams/group-chat-software), etc. are all promising options for team communication and would be easier for me to access than an Skype group chat that isn't in an isolated workspace. (Caveat: ["Microsoft Teams isn’t better than Slack, but it is freer."](https://www.vox.com/2019/7/9/20686206/microsoft-teams-slack-facebook-snapchat-copy) I am not too happy with Microsoft's tactics in this market.)

If we go with Twist we should also consider [Todoist](https://todoist.com/) (which is a Trello/Asana alternative).

veeara282:
Key question: What are our biggest pain points with existing collaboration software?

holta:
> Key question: What are our biggest pain points with existing collaboration software?

Educators, librarians and other humanitarians are effectively being excluded by GitHub's de facto barrier to entry, as a "developers, developers, developers" coding-centric environment.

[...]

PR #1757, opened 2019-06-21
Flat [for /usr/bin/iiab-diagnostics]

georgejhunt:
New solution to fieldback problem returning flat file. Replaces PR1596

Issue #1749, opened 2019-06-19
OpenVPN is very commonly needed when DNS is failing


holta:
@georgejhunt can [OpenVPN](https://github.com/iiab/iiab/tree/master/roles/openvpn) be made more resilient, so that it works when DNS is not available?

This would make OpenVPN an order of magnitude more useful, and allow IIAB to much more rapidly work through networking bugs, as are ever-present across OS's and evolving Internet standards, evolving OS's/protocols etc.

As OpenVPN's most important use case is troubleshooting complex situations when DNS is inevitably failing or absent for a wide assortment of reasons.

TeamViewer, for example, works when DNS is absent, for much the same reasons.

Do you know if this might be possible?

Refs: #1687, #1744, PR #1745, #1747 (etc)

holta:
Good news @georgejhunt believes it is likely possible to run OpenVPN when DNS is not working.

There is a concern than the IP address of IIAB's "OpenVPN cloud" might occasionally change however.

I've asked George if he has time to investigate various strategies to make OpenVPN substantially more resilient across many networking situations, and ideally several years into the future.  In Short: there might be contractual issues with our upstream "OpenVPN cloud" provider, whereby they might possibly change the IP address(es) from time to time &mdash; or possible another better/cheaper cloud provider becomes necessary in the future.

e.g. in any case, it might be possible to make OpenVPN into a substantially more robust diagnosis/troubleshooting tool, by checking **both** the "most recently known IP address" **and** also using DNS ***if*** indeed DNS happens to be working at that moment in time.

Noting that  OpenVPN-like tools are already a lifesaver and a lifeline for so many of these distant machines, where their background/context is so often extremely unclear, until a live connection is established...  Thank you to George and others who've made these OpenVPN tools possible!

ASIDE: #1596 is another key tool the IIAB community should add to its "arsenal/tool chest" in 2019, to strengthen all countries' collective wisdom here, building out a more rock solid community product.

georgejhunt:
Looks like we could have ip and dns references to the same server
https://unix.stackexchange.com/questions/183425/can-you-set-openvpn-to-auto-connect-to-fallback-ovpns-in-case-the-current-one

holta:
> Looks like we could have ip and dns references to the same server https://unix.stackexchange.com/questions/183425/can-you-set-openvpn-to-auto-connect-to-fallback-ovpns-in-case-the-current-one

[...]

PR #1763, opened 2019-06-22
WIP: iiab-diagnostics friendly community tooling, to build relationships between dev teams, implementers & educators

holta:
This is an early work-in-progress, building on #1575, #1749 and @georgejhunt's PRs #1596, #1757.

Related: 
#1759 Assess MediaWiki/Discourse/CRM/FB needs to scale up IIAB Community Support


holta:
Ansible stuff is not part of the mainline /usr/bin/iiab-diagnostics as for now it's (1) too slow, (2) too bloated in its output for pastebin (1/3 of a megabyte for the Ansible variable dump /etc/all-vars alone), and (3) its utility+usefulness appears highly limited within the context of most debugging situations we've faced in 2019 so far.

@georgejhunt LMK if we can fix some this 3 dilemmas above to revise this for the next rev in coming days/weeks?

PR #1765, opened 2019-06-23
Dnsmasq fixes

m-anish:
This includes Jerry's PR #1753 PLUS
  *  Changing the network iface list in Jerry's PR to include more docker-created interfaces
  *  Restarting dnsmasq whenever br0 comes up. This can allow hotplugging of the wired LAN later after the server has booted, and still dnsmasq will come up.

### DIFF
--- a/roles/network/tasks/computed_network.yml
+++ b/roles/network/tasks/computed_network.yml
@@ -157,7 +157,7 @@
 
 # so this works
 - name: Interface count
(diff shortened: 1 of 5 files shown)


holta:
What smoke-testing is complete / what smoke-testing is needed here?

(e.g. for RPi/Debian or particular networking topologies?)

m-anish:
LanController/Gateway mode testing on Deb9/10/Rpi might be handy, but I dont have bandwidth to do that. 

On 18.04 in Gateway mode, it works.

PR #1768, opened 2019-06-23
Do not copy dnsmasq restart script if we are on RPi

m-anish:


holta:
Ref: PR #1765 "Dnsmasq fixes"

PR #1774, opened 2019-06-25
Add capability in apache to globally limit connection speeds

m-anish:


m-anish:
NOT tested yet!

holta:
> NOT tested yet!

What's the plan here: does this need field testing in Zanskar Valley and then will this be documented for [IIAB 7.1](https://github.com/iiab/iiab/milestone/6) ?

(Or is a more aggressive schedule desired?)

m-anish:
I'm not sure. This is a simple enough PR, which we'll definitely put to use in Zanskar. If you think this is useful to have right now, go for it, otherwise it can wait.

Issue #1778, opened 2019-06-27
./iiab-network fails on Ubuntu 16.04 (and other OS's possibly?)


holta:
@jvonau & @m-anish can you help refine PR #1768 if this is indeed the same #1767 issue, presumably caused by PR #1765?

What other OS's should we be checking for?

Here's the error that arises every time on Ubuntu 16.04:

> TASK [network : Copy script to restart dnsmasq whenever br0 comes up] **********
fatal: [127.0.0.1]: FAILED! => {"changed": false, "checksum": "50de730947aa4cccc1586ff91d9f6bfc984b5127", "msg": "Destination directory /etc/networkd-dispatcher/routable.d does not exist"}

m-anish:
Are we going the dnsmasq route on 16.04 as well? I thought back in the day, we used bind?

Perhaps we should think which DNS protocol/server we should use on 16.04 before we fix this. Thoughts?

This is easy enough to patch, but we should address the larger question.

holta:
dnsmasq doesn't work perfectly on 16.04, but it's been mainline IIAB for 6 months now, so highly unlikely to be reverted at this point?

Ref: PR #1779

holta:
Solved by PR #1779

Issue #1767, opened 2019-06-23
RPi Prob: "Copy script to restart dnsmasq whenever br0 comes up" (Destination directory /etc/networkd-dispatcher/routable.d does not exist)


holta:
Error is:

> TASK [network : Copy script to restart dnsmasq whenever br0 comes up] **********
fatal: [127.0.0.1]: FAILED! => {"changed": false, "checksum": "50de730947aa4cccc1586ff91d9f6bfc984b5127", "msg": "Destination directory /etc/networkd-dispatcher/routable.d does not exist"}
>
> 
PLAY RECAP *********************************************************************
127.0.0.1                  : ok=551  changed=143  unreachable=0    failed=1    skipped=241  rescued=0    ignored=0

Machine/OS (10.8.0.30) background details: http://paste.debian.net/1088961/

Ref: PR #1765 "Dnsmasq fixes"

m-anish:
okay, there should be a check somewhere. ill take a look

holta:
Solved by PR #1768 it appears, Thanks @m-anish !

PR #1779, opened 2019-06-28
Fixing ./iiab-network on Ubuntu 16.04 (and other OS's possibly?)

holta:
Towards fixing #1778 

Do we also need to check for `is_debian_9` ?

Thanks @jvonau & @m-anish for reviewing.

Refs: PR #1765, #1767, PR #1768


holta:
@jvonau @m-anish wouldn't it be better for Ansible to test for the existence of *directory* /etc/networkd-dispatcher/routable.d — instead of guessing which OS's / situations permit this copying to go forward?

tim-moody:
it is way past time to stop writing one size fits all playbooks

holta:
Tested on 3+ OS's after trying to incorporate @jvonau suggestions.

Can be further revised later as nec.

PR #1780, opened 2019-06-28
InternetArchive: fix flags on launching to launch crawler as well

mitra42:
### Fixes Bug
internetarchive is starting as a server, but needs to start the crawler as well which requires the -c flag as well. 

Change flags from -s to -sc in internetarchive.server.j2

There are also some minor changes in README, 

Raspberry Pi current Raspbian

### DIFF
--- a/roles/internetarchive/README.md
+++ b/roles/internetarchive/README.md
@@ -58,13 +58,14 @@ first time, refresh after a little while and most should appear.
 
 ## Administration
 
-Administration is carried out through the same User Interface as browsing. 
+Administration is carried out mostly through the same User Interface as browsing. 
 
-Access [http://localhost:4244/local](http://localhost:4244/local) to see a
+Access [http://box.lan:4244/local](http://box.lan:4244/local) to see a
 display of local content, this interface is under development and various admin
-tools will be added here.  *At some point this will become the default page*.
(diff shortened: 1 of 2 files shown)


holta:
@mitra42 is there a chance this PR is related to the dramatic slowing down (much more than 3X slowdown, it appears) now being seen during `internetarchive`'s Ansible install?

Ref: https://github.com/iiab/iiab/pull/1806#issue-293486566

mitra42:
It shouldnt be, because the default crawl is empty until you click on something to crawl it. 

But if someone is doing a reinstall after playing with the pre-release then its possible they'll already have stuff setup to crawl (by clicking on the Crawl button), though normally it will already be downloaded in which case the crawl is super fast. 

If someone notice it, could they send me a copy of their `/root/dweb-mirror.config.yaml`
and ideally let me know at what point its slowed down. 

I think I might know the cause, I fixed the IPFS installer, which was enabled, but broken, and since by default its disabed noone would have noticed, while my end is fixed, IPFS's distribution servers are super flakey, so it can be really slow. I'll comment out the IPFS install, which I can do in my repo, it doesn't require changes to IIAB's distro.  I just pushed this to npm 

It would help if someone who experiences the slowdown could send me a log and mark where the slowdown occurred.

PR #1806, opened 2019-07-01
Ansible output fixed for internetarchive & Minetest

holta:
@mitra42 do you know why "yarn install to get needed modules" is taking so much longer than it used to?

https://github.com/iiab/iiab/blob/master/roles/internetarchive/tasks/main.yml#L21-L25

FWIW this is running on a much faster machine than in the past (RPi 4) and yet it still took 3X longer here!

(Not the end of the world of course, but several people have asked us to keep IIAB's mainline installs to less than an hour, so please help us keep an eye on it ;-)

### DIFF
--- a/roles/internetarchive/tasks/main.yml
+++ b/roles/internetarchive/tasks/main.yml
@@ -18,7 +18,7 @@
     state: directory
     owner: "root"
 
-- name: Run yarn install to get needed modules (CAN TAKE ~5 MINUTES)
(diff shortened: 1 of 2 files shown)


mitra42:
I just posted on this  in https://github.com/iiab/iiab/pull/1780. It would help to know WHERE in the yarn install it slows down, i.e. if its just a lot of modules (maybe someone added a sub-dependency) OR if its the problem I think I identified in that other comment in which case it will stop at a particular point which I've now disabled.

PR #1808, opened 2019-07-02
MediaWiki 1.32.2 -> 1.32.3

holta:


holta:
Smoke-tested on Ubuntu 18.04 & 16.04.

Remember that deleting the prior database is often necessary a priori (using "sudo mysql" then "DROP DATABASE nextcloud;") particularly if you have an older version of MediaWiki installed.

Already being replaced by PR #1809...

PR #1820, opened 2019-07-04
Check for PHP 7.1+ automatically, for Nextcloud 15/16

holta:
Resiliency enhancement, building off PR #1819 and Node.js auto-check @ https://github.com/iiab/iiab/blob/master/roles/nodejs/tasks/main.yml#L10-L51


holta:
Tested on Raspbian 9/10, Ubuntu 16/18 and Debian 10.

Issue #1815, opened 2019-07-03
box.lan does not resolve at the command-line (IIAB installed on Debian 10 Buster VM)


holta:
@jvonau do you have suggestions or ideas?

It has been rebooted several time.  Live at 10.8.0.30 for now:

```
root@box:~# more /etc/iiab/iiab.env
# This is a configuration file for IIAB
# It can be sourced in a shell script or read into an application
IIAB_BASE_PATH=/opt/iiab
IIAB_DIR=/opt/iiab/iiab
IIAB_RELEASE=7.0
IIAB_REVISION=0
IIAB_GATEWAY_ENABLED=False
LAN_IP=172.18.96.1
OS=debian
OS_VER=debian-10
WWWROOT=/library/www/html
STAGE=9
FQDN=box.lan
IIAB_USB_LIB_SHOW_ALL=True
IIAB_WAN_DEVICE="enp0s3"
IIAB_LAN_DEVICE="none"
HOSTAPD_ENABLED=True
HTTPCACHE_ON=False
```

holta:
@jvonau writes:
> quick fix is to edit /etc/hosts...  as /etc/resolv.conf is pointing upstream..

@holta writes:
> What playbook should do this?

holta:
@jvonau has raised concerns about the existence of `ifupdown` on Debian 10 "Buster", wondering if this might break some aspects of `systemd` ?  See these Debian 10 dependencies:.

```
root@box:/home/iiab-admin# apt-cache depends ifupdown
ifupdown
  Depends: libc6
  Depends: lsb-base
  Depends: adduser
  Depends: iproute2
  Breaks: systemd
 |Recommends: isc-dhcp-client
  Recommends: <dhcp-client>
    dhcpcanon
    dhcpcd5
    pump
    isc-dhcp-client
  Suggests: ppp
  Suggests: rdnssd
  Replaces: systemd
```

ASIDE: `ifupdown` also exists on Ubuntu 18.04 (Desktop version) FWIW.  But does not exist on Ubuntu 18.04's Server version:

```
root@box:~# apt -a list ifupdown
Listing... Done
ifupdown/bionic-updates 0.8.17ubuntu1.1 amd64
ifupdown/bionic 0.8.17ubuntu1 amd64
```

holta:
Curiously, I don't believe Raspbian Buster is affected by these Debian 10 Buster issues?

(But @jvonau & others can correct me if I'm wrong!)

holta:
[...]

PR #1829, opened 2019-07-07
Wan file ["fix WAN detection in /etc/network/interfaces"]

jvonau:
bug fix

### DIFF
--- a/roles/network/tasks/detected_network.yml
+++ b/roles/network/tasks/detected_network.yml
@@ -46,7 +46,7 @@
 - name: Setting wan_in_interfaces
   set_fact:
     wan_in_interfaces: True
-  when: is_debuntu and wan_file.stdout|int >= "0"
+  when: is_debuntu and (wan_file.stdout|int > 0)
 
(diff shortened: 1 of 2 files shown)


holta:
@jvonau does this also solve #1815?

i.e. "box.lan does not resolve at the command-line (IIAB installed on Debian 10 Buster VM)"

holta:
This PR #1829 tested on Raspbian Buster on RPi 4 w/ ./iiab-network per @jvonau's request.  Look good.

But back on Debian 10 pre-release 10.8.0.30, @jvonau asks if we might also need something like...?
```
root@box:~# more /etc/network/interfaces.d/patch_auto
auto enp0s3

holta:
> @jvonau does this also solve #1815?
> 
> i.e. "box.lan does not resolve at the command-line (IIAB installed on Debian 10 Buster VM)"

@jvonau responds: "no, that was /etc/hosts hack, but I can play with [fresh Debian 10.0.0 VM] and see what is needed."

PR #1834, opened 2019-07-08
Don't clobber http://box/maps redirect if it already exists :)

### LINKED ISSUE #1496: Make http://box/maps & http://box/map work with whatever maps the implementer has...most recently installed? [OSM]
@tim-moody @m-anish @georgejhunt see #877 which is our umbrella ticket of OSM needs & design issues.

holta:
Should fix #1496


holta:
Tested.

PR #1835, opened 2019-07-08
Tighten up PR #1834 [OSM redirect for http://box/maps, on initial install only!]

holta:
Builds on PR #1834, to solve #1496 more efficiently.


holta:
Tested in both situations...with and without existence of redirect file `/library/www/osm-vector-maps/index.html`

PR #1836, opened 2019-07-08
Enable http://box/map too, per #1496 + osm-vector-maps/README.md

holta:
Implements #1496 "Make http://box/maps & http://box/map work with whatever maps the implementer has...most recently installed? [OSM]"


holta:
Also adds a README, in keeping with our best Ansible playbooks:

https://github.com/iiab/iiab/blob/master/roles/osm-vector-maps/README.md

Issue #877, opened 2018-07-07
Package up vector-based OSM maps


holta:
@georgejhunt is working on this @ https://github.com/iiab/iiab-factory/pull/32

Related Demo Sites:
http://medbox.iiab.me/modules/en-osm-omt-min/ [shows STALE version, as of July/Aug 2018]
http://medbox.iiab.me/modules/en-osm-omt-central-am/ [shows STALE version, as of July/Aug 2018]

holta:
- [x] Excellent progress as written up in DRAFT https://github.com/iiab/iiab/wiki/IIAB-6.6-Release-Notes
- [x] OSM instructions (implementer doc[s], not just low-level doc #1033, PR iiab/iiab-factory#39) need review with a fresh set of eyes, to see if a newbie installer can understand, follow, succeed!  Three summary docs listed at the bottom of http://download.iiab.io/content/OSM/vector-tiles/
- [x] #1035 Can OSM Vector Maps fill the entire screen?
- [x] #1034 How many cities are searchable by new Vector-based OSM?
- [x] #662 Teachers want Accents to work when searching for cities in OpenStreetMap (what's our status? who should assist?)
- [x] #1036 Keep OSM from zooming past Level 10, where there's no data?
- [ ] #1170 Haiti Diaspora Map Pack: aggregate street detail from 12+ global cities/regions
  - [x] iiab/iiab-factory#52 Merge non-overlapping .mbtiles regions (into single .mbtiles file)
- [x] #1192 Overcrowding of OSM map labels at Zoom Levels 13 and 14 (in cities)
- [ ] #1423 Regenerate OSM custom tilesets [incl hillshading & contours?]
- [x] Apply for HOTosm.org [microgrant](https://www.hotosm.org/projects/microgrants_and_community_development) ([2019 invitation](https://www.hotosm.org/updates/2019-hot-microgrants-program-empowering-local-openstreetmap-communities/)) to assist India/UX/packaging work especially?
- [x] #1496 Make http://box/maps & http://box/map work with whatever maps the implementer has...most recently installed?
- [ ] #1568 Magrit: Cartography Teaching+Learning as part of IIAB [with OSM?]
- [ ] Easily download any "continent" !
  - [x] PR #1600 Regional Vector Tile selection
  - [x] PR #1602 Explain osm_vector vars in {MIN, MEDIUM, BIG} local_vars.yml
  - [x] PR iiab/iiab-admin-console#160 Regional Tile selection Osmap2
[...]

PR #1837, opened 2019-07-08
osm-vector-maps playbook: reload Apache at end

holta:
Allows `cd /opt/iiab/iiab ; ./runrole osm-vector-maps` to complete the job properly.

Builds on PRs #1834 #1835 #1836 

Ref: #877


holta:
Tested.

Issue #1830, opened 2019-07-07
Lokole admin user "settings" page does not work


m-anish:
### Expected behavior and actual behavior.

Admin settings page for lokole does not work. 

### Steps to reproduce the problem.

* Install `lokole` as part of IIAB install
* Login as administrator with Admin/changeme
* Access the Settings tab
* You get `Unexpected error. Please contact your administrator. ` message
* Output of systemctl status lokole is below
```

● lokole.service - Provides the Lokole Server
   Loaded: loaded (/etc/systemd/system/lokole.service; enabled; vendor preset: enabled)
   Active: active (running) since Sat 2019-07-06 13:40:11 IST; 24h ago
 Main PID: 1144 (bash)
    Tasks: 6 (limit: 4915)
   CGroup: /system.slice/lokole.service
           ├─1144 /bin/bash /home/lokole/state/webapp.sh
           ├─1202 /library/lokole/venv/bin/python3 /library/lokole/venv/bin/gunicorn --timeout=300 --workers=4 --bind=unix:/home/lokole/state/lokole_gunicorn.sock --log-level=error opwen_email_client.webapp:app
           ├─2020 /library/lokole/venv/bin/python3 /library/lokole/venv/bin/gunicorn --timeout=300 --workers=4 --bind=unix:/home/lokole/state/lokole_gunicorn.sock --log-level=error opwen_email_client.webapp:app
           ├─2034 /library/lokole/venv/bin/python3 /library/lokole/venv/bin/gunicorn --timeout=300 --workers=4 --bind=unix:/home/lokole/state/lokole_gunicorn.sock --log-level=error opwen_email_client.webapp:app
           ├─2042 /library/lokole/venv/bin/python3 /library/lokole/venv/bin/gunicorn --timeout=300 --workers=4 --bind=unix:/home/lokole/state/lokole_gunicorn.sock --log-level=error opwen_email_client.webapp:app
           └─2155 /library/lokole/venv/bin/python3 /library/lokole/venv/bin/gunicorn --timeout=300 --workers=4 --bind=unix:/home/lokole/state/lokole_gunicorn.sock --log-level=error opwen_email_client.webapp:app

Jul 06 13:40:11 box systemd[1]: Started Provides the Lokole Server.
Jul 07 14:05:19 box bash[1144]: [2019-07-07 14:05:19 +0530] [2020] [ERROR] TypeError: expected str, bytes or os.PathLike object, not NoneType
[...]

PR #1840, opened 2019-07-08
Add logic to install a given commit from the Lokole repository

veeara282:
### Fixes Bug
N/A

To make testing Lokole on IIAB easier, I've added an alternative play that installs an unreleased version of Lokole from [ascoderu/opwen-webapp](https://github.com/ascoderu/opwen-webapp) instead of a released version from PyPI. This play is activated by setting `lokole_commit` to the ID of the commit you want to install from. If `lokole_commit` is unset, the standard play to install a released version will run instead.

I anticipate that this will streamline addressing #1830 and future integration testing of Lokole &times; IIAB.

N/A, although I have used this hack on my development server, which runs Ubuntu 18.04.2 LTS.

@holta

### DIFF
--- a/roles/lokole/tasks/install.yml
+++ b/roles/lokole/tasks/install.yml
@@ -21,7 +21,25 @@
     extra_args: --no-cache-dir    # To avoid caching issues e.g. soon after new releases hit https://pypi.org/project/opwen-email-client/
   tags:
     - install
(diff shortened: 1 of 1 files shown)


holta:
Assists @m-anish with resolution of #1830

Issue #1406, opened 2019-01-17
is_rpi should test for RPi hardware (is_raspbian now tests for OS across diff CPU's)


holta:
Currently we test for debian-9 in at least 3 quite different ways, as seen below.  These scattershot approaches should be harmonized into a best practice across all IIAB playbooks.

NOTE: raspbian-9 is a subset of debian-9 according to the /opt/iiab/iiab/vars/<OS>.yml files below:
```
root@box:/opt/iiab/iiab# grep -r is_debian
vars/default_vars.yml:is_debian: False
vars/default_vars.yml:is_debian_10: False
vars/default_vars.yml:is_debian_9: False
vars/default_vars.yml:is_debian_8: False
vars/debian-8.yml:is_debian: True
vars/debian-8.yml:is_debian_8: True
vars/raspbian-9.yml:is_debian: True
vars/raspbian-9.yml:is_debian_9: True
vars/debian-9.yml:is_debian: True
vars/debian-9.yml:is_debian_9: True
vars/debian-10.yml:is_debian: True
vars/debian-10.yml:is_debian_10: True
vars/raspbian-8.yml:is_debian: True
vars/raspbian-8.yml:is_debian_8: True
roles/httpd/tasks/main.yml:  when: is_debian
roles/httpd/tasks/main.yml:  when: is_debian and ansible_distribution_major_version == "8"
roles/httpd/tasks/main.yml:  when: (is_debian and ansible_distribution_major_version == "9") or is_ubuntu_18
roles/httpd/tasks/php-stem.yml:  when: is_debian_9 and (ansible_machine == "x86_64")
roles/monit/tasks/main.yml:  when: is_debian and ansible_distribution_major_version == "8"
roles/monit/tasks/main.yml:  when: is_debian and ansible_local.local_facts.os_ver == "debian-8"
roles/network/tasks/debian.yml:  when: iiab_wan_iface != "none" and is_debian_9
roles/network/tasks/debian.yml:  when: wan_ip != "dhcp" and iiab_wan_iface != "none" and is_debian_9
roles/2-common/tasks/packages.yml:  when: internet_available and is_debian_9 and not is_raspbian_9
roles/nodered/templates/node-red.service.j2:{% if is_debian_8 or is_debian_9 or is_ubuntu_16 or is_ubuntu_17 %}
roles/nextcloud/tasks/main.yml:  when: is_debian
[...]

PR #1842, opened 2019-07-08
Ugly workaround so that IIAB installs on Debian 10 Buster (which suddenly removed monit)

holta:
Bypasses installation of 'monit' on Debian 10+ for now.

Not our fault, that Debian 10 suddenly yanked `monit` during the very final days prior to its 2019-07-06 release.

Still, IIAB variables `is_debian` and `is_debian_*` really should "mean what they say and say what they mean" in future, e.g. for IIAB 7.1 [*] just as when spoken out loud / in plain English!  But let's leave it as is just for now, to get IIAB 7.0 out the door!

Documented @ https://github.com/iiab/iiab/blob/master/roles/monit/tasks/main.yml

[*] Much like #1406, `is_rpi` should likewise speak the truth about RPi hardware regardless of OS.  In that case, the separate variable `is_raspbian` is established for purposes of identifying the Raspbian OS &mdash; no matter whether the underlying HW is RPi, PC or Mac.

### DIFF
--- a/roles/monit/tasks/install.yml
+++ b/roles/monit/tasks/install.yml
@@ -22,19 +22,19 @@
     group: root
     mode: 0600
 
-- name: Install config file /etc/monit.d/watchdog from template (NEVER RUNS, WHY?)
-  template:
-    src: watchdog
-    dest: /etc/monit.d/watchdog
-    owner: root
-    group: root
-    force: yes
-    mode: 0755
-  register: monit_config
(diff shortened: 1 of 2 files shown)


holta:
Tested on 3 kinds of OS's:
- (real!) Debian 10
- (fake!) Debian 10, specifically Raspbian Buster Lite
- Ubuntu

PR #1847, opened 2019-07-08
Make Ansible happy [so box.lan resolves on Debian 10 Buster]

jvonau:
### Fixes Bug
#1815
hand change on .30

### DIFF
--- a/roles/0-init/tasks/hostname.yml
+++ b/roles/0-init/tasks/hostname.yml
@@ -24,11 +24,12 @@
     mode: 0644
   when: is_redhat | bool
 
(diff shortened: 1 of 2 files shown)


holta:
@jvonau:

1) Is https://github.com/iiab/iiab/blob/master/roles/network/tasks/hosts.yml#L17-L26 really just a failsafe?

   Given that https://github.com/iiab/iiab/blob/master/roles/0-init/tasks/hostname.yml#L27-L35 should always have been run during install time?

2) Amazing that this seems to also fix name resolution of box.lan and box starting almost immediately (minutes if not seconds) after IIAB's installer begins, this is a very welcome addition (that we can hope works across all Linux distros!)

Issue #1437, opened 2019-01-31
[Debian 10] MongoDB not installable due to missing packages [for Sugarizer]


arky:
There is no package named 'mongodb' on Debian 10. More work needs to be done to enable mongodb service on Debian 10.

https://github.com/iiab/iiab/blob/c461767777ef6a69f8afcb9ea50114ce90d01df3/roles/mongodb/tasks/main.yml#L5

Refer: https://docs.mongodb.com/manual/tutorial/install-mongodb-on-debian/

holta:
Ref: #1387 #1449

deldesir:
mongodb package exists now on Debian 10 @arky.

holta:
> mongodb package exists now on Debian 10 @arky.

Do you have a link?  What version of MongoDB is currently being offered?

holta:
This appears to be MongoDB 3.4.18 for now, and might even work on RPi potentially ?

https://packages.debian.org/source/buster/mongodb
https://packages.debian.org/buster/mongodb
https://packages.debian.org/buster/mongodb-server

holta:
> This appears to be MongoDB 3.4.18 for now, and might even work on RPi potentially ?
> 
> https://packages.debian.org/source/buster/mongodb
> https://packages.debian.org/buster/mongodb
> https://packages.debian.org/buster/mongodb-server

Once again MongoDB was removed from Debian 10 "Buster" — the links above no longer work.

Also note Red Hat announced last week that it is removing MongoDB from RHEL 8.

@llaske is there an idea how Sugarizer will navigate this?

llaske:
@holta If MongoDB is no longer include as a Debian package, it's still possible to install it by hand.
On our side we're using a Docker container to avoid deployment issues.
See a schema of the architecture here: https://github.com/llaske/sugarizer-server/blob/dev/docs/architecture.md

floydianslips:
Mongo is still not available on Debian 10. Are there any suggestions on how to move forward? It would appear it is not coming back.

holta:
> Mongo is still not available on Debian 10. Are there any suggestions on how to move forward? It would appear it is not coming back.

Let's discuss during Thursday's call (http://minutes.iiab.io).

RHEL and therefore CentOS are also dropping MongoDB so we have a bigger dilemma growing here e.g. if Ubuntu drops it too?

(We may need a few months to understand what long-term strategy makes most sense?)

floydianslips:
[...]

PR #1852, opened 2019-07-08
Fix #1850: Fully bypass Sugarizer (AND MongoDB) on Debian 10+

holta:
Builds on PR #1850.  To mitigate #1437.

Also documents that:
- `mongodb_install` is completely ignored
- FYI `mongodb_enabled: False` works but is ineffective, as Sugarizer starts mongodb's systemd svc on its own


holta:
Tested on Debian 10 Buster.

Curiously not all 3 steps of Sugarizer's install are bypassed (the middle step below still runs on Debian 10!) but this will suffice for now:

1. MongoDB (bypassed)
2. Node.js (runs to completion, even if perhaps unecessarily)
3. Sugarizer (bypassed)

PR #1853, opened 2019-07-09
Activate local-only mode [Lokole "offline" email]

veeara282:
### Fixes Bug
#1830 -- even with the previous PR (#1844), the bug isn't actually fixed!

Configure the Lokole webapp with environment variable `OPWEN_SIM_TYPE='LocalOnly'` so that the Lokole offline instance knows that it is offline.

Here's a screenshot of the app with this change:
Screenshot of Lokole app when logged in as admin. Note that the navbar does not contain a settings menu item.

Confirmed that this works on my staging box.

@holta @m-anish


holta:
@aidan-fitz can you put a sentence into https://github.com/iiab/iiab/blob/master/roles/lokole/README.rst explaining how (and when!) to override variable `OPWEN_SIM_TYPE='LocalOnly'` ?

holta:
> @aidan-fitz can you put a sentence into https://github.com/iiab/iiab/blob/master/roles/lokole/README.rst explaining how (and when!) to override variable `OPWEN_SIM_TYPE='LocalOnly'` ?

e.g. some IIAB's are run with various kinds of (intermittent?) Internet connections, and some are run more fully offline.

In which circumstances do you recommend manually overriding this variable?

And is changing this possible at the command line, for less technical field operators — when the IIAB machine's placement/circumstances unexpectedly change after it's been installed — e.g. the village receives partial Internet of some kind?

holta:
A few 3G/4G LTE modem suggestions, that others have used with Lokole:
https://github.com/ascoderu/opwen-webapp#production-setup

veeara282:
@c-w Based on this passage from the README, if a sysadmin wanted to configure Lokole to work with a modem, they would have to use the installation script. They could not just set `OPWEN_SIM_TYPE=mkwvconf`. However, during installation, any USB modem plugged in would be configured automatically. Is this correct?

> There is a script to set up a new Lokole device. The script will install the email app in this repository as well as standard infrastructure like nginx and gunicorn. The script will also make ready peripherals like the USB modem used for data exchange, and set up any required background jobs such as the email synchronization cron job.

For context, some IIAB devices will be connected with USB modems.

veeara282:
@holta I'm thinking that we can add a separate playbook to configure an already installed Lokole software instance to use a modem plugged into the IIAB device, and to deconfigure it back to local-only mode.

Possible failure: the Lokole instance was configured to use a modem, but starts up with no modem or a different modem plugged in. Would it fail?

holta:
@m-anish @jvonau can you review the suggested IIAB + 3G/4G Modem approach below?

My own hunch (I could be quite wrong!) is that many/most technical implementers would prefer to have clean docs / HOWTO in this case?  Rather than an additional playbook that might be hard to maintain with all sorts of different/evolving USB "GSM" modems?

> @holta I'm thinking that we can add a separate playbook to configure an already installed Lokole software instance to use a modem plugged into the IIAB device, and to deconfigure it back to local-only mode.
> 
> Possible failure: the Lokole instance was configured to use a modem, but starts up with no modem or a different modem plugged in. Would it fail?

holta:
@aidan-fitz Part of my concern is that IIAB 7.0 should not be adding any playbooks at this very late stage in its release cycle.

But if that's definitely the best way to skin the cat, for IIAB 7.1 or an emergency exception to IIAB 7.0's current feature freeze or whatever, we'll certainly be understanding...of what's most practical+pragmatic for Lokole communities!

c-w:
There are quite a few things that need to happen for Lokole to use a USB modem and send live emails ([see our full install script](https://github.com/ascoderu/opwen-webapp/blob/master/install.py)) such as setting up WiFi (hostapd and dnsmasq), setting up modem tools (usb_modeswitch and wvdial) and setting up the MX records and Sendgrid account (requires a registration credential included in a http request to [opwen-cloudserver](https://github.com/ascoderu/opwen-cloudserver)). Additionally, we recently offloaded email exchange to a worker thread to increase reliability so the opwen-webapp daemons (celery worker, celery beat and restarter) would also have to be set up.

Once all of this machinery is set up, the app will take care of period or on-demand email exchange as well as configuring the supported modems. Modem dial-up information can either be pulled from a database via [mkwvconf](https://github.com/ascoderu/mkwvconf) or configured manually. Both of these tasks can be done in the admin UI so adding support for new SIM card configurations is essentially self-service. Adding support for new modems is mode involved since it requires knowing the correct magic bytes to switch the modem from storage mode to dial-up mode via usb_modeswitch so supporting new modems currently requires a code change.

While all of this is somewhat involved, it can all be done: we already solved all these problems in the Lokole project so it should be possible to port the code to IIAB. My bigger concern is how to manage client registration. Setting up a new Lokole email domain is fully automated but we don't have the support staff to handle the increased operational burden from unplanned/ungated registrations which is why we currently onboard new inboxes after discussing with the owner of the inbox. How would you propose to best share the workload between IIAB deployments that wish to use a live Lokole integration?

m-anish:
Sorry, but can someone outline:
1. I have a IIAB setup with the lokole with the issue
2. The IIAB will always be offline
3. What commands do I need to run?

m-anish:
Ignore this for now as the settings page seems to be irrelevant.

veeara282:
I don't want to delay this PR from being merged, so if there's anything y'all need me to do, I'll do it in a follow-up PR.

holta:
> I don't want to delay this PR from being merged, so if there's anything y'all need me to do, I'll do it in a follow-up PR.

I've lived through more than enough decades to know that documentation needs to happen _before_ merging, not _sometime later_ (-:

So call if it's easier.  As mentioned at the beginning of the week, https://github.com/iiab/iiab/blob/master/roles/lokole/README.rst needs to be updated, to explain `OPWEN_SIM_TYPE='LocalOnly'`

We should also say...something like:

- Lokole can be connected to the Internet on a nightly basis to deliver and receive emails globally.  However, arranging this is extremely complicated.  You would need an appropriate 3G/4G/5G modem and an Internet expert familiar with modem protocols and MX records, etc.  Ask that person to read https://github.com/ascoderu/opwen-webapp#opwen-webapp in its entirety, to help you understand if this might be realistic or not.

- Lokole and Internet-in-a-Box would welcome a business plan (whether volunteer-based, grant-based or for-profit) from someone willing to operationalize this — making it relatively hassle-free for schools, clinics, libraries and orphanages around the world — that generally do not have access to technical experts.  Please [contact us](http://FAQ.IIAB.IO#What_are_the_best_places_for_community_support.3F) if you have the capacity to help make such a social enterprise happen.

holta:
After https://github.com/iiab/iiab/blob/master/roles/lokole/README.rst communicates the very basics (as this PR should do) these other Lokole docs can sometime later be refined as necessary: _(not every doc below is preaching to the same choir of course, so the focus/audience/expectation of each can be quite different!)_

1. Lokole's main README, including the [Production setup](https://github.com/ascoderu/opwen-webapp/blob/master/README.rst#production-setup) section at the bottom:
   https://github.com/ascoderu/opwen-webapp/blob/master/README.rst
2. The 10-page Lokole-IIAB User Guide:
   https://github.com/iiab/iiab/blob/master/roles/lokole/The%20Lokole-IIAB%20User's%20Manual.pdf
3. Introductory Summaries at http://FAQ.IIAB.IO
   - Question/Answer 3:
   http://FAQ.IIAB.IO#What_can_I_do_with_E-books_and_Internet-in-a-Box.3F
   - Question/Answer 9:
   http://FAQ.IIAB.IO#What_services_.28IIAB_apps.29_are_suggested_during_installation.3F
4. Getting Started summaries that appear automatically on http://box IIAB home pages, for IIAB implementers who've chosen to install Lokole:
   - https://github.com/iiab/iiab-admin-console/blob/master/roles/js-menu/files/menu-files/menu-defs/en-lokole.json#L6
   - https://github.com/iiab/iiab-admin-console/blob/master/roles/js-menu/files/menu-files/menu-defs/en-lokole.html
5. What appears in /etc/iiab/iiab.ini that is very rarely read, but prob still should be accurate!
   https://github.com/iiab/iiab/blob/master/roles/lokole/tasks/install.yml#L144

veeara282:
@holta: I've added three paragraphs toward the end of the README about the default configuration and how Lokole could be configured to use the Internet. I've written it without making any assumption as to whether the IIAB device is using a USB modem, Ethernet cable, or SIM card, even though it seems like only USB modems are used in reality.

Issue #1658, opened 2019-05-13
Moodle 3.5.7 LTS released "2019-07-08" fails w/ PHP 7.3


holta:
ETA [2019-07-08](https://docs.moodle.org/dev/Releases#Moodle_3.5_.28LTS.29) but that might slip by a week or so...the date can often change, due to unforeseen circumstances...as Moodle folks try to warn us:
```
"All releases are preceded by a one week warning."
```

> Bad news is Moodle 3.5.x LTS is starting to show its age. Perhaps we should consider non-LTS Moodle 3.7.x when it's released in coming weeks — or Moodle 3.8.x when it's released by/around December 2019?
>
> Or possibly holding off til "Moodle 3.9.x LTS" in Q2 2020 ??
>
> ASIDE: people like @ericnitschke @kananigit @mikkokotila are starting to look at more nimble alternative LMS's even including WordPress, for more focused responses to very low-income community/school needs~

Ref: #1529

holta:
Draft release notes for Moodle 3.5.7 should appear here in coming days, prior to its release:

https://docs.moodle.org/dev/Moodle_3.5.7_release_notes

holta:
> Draft release notes for Moodle 3.5.7 should appear here in coming days, prior to its release:
> 
> https://docs.moodle.org/dev/Moodle_3.5.7_release_notes

Released/Confirmed!

Please all who can test Moodle 3.5.7 going forward.

georgejhunt:
When I bring up the install page, (rpi,buster,big), I get at the top of the validation page:
PHP version 7.3 is not supported.
Reading moodle docs at: https://docs.moodle.org/dev/Moodle_3.5_release_notes

<quite> 
PHP version: minimum PHP 7.0.0 Note: minimum PHP version has increased since Moodle 3.3. PHP 7.1.x and 7.2.x are supported too. PHP 7.x could have some engine limitations.
</quote>
and 
https://docs.moodle.org/dev/Moodle_and_PHP7
"PHP 7.3 can be used with Moodle 3.6.4, Moodle 3.7 and later releases." ( buster installs php7.3)

holta:
Thank @georgejhunt for catching this.

Can we consider dumping Moodle 3.5.x LTS in favor of 3.7.x ?

georgejhunt:
Or perhaps just dropping support for moodle entirely.

Do you know of any moodle users?

holta:
[...]

PR #1857, opened 2019-07-10
Moodle 3.5.x -> 3.7.x, for new OS's that contain PHP 7.3, for #1658

holta:
See #1658 PHP 7.x / 7.3 issue.


holta:
Smoke-tested on Raspbian Buster on RPi 4.

Issue #1851, opened 2019-07-08
When http://box/lokole doesn't load until an hour later, on 1st install of IIAB


holta:
@c-w & @aidan-fitz do you know why happens on fresh installs of IIAB (http://download.iiab.io) ?

This seems to happen on most (likely all) Linux distros, as if Apache isn't fully configured until about 30-45 minutes later.

This is confusing as this symptom does not happen to other IIAB Apps like http://box/nextcloud etc, which begin to function immediately after they are installed, while IIAB's install process continues.

Granted this is not a complete showstopper, as one can always wait an hour...

But what other/specific logs or diagnostics might help debug this, if you have any ideas?

FYI during IIAB's installation (typically about an hour long) the client machine's browser shows the following, when visiting http://box/lokole, or in this example http://10.8.0.14/lokole :
```
Internal Server Error
The server encountered an internal error or misconfiguration and was unable to complete your request.

Please contact the server administrator at [no address given] to inform them of the time this error occurred, and the actions you performed just before this error.

More information about this error may be available in the server error log.

Apache/2.4.38 (Debian) Server at 10.8.0.14 Port 80
```
At the same time, the IIAB server's `/var/log/apache2/error.log` shows:
```
[Mon Jul 08 16:16:41.713613 2019] [proxy:warn] [pid 27211] [client 192.168.0.11\
1:61911] AH01144: No protocol handler was valid for the URL /lokole/ (scheme 'h\
ttp'). If you are using a DSO version of mod_proxy, make sure the proxy submodu\
les are included in the configuration using LoadModule.
```

holta:
CLARIF: a reboot is not required.

I wish I knew what was required, but generally waiting about 30-45 minutes does the trick :)

Whereupon http://box/lokole starts to work &mdash; even before IIAB's own network stuff (Ansible role) has even begun to run (i.e. to tidy up IIAB networking configs, prior to field deployment etc).

holta:
Profound thanks to @jvonau for fixing this in PR #1859 !

PR #1859, opened 2019-07-11
Need to reload systemd daemon when adding apache config files [Lokole]

jvonau:
### Fixes Bug
#1851


holta:
Tangentially Related:

#1830, PRs #1844 #1853

PR #1872, opened 2019-07-15
kiwix reports size in K

tim-moody:


holta:
Close but no cigar?

When I installed this PR on RPi 4 and then ran `iiab-make-kiwix-lib`, all ZIM sizes were then displayed as "Size: null" on the IIAB home page (http://box).

PR #1880, opened 2019-07-16
Udev cleanup

jvonau:


holta:
@jvonau what tests does roles/2-common/tasks/udev.yml need here:
- `./runrole 2-common` ?
- a fresh IIAB install ?

Issue #1867, opened 2019-07-14
Need: a much more recent Raspberry Pi Guide than the 2012 Edition


holta:
The 152-page PDF offered by http://oer2go.org/viewmod/en-rpi_guide installable by http://box.lan/admin > `Install Content` > `Get OER2Go(RACHEL) Modules` is historically fascinating, as a 2012 manifesto and guide (http://www.cs.unca.edu/~bruce/Fall14/360/RPiUsersGuide.pdf) to what the Raspberry Pi movement became and is becoming.

This is what installs on IIAB's home page (http://box) with title "Rasp Pi User Guide" (and no description) from https://github.com/iiab/iiab-admin-console/blob/master/roles/js-menu/files/menu-files/menu-defs/en-rpi_guide.json

_But 7 years later we should find something much more relevant to 2019's developing world classroom/lab needs!_

holta:
@tim-moody can you arrange for the 315-page 28.5MB 2016 RPi User Guide (4th edition) to be automatically downloaded from https://dn.odroid.com/IoT/other_doc.pdf whenever requested by someone using the Admin Console?

holta:
The 244-page "Official" Raspberry Pi Beginner's Guide from 2018 (1st Edition) should probably also be included, e.g. for those who chose to download it from:
https://www.raspberrypi.org/magpi-issues/Beginners_Guide_v1.pdf

holta:
Ref: https://github.com/iiab/iiab-admin-console/issues/217 "Raspberry Pi User Guide (PDF)"

holta:
@georgejhunt might work on this, e.g. possibly offering IIAB implementers the option to download these 2 essential docs:
- https://dn.odroid.com/IoT/other_doc.pdf 315-page 28.5MB 2016 RPi User Guide (4th edition)
- https://www.raspberrypi.org/magpi-issues/Beginners_Guide_v1.pdf 244-page "Official" Raspberry Pi Beginner's Guide from 2018 (1st Edition)

Certainly most teachers would benefit if https://github.com/iiab/iiab/blob/master/roles/httpd/templates/refresh-wiki-docs.sh facilitated offline + impoverished communities having access to these within http://box/info

SJ Klein also strongly recommends we advertise this http://box/info (IIAB's offline docs) much more prominently near the top of newly installed IIAB home pages (http://box),

holta:
[...]

PR #1888, opened 2019-07-18
Download Raspberry Pi resources and make available offline

georgejhunt:
This has been tested on rpi4. 

It is coupled with a change to the iiab main wiki page (already done).

The main wiki page has two bullets for the two pi resources. The bullets are not translated by lynx correctly -- they are rendered inline.

But the links to the web are correctly changed to refer to /info/<rpi pdfs>

The update procedure for an already initialized pi involves:
```
./runrole httpd
iiab-refresh-wiki-docs
```


holta:
For #1867 "Need: a much more recent Raspberry Pi Guide than the 2012 Edition"

holta:
Thanks much @georgejhunt !

FYI the revised https://github.com/iiab/iiab/wiki splash page (i.e. what appears offline as http://box/info) makes this all possible &mdash; that I further tweaked to allow `pandoc` to digest this properly, generating actual bullet points & clean line breaks in the offline version.

Issue #1889, opened 2019-07-20
Tracking tkt for "Offline link to http://box/info (IIAB tech docs)"


holta:
https://github.com/iiab/iiab-admin-console/pull/244

PR #1891, opened 2019-07-20
Take 2: Add Lokole PDF (User's Guide) to http://box/info offline docs

holta:
Building off PR #1890

Works with #1889 ~= iiab/iiab-admin-console#244


holta:
Tested on Ubuntu 18.04

Issue #1394, opened 2019-01-15
IIAB fails to correctly autodetect VM networking setup


m-anish:
### Expected behavior and actual behavior.

I am setting up IIAB as per the specs below. While the install proceeds smoothly, the network mode chosen is gateway, and the dhcp server sits on the public bridged interface. Ideally, it should have auto-detected the dhcp server on that interface and set it self up as an appliance. Re-running iiab-network also does not fix the problem.

### Specifications like the version of the IIAB, operating system version, or hardware details.
I am testing the IIAB one line install with this as the latest commit
b2be3ccae19bb52be80b3f399989fcafd5edca49

Setup:
Vagrant, Ubuntu 18.04 (Bionic) VM, IIAB One line install with mininal vars file
Network setup contains two interfaces
1. default NAT
2. public bridged interface

holta:
Thanks @jvonau who's currently helping over live chat.  In reply, @m-anish has posted:
```
root@box:/opt/iiab/iiab/vars# ip r
default via 10.0.2.2 dev enp0s3 proto dhcp src 10.0.2.15 metric 100 
10.0.2.0/24 dev enp0s3 proto kernel scope link src 10.0.2.15 
10.0.2.2 dev enp0s3 proto dhcp scope link src 10.0.2.15 metric 100 
172.18.96.0/19 dev br0 proto kernel scope link src 172.18.96.1
```
And for `ip a` :
http://paste.ubuntu.com/p/qz5GrvgbK4/

And for `brctl show` :
http://paste.ubuntu.com/p/82xczbbvqw/

holta:
Chat except:

@jvonau: I'd just remove the internal Nat from the vagrant file..  or pre-populate /etc/iiab/local_vars.yml with the change mentioned above..

@m-anish: sure. i will do that.. 
but do you think this needs to be fixed sometime?

@jvonau: Or bug ansible's devs.. 
Nothing that I'm going to automate, the overrides are there to provide the functionality wanted.

@m-anish: okay.fair enough
but this is something that can break somebodys network

holta:
Chat excerpt:

@m-anish: do you think i could do something like this....

https://www.vagrantup.com/docs/networking/public_network.html#default-router

the "default router" section

@jvonau: What does your [vagrant] file look like?

@m-anish: https://termbin.com/6bn9

basically one extra line... config.vm.network "public_network"

the internet seems to suggest that removing the default nat interface is a) not possible or b) not a good idea

[...]

PR #1906, opened 2019-07-29
use 'not' in place of '!='

jvonau:
### Fixes Bug
should correct #1394 where setting “iiab-lan-enabled" had no effect, I suspect because of the use of '' in the when clauses(line 93, 155). The rest is to move to using not in place of != that I suspect is going away soon.


holta:
FYI @m-anish, who reported #1394

PR #1920, opened 2019-08-10
Node speed

jvonau:
Should offer some speed improvements

### DIFF
--- a/roles/nodejs/tasks/main.yml
+++ b/roles/nodejs/tasks/main.yml
@@ -69,6 +69,7 @@
   shell: curl -sL https://deb.nodesource.com/setup_{{ nodejs_version }} | bash -
   args:
     warn: no
+    creates: /etc/apt/sources.list.d/nodesource.list
   when: internet_available and is_debuntu
   #when: internet_available and (is_debian_8 or is_debian_9 or is_ubuntu_16 or is_ubuntu_17)
   # NOT NEC TO TEST FOR is_raspbian_8 OR is_raspbian_9 AS /opt/iiab/iiab/vars/<OS>.yml


holta:
I'm concerned this prevents https://github.com/iiab/iiab/tree/master/roles/nodejs from upgrading Node.js point releases.

While it's true that 'apt update; apt -y dist-upgrade' also upgrades Node.js -- not everone does that. 
 Hence it was intentional when I wrote the node role/playbook to rather forcefully install the latest Node in almost all circumstances.

Maybe there's a clever way to dial this back?  But #1920 seems to go too far i.e. preventing most all such upgrades of Node.js 10.x, which just doesn't seem right.

jvonau:
When you actually read [setup_10.x](https://deb.nodesource.com/setup_10.x) all it does is set location and the major version(10), point release increments should be handled within the apt packages in the 'package' stanza below. Nothing is preventing point version upgrades as nodejs_version_installed.stdout is only concerned with the major number(10x)

jvonau:
I reverted the second commit as I forgot 10.x is shipping with the distros now and would prevent setup_10.x from running.

holta:
@mitra42 please review this PR, which saves about 100 seconds on RPi 4 during subsequent runs of `./runrole nodejs`

...if we're in fact willing to skip the step that does `curl -sL https://deb.nodesource.com/setup_10.x | bash -` ?

[ Whereas re-running subsequent step "Install latest Node.js 10.x which includes /usr/bin/npm (debuntu)" takes only a few seconds, regardless if the curl-to-bash command is re-run or not. ]

mitra42:
The `curl -sL https://deb.nodesource.com/setup_10.x | bash - ?` lines is just to handle the various different platforms, I think we included this in our default install because of the challenge of getting it to run on the Rachel3+ box which is 32 bit. I have a feeling that without it at least some linux distro's refused to update leading to later problems.   If it works on all the platforms that IIAB supports then I don't have a problem with it being removed, but I'd suggest at least for now adding a comment about how to reenable it if it fails.

Issue #1812, opened 2019-07-03
Investigate @mitra42's suggestions re: severe recent slowdowns (which are related to 'internetarchive' ?)


holta:
This can be deferred to IIAB 7.1 if necessary, but several severe recent slowdowns in IIAB 7.0 pre-releases have been observed, that might possibly be related:

- MySQL (MariaDB 10.1.40) times out on every boot [never used to happen] #1802
- `yarn add @internetarchive/dweb-archive @internetarchive/dweb-mirror` can take 15+ minutes to install [used to by 5 minutes] https://github.com/iiab/iiab/pull/1806#issue-293486566
- See @mitra42's recommendations for monitoring these possible patterns here: https://github.com/iiab/iiab/pull/1780#issuecomment-507451425

holta:
From the user's perspective, slowness is a bug, so I'm escalating this one (whether we solve it for IIAB 7.0 or 7.1!)

holta:
Tangentially, see also this list that @mitra42 is compiling during his profiling to uncover where slowdowns are happening during (re)installs of 'internetarchive' etc: (similar to https://github.com/iiab/iiab/issues/1915#issuecomment-517874297)

Package dependency warnings {META}
internetarchive/dweb-mirror#234

holta:
@mitra42 @jvonau please LMK if the mitigations implemented by PRs #1920 #1921 #1923 #1927 are now sufficient to release IIAB 7.0, which needs to happen ASAP (Thanks!)

mitra42:
@holta  if you are seeing it install correctly then I don't see any reason to hold up 7.0

mitra42:
I think this can be closed now PR #1982  should significantly improve install speed.

PR #1921, opened 2019-08-10
Skip yarn add if installed, but allow yarn upgrade if set in local_vars.yml

jvonau:
### Fixes Bug #1812
Install only once to allow faster operation of [iiab-from-console](https://github.com/iiab/iiab/blob/master/iiab-from-console.yml) from the iiab admin page. Allows future upgrades with a fact that can be set via local_vars or some other logic to toggle the fact.

### DIFF
--- a/iiab-install
+++ b/iiab-install
@@ -124,6 +124,7 @@ if [ -f /etc/iiab/iiab.env ]; then
         echo "Removed /etc/iiab/iiab.env effectively resetting STAGE (counter)."
     elif [ "$1" == "--reinstall" ]; then
         STAGE=0
+        ARGS="$ARGS --extra-vars reinstall=True"
         sed -i 's/^STAGE=.*/STAGE=0/' /etc/iiab/iiab.env
         echo "Wrote STAGE=0 (counter) to /etc/iiab/iiab.env"
     elif [ "$STAGE" -ge 2 ] && [ "$1" == "--debug" ]; then
--- a/roles/0-init/tasks/first_run.yml
+++ b/roles/0-init/tasks/first_run.yml
@@ -1,2 +1,8 @@
+- name: Create symlink /usr/bin/iiab-diagnostics
+  file:
+    src: "{{ iiab_dir }}/scripts/iiab-diagnostics"
+    dest: /usr/bin/iiab-diagnostics
+    state: link
+
 - name: Create {{ iiab_ini_file }}
   include_tasks: iiab_ini.yml
--- a/roles/0-init/tasks/main.yml
+++ b/roles/0-init/tasks/main.yml
@@ -1,11 +1,4 @@
 # Initialize
-
-- name: Create symlink /usr/bin/iiab-diagnostics
-  file:
-    src: "{{ iiab_dir }}/scripts/iiab-diagnostics"
-    dest: /usr/bin/iiab-diagnostics
-    state: link
-
 - name: ...IS BEGINNING ============================================
   stat:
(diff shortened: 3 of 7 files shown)


mitra42:
@jvonau this something I added to my list yesterday so its great ... 
I think there are three cases
a) fresh install of IIAB + IA - we want it to do `yarn install`
b) repeat install of IIAB + IA - we want it to skip 
c) the update case - where I was planning to tell the user to do a `runrole internetarchive` and have it do a `yarn upgrade`.   

I think with your PR the `runrole internetarchive` would be a No-op ? 

(Apologies for any slow response, I'm leaving San Francisco for Australia later today, though I'll probably have a couple of internet windows during between now and arrival Monday evening Australia time (Sunday night US time)

jvonau:
You have to remember that iiab-install, runrole, and admin-console all share the same code from the roles directory. A & B work as is with this revision, while C just requires populating local_vars.yml (or config_vars if the code is in the admin-console for it) with 'internetarchive_upgrade: True' to run the yarn upgrade path with runrole. Warning while testing I did run into the same kind of failure as in #1915 but using yarn upgrade under runrole.

mitra42:
Yes - this all makes sense except asking people to edit config files ... which I suspect will lead to failures. Feel free to accept the PR thought, I can leave the instructions I give people as 
```
cd /opt/iiab/internetarchive && yarn upgrade
```
rather than telling them to use runrole. 

Note, I'm presuming there is no way to tell pass this arg to runrole.

jvonau:
Um, I changed runrole to take --reinstall as an optional argument and pass that info to ansible covering option C.  Without --reinstall runrole internetarchive will behave like A&B above but cover disabling/enabling the service quickly post-install. Now with  runrole --reinstall internetarchive you will run the yarn upgrade code path.

mitra42:
Thanks @jvonau - looks good (at least to my minimal ansible knowledge)

During some testing I noticed that `nodejs` is also reinstalling every time, even if its installed, and takes about 2 or 3 minutes even on a fast link - it might be worth using a similar technique there (and on `yarn`, though that only takes about a minute to reinstall each time)

jvonau:
> During some testing I noticed that `nodejs` is also reinstalling every time, even if its installed, and takes about 2 or 3 minutes even on a fast link - it might be worth using a similar technique there (and on `yarn`, though that only takes about a minute to reinstall each time)

There is a minor speedup with PR #1920 with skipping the setup of nodesources.list. Little background on ansible's package module's state: latest verses state: present, using 'present' just checks if the packages are installed and skips the install if already present while 'latest' will force an upgrade to the newest when the newest is not installed. We have to use latest in those roles to install the newest packages offered from the third party repos as the stock distro might have already installed a lessor version of the requested package.

mitra42:
Agreed regarding `latest`, but note I test by doing the runrole and then immediately repeating the runrole so that any time is purely 'wasted' time, rather than a legitimate upgrade.

mitra42:
@jvonau I've been investigating this more as I write install scripts for other platforms. (esp OrangePi/Armbian). I discovered the 'right' combo is .... 
If not already installed `yarn install`
If already installed `yarn install; yarn upgrade` 

The install runs really quickly if there is nothing to do, but catches cases where something new has been added or a version dependency changed. 

I'm not sure how to do this in the ansible? 
Note - I also posted a request for someone with a 1GB RPI4 to test with a config change to yarn - its in https://github.com/iiab/iiab/issues/1931, I don't have such a machine to test on.

PR #1938, opened 2019-09-02
Rpi static

jvonau:
### Fixes Bug
per chat testing


jvonau:
If you change the name of the var, you would need to change it in defaults/main.yml and start off as True

holta:
> If you change the name of the var, you would need to change it in defaults/main.yml and start off as True

Yes.  PR #1939 is my new attempt, trying to add documentation to make this much more clear.

PR #1939, opened 2019-09-02
Ethernet Static IP if DHCP not detected (for eth0 on RPi, link-local 169.154.* optional)

holta:
Builds on PR #1938

PR #1940, opened 2019-09-03
Ethernet DHCP IP (for field updates) then fallback to Static IP (for offline schools)

holta:
Builds on & tightens up PR #1939, arising from @jvonau's PR #1938

### DIFF
--- a/roles/network/defaults/main.yml
+++ b/roles/network/defaults/main.yml
@@ -6,24 +6,13 @@
 # ports_externally_visible: 3
 # https://github.com/iiab/iiab/wiki/IIAB-Networking#firewall-iptables
 
-# Ethernet - IF NECESSARY, CUSTOMIZE THESE 4+3 VARS IN /etc/iiab/local_vars.yml
-#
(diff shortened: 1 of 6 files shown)


holta:
Tested on Raspbian Lite 2017-09-10 in varying conditions.

Accompanying documentation finalized:
http://wiki.laptop.org/go/IIAB/FAQ#How_do_I_set_a_static_IP_address.3F

holta:
FYI this was also tested on Raspbian Desktop 2018-11-13 from last year, just to verify compatibility with an older version of dhcpcd.  Good news is this (also) works with `wan_try_dhcp_before_static_ip: True`, i.e. doing the right things in both cases:
- eth0 accepts a DHCP IP when DHCP is on the wire, or...
- eth0 falls back to the Static IP when DHCP is not detected on the wire...

As usual, this (generally) happens within about 20 seconds after the Ethernet cable is inserted (in both above cases).

However there are occasional times when this takes 30 seconds, or even more.

holta:
Merging now- huge thanks to @jvonau for all his help researching this & reviewing it over the past 4 days.

PR #1941, opened 2019-09-04
Nextcloud Liberation Front: Unblock Nextcloud in 192.168.1.* schools & all others!

holta:
Work In Progress.  In short: arbitrary security rules are hurting real users (e.g. in Chiapas, Mexico schools) so it's time for a completely new approach.


holta:
Tested on Raspbian Lite & Ubuntu 18.04

Issue #1980, opened 2019-10-04
Big install failing in step 4 - No package matching 'dansguardian' is available"


mitra42:
### Expected behavior and actual behavior.
I'm seeing a failure of the "big" install, with ` No package matching 'dansguardian' is available"` 

### Steps to reproduce the problem.

I'm running the standard install process, and selecting "Big" 

It fails on dansguardian, same if I repeat it. 

### Specifications like the version of the IIAB, operating system version, or hardware details.

This is running on an IIAB, against the master.  The only thing I've changed was to control-C out of the first install run, then replace the internetarchive role with one that I'm trying (and failing) to test, but internetarchive is in Stage 9 and the failure is in Stage 4, so I'm pretty sure its not that change. 

I also notice #1373 also talks about danguardian causing problems, but that was closed a long time ago.

holta:
Thank you @mitra42

To show us details & basic logs, can you run `iiab-diagnostic` (or `sudo iiab-diagnostics` if nec) ?

Then send us the resulting output (i.e. a pastebin URL please) Thanks!

holta:
1) Do you know what's possibly trying to install DansGuardian ?

   FYI it's extremely unusual for anybody to install this package in 2019, which is no longer part of Buster etc &mdash; especially when BIG-sized IIAB has those dansguardian_install and dansguardian_enabled variables very intentionally set to false in https://github.com/iiab/iiab/blob/master/vars/local_vars_big.yml#L156-L158

2) @mitra42 in general plz do identify your exact OS if poss!

   e.g. as shown by the date and "stage2" or "stage5" in /etc/rpi-issue

   e.g. stage2 means "Raspbian Lite", and stage5 means "Raspbian With Desktop"

mitra42:
dansguardian was turning up true in my vars. But I also saw a problem where somehow I had what appears to be an old version in /etc/iiab/local_vars.yml despite having new versions in /opt/iiab/iiab/vars/local_*_vars.yml . I can't see where that came from since it was a clean install, but its possible I broke something banging my head against trying to get it to run against my fork of the repo (github.com/mitra42/iiab#mitra42) rather than against github.com/iiab/iiab#master

holta:
Yes, odd since we haven't seen DansGuardian installed or enabled in a very long time.

[...]

PR #1983, opened 2019-10-05
Warn implementers: DansGuardian is NO LONGER AVAILABLE in Debian Buster etc

holta:
Ref #1980


holta:
Note also that the lastest stable release of DansGuardian is more than a decade old, from June 2019.

(This obsolete ship has sailed long ago!)

PR #1984, opened 2019-10-05
Check at install time for stale local_vars with DG enabled

jvonau:
Remove DG from local_vars and check at install time for stale local_vars with DG enabled


jvonau:
Closing should be a bug fix for release-7.0

PR #1990, opened 2019-10-07
Install common Apache modules in roles/httpd, for {http://box/kiwix, http://box/kolibri, http://box/nodered, etc}

### LINKED ISSUE #1986: Install essential Apache URL modules upfront, so {Kiwix, Kolibri, Node-RED, etc} playbooks become self-sufficient--so short URL's like http://box/nodered work immediately
As of now, http://box/nodered does not work until the Kiwix (or Kolibri?) playbook is later run.  Others (Kolibri until recently) suffer the very same problem.

The needed code is here...installing 4 Apache modules so URL's work:
https://github.com/iiab/iiab/blob/master/roles/kiwix/tasks/kiwix_install.yml#L74-L83

Or possibly this single Apache is sufficient?
https://github.com/iiab/iiab/blob/master/roles/kolibri/tasks/main.yml#L97-L101

QUESTION: can we fix this by installing the essential Apache modules right away in roles/httpd [during Stage 3](https://github.com/iiab/iiab/blob/master/roles/3-base-server/tasks/main.yml#L6-L10) ?

holta:
Resolves #1986, helping more playbooks' short URL's work immediately, i.e. right after running each IIAB app/service's Ansible playbook.

e.g. http://box/kiwix, http://box/kolibri, http://box/nodered, etc.

### DIFF
--- a/roles/activity-server/tasks/main.yml
+++ b/roles/activity-server/tasks/main.yml
@@ -77,6 +77,7 @@
             group=root
             mode=0644
 
+# SEE ALSO THE apache2_module SECTION IN roles/httpd/tasks/main.yml
 - name: enable mod_expires for debian
   command: a2enmod expires
   when: is_debuntu | bool
--- a/roles/awstats/tasks/install.yml
+++ b/roles/awstats/tasks/install.yml
@@ -18,6 +18,7 @@
   tags:
     - download
 
+# SEE ALSO THE apache2_module SECTION IN roles/httpd/tasks/main.yml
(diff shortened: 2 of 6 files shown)


holta:
Smoke-tested with a fresh install on Ubuntu Server 18.04.3 &mdash; where I confirmed that http://box/nodered now begins to work right away after the roles/nodered playbook completes &mdash; unlike in the past when that URL would not start working until much later, i.e. after the Kiwix role completed.

PR #1995, opened 2019-10-09
Build on + revise @jvonau's PR #1994 ("Combine iiab-support & install-support") for OpenVPN

holta:
2 primary changes, above & beyond PR #1994:

- I prefer `/usr/bin/iiab-support` in place (not just a symlink) so that it generally works even when `/opt/iiab/iiab` disappears for any reason e.g. if the implementer chose to conserve disk/storage space, or for any other reason.

- Central file `roles/openvpn/tasks/main.yml` and directory `roles/openvpn/templates/` have been extensively pruned of older/deprecated/unused "bitrot"


iiab-support
```diff
@@ -46,14 +49,17 @@ fi
...
-    echo -e "Please wait a few minutes as IIAB Stage 1 (1-prep) & OpenVPN are installed...\n"
+    echo -e "Plz wait a few minutes as sshd, iiab-admin & OpenVPN are confirmed/installed...\n"
     if grep -q '^openvpn_install:' /etc/iiab/local_vars.yml; then
         sed -i "s/^openvpn_install:.*/openvpn_install: True/" /etc/iiab/local_vars.yml
     else
         echo "openvpn_install: True" >> /etc/iiab/local_vars.yml
     fi
     cd /opt/iiab/iiab
-    ./runrole 1-prep
+    export ANSIBLE_LOG_PATH="/opt/iiab/iiab/iiab-install.log"
```
jvonau: if you use the full path the cd is not required

roles/openvpn/tasks/main.yml
```diff
@@ -82,19 +80,16 @@
...
-    # Obsolete & unused for ~2 years as of August 2018:
-    #- { src: 'iiab-vpn.j2', dest: '/usr/bin/iiab-vpn', mode: '0755' }
+    # 2019-10-09: Not recommended for over a year & no longer in use
+    # - { src: 'iiab-handle.j2', dest: '/usr/bin/iiab-handle', mode: '0755' }
+
+- name: Copy /opt/iiab/iiab/iiab-support to /usr/bin/iiab-support, in case git tree deleted e.g. on a smaller IIAB install
+  copy:
+    src: "{{ iiab_dir }}/iiab-support"
+    dest: /usr/bin/
+    mode: 0755
```
jvonau, line 94: if the git tree is deleted, then none of the ansible routines are available

holta: True.  If there are any especially important scripts that need to elevated from `/opt/iiab/iiab/scripts` or `/opt/iiab/iiab` to `/usr/bin` LMK and I will do that.

But we might possibly be ok as is?

iiab-support
holta: Good to know.  I've commented it out.

roles/openvpn/tasks/main.yml
jvonau: Think I see what you want with /usr/bin/iiab-support post install but the log directory must exist, a gratuitous "mkdir -p" in place of "cd" can't hurt.

iiab-support
jvonau: Here

roles/openvpn/tasks/main.yml
jvonau: I have to backtrack, the cd /opt/iiab/iiab is required to find $INVENTORY $PLAYBOOK

jvonau: I have to backtrack, the cd /opt/iiab/iiab is required to find $INVENTORY $PLAYBOOK

jvonau: I have to backtrack, the cd /opt/iiab/iiab is required to find $INVENTORY $PLAYBOOK

holta: Ok.  See https://github.com/holta/iiab/blob/support_yml/iiab-support#L59-L70

All code paths tested.

Should be ready to merge.

PR #2000, opened 2019-10-12
Enhance /usr/bin/iiab-diagnostics so (1) it can be run by non-root users (2) it prompts users to auto-publish to a pastebin (3) output much improved + more readable!

### LINKED ISSUE #1981: Make "iiab-diagnostics" work for all (w/o sudo) ...and auto-upload
What's the best way to simplify "sudo iiab-diagnostics" to "iiab-diagnostics" so newbies can type this in regardless whether they're logged in as root or not?

Are SUID/SGID bits a workable / reasonable approach &mdash; what recommended in general?

For now http://FAQ.IIAB.IO documents this as "sudo iiab-diagnostics" here:
http://wiki.laptop.org/go/IIAB/FAQ#What_are_the_best_places_for_community_support.3F

FYI here's the program:
https://github.com/iiab/iiab/blob/master/scripts/iiab-diagnostics

And its README / guide:
https://github.com/iiab/iiab/blob/master/scripts/iiab-diagnostics.README.md

holta:
Hopefully resolves #1981

Caveat...when `/usr/bin/iiab-diagnostics` is run by a non-root user, some diagnostics are omitted, as illustrated by these 3 glitches:

```
     netstat -natp  # Ports/Services in use
(Not all processes could be identified, non-owned process info
 will not be shown, you would have to be root to see it all.)

     iptables-save  # Firewall rules
Failed to list table names in /proc/net/ip_tables_names: Permission denied

     journalctl -u dnsmasq  # dnsmasq log
Hint: You are currently not seeing messages from other users and the system.
      Users in groups 'adm', 'systemd-journal' can see all messages.
      Pass -q to turn off this notice.
```


jvonau:
I don't understand why you don't want a symlink, git pull and /usr/bin/iiab-diagnostics gets auto updated without having to run ansible to put the file in /usr/bin/... Less code is better, if the user deletes /opt/iiab/iiab I suspect the script will stop at the first point where something referenced in /opt/iiab/iiab (git) is called. If you insist on this path create a role for iiab-diagnostics and move the code out of 0-init so you can use runrole to update the file in /usr/bin/

jvonau:
The one change below allows the pi user to run iiab-diagnostics without sudo
> diff --git a/scripts/iiab-diagnostics b/scripts/iiab-diagnostics
index 25d743d..04c1bec 100755
--- a/scripts/iiab-diagnostics
+++ b/scripts/iiab-diagnostics
@@ -20,9 +20,9 @@ if [ -z "$nickname" ]; then
 fi
   Build up a meaningful shared filename for DEV / IMPLEM / LEARNING team(s)
-outfile=/etc/iiab/diag/${IIAB_RELEASE}_${OS_VER}_${YMDT}_$nickname
+outfile=/tmp/iiab/diag/${IIAB_RELEASE}_${OS_VER}_${YMDT}_$nickname
  System "snapshots" (time-stamped output from this 'iiab-diagnostics' command) will be stored in this directory.  A bit like system logs, but only on request.
-mkdir -p /etc/iiab/diag
+mkdir -p /tmp/iiab/diag

The resulting file's output is http://paste.debian.net/1106248/. I see that netstat -natp and iptables-save are incomplete but prefixing the two command with sudo fixes that 

>@@ -161,8 +161,8 @@ cat_cmd 'ip addr' 'Network interfaces'
 cat_cmd 'ifconfig' 'Network interfaces (old view)'
 cat_cmd 'brctl show' 'Bridge for LAN side'
 cat_cmd 'netstat -rn' 'Routing table'
-cat_cmd 'netstat -natp' 'Ports/Services in use'
-cat_cmd 'iptables-save' 'Firewall rules'
+cat_cmd 'sudo netstat -natp' 'Ports/Services in use'
+cat_cmd 'sudo iptables-save' 'Firewall rules'

in http://paste.debian.net/1106252/

jvonau:
> -cat_file /usr/bin/iiab-gen-iptables
-cat_cmd 'netstat -natp' 'Ports/Services in use'
-cat_cmd 'iptables-save' 'Firewall rules'
+cat_cmd 'sudo netstat -natp' 'Ports/Services in use'
-echo -e "\n  5. Log Files: (last 100 lines of each)\n"
+echo -e "\n  5. Firewall Files\n"
+cat_file /usr/bin/iiab-gen-iptables
+cat_cmd 'sudo iptables-save' 'Firewall rules'
+echo -e "\n  6. Log Files: (last 100 lines of each)\n"

Moving a couple of things around yields better readability by grouping /etc/network/interfaces* output together, and firewall output together. http://paste.debian.net/1106258/

holta:
> The one change below allows the pi user to run iiab-diagnostics without sudo
...
> > -mkdir -p /etc/iiab/diag
> > +mkdir -p /tmp/iiab/diag

Avoiding `/tmp` is important, so diagnostics files are preserved across reboots (when working with untrained operators especially!)

> prefixing the two [actually 3] command with sudo fixes [incomplete results & associated warnings]

Let's give this a shot.  If the operator does not have sudo privs (or does not know their own password, as sometimes happens) this might run into trouble later, but let's see how it goes.

holta:
> Moving a couple of things around yields better readability by grouping /etc/network/interfaces* output together, and firewall output together.

Let's go with your suggestion above.  At the same time we cannot find a taxonomy/ordering of all 30+ items that will please everyone all the time, and let's make a point of _resisting_ changing the taxonomy/order regularly, every time the wind blows in a new direction.

> I don't understand why you don't want a symlink, git pull and /usr/bin/iiab-diagnostics gets auto updated without having to run ansible to put the file in /usr/bin/... Less code is better, if the user deletes /opt/iiab/iiab I suspect the script will stop at the first point where something referenced in /opt/iiab/iiab (git) is called.

Since 2017 there are more than a few Zero W implementers who've deleted their source trees, and we should not leave these "dieters" high and dry.  It's for this reason that I tested `/usr/iiab/diagnostics` which indeed works without `/opt/iiab/iiab`

> If you insist on this path create a role for iiab-diagnostics and move the code out of 0-init so you can use runrole to update the file in /usr/bin/

Humbly Disagree.  If we created a playbook for every 2 lines of code, we'd have 1000 playbooks (:

PR #1968, opened 2019-09-26
Install Kolibri w/ apt (.deb file) instead of pip (.pex file)

holta:
Per @jvonau recommendation.

PR #2001, opened 2019-10-14
(1) clarify Kolibri installation / migration requirements (2) eliminate manual chown step (per @benjaoming)

holta:
Building on PR #1978, Thanks @benjaoming for reviewing if poss!

Refs PRs: #1968 #1970 #1972 #1978

### DIFF
--- a/roles/kolibri/tasks/main.yml
+++ b/roles/kolibri/tasks/main.yml
@@ -25,23 +25,27 @@
     group: root
     mode: 0755
 
+# 2019-10-14: An alternative might be to put KOLIBRI_USER="kolibri" into
+# /etc/kolibri/conf.d/iiab.conf
 - name: Save kolibri_user ({{ kolibri_user }}) to /etc/kolibri/username
   copy:
-    content: "{{ kolibri_user }}"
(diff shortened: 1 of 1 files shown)


roles/kolibri/tasks/main.yml
```diff
@@ -96,6 +100,8 @@
   become_user: "{{ kolibri_user }}"
   when: kolibri_provision | bool
 
+# 2019-10-14: QUESTION FOR @benjaoming...is this stanza really necessary
+# (shouldn't migration & provisiondevice above be sufficient?)
```
benjaoming: It should not be, no. I also find this part very weird:

    group: "{{ apache_user }}"

The group is set to a user name?

Anyways, it doesn't seem like you are running Kolibri inside some different application worker with mod_wsgi or UWSGI, so I don't see why you'd want to create confusion about the permissions. They should work regardless of running `chmod` in the Ansible roles.

One more thing: When `kolibri-server` is installed, it should read the very same settings that you add in `/etc/kolibri/username`

holta: Thx for clarifying; I'll comment out the chown stanza!

benjaoming:
:+1:

PR #2004, opened 2019-10-15
calibre_via_debs: False [was True] + calibre_via_python: True [was False] in vars/debian-10.yml

holta:
So Debian 10.1 implementers have easy access to the latest Calibre (https://calibre-ebook.com/whats-new shows 4.1 for now) just like Ubuntu folk &mdash; instead of being locked into the distro's aging deb-based install of Calibre 3.48


holta:
Tested on top of a Debian 10.1 VM that was running IIAB 7.1/master...which now shows:

root@box:~# calibre --version
calibre (calibre 4.1)

Issue #1635, opened 2019-05-05
KA Lite language pack


tim-moody:
Why is registration and installation of en language pack no longer part of the kalite role?

tim-moody:
I someone tries to install kalite with runrole after an initial install they will not get installed.

holta:
> Why is registration and installation of en language pack no longer part of the kalite role?

As far as I know KA Lite registration & installation of its mandatory en language pack were never part of https://github.com/iiab/iiab/tree/master/roles/kalite

KA Lite's Ansible role/playbook (above) could certainly be improved as such.  Likewise @arky's [Transmission](https://github.com/iiab/iiab/tree/master/roles/transmission) (BitTorrent) role takes the next step of downloading KA Lite learning content.

Auto-provisioning of "ubiquitous" content for Kolibri is also requested by a growing many, that we should work with Learning Equality to address soon: https://github.com/iiab/iiab/issues/1545

holta:
On a related note, IIAB's 1-line installer (http://d.iiab.io) does basic provisioning of KA Lite as follows:

https://github.com/iiab/iiab-factory/blob/master/iiab#L22-L24
https://github.com/iiab/iiab-factory/blob/master/iiab#L215-L240

holta:
If this is a feature request to rework the KA Lite role, it should move to the IIAB 7.1 milestone, so that IIAB 7.0 can converge in coming weeks.

(Or if it's just a question, it should probably be closed now?)

jvonau:
#2739 #2928

PR #2014, opened 2019-10-20
KA Lite playbook cleaner / readability

holta:
Builds on: PR #2012

Related:
#1635 KA Lite language pack


holta:
Smoke-tested on Ubuntu 18.04 where KA Lite was not installed, using `./runrole kalite`

PR #2023, opened 2019-10-30
Gitea 1.9.4 -> 1.9.5

holta:
See https://github.com/go-gitea/gitea/releases/tag/v1.9.5

and https://github.com/go-gitea/gitea/releases


holta:
Smoke-tested on Ubuntu 18.04.3 -- http://box/gitea loads.

(The 1st account created after installation should automatically become an admin account per usual.)

PR #2025, opened 2019-10-31
Fix PR #2024: Install package python-virtualenv for KA Lite and Calibre-Web

holta:
See PR #2024 which is mostly about transitioning Ansible to Python 3.

Mandates kernel 4.19.79+ on Raspbian for now, to address #1993.

PR #2024, opened 2019-10-31
Python3-2 [Ansible]

jvonau:
creates python2 virtual environments using python3 tools removing the need to install python2 tools
builds on #2019

PR #2026, opened 2019-11-01
Only install python-virtualenv on older OS's (where does python3-virtualenv come from?)

holta:
Refines PR #2025 (fix for PR #2024) migrating Ansible to Python 3.

Background explanations under https://github.com/iiab/iiab/pull/2025#issuecomment-548669853


holta:
CLARIF: python3-virtualenv is now installed by https://github.com/iiab/iiab/blob/master/scripts/ansible#L79

PR #2027 seeks **Clarity of Purpose** around what these 7 python3-* packages are actually for &mdash; allowing for code maintainability and legility &mdash; avoiding long-term bitrot (and worse problems) carrying around packages year after year for no reason.

PR #2019, opened 2019-10-24
python3 [Ansible]

jvonau:
sample for #1810

PR #2027, opened 2019-11-01
Clarify of Purpose: what are these 7 python3-XYZ pkgs for in scripts/ansible ? (And install python-pip in 2-common for Admin Console?)

holta:
@jvonau thanks for your help here...plz give me a quick call so we cleanly document our intentions...so IIAB's code becomes readable and maintainable.

( Otherwise complex interdependencies spiral out of control, year after year :/ )

Building on PRs: #2019 #2024 #2025 #2026

### DIFF
--- a/roles/2-common/tasks/packages.yml
+++ b/roles/2-common/tasks/packages.yml
@@ -64,9 +64,11 @@
       - wget
     state: present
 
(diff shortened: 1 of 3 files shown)


holta:
Instead we went with PRs #2028 #2030

Issue #2013, opened 2019-10-20
Can python3-geojson be made to work, as required by OS's like Ubuntu 19.10+? [osm-vector-maps]


holta:
Thanks @georgejhunt for clarifying when you can!

Refs:
- #877 "Package up vector-based OSM maps"
- PR #2002 "httpd: php_version: 7.2 -> 7.3; lokole: bcrypt -> python3-bcrypt; osm-vector-maps: python-geojson -> python3-geojson for Ubuntu 19.10 [Node.js should install via curl soon after release on 2019-10-17; cmdsrv requires python-systemd -> python3-systemd ?]"
- PR #2011 "For now, keep python-geojson on OS's prior to Ubuntu 19.10 (can python3-geojson be made to work?)"

jvonau:
I don't think there is any python code that uses geojson directly and can be removed as pylint returns unused-import on /usr/bin/iiab-update-maps
> W:  4, 0: Unused FeatureCollection imported from geojson (unused-import)
W:  4, 0: Unused Point imported from geojson (unused-import)
W:  4, 0: Unused Polygon imported from geojson (unused-import)
W:  4, 0: Unused Feature imported from geojson (unused-import)
W:  5, 0: Unused import geojson (unused-import)

Commenting out those imports doesn't seem to affect running it but I don't have a data set to test with.

holta:
Recap: https://github.com/iiab/iiab/pull/2042#issuecomment-557616995 is another report installing an experimental version of IIAB (http://d.iiab.io/install-test.txt) onto Ubuntu 19.10, that ran into python-geojson package issues.

(Presumably we need to transition to python3-geojson instead?)

Possibly Related:
#2034 "Provide python2.7 virtual environment for iiab apps"

georgejhunt:
I agree with Jerry. I don't think python*-geojson needs to be installed. It
is used during map generation, but not deployment.

holta:
@georgejhunt is there any workaround possible here, that would allow Admin Console to install?

(When installing http://d.iiab.io onto Ubuntu Server 19.10, it fails as follows...)

> TASK [console : Install packages for console] **********************************
ok: [127.0.0.1] => (item=libapache2-mod-authnz-external)
ok: [127.0.0.1] => (item=pwauth)
failed: [127.0.0.1] (item=python-geojson) => {"ansible_loop_var": "item", "changed": false, "item": "python-geojson", "msg": "No package matching 'python-geojson' is available"}

[...]

Issue #1810, opened 2019-07-03
Python 2 -> Python 3 IIAB strategic planning to prepare for 2020+


holta:
https://pythonclock.org excerpts:

> Python 2.7 [will not be maintained past 2020](https://www.python.org/dev/peps/pep-0373/). Originally, there was no official date. Recently, that date has been updated to [January 1, 2020](https://github.com/python/devguide/pull/344).

> If the code you care about is still on Python 2, that's totally understandable. Most of PyPI's popular packages now [work on Python 2 and 3](https://python3wos.appspot.com/), and more are being added every day. Additionally, a number of critical Python projects have [pledged to stop supporting Python 2 soon](http://python3statement.org/). To ease the transition, the [official porting guide](https://docs.python.org/3/howto/pyporting.html) has advice for running Python 2 code in Python 3.

tim-moody:
biggest issue is switch from str and unicode types to bytes and str types.

traverseda:
Do we have a list of code that needs to be updated?

holta:
> Do we have a list of code that needs to be updated?

These are the 4+8=12 that matter, if you have time to assess any quick?

(1) Arising from `grep -rl '#!.*/python' /opt/iiab/iiab` we have:

- [roles/kiwix/templates/iiab-make-kiwix-lib.py](https://github.com/iiab/iiab/blob/master/roles/kiwix/templates/iiab-make-kiwix-lib.py)
- [roles/kiwix/templates/iiab-make-apache-config.py](https://github.com/iiab/iiab/blob/master/roles/kiwix/templates/iiab-make-apache-config.py)
- [roles/1-prep/templates/iiab_env.py.j2](https://github.com/iiab/iiab/tree/master/roles/1-prep/templates/iiab_env.py.j2)
- [roles/network/templates/captive-portal/captive-portal.py.j2](https://github.com/iiab/iiab/tree/master/roles/network/templates/captive-portal/captive-portal.py.j2)

(2) Arising from `grep -rl '#!.*/python' /opt/iiab/iiab-admin-console` we have:

- [roles/cmdsrv/files/iiab-cmdsrv](https://github.com/iiab/iiab-admin-console/blob/master/roles/cmdsrv/files/iiab-cmdsrv)
- [roles/cmdsrv/files/scripts/get_kiwix_lib](https://github.com/iiab/iiab-admin-console/blob/master/roles/cmdsrv/files/scripts/get_kiwix_lib)
[...]

PR #2034, opened 2019-11-10
provide python2.7 virtual environment for iiab apps

jvonau:
Keeps current python2.7 apps contained in virtualenv till porting can be completed, step to solve #2013 as the dependencies supplied by pip are in flux pending review and testing.

### DIFF
--- a/roles/1-prep/templates/iiab_env.py.j2
+++ b/roles/1-prep/templates/iiab_env.py.j2
@@ -1,4 +1,4 @@
-#!/usr/bin/python
+#!/opt/admin/venv/bin/python2.7
 # read iiab.env from python
 
 def get_iiab_env(name):
--- a/roles/2-common/tasks/pylib.yml
+++ b/roles/2-common/tasks/pylib.yml
@@ -8,3 +8,4 @@
   with_items:
     - { src: 'iiab_const.py.j2', dest: '{{ py3_dist_path }}/iiab/iiab_const.py' }
     - { src: 'iiab_lib.py', dest: '{{ py3_dist_path }}/iiab/iiab_lib.py' }
+    - { src: 'iiab_env.py.j2', dest: '{{ iiab_etc_path }}/iiab_env.py.j2' }
--- a/roles/2-common/templates/iiab_env.py.j2
+++ b/roles/2-common/templates/iiab_env.py.j2
@@ -1,4 +1,4 @@
-#!/opt/admin/venv/bin/python2.7
+#!/usr/bin/python3
 # read iiab.env from python
 
 def get_iiab_env(name):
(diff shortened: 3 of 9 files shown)


holta:
Ref: #1810

jvonau:
iiab-make-kiwix-lib.py iiab-update-maps ported to python3

holta:
@georgejhunt can you please review this now?

@jvonau @tim-moody what smoke/functional tests are most needed here?

georgejhunt:
Adam asked me to review this.
I don't understand what problem this is trying to solve.
If it is just geojson, I think this is overkill.
I think Tim has said that he has all his code running under py3

georgejhunt:
And as I read more, trying to educate myself about uWSGI and python3, it seems that many suggest that all significant programs should be housed in a venv. 
So if Tim's ok with it, since I don't know any different, I'm fine with #2034;
But don't we want to be shooting for py3 in the long term? Should we take the time now to get it running on py3?

georgejhunt:
I have succeeded in getting admin-console running on #2034 
Change uwsgi to:
[uwsgi]
	uid = www-data
	gid = www-data
	socket = /tmp/admin-console.sock
	chdir = /opt/admin/console
	wsgi-file = cmd-service.wsgi
	master = true
        plugins = python3 <=====
change uwsgi.unit to:
[Unit]
Description=uWSGI Service

[Service]
ExecStart=/usr/bin/uwsgi --ini /etc/uwsgi/apps-enabled/admin-console.ini <====
Restart=always
RestartSec=5
KillSignal=SIGQUIT
Type=notify
NotifyAccess=all

[Install]
WantedBy=multi-user.target

jvonau:
Wrong PR but I'll continue here to sum up the sequence of events, Tim asked for https://github.com/iiab/iiab/pull/2042/commits/0b0014df4074321eb549c0f9aeaecfe618e8bae7 but didn't provide the file in 263. Placing the file as need with the above edit results in a working admin-console backend on U-19.10 as 2042 does not address moving to python3 directly but is careful not to add python2 as a dependency.

georgejhunt:
Are you wanting to run admin-console under py2.7? If so why? Which iiab
apps do you think require py2.7?

jvonau:
> Are you wanting to run admin-console under py2.7? If so why? Which iiab apps do you think require py2.7?

Wrong PR, but when I started out there was no activity to port to python3 and Ubuntu-19.10 was failing due to python2 dependencies. Today it looks like the only thing left to port to python3 is captive-portal which I didn't address directly because the backend is Apache but your original PR never dealt with, but can provide a virtualenv to run python2 in for captive-portal to run in.

georgejhunt:
I've verified that captive portal does not error out under python3. But I
don't currently have the ability to test all the clients against py3, and
verify function.

jvonau:
replaced by 2048

PR #2043, opened 2019-11-19
Add py_lib [Python 3]

tim-moody:
creates python3 dist iiab dir
adds py_lib and const files to python3 dist
adds some variables to default


tim-moody:
passed smoke test, lib imports and a function works

holta:
@jvonau & @georgejhunt please review ASAP, on Tim's request.

jvonau:
re: iiab_etc_path might want to extend that to be used with iiab_local_vars_file, iiab_env_file and iiab_ini_file just below in default_vars

georgejhunt:
looks good to me

holta:
> passed smoke test, lib imports and a function works

What OS's was this tested on?

holta:
@tim-moody clarifies:
> Jerry I have rethought the migration to py3. If you have scripts converted as is I'm OK to use them until I finish the rewrite using libs. probably we will need both make_kiwix_lib.py (yours to be imported into other scripts) and make_kiwix_lib3.py (mine, which I would use in the wrapper script)

holta:
This is another PR helping to make the transition from Python 2.7 to 3: #1810

PR #2048, opened 2019-11-25
Temporary Fix [Python 3 for iiab-make-kiwix-lib, iiab-update-map etc]

jvonau:
### Fixes Bug
replaces #2034 
ports iiab-make-kiwix-lib.py iiab-update-map to python3
groups iiab-python files together.


holta:
Ref #1810

PR #2051, opened 2019-12-02
Use alternate hosts file /etc/hosts.dnsmasq [so dnsmasq DOESN'T send "127.0.0.1 box box.lan" from /etc/hosts to client machines]

jvonau:
### Fixes Bug
Mexico testing


holta:
FYI (explanation thanks to @jvonau) this IIAB bug has likely existed more than a year, since we incorporated dnsmasq in late 2018, but did not surface due to a lack of testing with Linux client machines.

PR #2058, opened 2019-12-10
Include a php clause in NGINX conf [for osm-vector-maps]

georgejhunt:
I cannot runrole on this role because of the hyphens in the role name.
But it has only line changes in the nginx config file. So I'm fairly sure it will smoke test.
The central-america map displays after rebooting the client, and clearing browser cache.
I'm willing to do more testing, in the next few days. Or you can just merge it, since the current master is broken, and this has a fairly high probability of fixing it.


tim-moody:
I copied it to a broken install and it works for me. agree to merge

holta:
This stands a good chance of fixing #2013 so I'm merging it in a moment.

PR #2061, opened 2019-12-11
[Remove apt package avahi-discover as it...] drags in desktop and python2

jvonau:
was part of #1937


holta:
What tests are needed prior to merging this, if any?

(e.g. any possible ramifications that we need to uncover, resulting from removing pkg avahi-discover ?)

holta:
@jvonau writes "nothing uses it, was part of the XO days"

Issue #1993, opened 2019-10-08
Kiwix install fails as part of install of current master [segmentation fault, due to Raspbian or Ubuntu 19.10 kernel on RPi]


tim-moody:
TASK [kiwix : Enable & Restart 'kiwix-serve' service] ****************************************************************************************************************************************
fatal: [127.0.0.1]: FAILED! => {"changed": false, "msg": "Unable to restart service kiwix-serve: Job for kiwix-serve.service failed because a fatal signal was delivered to the control process.\nSee \"systemctl status kiwix-serve.service\" and \"journalctl -xe\" for details.\n"}

tim-moody:
fails again on rerun of install

tim-moody:
but the services is actually running

tim-moody:
ran a third time and it succeeded

tim-moody:
```
-- A start job for unit kiwix-serve.service has begun execution.
--
-- The job identifier is 2362.
Oct 08 21:16:11 box.lan systemd[1]: kiwix-serve.service: Control process exited, code=killed, status=11/SEGV
-- Subject: Unit process exited
-- Defined-By: systemd
-- Support: https://www.debian.org/support
--
-- An ExecStart= process belonging to unit kiwix-serve.service has exited.
--
-- The process' exit code is 'killed' and its exit status is 11.
Oct 08 21:16:11 box.lan systemd[1]: kiwix-serve.service: Failed with result 'signal'.
-- Subject: Unit failed
-- Defined-By: systemd
-- Support: https://www.debian.org/support
--
-- The unit kiwix-serve.service has entered the 'failed' state with result 'signal'.
Oct 08 21:16:11 box.lan systemd[1]: Failed to start Provides the Kiwix ZIM File Server.
-- Subject: A start job for unit kiwix-serve.service has failed
-- Defined-By: systemd
-- Support: https://www.debian.org/support
--
-- A start job for unit kiwix-serve.service has finished with a failure.
--
-- The job identifier is 2362 and the job result is failed.
Oct 08 21:16:16 box.lan systemd[1]: kiwix-serve.service: Service RestartSec=5s expired, scheduling restart.
Oct 08 21:16:16 box.lan systemd[1]: kiwix-serve.service: Scheduled restart job, restart counter is at 1.
-- Subject: Automatic restarting of a unit has been scheduled
-- Defined-By: systemd
-- Support: https://www.debian.org/support
--
[...]

PR #2063, opened 2019-12-13
Does 'kiwix-serve --threads 8' help more users search Wikipedia simultaneously? ["capacity problems" in Mexico schools]

holta:
We still don't know the root cause of students being locked out of Wikipedia, as commonly happens (for about 2 min at time) in almost every Mexican school where [kiwix-serve 3.0.1-6](http://download.kiwix.org/release/kiwix-tools/) was deployed on RPi 3 last week.

But Emmanuel Engelhart suggests kiwix-serve '--threads 8' as a workaround (instead of the '--threads 4' default) for now.

Refs: #1993, kiwix/kiwix-tools#345

### DIFF
--- a/roles/kiwix/templates/kiwix-serve.service.j2
+++ b/roles/kiwix/templates/kiwix-serve.service.j2
@@ -4,7 +4,7 @@ After=syslog.target network.target local-fs.target
 
 [Service]
 Type=forking
-ExecStart={{ iiab_base }}/kiwix/bin/kiwix-serve --daemon --port {{ kiwix_port }} --nolibrarybutton --library {{ kiwix_library_xml }} --urlRootLocation={{ kiwix_url }}
+ExecStart={{ iiab_base }}/kiwix/bin/kiwix-serve --daemon --threads 8 --port {{ kiwix_port }} --nolibrarybutton --library {{ kiwix_library_xml }} --urlRootLocation={{ kiwix_url }}
 TimeoutStartSec=180
 Restart=on-abort
 RestartSec=5s


tim-moody:
I wrote a puppeteer script to test the kiwix load on an rpi 4 with 2 G RAM.
It issues a set of 10 predetermined search urls and 10 requests to random and does it 6 times for 60 search and 60 random requests, all issued asynchronously within a couple of seconds over a wired connection.

With kiwix proxied through nginx I start to get gateway timeouts about half way through the run. If I change the requests to go directly to port 3000 I get no 504 errors.

The 4 kiwix threads are pinned at nearly 100%, but tail off before all the results have been sent, even directly from port 3000. When proxied, it took about 40 seconds before any results appeared and the whole run took 2 minutes and 10 seconds. Each thread took 8 to 10% of memory, so I think not memory bound.

I tried 8 threads and they ran at about 50% cpu, so I think no benefit.

holta:
Is https://www.scalescale.com/tips/nginx/504-gateway-time-out-using-nginx/ at all relevant, as mentioned by @jvonau ?

> For **Nginx as Proxy** for Apache web server, this is what you have to try to fix the 504 Gateway Timeout error:
> 
> Add these variables to nginx.conf file:
>
>   proxy_connect_timeout       600;
  proxy_send_timeout          600;
  proxy_read_timeout          600;
  send_timeout                600;
>
> Then restart nginx...

holta:
@tim-moody & @jvonau did several tests over VPN (to RPi 3 10.8.0.94's Apache proxy, and then direct to its kiwix-serve's port 3000) that I hope they can explain here or on https://github.com/kiwix/kiwix-tools/issues/345 ?

Tests were done using @tim-moody's https://github.com/iiab/iiab-factory/blob/master/testing/kiwix-load-test.js

holta:
1) fyi `--verbose --threads 2` was removed from [/etc/systemd/system/kiwix-serve.service](https://github.com/iiab/iiab/blob/master/roles/kiwix/templates/kiwix-serve.service.j2) soon after the above-mentioned testing began.  In order to restore kiwix-serve's defaults (i.e. `--threads 4`) as was deployed in Oaxaca State, Mexico earlier this month.

   And then I ran `systemctl daemon-reload` followed by `systemctl restart kiwix-serve`

2) Later on @jvonau commented out the `Timeout 120` line from [/etc/apache2/sites-available/010-iiab.conf](https://github.com/iiab/iiab/blob/master/roles/httpd/templates/010-iiab.conf.j2#L41) raising Apache's timeout from 120 seconds (2min) to what is apparently their 5min default:

   ```
   # Timeout: The number of seconds before receives and sends time out.
   #
   #Timeout 120
   ```

   Not directly related, but fyi this file also contains timeout setting:

   ```
   # KeepAliveTimeout: Number of seconds to wait for the next request from the
   # same client on the same connection.
   #
   KeepAliveTimeout 1
   ```

jvonau:
The noted 502's observed while using load-kiwix.js with the lower timeout value went away when the increased timeout value was used.

holta:
So far the consensus is that `kiwix-serve --threads 8` (instead of the default of 4) would not help.

Should we close this ticket as a result?

holta:
See the many updates added to https://github.com/kiwix/kiwix-tools/issues/345#issuecomment-572499329 over the past 36 hours:

1) Emmanuel closed that issue, basically saying these issues that Matthieu Gautier uncovered are interesting but known to date back many years, i.e. generally too hard to fix for now.
2) As such IIAB needs to continue refining usability of Kiwix searches &mdash; when multiple/many students are trying to search Wikipedia at the same time &mdash; so that...

   **a) Proxying doesn't shut out users!**  (Regardless whether Apache/NGINX are involved or not.)

   b) Schools/teachers/students get clear instructions as to what's truly realistic and what's not, on...
      1) RPi Zero W
      2) RPi 3
      3) RPi 4

holta:
@jvonau writes:

```
grep Timeout /etc/apache2/apache2.conf

# Timeout: The number of seconds before receives and sends time out.
Timeout 300
# KeepAliveTimeout: Number of seconds to wait for the next request from the
KeepAliveTimeout 5
```

300 seconds seems reasonable for IIAB in general (until we hear otherwise!)

So I'll create a PR to comment out `Timeout 120` from [010-iiab.conf](https://github.com/iiab/iiab/blob/master/roles/httpd/templates/010-iiab.conf.j2#L41) explaining that we're deferring to this Apache default...

jvonau:
> Jerry, 1:20 PM Canoe 10.8.0.38 has the apache MX config on it? Tim to sum up your testing going to port 3000 doesn't generate timeout messages while accessing /kiwix does generate 504 errors when running with nginx? 
Canoe, 1:22 PM Do not use 10.8.0.38 -- the Mexico RPi 3 duplicating early December's failures is 10.8.0.94 Recap: .94 is the exact MX duplicate I created about a week ago -- feel free to test against it if nec.
1:25 PM I was hoping to see if Tim's nodejs testing script could be run against .94 and see what the logs catch.. 
Canoe, 1:26 PM Good idea.
Tim, 2:44 PM pushed to factory
Canoe, 2:47 PM "Testing: throws multiple requests at kiwix" https://github.com/iiab/iiab-factory/pull/87 Emmanuel may be stating the obvious... "Simplify the problem: try directly to deal with kiwix-serve" https://github.com/kiwix/kiwix-tools/issues/345#issuecomment-566717558
Tim, 2:51 PM why are my comments not there
Canoe, 2:52 PM Click within https://github.com/kiwix/kiwix-tools/issues/345#issuecomment-566704649
Jerry, 2:56 PM I don't think it's kiwix itself, direct to port 3000 as Tim reports has no errors... It's a worker timeout issue from the front-end with nginx or apache 
Tim, 3:32 PM the timeouts will cascade, though. If you make the kiwix proxy a longer timeout other proxied services, kalite for example, will start to timeout as the server gets busier and busier. kiwix can already pin the cpu at 100%any way to bind a wifi adapter to a browser session?I would be very interested in how many of the puppeteer requests can be serviced by an rpi 3 or 3+. The anecdote was that 'only a couple of kids' were on.
Jerry, 3:35 PM point the script at 10.8.0.94?
Tim, 3:46 PM the vpn is pretty slow, but I can try it.so .94 is my target, right?
Jerry, 3:47 PM yup /home is in Spanish... 
Tim, 3:48 PM running
Canoe, 3:48 PM Yes, 10.8.0.94 is the RPi 3 I cloned about a week ago from the early Dec work in Mexico.
Tim, 3:49 PM you need to watch it. I'm not set up to connect except to run the script
Jerry, 3:50 PM tailing the access.log
Tim, 3:50 PM first results back
Jerry 3:50 PM 502's
Tim, 3:50 PM 502 Proxy Error502 Proxy Errorrandom didn't work for some reason , but search did
Jerry, 3:51 PM saw a few 404's
Tim, 3:51 PM those are randommy random is against the med wiki, which is probably not installed
Canoe, 3:54 PM Definitely not installed. http://10.8.0.94/kiwix/wikipedia_es_all_maxi_2019-09/ is the one that matters in MX.
Tim, 3:54 PM the 60 searches were still valid and have now completedso about 5 minutestestUrls = [ '/kiwix/search?content=wikipedia_es_all_maxi_2019-09&pattern=mexico+', '/kiwix/search?content=wikipedia_es_all_maxi_2019-09&pattern=futbol+', '/kiwix/search?content=wikipedia_es_all_maxi_2019-09&pattern=amlo+', '/kiwix/search?content=wikipedia_es_all_maxi_2019-09&pattern=gobierno+', '/kiwix/search?content=wikipedia_es_all_maxi_2019-09&pattern=pais+', '/kiwix/search?content=wikipedia_es_all_maxi_2019-09&pattern=nuevo+', '/kiwix/search?content=wikipedia_es_all_maxi_2019-09&pattern=juego+', '/kiwix/search?content=wikipedia_es_all_maxi_2019-09&pattern=oaxaca+', '/kiwix/search?content=wikipedia_es_all_maxi_2019-09&pattern=gastronomia+', '/kiwix/search?content=wikipedia_es_all_maxi_2019-09&pattern=pueblo+' Canoe, 3:56 PM/etc/systemd/system/kiwix-serve.service contains '--verbose --threads 2' since 4 days ago.
Tim, 3:56 PM that would slow it down
Canoe, 3:56 PM Should I revert to remove that, as it was in MX?
Tim, 3:57 PM don't know if it matters. we saw 502s when it was busyI can run it against 3000 to see if they disappear
Jerry3:57 PM please..
Canoe, 3:58 PM Do I remove '--verbose --threads 2' now or later?
Tim, 3:59 PM verbose will definitely slow it down. sure. let's go to stock with 4 threads and no verbose
Canoe, 3:59 PM Ok, 1minTim, 4:00 PMverbose is pretty useless as is way too verbose
Jerry, 4:00 PM rerun the same test first with 4 treads before checking 3000
Tim, 4:00 PM OK
Canoe, 4:00 PM Done: reverted to stock MX / stock kiwix-serve as is deployed in Oaxaca State earlier this month.daemon-reload also done & kiwix-serve restarted.
Jerry, 4:01 PM 502's suggest a timeout was hit in apache
Tim, 4:01 PM ready to run
Canoe, 4:01 PM Go for it.
Tim, 4:01 PMstartedfixed randomdoes the rpi3 have fewer cores? the 4 threads are at 50% averagenow climbing
Jerry 4:03 PM 302's
Tim, 4:03 PM are a redirectgetting results502 Proxy Error
Jerry, 4:04 PM 502'slol..think I have a quick fix...
Tim, 4:05 PM the fix is the proxy timeout
Canoe, 4:05 PM Hypothetically this VPN slowness/noise might possibly be a decent stand-in for classroom WiFi slowness/noise...
Tim, 4:05 PM done
Jerry, 4:05 PM yup.. on both
Tim, 4:06 PM ready for 2nd run?how do I tell htop just to track kiwix
Jerry, 4:06 PM I can test an apache change or you want to try port 3000
Tim, 4:07 PM let's run 3000
Canoe, 4:07 PM Go for it
Jerry, 4:07 PM sure..
Tim, 4:07 PM startedfirst results
Jerry, 4:08 PM so it's the apache workers limiting response times?
Tim, 4:09 PM but not as the load declinesall 4 threads in the 90s
Jerry, 4:12 PM so it's the apache workers limiting response times? opps erased the old comment
Tim, 4:12 PM doneagain 5 minutes
Jerry, 4:15 PM can we try the /kiwix path again? I bumped an apache timeout value want to see if that helps
Tim, 4:16 PM startedassume you're looking at https://httpd.apache.org/docs/2.4/mod/mod_proxy.htmlhttps://httpd.apache.org/docs/2.4/mod/mod_proxy.html
Jerry, 4:19 PM yes, the timeout values that could be used..020-iiab.conf sets Timeout to 2min while the default is 5min... so I # that out..
Tim, 4:22 PM finished
Jerry, 4:22 PM no 502 seen...
Tim, 4:23 PM if you are modifying a larger scope than kiwix.conf there may be unintended consequences
Jerry, 4:25 PM no change was to 020-iiab.conf which was producing an override of the default Timeout value of 5mins... not saying 5mins is good but 2mins is not enough... given the 502 now went away..
Canoe, 4:26 PMroot@box:~# grep -in timeout /etc/apache2/sites-available/010-iiab.conf 33:# Timeout: The number of seconds before receives and sends time out. 35:#Timeout 120 51:# KeepAliveTimeout: Number of seconds to wait for the next request from the 54:KeepAliveTimeout 1
Tim, 4:26 PM well. it's a data point. academic after we move to nginxsome commenters think the problem is keepalive
Jerry, 4:27 PM you face the same problem there with the 504's
Tim, 4:27 PM and have to make a decision about where to handle it. and we don't have olpc legacy configs


This proxy timeout issue will need to be revisited under nginx as I think the proxy timeout is set to 60 seconds by default.

holta:
> This proxy timeout issue will need to be revisited under nginx as I think the proxy timeout is set to 60 seconds by default.

Indeed.

@jvonau wrote:
> http://nginx.org/en/docs/http/ngx_http_proxy_module.html#proxy_read_timeout

Where should we change "proxy_read_timeout 60s;" to a higher value?

Currently I don't see proxy_read_timeout explicitly set anywhere within IIAB or the OS's NGINX installation, so this should be increased (somewhere) for things like Kiwix especially...

@jvonau wrote:
> I'd suggest /etc/nginx/proxy_params

I'll put 300 seconds in the PR, unless there's a known danger there?

holta:
@jvonau offered several recommendations for raising NGINX timeout(s) that we should incorporate into IIAB soon, that I've outlined in the now-merged PR #2151.  Eventually the `/etc/apache2/sites-available/010-iiab.conf` file might go away if/as Apache becomes less important to IIAB in coming quarters?  But currently @jvonau's and my recommendations are summarized here: https://github.com/iiab/iiab/blob/master/roles/httpd/templates/010-iiab.conf.j2#L46-L56

holta:
Very unlikely to help on CPU's that only have 4 cores, as discussed on our IIAB call today.

holta:
Let's move the discussion to #2152 to see if that helps solve this/these quite serious ongoing Kiwix problem(s),  that basically every school seems to be experiencing!

#2152 'Raise NGINX proxy timeout(s) from 60 sec--similar to Apache--to reduce "capacity" failures during Kiwix search etc'

Issue #2064, opened 2019-12-15
Master is broken by incorrect /etc/hosts.dnsmasq


georgejhunt:
On clean install of lite via one liner using local-min.
A client accessing via http://box.lan gets an error when connected via wifi.
/etc/hosts:
127.0.0.1     box.lan localhost.localdomain localhost box box box.lan
::1      localhost ip6-localhost ip6-loopback
ff02::1     ip6-allnodes
ff02::2     ip6-allrouters

127.0.1.1   raspberrypi

A client is not directed to 172.18.96.1 as she should be
~                                                                               
~Now I see that hosts is deprecated.
But still I observe that "ping box.lan" does not resolve. Where is the new config that substitutes for hosts?

holta:
Likely due to PR #2051.

georgejhunt:
uncommenting L668 in /etc/dnsmasq.conf does not fix the problem for me.
I'm stuck.

holta:
> uncommenting L668 in /etc/dnsmasq.conf does not fix the problem for me.

What is L668 ?

In any case, I've asked @jvonau to look into this as soon as he can.

georgejhunt:
I think 2065 fixes it

PR #2065, opened 2019-12-15
hosts file wrong sequence

georgejhunt:
smoke tested.
Functions after clearing browser cache, and reboot.

### DIFF
--- a/roles/network/templates/network/hosts-dnsmasq.j2
+++ b/roles/network/templates/network/hosts-dnsmasq.j2
@@ -1,3 +1,3 @@
 # Supplied by IIAB sourced by /etc/dnsmasq.d/iiab.conf
-{{ iiab_hostname }}   {{ lan_ip }}
-box   {{ lan_ip }}
+{{ lan_ip }}  {{ iiab_hostname }}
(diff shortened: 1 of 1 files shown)


holta:
Shouldn't these 2 lines really be...on the same line?

Refs: PR #2051, #2064

georgejhunt:
If the host name is something other than box, I think we still want to box
to resolve

holta:
> If the host name is something other than box, I think we still want to box to resolve

Indeed.  But the way to implement that would (normally, to date) be as follows:

```
{{ lan_ip }}   box {{ iiab_hostname }}
```

@jvonau can you confirm?

jvonau:
I blew it, it's backwards. I had that corrected locally but forgot to push the commit up to the branch.

jvonau:
I have noticed that the url in the browser on my cell changes from box.lan/home to 172.18.96.1/home, can you check for this behavior in your testing. Seems like a redirect is occurring on the nginx side.

holta:
PR #2066 tested with @jvonau, and then merged (serving the same function).

Issue #2067, opened 2019-12-16
master broken: box.lan changes to 172.18.96.1 in the browser


jvonau:
### Expected behavior and actual behavior.
When I access box.lan on the my cell phone the url displayed in the browser changes to 172.18.96.1, should remain box.lan

### Steps to reproduce the problem.
install master [and add #2065 NO LONGER NEC, run ./iiab-network NO LONGER NEC], then access box.lan over wifi

### Specifications like the version of the IIAB, operating system version, or hardware details.
think it's nginx related, this line: rewrite ^/$ $scheme://$server_addr/home/; in /etc/nginx/server.conf is forcing the use of the ip address

holta:
@georgejhunt can you take a look at this?

jvonau:
/home should really be defined by the variable {{ iiab_home_url }}

holta:
@georgejhunt: @jvonau suggests you're probably the best person to solve this?

But if not please unassign yourself, Thanks!

georgejhunt:
Function > form for me. But it's true that I've done quite a bit of the
nginx migration.
It comes down to reading docs, and trial and error.
And since it is cosmetic, will be relatively low on my priority list.
You can leave it on my list, until someone appears who will make it a
higher priority.
Sometimes I'm bored, and looking for a small task that might give me a
sense of accomplishment.

holta:
Please review @jvonau's proposed solution at PR #2068.

Thanks all for bringing these many NGINX regressions under control.

We're not there yet (http://minutes.iiab.io) but hopefully St Nick will help ;)

PR #2066, opened 2019-12-16
Alternative /etc/hosts.dnsmasq fix for PR #2051

holta:
Concise alternative to PR #2065, to fix #2064, arising from PR #2051.

PR #2068, opened 2019-12-17
Home page [NGINX should preserve box.lan etc, instead of redirecting to 172.18.96.1]

jvonau:
### Fixes Bug
#2067 correct rewrite with ip addresses in browser's address bar. 
consolidate modules.conf usb-lib.conf into iiab.conf.j2,  restore iiab_home_url,

### DIFF
--- a/roles/nginx/tasks/main.yml
+++ b/roles/nginx/tasks/main.yml
@@ -6,7 +6,7 @@
     - { src: "server.conf",dest: "/etc/nginx/" }
     - { src: "nginx.conf",dest: "/etc/nginx/" }
     - { src: 'ports.conf' , dest: '/etc/{{ apache_service }}/' , mode: '0644' }
-    - { src: "iiab.conf",dest: "/etc/nginx/conf.d/" }
+    - { src: "iiab.conf.j2",dest: "/etc/nginx/conf.d/iiab.conf" }
   when: nginx_enabled
 
 - name: Insure that apache2 is not running -- we may need port swap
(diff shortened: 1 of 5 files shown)


georgejhunt:
Seems reasonable. There is no comment in the PR about whether it passes smoke test, and functions.

tim-moody:
I prefer
location = / {
        return 301 {{ iiab_home_url }};
    }

holta:
Prior to our 11AM EST call tmrw [ Thursday, http://minutes.iiab.io ] what next steps are needed here?

Tangentially related: PR #2066

holta:
> I prefer
> location = / {
> return 301 {{ iiab_home_url }};
> }

1) Today's call (http://minutes.iiab.io) consensus is that the above [redirect instead of rewrite] is a good idea (though possibly to merged after this PR #2068?)

2) @tim-moody suspects the rewrite might be a redirect behind the scenes anyway.

holta:
@jvonau tested this over WiFi [LAN] on Raspbian (does the right thing over WAN, and over tunnel/VPN [which is neither WAN not LAN]).

Aside: iptables does not affect tunnel/VPN.

holta:
@jvonau this looks like a possible glitch in PR #2068 here, as merged this morning... 161-u19-srv-BIG-1219 = http://10.8.0.42 redirects to http://10.8.0.42//home/

Likewise http://192.168.0.161 redirects to http://192.168.0.161//home/

Can you tell why?

[Posted to #2073]

PR #2070, opened 2019-12-18
Cap3.1 on NGINX and Python 3 [Captive Portal]

georgejhunt:
1. Smoke tested on Raspbian lite-min. 
2. Functions on android 5.7, Ios, mac, windows 10, fails on windows 7.
3. Not tested as an upgrade to previous captive-portal -- since the name of the role is changed, there may be detritus folders.


holta:
> 2. Functions on android 5.7, Ios, mac, windows 10, fails on windows 7.

I assume you mean Android 5 & 7?  Alongside iOS, macOS, Windows 10.

Windows 7 (legal or not!) is on almost every client PC in low-income schools like those in Chiapas, Mexico — so we need to face this too — where/when possible.

roles/captiveportal/templates/captiveportal.ini
```diff
@@ -0,0 +1,12 @@
+[uwsgi]
+   uid = www-data
+   gid = www-data
+   http-socket = :9090
```
jvonau: {{ apache_user }} {{ captiveportal_port }} and make this file a .j2

roles/captiveportal/tasks/main.yml
```diff
@@ -0,0 +1,110 @@
...
+  args:
+      creates: /etc/nginx/sites-available/capture.conf
+  
+- name: Enable nginx to service the sites in checkurls list
+  file:
+    src: /etc/nginx/sites-available/capture.conf
+    path: /etc/nginx/sites-enabled/capture.conf
+    state: link
+  when: captiveportal_enabled | bool
+
```
jvonau, line 77: Needs file absent when: not captiveportal_enabled for sites-enabled/capture.conf

roles/nginx/tasks/install.yml
```diff
@@ -25,3 +25,11 @@
     - { src: "roles/nginx/templates/server.conf", dest: "/etc/nginx/" }
     - { src: "roles/nginx/templates/nginx.conf", dest: "/etc/nginx/" }
     - { src: 'roles/nginx/templates/ports.conf', dest: '/etc/{{ apache_service }}/' , mode: '0644' }
+    - { src: 'roles/nginx/templates/uwsgi.service', dest: '/etc/systemd/system/' , mode: '0644' }
```
jvonau, line 28: This will override the default uwsgi.service which parses /etc/uwsgi/apps-enabled/ for any .ini files out of the box which is used by https://github.com/iiab/iiab-admin-console/blob/master/roles/console/tasks/main.yml without needing to supply this service file that would limit uwsgi to only run admin-console.ini

roles/nginx/tasks/install.yml
```diff
@@ -25,3 +25,11 @@
...
+
+- name: Let uwsgi running as {{ apache_user }} write log files
+  file: 
+      path: /var/log/uwsgi/app
+      state: directory
+      owner: "{{ apache_user }}"
+
```
jvonau, line 35: This should go into captiveportal/main.yml to ensure the log directory is created for captiveportal.log when runrole is used.

roles/captiveportal/templates/captiveportal.ini
georgejhunt: agreed

roles/captiveportal/tasks/main.yml
georgejhunt: done

roles/nginx/tasks/install.yml
georgejhunt: I did it the way you suggest first and could not get uwsgi to come up properly after reboot -- maybe because admin-console was using socket, and captive portal was using port. The docs suggest that the way to solve is with uwsgi emperor. I chose to kiss. 
I wanted to keep captive portal on a port for debugging simplicity -- see https://github.com/georgejhunt/iiab/blob/cap3.1/roles/captiveportal/README.md

roles/nginx/tasks/install.yml
georgejhunt: uwsgi is a dependency for both captive portal, and admin Console. I think it belongs where it is.

georgejhunt:
final testing of this PR requires box.lan to resolve properly, which is in another PR.
I should soft code the captive portal target URL, and test.

roles/nginx/tasks/install.yml
jvonau: With the recent commit the logging moved to apache2 for captive portal and Admin Console doesn't use a log directory.

roles/nginx/tasks/install.yml
jvonau: I'm missing something, admin-console.ini is installed to /etc/uwsgi/apps-enabled/ is supplied by iiab-admin-console and the stock uwsqi.service is enabled there. There is no uwsqi service enabled for nginx's role, it's been moved.

holta:
> final testing of this PR requires box.lan to resolve properly, which is in another PR.
> I should soft code the captive portal target URL, and test.

This was resolved by PR #2066 (i.e. 3 days ago) as mentioned during our http://minutes.iiab.io call earlier today.

Tangentially related: PR #2068

roles/2-common/tasks/packages.yml
```diff
@@ -33,7 +33,7 @@
     state: present
   when: is_debuntu | bool
 
-- name: "Install 23 common packages: acpid, bridge-utils, bzip2, curl, gawk, hostapd, htop, i2c-tools, logrotate, make, mlocate, netmask, net-tools, ntfs-3g, pandoc, pastebinit, rsync, sudo, tar, unzip, usbmount, usbutils, wget"
+- name: "Install 24 common packages: acpid, bridge-utils, bzip2, curl, gawk, hostapd, htop, i2c-tools, logrotate, make, mlocate, netmask, net-tools, ntfs-3g, pandoc, pastebinit, rsync, sqlite3,sudo, tar, unzip, usbmount, usbutils, wget"
```
holta, line 36:
```suggestion
- name: "Install 24 common packages: acpid, bridge-utils, bzip2, curl, gawk, hostapd, htop, i2c-tools, logrotate, make, mlocate, netmask, net-tools, ntfs-3g, pandoc, pastebinit, rsync, sqlite3, sudo, tar, unzip, usbmount, usbutils, wget"
```

holta:
A big thank you to @georgejhunt for beginning to publish a Captive Portal README:
https://github.com/georgejhunt/iiab/blob/cap3.1/roles/captiveportal/README.md

Which will presumably appear here soon:
https://github.com/iiab/iiab/tree/master/roles/captiveportal/README.md
...or possibly here:
https://github.com/iiab/iiab/tree/master/roles/captive-portal/README.md

(Or this could later move to one of our more prominent wikis &mdash; which is NOT to say IIAB can solve the world's Captive Portal technical quagmires ever-arising from each new OS/browser/update/etc &mdash; BUT just to invite streamlined "technical fieldback" from thoughtful implementers around the globe...)

holta:
README is now published:

https://github.com/iiab/iiab/blob/master/roles/captiveportal/README.md

(Please all who can test with diverse smartphones/tablets/laptops/browsers/OS's and [contact us](http://internet-in-a-box.org/pages/contributing.html) with any results!)

Issue #2005, opened 2019-10-15
Transition from Node.js 10 to 12 as it becomes LTS on 2019-10-21 [Node.js 10 loses full LTS status on 2020-03-31]


holta:
@m-anish @llaske @mitra42 do you forsee any problems upgrading Sugarizer, Node-RED and "Internet Archive Offline" etc from Node.js 10 to 12?

As Node.js 12 becomes LTS starting Monday in 6 days from now &mdash; please see:

- https://nodejs.org/en/about/releases/
- https://nodejs.org/en/download/releases/

Aside: the new Ubuntu 19.10 apt installs Node.js v10.15.2 and npm 5.8.0, both of which are a bit dated, but just happen to be the easiest way to test this OS for now &mdash; given that 19.10's official OS release is ~48h away (on Thur 2019-10-17) &mdash; whereupon Node.js's [official/recommended curl-based installation technique](https://github.com/nodesource/distributions#installation-instructions) should begin to work:

```
# Run as root

curl -sL https://deb.nodesource.com/setup_12.x | bash -
apt-get install -y nodejs
```

mikkokotila:
I think Node 10 is totally ok for now. Unless there is a specific reason to move to 12 already.

mitra42:
I have seen no problems with v12 all of our code is happy with v10 or v12 so I think you could upgrade when you get confirmation from the other packages. 

What I did in one of my previous PR's was change the `node` code in the `internetarchive` role to be happy with either 10.x or 12.x . 

I think (but can't be positive I'm correctly remembering) that I made that change because an installation broke. Which I believe was because the nodejs role gets the latest support release and so probably automatically gets 12.x now.
```
- name: FAIL (STOP INSTALLING) IF nodejs_version is not set to 10.x or 12.x
  fail:
    msg: "Internet Archive install cannot proceed, as it currently requires Node.js 10.x or 12.x, and your nodejs_version is set to {{ nodejs_version }}.  Please check the value of nodejs_version in /opt/iiab/iiab/vars/default_vars.yml and possibly also /etc/iiab/local_vars.yml"
  when: internetarchive_install and (nodejs_version != "10.x") and (nodejs_version != "12.x")
```

mitra42:
Looking at Sugarizer it doesnt appear to be installing node (so I hope it only installs after internetarchive or something else that does)

mitra42:
[...]

PR #2075, opened 2019-12-21
Node.js 10.x -> 12.x

holta:
We're almost halfway through the ~5 month transition from Node.js 10.x LTS to 12.x LTS, as per #2005, so let's now broaden testing to confirm we're ready for 2020-3-31 (when Node.js 10.x is no longer full LTS) with many more users.

https://nodejs.org/en/about/releases/


holta:
Testing is currently underway with a fresh install of BIG-sized http://download.iiab.io @ 10.8.0.46 = Desktop version of Ubuntu 18.04.3 LTS.

PR #2087, opened 2019-12-30
kalite - setup.yml needs kalite_install boolean

jvonau:
to suppress setup.yml when kalite_install is set to False

### Fixes Bug
chat message


tim-moody:
looks right

Issue #1562, opened 2019-03-12
Don't hard code captive_portal url


m-anish:
[This](https://github.com/iiab/iiab/blob/d2b7bb78a776bd21cad88a39eb0e2407f32f737d/roles/captive-portal/templates/capture-wsgi.py#L188) line seems to want to hard code the captive portal url on first glance. Perhaps we should inherit this from a variable. 

I think this might throw up a bug in case somebody chooses a hostname other than box?

holta:
@georgejhunt or @m-anish can you propose a PR?

Or ask @tim-moody & @jvonau if you're unsure how?

m-anish:
@holta this is potentially a bug rather than an enhancement :)

holta:
@georgejhunt can you please update this tkt with the latest?

And close it ideally?

holta:
@georgejhunt can you confirm this tkt is now closeable — as a result of just-merged PR #2070 ?

georgejhunt:
There is now a variable captiveportal_splash_page initialized to '/'. The IIAB already lets the user redefine whether the '/' goes to /home or to wordpress.
Initialized to some other value than '/', captive portal will go there.

holta:
> There is now a variable captiveportal_splash_page initialized to '/'. The IIAB already lets the user redefine whether the '/' goes to /home or to wordpress.
> Initialized to some other value than '/', captive portal will go there.

Thanks.  This needs to be in local_vars.yml and http://FAQ.IIAB.IO — let's not forget prior to IIAB 7.1 release, around 2020-1-30.

holta:
> > There is now a variable captiveportal_splash_page initialized to '/'. The IIAB already lets the user redefine whether the '/' goes to /home or to wordpress.
> > Initialized to some other value than '/', captive portal will go there.
> 
> Thanks. This needs to be in local_vars.yml and http://FAQ.IIAB.IO — let's not forget prior to IIAB 7.1 release, around 2020-1-30.

Followup is happening here:

PR #2094 "How does one use vars iiab_home_url & captiveportal_splash_page ?"

PR #2094, opened 2020-01-03
How does one use vars iiab_home_url & captiveportal_splash_page ?

holta:
@georgejhunt please suggest implementer examples so I can write this up in http://FAQ.IIAB.IO, Thanks!

Ref: #1562 "Don't hard code captive_portal url"


holta:
> @georgejhunt please suggest implementer examples so I can write this up in http://FAQ.IIAB.IO, Thanks!

_If this is easier to explain on a live call, let's do it that way, so these variables are truly understood by the broader implementer community-_

(Merging now so that broader January testing for IIAB 7.1 begins now!)

Issue #2062, opened 2019-12-11
Runrole regressions [overstrict checking of local_vars.yml & let's restore dashes in role names]


georgejhunt:
At a minimum, ./runrole should check vars/default_vars.yml, as well as /etc/iiab/local_vars.yml.

I cannot change the config of nginx, without running ./iiab-install --reinstall.

jvonau:
For local testing just add nginx_install: True and nginx_enabled: True to your local_vars.yml file to use runrole. 
You could also try running ./iiab-configure which would be the equivalent of using the admin-console's 'save configuration' & 'install configured options' in place of ./iiab-install --reinstall.

holta:
Whatever's decided, let's please document these practices within http://FAQ.IIAB.IO !

jvonau:
or add nginx_install: True and nginx_enabled: True to the 3 local_vars.yml files

holta:
Under discussion during today's http://minutes.iiab.io call...

@jvonau is trying to avoid the counter being reset when `./runrole` invokes one of the 9-10 aggregate roles: (stages)

https://github.com/iiab/iiab/wiki/IIAB-Contributors-Guide#ansible

@tim-moody suggests `./runrole` include a `--force` flag for people who really want to run one of those 9-10 aggregate roles / stages.

holta:
Call Summary: @jvonau is mulling over a redesign, so power users (advanced implementers etc) aren't impeded by overstrict checking.

holta:
Ref:

#1557 ./runrole options

holta:
ASIDE: "./runrole gitea" used to just work, e.g. to install a newer or older version etc!

Now one also needs to remove the line "gitea_installed: True" from /etc/iiab/iiab_state.yml

(What's the best way to bring back operator simplicity here?)

Ref: #2057

holta:
> ASIDE: "./runrole gitea" used to just work, e.g. to install a newer or older version etc!
> 
> Now one also needs to remove the line "gitea_installed: True" from /etc/iiab/iiab_state.yml

Instead of the above, one can apparently run:

```
cd /opt/iiab/iiab
./runrole --reinstall gitea
```

holta:
PR #2126 "runrole - remove installed lockout" resolves half of this (overstrict checking is now removed).

_...but does [./runrole](https://github.com/iiab/iiab/blob/master/runrole) now allow for dashes-in-role-names ?_

holta:
[...]

Issue #2078, opened 2019-12-21
NGINX somehow prevents rapid testing of IIAB apps/services [e.g. during IIAB installation]


holta:
With Apache it was an absolute lifeline that I could actively test all apps during IIAB installation, using URL's like...

http://10.8.0.46/gitea
http://10.8.0.46/lokole
http://10.8.0.46/nodered
http://10.8.0.46/mediawiki
(etc)

But now with NGINX I can no longer, until an hour later after reboot or such, which makes efficient testing (i.e. a rapid feedback loop) absolutely impossible :(

FYI the equivalent URL's with http://192.168.0.191/* also fail with NGINX 404's.

Any idea why this is happening?

Tangentially Related:

#2055 'AWStats URL's erroneously redirect to http://box.lan/cgi-bin/awstats.pl?config=schoolserver'
#2056 'http://box/wiki (DokuWiki) erroneously redirects to http://127.0.0.1:8090/wiki/'
#2057 'http://box/gitea displays 404 and broken images'
#2062 'Runrole regressions [overstrict checking of local_vars.yml & let's restore dashes in role names]'
#2077 'NGINX: http://box/mediawiki erroneously redirects to http://localhost/mediawiki/index.php/Main_Page'
#2079 'NGINX: http://box/maps "404 Not Found"'

tim-moody:
so is the problem that you get 404s or that a reboot (and a wait) are required for the links to work?
did you clear your cache?

tim-moody:
we need to document which services are nginx direct and which have shims

holta:
> so is the problem that you get 404s or that a reboot (and a wait) are required for the links to work?

The hourlong is the problem.  The above URL's stop showing 404s and being to work "an hour or more later" &mdash; but this destroys the rapid feedback look needed for efficient/iterative testing.

holta:
@jvonau is the solution something as simple as running `systemctl restart nginx` (or similar) near the end of each playbook?

jvonau:
> @jvonau is the solution something as simple as running `systemctl restart nginx` (or similar) near the end of each playbook?

No, not that simple, the above 4 urls have apache as the backend (think you could add /nextcloud /wordpress /elgg to the list) without native support so the nginx 'shims' would need to be installed in the role in addition to adding systemctl restart nginx.

> we need to document which services are nginx direct and which have shims

 It's been started see https://github.com/iiab/iiab/blob/master/roles/nginx/files/README.md

holta:
[...]

Issue #224, opened 2017-09-05
NGINX alongside Apache for better analytics (& Captive Portal too?)


holta:
Ref #210 @georgejhunt & @tim-moody are exploring.

If this works in the end, Apache might possibly be preserved alongside.

(Compare @JulianHarty's web-proxy-for-analytics approach?)

holta:
See @georgejhunt's ongoing work here for [IIAB 6.5](https://github.com/iiab/iiab/milestone/2) or [IIAB 6.6](https://github.com/iiab/iiab/milestone/3) &mdash; which MIGHT even permit Moodle/WordPress/Elgg to work(*) cleanly campuswide:

#285 Tinynginx
#286 AWStats nginx
#287 KA Lite nginx
#288 Kiwix nginx
#289 Munin nginx
#290 Nextcloud nginx
#291 OSM nginx
#292 WordPress nginx

(*) via WAN side of IIAB server, if NGINX is essentially translating headers of most all http traffic from http://localhost/x to http://campuswide-server-name/x etc).

holta:
FYI there have been many attempts here, since this ticket was opened 26 months ago.

Here are some of the latest:
- PR #2009 "Nginx shim for local_vars_big" (by @georgejhunt)
- PR #2017 (refines the above, by @jvonau)

holta:
Time to close this soon, if NGINX is converging properly for [IIAB 7.1](https://github.com/iiab/iiab/milestone/6) anticipated to be released 2020-01-30 ?

Note that Apache will supported alongside, for the indefinite future, or at least for a while to come~

Refs:
#1268 AWStats community patterns/analytics/statistics/metrics
#1762 IIAB Community Analytics framework, above & beyond AWStats (likely Matomo?)
#2052 NGINX (main PR)

holta:
TO DO: substantially clean up NGINX's README.md here...
https://github.com/iiab/iiab/tree/master/roles/nginx#transition-to-nginx

Starting with things like PR #2141's https://github.com/jvonau/iiab/blob/restarts/roles/nginx/README.md

holta:
@jvonau @georgejhunt @tim-moody 

Is it almost time to declare victory now, closing this ticket in favor of smaller NGINX-associated cleanup tickets going forward?

holta:
Not sure whether to declare victory.

So I'll move this to IIAB 7.2's milestone for now.

jvonau:
[...]

PR #2102, opened 2020-01-05
NGINX Part 2 [for ./runrole & for rapid testing of IIAB apps/services e.g. during IIAB installation]

jvonau:
### Fixes Bug
#2062 #2078

### DIFF
--- a/roles/0-init/tasks/main.yml
+++ b/roles/0-init/tasks/main.yml
@@ -39,6 +39,11 @@
     phplib_dir: "{{ ansible_local.local_facts.phplib_dir }}"
     iiab_stage: "{{ ansible_local.local_facts.stage }}"
 
+- name: Set 'installing' fact to True based on stage
+  set_fact:
+    installing: True
+  when: iiab_stage|int < 9
+
 - name: Discover if this is running Raspbian -- if so assume it is an RPi
   set_fact:
     rpi_model: "rpi"
--- a/roles/2-common/tasks/main.yml
+++ b/roles/2-common/tasks/main.yml
@@ -6,8 +6,6 @@
 - name: Create IIAB directory structure ("file layout")
   include_tasks: fl.yml
 
-- include_tasks: pylib.yml
-
 - include_tasks: centos.yml
   when: ansible_distribution == "CentOS"
 
--- a/roles/2-common/tasks/pylib.yml
+++ b/roles/2-common/tasks/pylib.yml
@@ -6,6 +6,6 @@
     group: root
     mode: 0644
   with_items:
-    - { src: 'iiab_const.py.j2', dest: '{{ py3_dist_path }}/iiab/iiab_const.py' }
-    - { src: 'iiab_lib.py', dest: '{{ py3_dist_path }}/iiab/iiab_lib.py' }
-    - { src: 'iiab_env.py.j2', dest: '{{ iiab_etc_path }}/iiab_env.py' }
+    - { src: 'roles/2-common/templates/iiab_const.py.j2', dest: '{{ py3_dist_path }}/iiab/iiab_const.py' }
+    - { src: 'roles/2-common/templates/iiab_lib.py', dest: '{{ py3_dist_path }}/iiab/iiab_lib.py' }
+    - { src: 'roles/2-common/templates/iiab_env.py.j2', dest: '{{ iiab_etc_path }}/iiab_env.py' }
--- a/roles/4-server-options/tasks/main.yml
+++ b/roles/4-server-options/tasks/main.yml
@@ -23,12 +23,6 @@
 - name: Install iiab python files
   include_tasks: roles/2-common/tasks/pylib.yml
 
-- name: USB-LIB
-  include_role:
-    name: usb-lib
-  when: usb_lib_install | bool
-  tags: usb-lib
-
 # MANDATORY SO PERHAPS THIS BELONGS IN 3-BASE-SERVER ?
 - name: HOMEPAGE
   include_role:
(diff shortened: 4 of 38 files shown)


holta:
A BIG-sized fresh install of this PR (#2102) is now in Stage 4 of installing on 161-ub1804-srv-jv-wp2-BIG-0104 = 10.8.0.66

Refs: #224 #2052

holta:
The `./iiab-configure` command runs Stage 0, followed by Stages 4-9 (e.g. if several IIAB Apps are enabled or disabled within [/etc/iiab/local_vars.yml](http://wiki.laptop.org/go/IIAB/FAQ#What_is_local_vars.yml_and_how_do_I_customize_it.3F)) but does Not do the NGINX/Apache restarts (by design, this is different from ./runrole and ./iiab-install).

Should this be documented within the command similar to https://github.com/iiab/iiab/blob/master/iiab-install#L55-L63 and https://github.com/iiab/iiab/blob/master/iiab-install#L133-L139 ?

georgejhunt:
Adam texted me, an asked for me to review this PR.

1. I have expressed my disapproval about massive PRs such as this, many times. I often feel ignored. I realize that Jerry has put a lot of work into this PR, and Adam has been willing to work on testing and fixing massive changes in the past.
2. I think  merging is a bad idea, and will delay the release.  
3. I am not willing to spend my time getting it all to work again after the merge. I'd rather spend my time improving the maps for 7.1. 
4. I am willing to fix captive portal, and improve the Docs for it. And maybe some nginx shims will need to be improved.

As far as I can tell, most of this PR restructures functionality that is already working correctly. So the up side potential benefit is low, and the down side risk is high. Adam has asserted that pretty code helps the world take IIAB more seriously.  Maybe he's right. I think the jury is still out on that. Will the changes in this PR be seen as more attractive to our  apparent audience -- the technical assistants to fledgling deployments? I doubt it.

tim-moody:
Impossible to review all this, but a couple of things:

- Why was bluetooth, cups, etc. moved to 5-xo-services?
- Turn on Apache for (dokuwiki elgg lokole moodle nodered nextcloud) does not belong in 0
- why add the complication of explicitly mentioning role/2-common in roles/2-common/templates/iiab_const.py.j2', dest: '{{ py3_dist_path }}/iiab/iiab_const.py' }
- what is when: installing? when are we not installing?
- what is the advantage of supporting multiple web servers?

holta:
`| bool` (casting to boolean) should be removed, except when there are bare vars i.e. `when: var | bool`

It's an *extremely* annoying situation introduced by Ansible 2.8.

But at least @jvonau and I are understanding it much better now, as we've written up here: https://github.com/iiab/iiab/issues/1632#issuecomment-570983047

jvonau:
> Impossible to review all this, but a couple of things:
> 
>     * Why was bluetooth, cups, etc. moved to 5-xo-services?

Just to give a clear look on what needs to be addressed in stage 4, there are duplicate/unused code that should be culled.
> 
>     * Turn on Apache for (dokuwiki elgg lokole moodle nodered nextcloud) does not belong in 0
> 

This was discussed in the chat meeting and that is what you suggested, don't enable apache if there are no apps that require apache.

>     * why add the complication of explicitly mentioning role/2-common in roles/2-common/templates/iiab_const.py.j2', dest: '{{ py3_dist_path }}/iiab/iiab_const.py' }
> 

Though it would be nice for ICO to update code in stage 4 instead of having to do a reinstall to get the latest code churn.

>     * what is when: installing? when are we not installing?
> 

When ICO or iiab-configure runs. Or don't you care about how iiab-from-console.yml runs? 

>     * what is the advantage of supporting multiple web servers?

Again in the chat meeting you suggested that apache is not going anywhere soon..

holta:
Thank you @tim-moody & @jvonau for moving these critical-but-difficult discussions forward.

The fundamental purpose of [IIAB 7.1](https://github.com/iiab/iiab/milestone/6) is to provide a clean and stable testing environment for our growing community of integrators and devs (and those they represent around the world) to build on &mdash; even if some of them consider NGINX a distraction from their educational goals &mdash; but still we have to march forward now figuring this out, putting it into a stable and understandable form.

Even if this PR #2102 isn't final obviously, *Rome Was Not Built In A Day.*

e.g. we could possibly break this PR (#2102) apart into separate PR's to refine it more understandably, should that be necessary in the end.

tim-moody:
> * Turn on Apache for (dokuwiki elgg lokole moodle nodered nextcloud) does not belong in 0

> This was discussed in the chat meeting and that is what you suggested, don't enable apache if there are no apps that require apache.

I was thinking it didn't need to be done in 0, but I guess runrole needs it there.

> Again in the chat meeting you suggested that apache is not going anywhere soon..

I thought we are supporting apache as used by nginx only, not without nginx

> When ICO or iiab-configure runs

We are not installing?

> * why add the complication of explicitly mentioning role/2-common in roles/2-common/templates/iiab_const.py.j2', dest: '{{ py3_dist_path }}/iiab/iiab_const.py' }

> Though it would be nice for ICO to update code in stage 4 instead of having to do a reinstall to get the latest code churn.

I don't get that. code in 2 has full reference so that it can be run in 4?

jvonau:
was tested as is, but will break down into smaller bits for further digestion.

tim-moody:
I think I don't understand the difference between iiab-install and iiab-configure. Is the latter meant only to be run after the former? Is it meant only to change enablement?

jvonau:
> I thought we are supporting apache as used by nginx only, not without nginx
> 

The original PR was a POC that needed further polish, as such my thought was to be optional and could be reverted. Now that nginx is now enforced it makes sense to move the templates to the roles and add the needed nginx restarts.

> > When ICO or iiab-configure runs
> 
> We are not installing?
> 

Well yes and no, if there are no changes to *_install in local_vars after install one would be just adjusting services by virtue of 'installed' being present and running just enabled.yml in any roles' main.yml. Should someone adjust *install to True in local_vars.yml then that role would install while the others would just run enabled.yml.

> > * why add the complication of explicitly mentioning role/2-common in roles/2-common/templates/iiab_const.py.j2', dest: '{{ py3_dist_path }}/iiab/iiab_const.py' }
> 
> > Though it would be nice for ICO to update code in stage 4 instead of having to do a reinstall to get the latest code churn.
> 
> I don't get that. code in 2 has full reference so that it can be run in 4?

Could of moved the templates but yes, by using an absolute path you gain portability to any other role.

jvonau:
> I think I don't understand the difference between iiab-install and iiab-configure. Is the latter meant only to be run after the former? Is it meant only to change enablement?

Correct on the first part. iiab-configure and iiab-from-console.yml share the same code stage 4+  but without network added at the end.

holta:
@jvonau explained extensive progress here as to how he will move forward.

Apologies my note-taking did not keep up during our weekly IIAB call that just wrapped up here: http://minutes.iiab.io

jvonau:
For the most part implemented via the sub PR's

PR #2121, opened 2020-01-09
logrotate daily disables captive portal after first day

georgejhunt:
### Fixes Bug

Logrotate for apache2 installed by awstats makes ownership root www-data with permissions 640 That means that captiveportal cannot write a log after the first day, and errors out after first day.

Perhaps this should wait until all of nginx logs get moved out of /var/log/apache2.

### DIFF
--- a/roles/awstats/templates/logrotate.d.apache2
+++ b/roles/awstats/templates/logrotate.d.apache2
@@ -5,7 +5,7 @@
 	compress
 	delaycompress
 	notifempty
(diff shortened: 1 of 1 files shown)


holta:
In general, do we want to continue parametrizing www-data as {{ www-data }} ?

(Most every playbook has done that so far.)

holta:
@georgejhunt will replace this after #2124

PR #2125, opened 2020-01-10
Softcode - subset of #2102

jvonau:
subset of #2102 -  cleanup, softcode and bools

PR #2127, opened 2020-01-10
Refine @jvonau's softcode - subset of #2102 (WIP)

holta:
Building on PR #2125 with standardized bare vars cast to boolean ("when: var | bool") per Ansible recommendation off of #1632.


holta:
We should shoot for 1 or 2 standardized Ansible playbooks that everyone can use as models to create IIAB Apps/Services.

At the moment it's a bit chaotic with some playbooks installing Apache's .conf files within install.yml, and others within enable.yml

Also we should determine if this common code snippet below is optional/unnecessary or not &mdash; some playbooks include it when installing Apache/NGINX **templates** &mdash; and many others do not:
```
owner: root
group: root
mode: '0644'
```

holta:
Smoke-tested & quickly functionality-tested (across the dozen+ relevant IIAB apps/services) but needs a followup PR to move kiwix_enable.yml to enable.yml

OS: Ubuntu 19.10

PR #2134, opened 2020-01-12
BUGFIX - [also tests if Node.js version too old?] Avoid dups - speed

jvonau:
subset of #2102, NEW - enforces nodesource update on older installs when reinstalling.


roles/nodejs/tasks/main.yml
```diff
@@ -41,6 +41,12 @@
 #      /etc/iiab/local_vars.yml -- then re-run this IIAB installer.
 #  when: nodejs_version_installed is defined and nodejs_version_installed.stdout != nodejs_version and nodejs_version_installed.stderr == ""
 
+- name: Remove /etc/apt/sources.list.d/nodesource.list if nodejs_version_installed.stdout is less than {{ nodejs_version }}
+  file:
+    state: absent
+    path: /etc/apt/sources.list.d/nodesource.list
+  when: nodejs_version_installed is defined and nodejs_version_installed.stdout != nodejs_version and nodejs_version_installed.stdout != ""
```
holta, line 49: Per the description 4 lines above, let's actually test to see if the installed Node.js version is **less than** desired (i.e. too old rather than too new) just like we use Ansible to determine if "PHP version is too old" for some versions of Nextcloud here: (https://github.com/iiab/iiab/blob/master/roles/nextcloud/tasks/install.yml#L34)

```
- name: Set var 'php_new' indicating if installed version of PHP ({{ php_version_installed.stdout }}) >= 7.1, as required by Nextcloud 16
  set_fact:
    php_new: "{{ php_version_installed.stdout is version('7.1', '>=') }}"
    # Ansible's Version Comparison routine:
    # https://docs.ansible.com/ansible/latest/user_guide/playbooks_tests.html#version-comparison
  when: php_version_installed.stdout != ""    # i.e. IF ABOVE 'php -v' WORKED
```

jvonau: I just reused the exact same logic that the stanza below uses to uninstall nodejs when the version is not correct. If you wish to use an integer comparison please alter the nodejs_version_installed stanza to suit your taste, change this stanza and the one below please.

jvonau:
BUGFIX - stop sugarizer service before stopping the database backend.
Also allows the PG-db backend to be disabled if the service is disabled to help free up memory if needed.

holta:
Now smoke-testing on 192-u1910-desk-BIG-0123-PR2134

roles/4-server-options/tasks/main.yml
```diff
@@ -32,28 +32,12 @@
...
-# NETWORK moved to the very end, after Stage 9 (9-LOCAL-ADDONS)
-# It can also be run manually using: cd /opt/iiab/iiab; ./iiab-network
-#
-#- name: NETWORK
-#  include_role:
-#    name: network
-#  # has no "when: XXXXX_install" flag
-#  tags: base, network
-
-# MANDATORY SO PERHAPS THIS BELONGS IN 3-BASE-SERVER ?
```
holta, line 44: I'll add a comment explaining this is in Stage 4-SERVER-OPTIONS because variable iiab_home_url change, and so this may need to be re-run in the field.

PR #2138, opened 2020-01-12
25 of 26 install.yml's + 10 others tasks/*.yml now write to /etc/iiab/iiab_state.yml

holta:
Summary:

1. 4 install.yml files (httpd/tasks/install.yml, monit/tasks/install.yml, nginx/tasks/install.yml, teamviewer/tasks/install.yml) are hereby added to the list of files that write their respective `XYZ_installed: True` to /etc/iiab/iiab_state.yml
 
   So of all 26 files named `*/tasks/install.yml`, [roles/kalite/tasks/install.yml](https://github.com/iiab/iiab/blob/master/roles/kalite/tasks/install.yml) remains the 1-and-only exception that doesn't yet write its own `XYZ_installed: True` to /etc/iiab/iiab_state.yml

   ([roles/kalite/tasks/setup.yml](https://github.com/iiab/iiab/blob/master/roles/kalite/tasks/setup.yml) does that job, for now anyway...possibly this should change in future, such that `install.yml` always lives up to its name...or at least tried...and upon completion writes `kalite_installed: True` to /etc/iiab/iiab_state.yml like all the others?)

2. The 10 _other_ `*/tasks/*.yml` files (those not named `install.yml`, that also write their respective `XYZ_installed: True` to /etc/iiab/iiab_state.yml) are currently:

   1. captiveportal/tasks/main.yml
   2. cups/tasks/main.yml
   3. kalite/tasks/setup.yml (double-counting kalite, as it's also in the 25+1 install.yml's above!)
   4. minetest/tasks/provision.yml
   5. network/tasks/dansguardian.yml
   6. network/tasks/dhcpd.yml
   7. network/tasks/named.yml
   8. network/tasks/squid.yml
   9. network/tasks/wondershaper.yml
   10. samba/tasks/main.yml

   These 10 above should probably be renamed-or-worked-into `XYZ/tasks/install.yml` or if that's not possible `XYZ/tasks/XYZ-install.yml` e.g. if this is not viable for a playbook like network/tasks/install.yml

   _In any case, there are now 25 + 10 = 35 total (5 of which explicitly contained within network/tasks) playbooks-or-sub-playbooks writing the /etc/iiab/iiab_state.yml = {{ iiab_state_file }}.

3. The `state: present` lines were gratuitous, so I removed these unnecessary from all these 35 `lineinfile` stanzas.

### DIFF
--- a/roles/awstats/tasks/install.yml
+++ b/roles/awstats/tasks/install.yml
@@ -76,8 +76,8 @@
   shell: /usr/bin/perl /usr/lib/cgi-bin/awstats.pl -config=schoolserver -update
   # when: awstats_enabled | bool
 
-- name: "Add 'awstats_installed: True' to {{ iiab_state_file }}"    # /etc/iiab/iiab_state.yml
+- name: "Add 'awstats_installed: True' to {{ iiab_state_file }}"
   lineinfile:
-    dest: "{{ iiab_state_file }}"
+    dest: "{{ iiab_state_file }}"    # /etc/iiab/iiab_state.yml
     regexp: '^awstats_installed'
     line: 'awstats_installed: True'
--- a/roles/awstats/tasks/install.yml
+++ b/roles/awstats/tasks/install.yml
@@ -81,4 +81,3 @@
     dest: "{{ iiab_state_file }}"
     regexp: '^awstats_installed'
     line: 'awstats_installed: True'
-    state: present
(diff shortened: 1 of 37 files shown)


holta:
Smoke-tested on Ubuntu 18.04

holta:
Smoke tests successful &mdash; both with and without deleting `/etc/iiab/iiab_state.yml` (to force more complete reinstalls of IIAB apps/services).

PR #2140, opened 2020-01-13
moodle/tasks/enable.yml for NGINX & WordPress tweaks (WIP)

holta:
Towards restoring Moodle on RPi, and in general?

ASIDE...Moodle 3.8.1 should be released within about 24h:
https://docs.moodle.org/dev/Moodle_3.8.1_release_notes


holta:
Tested: the standard enable.yml template works to enable http://box.lan/moodle via NGINX on RPi 4 test machine 10.8.0.58

PR #2142, opened 2020-01-13
(1) Quick clean high-level roles: NGINX, Apache, network, OpenVPN, numbered roles (stages) & defaults_vars.yml (2) Add vars/ubuntu-20.yml for Ubuntu 20.04 pre-releases (3) Fix http://box/mediawiki

holta:
Also experimentally comments out tags within these high-level roles.

WIP.

PR #2149, opened 2020-01-14
Fix for Admin Console login popup loop: install NGINX after Apache & MySQL

holta:
Hopefully @jvonau & @tim-moody (or @georgejhunt ?) can explain why NGINX cannot currently be installed prior to Apache &/or MySQL!

CONTEXT: installing NGINX prior to Apache & MySQL (during an IIAB install, in [3-base-server](https://github.com/iiab/iiab/blob/master/roles/3-base-server/tasks/main.yml)) currently causes Admin Console logins to ignore the username/password you type into the browser popup (typically iiab-admin/g0adm1n) and then another login popup keeps appearing every time (about 1-2 seconds after each username/password attempt).

FYI this failure occurred regardless of OS {Ubuntu 18.04, 19.10 and Debian 10.2}.

Ref: #2142

### DIFF
--- a/roles/3-base-server/tasks/main.yml
+++ b/roles/3-base-server/tasks/main.yml
@@ -3,10 +3,6 @@
 - name: ...IS BEGINNING =====================================
   command: echo
 
-- name: Install NGINX (configured LATER, in Stage 9-LOCAL-ADDONS)
-  include_tasks: roles/nginx/tasks/install.yml
-  when: nginx_install | bool
-
 - name: HTTPD (APACHE)
   include_role:
     name: httpd
@@ -18,6 +14,10 @@
     name: mysql
   when: mysql_install | bool
   #tags: base, mysql
+
+- name: Install NGINX (configured LATER, in Stage 9-LOCAL-ADDONS)
+  include_tasks: roles/nginx/tasks/install.yml
+  when: nginx_install | bool
 
 - name: Install dnsmasq
   include_tasks: roles/network/tasks/dnsmasq.yml


tim-moody:
hard for me to see what apache has to do with this as it was formerly possible to run adm cons with apache stopped.

holta:
Once we understand this Apache-MySQL-NGINX install order problem better, we can/should definitely fine tune.

(It's not like Admin Console or anybody really cares about alphabetical order in 3-base-server !)

holta:
@tim-moody suggests "as long as apache is needed it should be installed before nginx as our nginx proxies are dependent on it"

That's confusing to me, as my observation (when installing NGINX prior to Apache & MySQL on Ubuntu 18.04, 19.10 and Debian 10.2) is that this did not adversely affect [IIAB's other 20+ core http://box/ABC apps/services](http://wiki.laptop.org/go/IIAB/FAQ#What_services_.28IIAB_apps.29_are_suggested_during_installation.3F) during quick testing of each.

_(It's almost as if authentication [the browser's username/password popup, that Admin Console depends on] was the only thing that was broken possibly...which only became visible when trying to login to http://box/admin ?)_

jvonau:
How was this [change](c396f7bb75975b492b106f8259340fb5a3da3e61) tested before merging?  On a fresh install having nginx install before apache should of stopped the install dead when the nginx role installs the ports.conf into apache's configuration directory which should not exist if apache is installed later.

holta:
> On a fresh install having nginx install before apache should of stopped the install dead when the nginx role installs the ports.conf into apache's configuration directory which should not exist if apache is installed later.

That's what I thought, but that's not how it works:

`/etc/apache2/ports.conf` is installed by [nginx/tasks/main.yml#21](https://github.com/iiab/iiab/blob/master/roles/nginx/tasks/main.yml#L21) regardless of whether Apache was already installed or not, as tested on 3 different OS's {Ubuntu 18.04, Ubuntu 19.10, Debian 10.2}.

Does this mean directory `/etc/apache2` already exists for some mysterious reason?

(Or does it mean that Ansible is somehow forcing creation of non-existent parent directories when nec?)

jvonau:
From past experience with file and errors that popped when the directory didn't exist I didn't think the use of 'file' would auto create the directory but perhaps 'file' is smarter now. Look at the time stamps of the files in /etc/apache2?

holta:
> From past experience with file and errors that popped when the directory didn't exist I didn't think the use of 'file' would auto create the directory but perhaps file is smarter now. Look at the time stamps of the files in /etc/apache2?

I don't have the answer as [Ansible's template module](https://docs.ansible.com/ansible/latest/modules/template_module.html) [*] has not created parent directories in the past.

[*] as used here by [nginx/tasks/main.yml#L21](https://github.com/iiab/iiab/blob/master/roles/nginx/tasks/main.yml#L21)

But FYI/FWIW on 10.8.0.70 = 162-deb102-srv-BIG-0114 (a Debian 10.2 VM where NGINX was installed first) ...this auto-creation of parent directories would (appear!) to be happening:

> root@box:/etc# ls -ld apache2 apache2/ports.conf
drwxr-xr-x 8 root root 4096 Jan 14 13:44 apache2
-rw-r--r-- 1 root root  338 Jan 14 13:44 apache2/ports.conf

holta:
Maybe this is an undocumented behavior of https://docs.ansible.com/ansible/latest/modules/template_module.html

Certainly https://docs.ansible.com/ansible/latest/modules/file_module.html says:

> If [ `state: directory` ] , all intermediate subdirectories will be created if they do not exist

Conversely https://docs.ansible.com/ansible/latest/modules/copy_module.html says:

> If `src` and `dest` are files, the parent directory of `dest` is not created and the task fails if it does not already exist.

PR #2158, opened 2020-01-16
Cleanup of NGINX/Apache log file PRs #2155, #2156 & begin streamlining Munin playbook (WIP)

holta:
Includes & builds on @georgejhunt's PRs #2155, #2156 and my PR #2147.

PR #2159, opened 2020-01-17
Mediawiki 1.34.0 strictly disallows password 'changeme' so for now let's use 'changeme2020'

holta:
Until we come up with something better!

Builds on PR #1973 from 2019-09-30, and PR #2158 from 2020-01-16.

### DIFF
--- a/roles/mediawiki/defaults/main.yml
+++ b/roles/mediawiki/defaults/main.yml
@@ -13,11 +13,12 @@ mediawiki_src: "mediawiki-{{ mediawiki_version }}.tar.gz"
 
 mediawiki_db_name: iiab_mediawiki
 mediawiki_db_user: iiab_mediawiki_user
-mediawiki_db_user_password: changeme
+mediawiki_db_user_password: changeme2020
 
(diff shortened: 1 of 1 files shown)


holta:
Important: needs to be documented in http://FAQ.IIAB.IO -> "13 What are the default passwords?" after this PR merges!

tim-moody:
I have not seen this in my testing. What evidence is there?

holta:
> I have not seen this in my testing. What evidence is there?

'changeme' failed on 3 OS's Ubuntu 18.04, 19.10, Raspbian Lite on RPi 4.

I changed the password to 'changeme2020' per this PR, and the problem went away.

holta:
We can change the password to something else later, if there's something better than `changeme2020`

Or perhaps implement an override within MediaWiki that allows weak passwords, if you can find such a setting @tim-moody, and we're confident that will remain supported in future versions of MediaWiki ?

holta:
> Important: needs to be documented in http://FAQ.IIAB.IO -> "13 What are the default passwords?" after this PR merges!

Done.

Issue #2139, opened 2020-01-13
KA Lite fails to install on RPi 4: piwheels.org's "setuptools requires Python '>=3.5' but the running Python is 2.7.16"


holta:
Failure (pasted in below) is reminiscent of https://github.com/iiab/iiab/issues/392 in late 2017:

Background: a BIG-sized install of IIAB (http://download.iiab.io) is being attempted on a 1GB RAM Raspberry Pi 4 and the OS is Raspbian Lite 2019-09-26 with all apt updates applied, and then rpi-update firmware/kernel (4.19.93) to work around #1993.

> TASK [kalite : Install KA Lite static to /usr/local/kalite/venv, using pip (debuntu)] ***
[...]

PR #2160, opened 2020-01-17
KA Lite temp patch so IIAB once again installs on RPi: mv /etc/pip.conf /etc/pip.conf.see-iiab-issue-2139

holta:
See #2139 & https://github.com/piwheels/packages/issues/74


holta:
@jvonau @benjaoming

IIAB doesn't need to patch https://raw.githubusercontent.com/learningequality/ka-lite/master/requirements.txt immediately do we, as part of this current emergency patch?

(Per https://github.com/learningequality/ka-lite/pull/5626/files will the above `requirements.txt` likely be patched within KA Lite's master branch in the coming month or so?)

PR #2166, opened 2020-01-18
Wordpress & Munin scaffolding (WIP)

holta:
Preliminary scaffolding building on PR #2158, exploring options for tighter code maintainability ~ code readability ~ code safety.

Focusing on Modularity, and especially targetting shorter/safer & far more understandable [when:](https://docs.ansible.com/ansible/latest/user_guide/playbooks_conditionals.html#the-when-statement) clauses.

In preparation for things like @jvonau's PRs #2134 #2141 #2143 #2144.

PR #2167, opened 2020-01-18
Wordpress/Munin Scaffolding: tiny enhancements

holta:
Tiny enhancements:

1) Output during WordPress installation is now substantially cleaner, building on PR #2166.

2) For now these 3 below are saved into `/etc/iiab/iiab_state.yml` (upon installation of each, [so now a total of 36 instead of 35 markers](https://github.com/iiab/iiab/pull/2138#issue-361880590)) even if the option to (or need to) reinstall - enable - disable these 3 is still quite unusual today:

   - `apache_installed: True`
   - `nginx_installed: True`
   - `osm_vector_maps_installed: True`

   Still: 'iiab_state.yml' as an authoritative registry (showing whether these and other major apps/services have been installed of not) appears to be incredibly useful going forward...

### DIFF
--- a/roles/osm-vector-maps/tasks/main.yml
+++ b/roles/osm-vector-maps/tasks/main.yml
@@ -98,3 +98,9 @@
   systemd:
     name: nginx
     state: reloaded
+
+- name: "Add 'osm_vector_maps_installed: True' to {{ iiab_state_file }}"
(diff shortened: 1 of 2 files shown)


holta:
Smoke-tested on Ubuntu 19.10

PR #2183, opened 2020-01-22
Validate [DISALLOW "XYZ_install: False" WHEN "XYZ_installed is defined" IN /etc/iiab/iiab_state.yml]

jvonau:
install False installed True comparison. Works as I expect.


jvonau:
should address https://github.com/iiab/iiab/pull/2142#issuecomment-576399949

Issue #2164, opened 2020-01-17
NGINX: MediaWiki has broken & quirky links


holta:
1) After you enter Admin/changeme2020 to try to login to http://box/wiki it erroneously redirects to broken link http://localhost/wiki/Main_Page
   - If you then manually force your browser to http://box/wiki/Main_Page (for any valid value of 'box') most MediaWiki links then work, however...
2) The 'Random page' link in the left margin erroneously redirects from http://box/wiki/Special:Random to broken link http://localhost/wiki/Main_Page as in (1) above.
3) The following links appear quirky...containing "/mwlink/" instead of "/wiki/" in the middle of the URL... _is this really what we want?_
   - Top...
      - Edit http://box/mwlink/index.php?title=Main_Page&action=edit
      - View history http://box/mwlink/index.php?title=Main_Page&action=history
   - Lower Left...
      - Printable version http://box/mwlink/index.php?title=Main_Page&printable=yes
      - Permanent link http://box/mwlink/index.php?title=Main_Page&oldid=1
      - Page information http://box/mwlink/index.php?title=Main_Page&action=info
   - Lower Right...
      - Mark this page as patrolled http://box/mwlink/index.php?title=Main_Page&action=markpatrolled&rcid=1

See:
https://github.com/iiab/iiab/blob/master/roles/mediawiki/templates/mediawiki-nginx.conf.j2

Refs:
<strike>PR #2080 "change services to nginx and simplify php"</strike>
PR #2086 "Native nginx services"
PR #2142 "Fix http://box/mediawiki"
PR #2159 "MediaWiki [...] disallows password 'changeme' so for now let's use 'changeme2020'"

holta:
Discussed on our weekly call (http://minutes.iiab.io)

@tim-moody will look into this...

tim-moody:
probably the issue is in /library/mediawiki-1.34.0/LocalSettings.php:

## The protocol and server name to use in fully-qualified URLs
$wgServer = "http://localhost";

tim-moody:
https://www.mediawiki.org/wiki/Manual:$wgServer 

> Please make sure that $wgServer is not set to http://localhost or //localhost! If it is, a number of things will break, e.g. links inside your wiki will not work, stylesheets will not display properly and many more. Set it to the actual base URL instead!

tim-moody:
$wgServer = "//" . $_SERVER["HTTP_HOST"];

holta:
[...]

PR #2188, opened 2020-01-23
make mw use current host instead of localhost [MediaWiki]

tim-moody:
#2164


tim-moody:
tested on ub 19.10 vm

PR #2193, opened 2020-01-24
Move 19 roles into roles/0-DEPRECATED-ROLES (etc)

holta:
List of 19 unmaintained/deprecated roles remains at:


holta:
Smoke-tested on Ubuntu 19.10

Issue #2186, opened 2020-01-23
osm-vector-maps install breaks when no internet available and already installed.


jvonau:
### Expected behavior and actual behavior.
not to contact internet when already installed and operating offline.
### Steps to reproduce the problem.
disconnect from internet and run iiab-configure or ICO post install.

TASK [OSM-VECTOR-MAPS] ********************************************************************************

TASK [osm-vector-maps : Ensure directory /library/www/osm-vector-maps/maplist/assets exists] **********
ok: [127.0.0.1]

TASK [osm-vector-maps : Download map catalog http://download.iiab.io/content/OSM/vector-tiles/maplist/hidden/assets/regions.json to /library/www/osm-vector-maps/maplist/assets/] ***
fatal: [127.0.0.1]: FAILED! => {"changed": false, "dest": "/library/www/osm-vector-maps/maplist/assets/", "elapsed": 0, "gid": 33, "group": "www-data", "mode": "0755", "msg": "Request failed: <urlopen error [Errno 101] Network is unreachable>", "owner": "www-data", "size": 4096, "state": "directory", "uid": 33, "url": "http://download.iiab.io/content/OSM/vector-tiles/maplist/hidden/assets/regions.json"}

PLAY RECAP ********************************************************************************************
127.0.0.1                  : ok=182  changed=59   unreachable=0    failed=1    skipped=109  rescued=0    ignored=1   

Needs the install/enabled-or-disabled split applied to the role to ensure off-line operation.

tim-moody:
There are a number services that try to refresh external data during configuration. I suppose should be conditional on {{ internet_available }}, but may have unintended consequences.

holta:
@avni & I might re-test this tmrw...and implement @jvonau's recommended fix by splitting main.yml into {install.yml online, enable-or-disable.yml offline}

Issue #2197, opened 2020-01-24
OSM Install Fails and gives the following Errors: "refusing to convert from file to symlink for /library/www/html/common/assets/regions.json" & "/usr/bin/iiab-update-map: not found"


holta:
@georgejhunt OSM install routines have proven fragile below.  Let's discuss with @tim-moody how to make this more resilient?

Thanks to @jvonau who figured out (below) what seems to be happening when './iiab-install --reinstall' (and './runrole osm-vector-maps') both fail every time on BIG-sized master when layered on top of 182-rpi4-2g-MIN-0122-PR2171 = 10.8.0.66 that had no osm-vector-maps installed originally:

> TASK [osm-vector-maps : Symlink catalog /library/www/html/common/assets/regions.json -> /library/www/osm-vector-maps/maplist/assets/regions.json] ***
fatal: [127.0.0.1]: FAILED! => {"changed": false, "gid": 0, "group": "root", "mode": "0644", "msg": "refusing to convert from file to symlink for /library/www/html/common/assets/regions.json", "owner": "root", "path": "/library/www/html/common/assets/regions.json", "size": 21, "state": "file", "uid": 0}

Chat Log:

AH: OSM Error Context: I re-cloned from master and put a brand new BIG-sized /etc/iiab/local_vars.yml in place, before I ran "./iiab-install --reinstall"

JV: did you install a real map prior to --reisntall ?

AH: No.  (I did not install any maps or content.)

JV: https://github.com/iiab/iiab-admin-console/blob/master/roles/js-menu/tasks/main.yml#L132  starts some osm routines about regions.json

replacing the symlink is noted.

AH: Should I re-run './iiab-install' or './runrole osm-vector-maps' ?

JV: figure out which playbook is running the show first..
iiab or iiab-admin-console

AH: Not sure what that means in terms of next steps...

JV: you need to do the install/enabled split first
keep the symlink creation in install.yml as it can changed later as you see now..

AH: You mean https://github.com/iiab/iiab/blob/master/roles/osm-vector-maps/tasks/main.yml ?

JV: yup.. just another service

AH: Why didn't this happen before, do you know?

JV: and should be treated as such..
I can guess at incomplete testing, like offline behavior I found. [#2186 "osm-vector-maps breaks when no internet available and already installed"]

doing a admin-console install over prior install "Clone js-menu-files" doesn't run but "Otherwise sync local menu defs" does it's sync thing... taking a while.

[...]

PR #2199, opened 2020-01-26
Osm split

jvonau:
### Fixes Bug
#2186

### DIFF
--- a/roles/osm-vector-maps/tasks/enable-or-disable.yml
+++ b/roles/osm-vector-maps/tasks/enable-or-disable.yml
@@ -2,19 +2,6 @@
 #- name: Run the script that does osm-vector-maps housekeeping
 #  shell: /usr/bin/iiab-update-map
 
-#- name: Does the {{ vector_map_path }}/index.html redirect already exist?
-#  stat:
-#    path: "{{ vector_map_path }}/index.html"
-#  register: osm_redirect
-
-#Copy the redirect to the test page -- delete this later if more than one map
-- name: Install {{ vector_map_path }}/index.html redirect for http://box/maps -> http://box/osm-vector$
-  copy:
-    force: no
-    src: test-index.redirect
-    dest: "{{ vector_map_path }}/index.html"
-  #when: not osm_redirect.stat.exists
-
 - name: Copy countries.json (geojson) to {{ vector_map_path }}/maplist/assets
   copy:
     src: countries.json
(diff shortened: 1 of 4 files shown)


holta:
What tests are most appropriate to test this?

tim-moody:
why the departure from the enable.yml naming convention?

jvonau:
apply the PR
`sudo ip r del default ` kills internet access but not local WAN/LAN
ping somewhere on the internet to confirm access is not possible
`sudo ./runrole osm-vector-maps`
note that operation completed without internet access
reboot to restore default gateway that was removed for testing

holta:
> why the departure from the enable.yml naming convention?

"Do as you say, say as you do"

(enable-or-disable.yml is far more accurate description of what is really happening)

jvonau:
> why the departure from the enable.yml naming convention?

wonderchild wants it like that...

tim-moody:
Since we have a flag that is _enable(d) T/F that drives the process I think most people get that this yml deals with both states.
I am 100% opposed to the application churn required to retrofit this new name to all roles.

holta:
> Since we have a flag that is _enable(d) T/F that drives the process I think most people get that this yml deals with both states.

Disagree: install.yml only installs and cannot uninstall.  Which is fine.  But we should "say as we do, do as we say" wherever humanly possible if we want understanding to grow towards onboarding of new IIAB people demanding tangible clarity in our actions, which is imperative for our survival.

holta:
The goal for IIAB 7.1 is not to entrench the past, but to set up more cleanly understandable playbooks for the future.  I will do the work to make sure this happens.

holta:
Also note that both proposed enable.yml and enable-or-disable.yml are disappearing entirely from many playbooks, see for example:
- https://github.com/iiab/iiab/blob/master/roles/wordpress/tasks/main.yml
- https://github.com/iiab/iiab/blob/master/roles/munin/tasks/main.yml

jvonau:
> Also note that both proposed enable.yml and enable-or-disable.yml are disappearing entirely from many playbooks, see for example:
> 
>     * https://github.com/iiab/iiab/blob/master/roles/wordpress/tasks/main.yml
> 
>     * https://github.com/iiab/iiab/blob/master/roles/munin/tasks/main.yml

Any role that has a service behind it that would need to be turned off|on does require something more than the wordpress press example. If you are looking to move the restarting of the service back into the main.yml as munin is setup, go for it. You are on your own to change all the existing playbooks in short order for the proposed release time frame.

holta:
Background: My presumption is that [IIAB will not support uninstalls](https://github.com/iiab/iiab/blob/master/vars/default_vars.yml#L7-L8) of its [~30 apps/services](http://wiki.laptop.org/go/IIAB/FAQ#What_services_.28IIAB_apps.29_are_suggested_during_installation.3F) for the foreseeable future.  But that could of course change possibly.

holta:
I presume this PR does not address OSM install issue #2197 ...is that correct?

jvonau:
> I presume this PR does not address OSM install issue #2197 ...is that correct?

Not directly but the index.files are contained in enable-or-disable.yml for a clearer look at them.

holta:
> apply the PR
> `sudo ip r del default ` kills internet access but not local WAN/LAN
> ping somewhere on the internet to confirm access is not possible
> `sudo ./runrole osm-vector-maps`
> note that operation completed without internet access
> reboot to restore default gateway that was removed for testing

Above test worked (i.e. running `./runrole osm-vector-maps` before and after disabling Internet) after this PR was applied to 181-rpi4-1g-lite-BIG-0120-PR2171 = 10.8.0.74

Issue #2234, opened 2020-02-04
yarn update failures due to IIAB / internetarchive on Ubuntu 18.04, 16.04, etc


holta:
@mitra42 these errors have been ongoing for a day or two now...I was hoping they would just go away, pending yarn maintainers cleaning up their keys/signatures...but so far no.

Are you aware and/or can you recommend any corrective action?

Pastes from 2 different OS's below, in case that helps...

```
root@box:~# cat /etc/issue.net
Ubuntu 18.04.4 LTS

root@box:~# cat /etc/apt/sources.list.d/yarn.list
deb http://dl.yarnpkg.com/debian/ stable main

root@box:~# apt update
Get:1 http://dl.yarnpkg.com/debian stable InRelease [17.1 kB]
Hit:2 http://us.archive.ubuntu.com/ubuntu bionic InRelease
Get:3 http://linux.teamviewer.com/deb stable InRelease [9,388 B]
Hit:4 http://us.archive.ubuntu.com/ubuntu bionic-updates InRelease
Hit:5 http://us.archive.ubuntu.com/ubuntu bionic-backports InRelease
Hit:6 https://deb.nodesource.com/node_10.x bionic InRelease
Hit:7 http://ppa.launchpad.net/ansible/ansible/ubuntu bionic InRelease
Hit:8 http://security.ubuntu.com/ubuntu bionic-security InRelease
Err:1 http://dl.yarnpkg.com/debian stable InRelease
  The following signatures were invalid: EXPKEYSIG 23E7166788B63E1E Yarn Packaging <yarn@dan.cx>
Fetched 26.5 kB in 1s (20.3 kB/s)
Reading package lists... Done
Building dependency tree
Reading state information... Done
All packages are up to date.
W: An error occurred during the signature verification. The repository is not updated and the previous index files will be used. GPG error: http://dl.yarnpkg.com/debian stable InRelease: The following signatures were invalid: EXPKEYSIG 23E7166788B63E1E Yarn Packaging <yarn@dan.cx>
W: Failed to fetch http://dl.yarnpkg.com/debian/dists/stable/InRelease  The following signatures were invalid: EXPKEYSIG 23E7166788B63E1E Yarn Packaging <yarn@dan.cx>
W: Some index files failed to download. They have been ignored, or old ones used instead.
```

```
root@box:~# cat /etc/issue.net
Ubuntu 16.04.6 LTS

root@box:~# cat /etc/apt/sources.list.d/yarn.list
[...]

Issue #2210, opened 2020-01-28
'./runrole pbx' fails due to predefined var 'pbx_installed' in pbx/defaults/main.yml -- illustrates weakness of var validation when invoked from ./iiab-install (instead of from ./runrole)


holta:
`./runrole pbx` fails validation in [0-init/tasks/main.yml](https://github.com/iiab/iiab/blob/master/roles/0-init/tasks/main.yml#L36-L37) for several quite subtle reasons...

Thanks @jvonau for uncovering this complex error, exacerbated by [./iiab-install](https://github.com/iiab/iiab/blob/master/iiab-install)'s Ansible syntax which fails to instantiate/get variable values within every roles/*/defaults/main.yml (whereas [./runrole](https://github.com/iiab/iiab/blob/master/runrole) DOES instantiate/get these variable values for the single role it's trying to run, and hence validates vars correctly, as defined across the many files where they might originate!)

It'll take some time/thought/engineering to resolve this multi-layered Ansible problem.  Thanks @jvonau working on that in coming days...

@m-anish: it may be the case (likely?) that we'll have to rename your variables from `pbx_installed` and `freepbx_installed` to something else?  Given that all such `<ROLE NAME>_installed` variable names are now reserved for /etc/iiab/iiab_state.yml (sorry if that's what happens in the end!?)

Anyway at least we now know the root cause, which boils down to [Line 17 of pbx/defaults/main.yml](https://github.com/iiab/iiab/blob/master/roles/pbx/defaults/main.yml#L17) here:

> pbx_installed: False

Which thankfully uncovered several much deeper Ansible issues...making clear that our new 0-init [variable validation](https://github.com/iiab/iiab/blob/master/roles/0-init/tasks/validate_vars.yml) is not quite complete/sufficient when invoked from [./iiab-install](https://github.com/iiab/iiab/blob/master/iiab-stages.yml) (but works wonderfully and comprehensively when invoked from [./runrole](https://github.com/iiab/iiab/blob/master/run-one-role.yml) !)

holta:
Related:
PR #2211 "Var Validation Explained: Error msgs that tell you what's happening & what to do"

holta:
Some of this Error Messaging should be further improved beyond PR #2211 to mention:

```
cd /opt/iiab/iiab
./runrole --reinstall <ROLE NAME>
```

(And where to get more help about the `./runrole` command, as it and `./iiab-install` are both evolving, with this #2210 etc.)

[...]

Issue #2204, opened 2020-01-27
Internet Archive http://box:4244 isn't working on fresh install of IIAB on Ubuntu 18.04


holta:
@mitra42 would you possibly know the reason?  And/or what I should investigate to find the cause?

Internet Archive's port (4244) does not appear in `netstat -ltup | grep 4244` whereas most other [IIAB ports](https://github.com/iiab/iiab/wiki/IIAB-Networking#list-of-ports--services) do appear there.

FYI this machine is running the latest IIAB 7.1 pre-release (installed using http://download.iiab.io) on Ubuntu Desktop 18.04 in preparation for Afghan, Mideast & African refugees.

Aside: machine nickname is 191-u1804-desk-BIG+calib-0127 = 10.8.0.14

jvonau:
> Jan 27 11:04:28 box.lan internetarchive[27069]: /opt/iiab/internetarchive/node_modules/@internetarchive/dweb-transports/Errors.js:38
Jan 27 11:04:28 box.lan internetarchive[27069]: }
Jan 27 11:04:28 box.lan internetarchive[27069]: ^
Jan 27 11:04:28 box.lan internetarchive[27069]: SyntaxError: Unexpected token '}'
Jan 27 11:04:28 box.lan internetarchive[27069]:     at Module._compile (internal/modules/cjs/loader.js:891:18)
Jan 27 11:04:28 box.lan internetarchive[27069]:     at Object.Module._extensions..js (internal/modules/cjs/loader.js:991:10)
Jan 27 11:04:28 box.lan internetarchive[27069]:     at Module.load (internal/modules/cjs/loader.js:811:32)
Jan 27 11:04:28 box.lan internetarchive[27069]:     at Function.Module._load (internal/modules/cjs/loader.js:723:14)
Jan 27 11:04:28 box.lan internetarchive[27069]:     at Module.require (internal/modules/cjs/loader.js:848:19)
Jan 27 11:04:28 box.lan internetarchive[27069]:     at require (internal/modules/cjs/helpers.js:74:18)
Jan 27 11:04:28 box.lan internetarchive[27069]:     at Object.<anonymous> (/opt/iiab/internetarchive/node_modules/@internetarchive/dweb-transports/Transports.js:3:28)
Jan 27 11:04:28 box.lan internetarchive[27069]:     at Module._compile (internal/modules/cjs/loader.js:955:30)
[...]

PR #2238, opened 2020-02-05
Cleaner Dependencies for Sugarizer, Moodle, Node-RED, PBX, Internet Archive, Elgg

holta:
Working to make dependencies more reliable, more understandable and where possible faster.  Not the final word but way better and far less chaotic than where we were a month ago.  Thanks to some thoughtful structure @jvonau & others helped lay out in recent months.

BIG-sized IIAB install smoke-tested on:
- Ubuntu 19.10 (fresh install i.e. http://download.iiab.io)
- Debian 10.2 (./iiab-install --reinstall)

_@mitra42 see you when we meet in person in about 24h &mdash; I can work with you to further revise any of this [roles/internetarchive](https://github.com/holta/iiab/tree/cleaner-deps/roles/internetarchive) code herein as nec!  (While changes go way beyond the #2234 yarn patch from http to https &mdash; this PR is intending to make some of the larger IIAB roles a bit more modern+maintainable &mdash; few if any functional changes are included.)_


roles/internetarchive/tasks/main.yml
```diff
@@ -1,32 +1,93 @@
...
+      state: stopped
+    #when: internetarchive_enabled and internetarchive_upgrade
+
+  - name: Update pre-existing install (yarn upgrade) if internetarchive_upgrade
+    shell: yarn config set child-concurrency 1 && yarn install && yarn upgrade
+    args:
+      chdir: "{{ internetarchive_dir }}"
+    #when: internetarchive_enabled and internetarchive_upgrade
+
+  when: internetarchive_installed is defined and internet_available   # END BLOCK
```
jvonau, line 42: Painful always running upgrade when ansible runs, and risky as unplanned changes could be introduced like what occurred in #2204.

roles/internetarchive/tasks/main.yml
```diff
@@ -1,32 +1,93 @@
...
+
+- block:    # UPDATE IF... internetarchive_installed is defined and internet_available
+
+  - name: Stop 'internetarchive' systemd service, if internetarchive_upgrade
+    systemd:
+      name: internetarchive
+      daemon_reload: yes
+      state: stopped
+    #when: internetarchive_enabled and internetarchive_upgrade
+
```
jvonau, line 35: This change should cause the install to fail as the service file is not installed yet, that is done later in install.yml but ansible might silently discard the `Failed to stop foo.service: Unit foo.service not loaded` error emitted by systemd, I think that would be a bug in ansbile.

roles/internetarchive/tasks/main.yml
holta: Excellent point.  @mitra42 and I should discuss Community QA strategy when we meet in person here in Arizona in a few hours.

"Customers" who aren't paying should definitely help with QA (paying it forward) but at same time we/they/all need to bracket & mitigate the worst risks in creative ways.

So this is a really great discussion point towards tuning up near-fieldback not just distant-fieldback...

roles/internetarchive/tasks/main.yml
holta: Install doesn't fail fyi, as confirmed in my testing a fresh install.

See the enclosing `block:` (indented part) which prevents those 2 stanzas from running if internetarchive_installed is not yet defined.

jvonau: I see, something new one big when clause in place of using two whens, missed that.

roles/pbx/tasks/install.yml
```diff
@@ -0,0 +1,72 @@
...
+
+- name: FAIL (STOP THE INSTALL) IF 'nodejs_installed is undefined'
+  fail:
+    msg: "PBX install cannot proceed, as Node.js is not installed."
+  when: nodejs_installed is undefined
+
+- name: FAIL (STOP THE INSTALL) IF 'nodejs_version != "10.x"'
+  fail:
+    msg: "PBX install cannot proceed, as it currently requires Node.js 10.x, whereas nodejs_version is set to {{ nodejs_version }}.  Please check the value of nodejs_version in /opt/iiab/iiab/vars/default_vars.yml, /etc/iiab/local_vars.yml, /opt/iiab/iiab/roles/nodejs, etc."
+  when: nodejs_version != "10.x"
```
jvonau, line 18: If you are touching pbx perhaps it would be a good time to address #2210 and correct this test.

holta: I don't know how to solve #2210 so am keeping that separate.

As important as it is.

@tim-moody may have ideas off of PR #2239 possibly.

roles/yarn/tasks/main.yml
```diff
@@ -1,41 +1,48 @@
...
+    fail_msg: "PLEASE GIVE VARIABLE 'yarn_enabled' A PROPER (UNQUOTED) ANSIBLE BOOLEAN VALUE e.g. IN: /etc/iiab/local_vars.yml"
+    quiet: yes
+
+- debug:
+    var: yarn_install
+- debug:
+    var: yarn_enabled
+- debug:
+    var: yarn_installed
+
```
jvonau, line 27: I see debug was removed elsewhere, going to release with debug enabled?

runrole
```diff
@@ -61,7 +61,7 @@ fi
 #fi
 
 if [ "$REINSTALL" == "1" ]; then
-    if [ ! $1 == "internetarchive" ]; then # special handling
+    #if [ ! $1 == "internetarchive" ]; then # special handling
```
jvonau, line 64: This is a change @mitra42 wants? Removing the installed flag results in #1921 being reverted.

tim-moody:
no to special handling.

roles/yarn/tasks/main.yml
holta: For now yes, intentionally displaying the 3 core variables within dependencies for the coming months.  I think I've got these 5 covered, but might want to add a few others later?!

- MongoDB
- MySQL
- PostgreSQL
- Node.js
- Yarn

Even later this year this kind of thing might still make sense to educate new implementers.

runrole
holta: Seems like a very positive change to standardize most all IIAB roles so implementers stay sane.

But @mitra42 is here with me in Arizona and we might call you later to see if other options are better...

jvonau:
> no to special handling.

ARGS="$ARGS -e reinstall=True" was really misnamed but was the prototype for future --upgrade handling, just not exposed to the user.

runrole
mitra42: To be clear - I haven't requested any change - We finally got a way about two weeks ago to do an update. (It was there before, but without documentation weren't able to use it) 
```
cd /opt/iiab/iiab
git pull
./runrole --reinstall internetarchive
```

I'm fine if that changes to some other command, though something like ..

`ARGS="$ARGS -e reinstall=True" ` 

Plus whatever else is required (I'm sure this isn't all),  seems highly prone to mistyping, so I much prefer the former.

mitra42:
What's the complete command to upgrade for example internetarchive ? I'm sure it isn't just 
`ARGS="$ARGS -e reinstall=True"`
Which alone seems HIGHLY prone to mistyping, we need a simpler way to instruct someone we are providing support to how to upgrade.

jvonau:
> What's the complete command to upgrade for example internetarchive ? I'm sure it isn't just
> `ARGS="$ARGS -e reinstall=True"`
> Which alone seems HIGHLY prone to mistyping, we need a simpler way to instruct someone we are providing support to how to upgrade.

`./runrole --reinstall internetarchive` used to set an ansible 'reinstall=True' variable which the role relied upon to fire off `yarn config set child-concurrency 1 && yarn install && yarn upgrade` bypassing the node and yarn installs when requested not every time the role is called.

mitra42:
Sorry, having trouble parsing that last sentence.

When you say "used to set" do you mean "used" as in " it is used to set..." or as in "In the past it used to ... but not any more" .  ...

holta:
> "In the past it used to ... but not any more"

That was the past, indeed.  But let's get rid of special cases now.  So that all (or most every) IIAB role is reinstalled using this same syntax:
```
./runrole --reisntall internetarchive
```

roles/internetarchive/tasks/main.yml
holta: Definitely not the final word, but I'd like to merge now, while at the same time let's keep thinking thru these `--reinstall` and `--upgrade` possibilities to get this right in 2020.

jvonau: I'll go with this role is the only role forcing an upgrade to be preformed post-install deviating from the behavior of all the other roles. This results in a large increase(50%) in post-install runtime of ICO or iiab-configure was ~9mins on large, now at ~15mins. footnote while testing #2200 which had a runtime of 6.5mins or less and cracked 6min a couple of times so this is going the wrong way IMHO.

jvonau:
> > "In the past it used to ... but not any more"
> 
> That was the past, indeed. But let's get rid of special cases now. So that all (or most every) IIAB role is reinstalled using this same syntax:
> 
> ```
> ./runrole --reisntall internetarchive
> ```
                   ^^ --reinstall
The routine always did follow what upstream wanted but was not documented well, me bad, you bad whatever...  Kind of need that 'hey it works the way I like' type of feedback to put the effort into docs but you changed what --reinstall does now so it's not on my plate.

jvonau:
> Sorry, having trouble parsing that last sentence.
> 
> When you say "used to set" do you mean "used" as in " it is used to set..." or as in "In the past it used to ... but not any more" . ...

The second one.

roles/internetarchive/tasks/main.yml
holta: > this role is the only role forcing an upgrade to be preformed post-install

Many roles traditionally have upgraded after installation, for example WordPress, and for very good reasons — even if deciding when security is a priority, when ongoing community dogfooding (community testing) is a priority, when databases should be wiped completely clean — are tough decisions where the answer will not always be the same for each app.

Etc.  Among other post-install desires, which are the reason that flags like `--upgrade` are not at all an easy topic — but some/such will eventually be needed!

holta: Just to restate what was said earlier on this tkt and others — internetarchive and/or others will likely soon want their Ansible roles to run in 3-or-4-or-5-or-more different ways, e.g.:
1) `--reinstall` while preserving all of the app's db's, content & metrics/logs
2) `--reinstall` after WIPING all app's db's, content (and ideally wiping all of its associated privacy/analytics/metrics/logs)
3) `--upgrade` an app
4) Disable-or-enable an already installed app
5) Other scenarios that will surely emerge...both online and offline

mitra42: I think this is an important conversation that shouldnt be buried in a comment on a MR that is already merged :-) 

I agree with @jvonau that we dont want an automatic update of Internet Archive's code to slow down other completely unrelated tasks. 

I also agree with @holta that we should think about what flags are needed so there is a template and consistent way to express different requirements. I'm not quit sure of the nuances of the differences between his options 1 and 2 above, but I certainly see three completely difference scenarios 
a) Upgrade - make sure the box is running the latest code, this is particularly important when bugs are getting fixed rapidly (Wordpress is a good example of this). 
b) Reinstall - something went wrong, lets clean up as much as we can and try again. Typically I'm going to want to preserve as much data as possible in that case. 
c) Deinstall - take the app off, and wipe the data. 

And especially "a" and "b" need to be one-liners, since its the kind of thing I'd expect to be doing via remote support, when I certainly dont want to be telling someone to go and edit this variable, or delete this directory etc etc. 

One impressive thing on the OLIP demo was the way it was easy to add / delete / update apps from a UI.

holta: > c) Deinstall - take the app off, and wipe the data.

Yes, we might start experimenting with this earlier than I expected.  IIAB file `/etc/iiab/iiab_state.yml` now makes this opportunity possible.

> One impressive thing on the OLIP demo was the way it was easy to add / delete / update apps from a UI.

Yes.  That was one aspect of OLIP that appeared to work extremely well — somewhat similar to Kolibri — offering a very clean and simple GUI that just works as people expect (requiring extensive community listening to refine & get right!) as is becoming critical in this new decade...

holta: Thanks to @mitra42 for calling today, further evolving this preliminary design discussion, around what folks truly need from [./runrole](https://github.com/iiab/iiab/blob/master/runrole) and similar post-install scenarios.  He explained that internetarchive requires only 2 code paths for now (for IIAB 7.1) and that is more than sufficient for now.

But that we all agree to think about 3 code paths e.g. something like `./runrole --upgrade internetarchive` in future if that proves necessary.

As summarized below, there is usually no significant delay to running `yarn upgrade` every time an implementer is online (as documented at PR #2248) and regardless even if there was, any larger delay would be most worth it at this phase in [internetarchive](https://github.com/iiab/iiab/tree/master/roles/internetarchive/)'s life cycle &mdash; as this increases testing of the latest code towards it becoming a more mature product:  https://github.com/iiab/iiab/blob/master/roles/internetarchive/tasks/main.yml#L22-L30

PR #2242, opened 2020-02-07
Bug fix: Split http/apache

jvonau:
### Fixes Bug
Setting apache_install: False would suppress running html.yml breaking menus and/or admin-console
splits apache/httpd 
subset of #2200 testing

### DIFF
--- a/roles/3-base-server/tasks/main.yml
+++ b/roles/3-base-server/tasks/main.yml
@@ -3,21 +3,24 @@
 - name: ...IS BEGINNING =====================================
   command: echo
 
-- name: HTTPD (APACHE)
+- name: HTTPD
   include_role:
     name: httpd
-  when: apache_install | bool
 
 - name: MYSQL
   include_role:
     name: mysql
   when: mysql_install | bool
 
-- name: Install NGINX (configured LATER, in Stage 9-LOCAL-ADDONS)
+- name: Install APACHE (configured LATER, in Stage 4)
+  include_tasks: roles/httpd/tasks/install.yml
+  when: apache_install | bool
+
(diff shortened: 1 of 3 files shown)


tim-moody:
I have thought for some time that adm console should install its own directories in docroot, any not used by something else.

holta:
Can we move all [3 tasks](https://github.com/iiab/iiab/blob/master/roles/httpd/tasks/main.yml#L13-L20) (currently within Apache role [httpd](https://github.com/iiab/iiab/tree/master/roles/httpd)) if it's confirmed most-or-all are no longer exclusively related to Apache?

- [html.yml](https://github.com/iiab/iiab/blob/master/roles/httpd/tasks/html.yml)
- [php-stem.yml](https://github.com/iiab/iiab/blob/master/roles/httpd/tasks/php-stem.yml)
- [iiab-refresh-wiki-docs](https://github.com/iiab/iiab/blob/master/roles/httpd/templates/refresh-wiki-docs.sh)

If so, can we move all 3 into a new role, in support of NGINX-and-Apache both?

(What's a decent name for such a new role if so?!)

jvonau:
Everything other than enable|install.yml in httpd is shared between apache/nginx, be easier to move just the install|enable.yml to a new apache role? This PR is the quickest fix without major renaming anything for now. 

I'm not a fan of having when *_install in 3-common, optional stuff is what 4-server is intended for and the need for tracking. Allowing mysql to be user optional could result in missing some php packages that are installed in mysql's role. Might want to split that up but there is no need to really track if the state is installed when we install in stage3 as that should be considered always installed. Being able to shutdown mysql by the user should be allowed for those users who might of slimmed down the install and nothing is using the DB backend.

Issue #2247, opened 2020-02-11
Internetarchive --reinstall not working


mitra42:
### Expected behavior and actual behavior.
```
cd /opt/iiab
sudo git pull
sudo ./runrole --reinstall internetarchive
```
I'd expect that it does a `yarn upgrade` at some point as it used to. 

But ... the upgrade code seems to have been pulled out during that previous MR - so it doesn't work

### Specifications like the version of the IIAB, operating system version, or hardware details.
Git pull including up to the MR 2245 from @holta on 10th Feb

holta:
> `cd /opt/iiab`

You mean `cd /opt/iiab/iiab` I'm pretty sure?  I'll look into it!

holta:
> upgrade code seems to have been pulled out

Upgrade/update code was preserved here: https://github.com/iiab/iiab/blob/master/roles/internetarchive/tasks/main.yml#L27-L42

As of now this runs when the operator/implementer does:

```
sudo su -
cd /opt/iiab/iiab
./runrole internetarchive
```

As opposed to...

```
sudo su -
cd /opt/iiab/iiab
./runrole --reinstall internetarchive
```

But let's not get losts in the reeds.  More important I want to call @mitra42, so we all get on the same page first, helping everyone understand what's actually expected/useful/best going forward, so the most common cases (if not every obscure desire!) are supported in very-clear-and-documented ways.

Whether or not following outlines (like!) https://github.com/iiab/iiab/pull/2238#discussion_r376168178

mitra42:
Yes - in https://github.com/iiab/iiab/pull/2238#issuecomment-582940866 it was said that:
```
That was the past, indeed. But let's get rid of special cases now. So that all (or most every) IIAB role is reinstalled using this same syntax:
./runrole --reisntall internetarchive
```
(note there is an obvious typo in that line).

but I took that to mean that update is done only when I give the arg `--reinstall` rather than only when I DO NOT give the `--reinstall` arg.

I also cant find anywhere that `interetarchive_upgrade` is set, since the `set_fact` for that has been commented out. 

I don't really care what the outcome is, except that I *do* need a way to tell it to do an upgrade !

holta:
Thanks @mitra42 for calling me just now.  Conclusions:

[...]

PR #2248, opened 2020-02-11
internetarchive: wipe /opt/iiab/internetarchive/node_modules prior to --reinstall

holta:
Also cleans up code per discussion with @mitra42 at #2247

Tested on Ubuntu 19.10:
- `./runrole internetarchive` completed in 37 seconds (both times)
- `./runrole --reinstall internetarchive` completed in 42 sec (1st time) or 41 sec (2nd time)

(Interestingly the vast majority of both 2 above times appears to be consumed by IIAB's [0-init](https://github.com/iiab/iiab/tree/master/roles/0-init) stage, so there's certainly room to optimize that later, if @jvonau or someone else has time!)

### DIFF
--- a/iiab-install
+++ b/iiab-install
@@ -159,7 +159,7 @@ if [ -f /etc/iiab/config_vars2.yml ]; then
     fi
 
 # TEMPORARY: another change to account for
-    sed -i -e 's/pan_bluetooth/bluetooth/' $IIAB_STATE_FILE
+#    sed -i -e 's/pan_bluetooth/bluetooth/' $IIAB_STATE_FILE
 
     if [ "$STAGE" -eq 2 ]; then
         echo -e "\nCompleting Stage 3 from IIAB image (starts systemd service iiab-setup-db to run the 'mysql' role)."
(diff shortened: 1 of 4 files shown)


jvonau:
time sudo ./runrole internetarchives gives me 2m57.219s on rpi4
time sudo ./runrole --reinstall internetarchives gives me 5m10.179s on rpi4
I can't see how 30-40ish seconds is even possible unless offline.

holta:
> time sudo ./runrole internetarchives gives me 2m57.219s on rpi4
> time sudo ./runrole --reinstall internetarchives gives me 5m10.179s on rpi4
> I can't see how 30-40ish seconds is even possible unless offline.

The above numbers are extremely useful to add to the earlier tests from Ubuntu 19.10 VM.

(And of course all these numbers are heavily dependent the time it takes `yarn add` to download what becomes 112+ MB, or the time it takes `yarn upgrade` to download whatever incremental bits.)

mitra42:
The one number we need in addition is running `./runrole --reinstall internetarchive` immediately after a previous `./runrole --reinstall internetarchive` its quite possible that the first one actually did an update, though its not as instantaneous as I'd like because yarn rebuilds packages unneccessarily (I've posted that bug on the yarn site twice now).

PR #2249, opened 2020-02-12
Apache & NGINX made more readable. Add roles/www_back_end (in Stage 3 for now) & roles/www_front_end (in Stage 4 for now)

holta:
Building on @jvonau's PR #2242

Includes an interim solution replacing roles/homepage with web_support/tasks/main.yml & httpd/tasks/install.yml &mdash; as the above PR #2242 caused Admin Console installation to fail &mdash; as a result of /library/www/html/home being missing.

Later on we should properly softcode `iiab_home_url` across the board, when it's more clear how to do that in service of genuine field needs.

Tested on Ubuntu 18.04 & 19.10


holta:
1) CORRECTION: this PR is building on @jvonau's PR #2242 (not 2204, as GitHub's web autocorrect erroneously changed it to earlier).

2) I'd love to incorporate ideas from different people, if others have time to review this today ideally.

jvonau:
What a bunch of BS move 70 historical files around for what reason? 

> Later on we should properly softcode `iiab_home_url` across the board, when it's more clear how to do that in service of genuine field needs.

`iiab_home_url` is already [softcoded](https://github.com/iiab/iiab/search?q=iiab_home_url&unscoped_q=iiab_home_url). The reason the install/enable tasks are in different stages is to be able to toggle things in the field. Moving iiab_home_url assignment into stage3 is a step backwards, with this one has to --reinstall to be able to pass through nginx/apache's enable* tasks where  iiab_home_url  is used in the templates.

jvonau:
> as the above PR #2242 caused Admin Console installation to fail — as a result of /library/www/html/home being missing.

so where is this comment in 2242? and is now corrected

roles/nginx/tasks/install.yml
```diff
@@ -8,6 +8,12 @@
     - { path: "{{ nginx_conf_dir }}/usb-lib.conf" }
     - { path: "{{ nginx_conf_dir }}/modules.conf" }
 
+- name: Stop '{{ apache_service }}' systemd service
+  systemd:
+    name: "{{ apache_service }}"    # apache2 or httpd, per /opt/iiab/iiab/vars/<OS>.yml
+    state: stopped
+  ignore_errors: yes
+
```
jvonau, line 16: That is why httpd/tasks/install.yml had(before you disabled) the final systemd call set to stopped in that code block. Things were a certain way for a reason, to avoid using 'ignore_errors' needlessly. Don't forget to use apache_installed if you want to avoid ignore_errors red warning when apache may not be installed in the future.

holta:
BIG-sized install smoke-tested with basic functionality testing of most apps during a fresh install onto Ubuntu 18.04

holta:
Also smoke-tested on RPi 4 w/ 2GB RAM.

holta:
Today there are few-to-no requests from offline communities to change things like `/library/www/html/home` as this is not something teacher/medics care about.

However: @jvonau's general notion that some TBDiscovered things need to be "instantly" changeable in the offline/field is right on target.

So for now I've implemented [roles/www_back_end](https://github.com/iiab/iiab/tree/master/roles/www_back_end) (runs in Stage 3 for now) & [roles/www_front_end](https://github.com/iiab/iiab/tree/master/roles/www_front_end) (runs in Stage 4 for now) each of which can be renamed and reapportioned later as nec.

So as to honor the general spirit of @jvonau's suggestion (that various TBDiscovered UX-related Apache/NGINX settings not be locked into Stage 3-BASE-SERVER, permitting such to be rapidly field-changeable in various ways with GUI and non-GUI, regardless whether online or offline...)

i.e. if and when a teacher-friendly / smartphone-friendly GUI does mature (for NON-technical operators!) things like an IIAB Internet ON/OFF switch, and maybe a Wi-Fi SSID-changer (etc) might possibly be in very high demand...

Issue #2222, opened 2020-01-31
iiab_admin_user_install: False trap


tim-moody:
If you set iiab_admin_user_install: False, iiab-install will fail in stage 4. However, it is not enough to go back and set it to True, because the stage in which it is to be set has already been marked as complete. Therefore you must begin again with iiab-install --reinstall.

holta:
> Therefore you must begin again with iiab-install --reinstall.

Tangentially Related, is a design question about whether './iiab-install --reinstall' should 'rm /etc/iiab/iiab_state.yml' (or most/all of its contents?)

Background:
- './runrole --reinstall [ROLE NAME]' deletes (1 line) from /etc/iiab/iiab_state.yml
- './iiab-install --reinstall' does not currently delete (any lines) from /etc/iiab/iiab_state.yml

Somewhat related to:

#2210 "'./runrole pbx' fails due to predefined var 'pbx_installed' in pbx/defaults/main.yml -- illustrates weakness of var validation when invoked from ./iiab-install (instead of from ./runrole)"

jvonau:
Great bug report 'stage 4 failure', recall which sub-role was involved? I'll go out on a limb and suggest it was iiab_lib as #1228 suggest that iiab_admin_user is the owner of the content directory.

holta:
> Great bug report

#2222 is easy to remember too (two too to... ;)

tim-moody:
> recall which sub-role was involved?

no. which is why I didn't speculate.

holta:
I'm assuming this was closed by accident?

(Seems like something that should be fixed eventually.)

jvonau:
> If you set iiab_admin_user_install: False, iiab-install will fail in stage 4. However, it is not enough to go back and set it to True, because the stage in which it is to be set has already been marked as complete. Therefore you must begin again with iiab-install --reinstall.



> > Therefore you must begin again with iiab-install --reinstall.
> 
> Tangentially Related, is a design question about whether './iiab-install --reinstall' should 'rm /etc/iiab/iiab_state.yml' (or most/all of its contents?)
> 
> Background:
> 
>     * './runrole --reinstall [ROLE NAME]' deletes (1 line) from /etc/iiab/iiab_state.yml
> 
>     * './iiab-install --reinstall' does not currently delete (any lines) from /etc/iiab/iiab_state.yml
> 
[...]

Issue #1228, opened 2018-10-15
/library/www/html/local_content group perm: {{ iiab_admin_user }} instead of {{ apache_user }} for a reason? NGINX?


holta:
Am just trying to understand if this is truly needed (?)

See [roles/usb-lib/tasks/main.yml#L5-L6](https://github.com/iiab/iiab/blob/master/roles/usb-lib/tasks/main.yml#L5-L6)

holta:
Weird: different playbooks of ours set different permissions and different ownership for /library/www/html/local_content ...so when you rerun 2-common's [fl.yml](https://github.com/iiab/iiab/blob/master/roles/2-common/tasks/fl.yml) it briefly reverts to:

> root:root ownership with 07**5**5

And then some other later playbook re-reverts it to:

> www-data:iiab-admin ownership with 07**7**5

holta:
@tim-moody is this ok as is?

Or can you make a recommendation here, towards closing this issue?

holta:
> Am just trying to understand if this is truly needed (?)
> 
> See [roles/usb-lib/tasks/main.yml#L5-L6](https://github.com/iiab/iiab/blob/master/roles/usb-lib/tasks/main.yml#L5-L6)

Code is now here:
https://github.com/iiab/iiab/blob/master/roles/usb_lib/tasks/main.yml#L11-L12

(...as we're discussing right now during this week's http://minutes.iiab.io call!)

@tim-moody & @jvonau suggest we change this now to www-data:www-data

holta:
Solved by PR #2250

PR #2250, opened 2020-02-13
Modularize roles/usb_lib + change group ownership of /library/www/html/local_content

holta:
For #1228, #2222 etc.


holta:
Smoke-tested both code paths using `./runrole usb_lib` and `./runrole --reinstall usb_lib` (on Ubuntu 18.04).

holta:
Documentation (i.e. README.rst) is getting quite stale @ https://github.com/iiab/iiab/tree/master/roles/usb_lib

But IIAB still leads (compared to many other offline libraries products) here so...I think I'll leave it there for now...as a technical reference to supplement http://FAQ.IIAB.IO -> "4 Can teachers display their own content?"

PR #2240, opened 2020-02-06
Cal py3 [Calibre-Web: move to Python 3] ...uncovers 2 Calibre-Web issues: slow shutdown & "Perform Update" button risks

jvonau:
### Fixes Bug
move to python3
### Description of changes proposed in this pull request.
move calibre-web to python3 #2231 
### Smoke-tested in operating system.
raspbian-10

Issue #2195, opened 2020-01-24
NGINX: Can someone help fix http://box/livres, http://box/libros etc? [alongside Calibre-Web's http://box/books]


holta:
This used to work with Apache!

See both Apache and NGINX's .conf files here:
https://github.com/iiab/iiab/tree/master/roles/calibre-web/templates

The 3 current URL's are defined here:
https://github.com/iiab/iiab/blob/master/roles/calibre-web/defaults/main.yml
...which really means:
https://github.com/iiab/iiab/blob/master/vars/default_vars.yml#L586-L594

holta:
Huge thanks to @avni helping us all with testing as this hopefully comes together soon for [IIAB 7.1](https://github.com/iiab/iiab/wiki/IIAB-7.1-Release-Notes) !

PR #2253, opened 2020-02-14
NGINX: Add http://box/libros & http://box/livres (in addition to http://box/books, for Calibre-Web)

holta:
Appears to work during quick testing on Ubuntu 19.10

Per @jvonau recommendation, to fulfill #2195

Related: #2240 "Cal py3 [Calibre-Web: move to Python 3]"


jvonau:
Works here with ./runrole calibre-web to apply, no need for --reinstall as the nginx config file is supplied outside of install.yml.

PR #2263, opened 2020-02-19
Update RPi minetest to 5.1.1 for compatibility with latest raspbian

tim-moody:


holta:
Is this tested & ready to merge?

tim-moody:
passed smoke test. passed function test that a 5.1.1 client can connect and navigate through the virtual world. not tested for all virtual world functionality.
bottom line is that there is little risk in merging as it replaced something that doesn't work at all and does not seem to break the playbook.

PR #2267, opened 2020-02-26
Move calling iiab-divert-to-nginx [fix to enable Captive Portal]

jvonau:
### Fixes Bug
live testing in MX


holta:
Thanks @jvonau for noticing that `/etc/dnsmasq.d/capture` was missing during RE-enabling of Captive Portal.

(Please test, all who can!)

PR #2275, opened 2020-03-04
Kill off is_rpi [in favor of is_raspbian and rpi_model]

jvonau:


jvonau:
Writes model of rpi to iiab.ini
> changed: [127.0.0.1] => (item={u'option': u'is_rpi', u'value': u'Raspberry Pi 4 Model B Rev 1.1'})

holta:
@jvonau out of curiosity is `/proc/device-tree/model` that you're using on Line 60 of /opt/iiab/iiab/scripts/local_facts.fact any different[*] than `/sys/firmware/devicetree/base/model` that's currently used on Line 111 of /opt/iiab/iiab/scripts/iiab-diagnostics ?

[*] Better?  Or worse?  (Or hypothetically perhaps both are guaranteed to be identical on all OS's ??)

tim-moody:
The problem is that rpi refers both to a hw device and in our code to a presumed os. This change assumes it is only an os. So, for example, bluetooth would not get installed on rpi under ubuntu. There are many services that may or may not work under ubuntu on rpi due to these changes.

holta:
> The problem is that rpi refers both to a hw device and in our code to a presumed os. This change assumes it is only an os. So, for example, bluetooth would not get installed on rpi under ubuntu. There are many services that may or may not work under ubuntu on rpi due to these changes.

Great question.

It looks like @jvonau is proposing to initialize `rpi_model: none` on Line 9 of roles/0-init/defaults/main.yml as a de facto substitute for the `is_rpi` boolean variable, to allow conditional testing of Raspberry Pi hardware.

With the string value `none` being used in much the same way as *_install and *_enabled Ansible variables derived from the contents of /etc/iiab/iiab_state.yml i.e. when they exist in that file.  (Whereas we run tests like `when: moodle_install undefined` much like @jvonau appears to be proposing here that we run tests like `when: rpi_model == "none"`).

jvonau:
> /sys/firmware/devicetree/base/model

I'd say yes 
> pi@box:/opt/iiab/iiab $ ls -la /proc/device-tree
lrwxrwxrwx 1 root root 29 Mar  8 10:04 /proc/device-tree -> /sys/firmware/devicetree/base

jvonau:
> > The problem is that rpi refers both to a hw device and in our code to a presumed os. This change assumes it is only an os. So, for example, bluetooth would not get installed on rpi under ubuntu. There are many services that may or may not work under ubuntu on rpi due to these changes.
> 
> Great question.
> 
> It looks like @jvonau is proposing to initialize `rpi_model: none` on Line 9 of roles/0-init/defaults/main.yml as a de facto substitute for the `is_rpi` boolean variable, to allow conditional testing of Raspberry Pi hardware.
> 
> With the string value `none` being used in much the same way as *_install and *_enabled Ansible variables derived from the contents of /etc/iiab/iiab_state.yml i.e. when they exist in that file. (Whereas we run tests like `when: moodle_install undefined` much like @jvonau appears to be proposing here that we run tests like `when: rpi_model == "none"`).

The [conditional](https://github.com/iiab/iiab/blob/master/roles/1-prep/tasks/main.yml#L119-L120) that fires off raspberry_pi.yml is already != none, so that is not a change, just based on better information.

jvonau:
> The problem is that rpi refers both to a hw device and in our code to a presumed os. This change assumes it is only an os. So, for example, bluetooth would not get installed on rpi under ubuntu. There are many services that may or may not work under ubuntu on rpi due to these changes.

Ubuntu and bluetooth has never been offered so that is not a change but I think bluetooth might work under ubuntu or debian. If something doesn't work with ubuntu on rpi the cause would not be this PR. This is a straight replacement of is_rpi with is_raspbian both defined in the OS.yml files. Now rpi_model is gaining better info, and raspberry_pi.yml being run, only risky new change, should be safe when ubuntu is on the rpi.

Mortvert:
I have a new one for you.

obraz

I run my Pi 3 off a USB HDD (faster than microsd for some reason) and when I tried to run the install script - you get what's above.

@holta Sorry if this doesn't belong here but it looks like it? If not, I'll create a new issue.

holta:
@Mortvert you can create an issue if you prefer.

The core of the issue is that installing to microSD is the (supported) norm on Raspberry Pi.

Whereas you're installing in a non-standard way, so you likely need to comment out or modify [Line 177-180 of /usr/bin/iiab](https://github.com/iiab/iiab-factory/blob/master/iiab#L177-L180) &mdash; before you run `sudo iiab`

jvonau:
Feels like a bug, the root filesystem on usb option is pretty common given boot from usb works on rpi3 out of the box(rpi4 coming soon). Why have a user edit the iiab script? The script could (should?) check if /dev/mmcblk0p2 is mounted before running tune2fs.

holta:
> Feels like a bug, the root filesystem on usb option is pretty common given boot from usb works on rpi3 out of the box(rpi4 coming soon). Why have a user edit the iiab script? The script could (should?) check if /dev/mmcblk0p2 is mounted before running tune2fs.

I'm sympathetic, if this is easy to implement, yes.

Issue #2276, opened 2020-03-05
./iiab-network failed to set a Fixed IP, due to double quotes around eth0 in /etc/dhcpcd.conf


holta:
Running `cd /opt/iiab/iiab` then `./iiab-network` on RPi 3 B+ (the IIAB microSD was originally built in a RPi 4) appears to have put the following into /etc/dhcpcd.conf &mdash; after a Fixed IP Address was specified in local_vars.yml:

```
interface "eth0"
```

The specified Fixed IP Address never appeared (even after reboot) as a result of the double quotes above.  So I manually changed it to:

```
interface eth0
```

And then rebooted...at that point eth0's Fixed IP Address began to work.  @jvonau may have ideas why.

Aside: Being extremely offline, thankfully I had an offline copy of FAQ.IIAB.IO -> "22 How do I set a static IP address?" on the IIAB itself at http://box/info (telling me what to do manually) to rescue the situation in front of 20 teachers & 5 Chiapas govt inspectors!

Related:

- http://FAQ.IIAB.IO -> "22 How do I set a static IP address?"
- #1965 "eth0" appears (with quotes, sometimes alone on its own line) in /etc/iiab/iiab.ini

jvonau:
Without a gateway present the recorded IIAB_WAN_DEVICE is used as a seed value via prior_gw -> device_gw -> lines 120-123 of computed_network.yml

holta:
@jvonau's PR #2278 should hopefully fix this, please take a look those who can!

holta:
This appears fixed by @jvonau's PR #2278, now merged.

Issue #1965, opened 2019-09-26
"eth0" appears (with quotes, sometimes alone on its own line) in /etc/iiab/iiab.ini


holta:
This occurred on a fresh install being worked on 1 month ago in Mexico ("eth0" appeared in /etc/iiab/iiab.ini, including the quotation marks).  Concerning that the double-quotes appear (and should not).

Even more concerning is "eth0" that appeared twice in /etc/iiab/iiab.ini in South Africa on this IIAB machine built in April 2019 that was reinstalled in early September 2019: (again with double-quotes, but even more seriously when there is not variable name on the left, on the 4th line below)

```
...
[detected_network]
has_ifcfg_gw = none
prior_gateway_(device_gw2) = "eth0"
"eth0"
dhcpcd_result = enabled
network_manager_active = False
systemd_networkd_active = False
wan_in_interfaces = False
wireless_list_1(wifi1) = not found-1
wireless_list_2(wifi2) = wlan0
num_wifi_interfaces = 1
discovered_wireless_iface = wlan0
discovered_wired_iface = none
num_lan_interfaces = 1
gui_static_wan = False
iiab_lan_iface = br0
iiab_wan_iface = eth0
...
```

As excerpted from http://paste.debian.net/1102439/ which shows /etc/iiab/iiab.env as follows:

```
# This is a configuration file for IIAB
# It can be sourced in a shell script or read into an application
IIAB_BASE_PATH=/opt/iiab
IIAB_DIR=/opt/iiab/iiab
IIAB_RELEASE=7.0
IIAB_REVISION=0
IIAB_GATEWAY_ENABLED=True
LAN_IP=172.18.96.1
OS=raspbian
OS_VER=raspbian-9
WWWROOT=/library/www/html
STAGE=9
FQDN=box.lan
IIAB_USB_LIB_SHOW_ALL=True
IIAB_WAN_DEVICE="eth0"
IIAB_LAN_DEVICE="br0"
HOSTAPD_ENABLED=True
HTTPCACHE_ON=True
```

This results in Admin Console failing to install with the "eth0" error shown at bottom here:

[...]

PR #2278, opened 2020-03-05
remove quotes

jvonau:
### Fixes Bug
#2276
quote part of #1965

### DIFF
--- a/roles/network/tasks/computed_network.yml
+++ b/roles/network/tasks/computed_network.yml
@@ -176,14 +176,14 @@
   lineinfile:
     path: "{{ iiab_env_file }}"
     regexp: '^IIAB_WAN_DEVICE=*'
-    line: 'IIAB_WAN_DEVICE="{{ iiab_wan_iface }}"'
+    line: 'IIAB_WAN_DEVICE={{ iiab_wan_iface }}'
   when: not installing   #REMOVE THIS LINE IF installing IS ALWAYS false AS SET IN roles/0-init/defaults/main.yml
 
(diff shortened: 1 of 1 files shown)


holta:
How safe is this PR?

(Any specific/quick tests that should be fun [run] before this is merged?)

holta:
I just tried this PR on an offline RPi 3 B+ &mdash; for some reason it continues with the old/erroneous behaviore &mdash; putting double quotes into /etc/dhcpcd.conf within this line:

```
interface "eth0"
```

If it helps to debug where this is going wrong, I'd attached the last 975 lines of the offline machine's /etc/iiab/iiab/iiab-network.log here:

https://paste.ubuntu.com/p/rwbDjzzNXs/

jvonau:
would like to see /etc/iiab/iiab.env also

holta:
I forgot that this implements "fallback" to a Static/Fixed IP Address on eth0, which requires an Ethernet cable plugged in, with link-light etc.

We should state this more strongly in http://FAQ.IIAB.IO -> "22 How do I set a static IP address?"

In any case, this PR appears to be correct for new installs &mdash; so that eth0 falls back to the Fixed IP (as set in /etc/dhcpcd.conf) after about 30 to 45 seconds.

holta:
Tested on Raspbian Desktop 2020-02-13.

holta:
> We should state this more strongly in http://FAQ.IIAB.IO -> "22 How do I set a static IP address?"

Done!

Issue #2270, opened 2020-03-03
Ubuntu 20.04 aarch64 / arm64 testing [on Raspberry Pi]


georgejhunt:
I've had fun setting up a development branch ubuntu 20.04 64bit ubuntu on rpi4 from the available install iso. 
Our oneline install script stalled on python3-pip not available yet.  
We may find that they do not meet their release target -- April 23 (current date Mar 2)

georgejhunt:
## Recipe for using daily build iso on rpi:
(Please disregard this item. Jerry's idea of upgrading is much better -- see below)
1. Source for daily build: http://cdimage.ubuntu.com/ubuntu-server/daily-live/current/focal-live-server-arm64.iso
1. I started with ubuntu 19.10 image. But it might have been faster to partition and copy /boot (deleting all of partition2 took a long time)
2. Copy kernel and initrd to /boot (the kernel seq number needs to be max of what/s in /boot).
3. Mount the squashfs from the daily build iso, and copy it to userland, to /dev/mmcblk0p2.
4. Power on, and add "restricted" and "universe" to /etc/sources.list.
5. I always apt-get install vim git tree
5. Needs a /etc/netplan/config.yaml to bring up eth0, set a /etc/resolv.conf that points to a good dns.
5. Expand the /dev/mmcblk0p2 beyond 2GB using resize2fs, after making larger partition with fdisk.
5. Install openssh-server
6. Without a functioning /etc/fstab, /dev/mmcblk0p2 needs to be remounted rw after every boot.
7. We need a real is_rpi. I just put it in ./vars/ubuntu-20 for temporary.
7. Get wget, and get the one line installer.
8. Dnsmasq does not want to start if br0 is missing -- comment it out in /etc/dnsmasq.d/iiab
9. Apparently netplan only works on wifi with NetworkManager, but we use systemd.networkd. So getting br0 going via hostapd needs work.
9. Changes to master for ubuntu 20.04 at https://github.com/georgejhunt/iiab/tree/u2004.

holta:
> I started with ubuntu 19.01 image

There is no 19.01 per se.  (There are 19.10, 19.10.1 and 20.04 daily builds of different kinds.)

To avoid all confusion, @georgejhunt  can you please provide a complete/example URL to an ISO you'd use ?

jvonau:
[...]

PR #2291, opened 2020-03-15
ubuntu 20 ansible workaround

jvonau:
### Fixes Bug
#2270 
untested more notes than working code

### DIFF
--- a/scripts/ansible
+++ b/scripts/ansible
@@ -67,7 +67,8 @@ elif [ -f /etc/debian_version ]; then    # Includes Debian, Ubuntu & Raspbian
 
     echo -e "\napt update; install dirmngr; PPA to /etc/apt/sources.list.d/iiab-ansible.list\n"
     apt update
-    if ! [ `cat /etc/os-release | grep UBUNTU_CODENAME= | awk -F "=" '{print $2}'` == "focal"]; then
+    IS_U20?=`cat /etc/os-release | grep UBUNTU_CODENAME= | awk -F "=" '{print $2}'`
+    if [ "$IS_20?" != "focal" ]; then
         apt -y install dirmngr
         echo "deb http://ppa.launchpad.net/ansible/ansible/ubuntu bionic main" \
          > /etc/apt/sources.list.d/iiab-ansible.list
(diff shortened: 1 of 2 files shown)


jvonau:
Now tested on raspbian OS works as expected installing the key and ppa. 2.9.4 should be installed on Ubuntu-20.04 without installing the repo key and ppa for ansible.

scripts/ansible
```diff
@@ -64,16 +64,17 @@ elif [ -f /etc/centos-release ]; then
...
+        apt -y install dirmngr
+        echo "deb http://ppa.launchpad.net/ansible/ansible/ubuntu bionic main" \
          > /etc/apt/sources.list.d/iiab-ansible.list
 
-    echo -e '\nIF YOU FACE ERROR "signatures couldn'"'"'t be verified because the public key is not available" THEN REPEATEDLY RE-RUN "sudo apt-key adv --keyserver hkp://keyserver.ubuntu.com:80 --recv-keys 93C4A3FD7BB9C367"\n'
-    apt-key adv --keyserver hkp://keyserver.ubuntu.com:80 --recv-keys 93C4A3FD7BB9C367
-
+        echo -e '\nIF YOU FACE ERROR "signatures couldn'"'"'t be verified because the public key is not available" THEN REPEATEDLY RE-RUN "sudo apt-key adv --keyserver hkp://keyserver.ubuntu.com:80 --recv-keys 93C4A3FD7BB9C367"\n'
+        apt-key adv --keyserver hkp://keyserver.ubuntu.com:80 --recv-keys 93C4A3FD7BB9C367
+    fi
```
holta, line 81: Can ` apt install ansible` be considered on Ubuntu 20.04, perhaps with an else clause here?

scripts/ansible
```diff
@@ -91,12 +92,11 @@ elif [ -f /etc/debian_version ]; then    # Includes Debian, Ubuntu & Raspbian
     #wget http://download.iiab.io/packages/ansible_2.4.2.0-1ppa~xenial_all.deb
     #apt -y --allow-downgrades install ./ansible_2.4.2.0-1ppa~xenial_all.deb
 
-    echo -e 'PPA source "deb http://ppa.launchpad.net/ansible/ansible/ubuntu bionic main"'
-    echo -e "successfully saved to /etc/apt/sources.list.d/iiab-ansible.list\n"
-
-    echo -e "IF *OTHER* ANSIBLE SOURCES APPEAR BELOW, PLEASE MANUALLY REMOVE THEM TO"
-    echo -e 'ENSURE ANSIBLE UPDATES CLEANLY: (then re-run this script to be sure!)\n'
-    grep '^deb .*ansible' /etc/apt/sources.list /etc/apt/sources.list.d/*.list | grep -v '^/etc/apt/sources.list.d/iiab-ansible.list:' || true    # Override bash -e (instead of aborting at 1st error)
+#    echo -e 'PPA source "deb http://ppa.launchpad.net/ansible/ansible/ubuntu bionic main"'
```
holta: Lines 95-99 help people who've ended up putting too many different Ansible sources into /etc/apt/sources.list.d/*.list — giving them important guidance on how to get back on track.

So can we move these 5 lines, so they live just prior to the `fi` on Line 77 of this PR?

scripts/ansible
jvonau: Why? The PPA and required key are not being used to install, ansible is being provide by Ubuntu, just setting up the PPA and key are skipped when 'focal' is matched. What do you propose?

scripts/ansible
jvonau: ok prior to 77 works, was looking to limit misleading information.

scripts/ansible
holta: I was assuming @georgejhunt manually ran something like `apt install ansible` on Ubuntu 20.04 pre-releases, and that scripts/ansible here could emulate his approach?

(But perhaps I have the details wrong...and there's a better way?)

jvonau: Think that might be the case and perhaps left out the ansible dependencies, only to install them later in roles/...

holta:
@jvonau smoke-tested this on RPi 4 with both OS's: Ubuntu 20.04 daily pre-release & Raspbian.

Thanks also to @georgejhunt who informed this PR as it evolved.

Issue #2297, opened 2020-03-20
lokole does not install on 64bit ubuntu on rpi


tim-moody:
ubuntu 19.10.1 64 bit on rpi 4 2G

TASK [lokole : DEFAULT: pip install opwen_email_client (Lokole, latest available version) from PyPI to /library/lokole/venv, if above vars both UNdefined] ***********************************
[...]

Issue #2296, opened 2020-03-20
gitea does not install on 64 bit ubuntu on rpi


tim-moody:
rpi 4 2G with ubuntu 19.10.1 64 bit

TASK [gitea : Fail if we detect unknown architecture] ****************************************************************************************************************************************
fatal: [127.0.0.1]: FAILED! => {"changed": false, "msg": "Could not find a binary for the CPU architecture \"aarch64\""}

holta:
@aidan-fitz can you look into this Gitea bug/issue, please if possible?  FYI:

1) Many people are now installing IIAB on Ubuntu 19.10.1 (and the daily pre-release builds of Ubuntu 20.04 LTS) on Raspberry Pi 4, as an alternative to the Raspbian OS &mdash; as Ubuntu on Raspberry Pi has proven to allow 32 simultaneous WiFi users &mdash; instead of just 18 as offered by the Raspbian OS: #823

2) Direct OS links here, including links to Ubuntu 20.04 LTS pre-releases: https://github.com/iiab/iiab/wiki/IIAB-Platforms

3) Perhaps even better: https://ubuntu.com/blog/how-to-install-ubuntu-with-the-new-raspberry-pi-imager

4) Or, if you need a direct VPN-like connection to one of the above OS's, please contact me directly!

tim-moody:
@aidan-fitz to be clear this is 64 bit and there is no aarch64 package.

tim-moody:
with kiwix we were able to run the 32 bit exec on 64 bit os

holta:
Aside: [Lokole](https://github.com/iiab/iiab/tree/master/roles/lokole) also does not install/run on Ubuntu on Raspberry Pi.

Hopefully @tim-moody (or someone!) can post the Lokole error message to a separate ticket in coming days.

holta:
Should be solved by @jvonau's PR #2301 -- thanks to everyone who can help with ongoing testing.

PR #2301, opened 2020-03-25
Aarch64

jvonau:
### Fixes Bug
#2297 #2296 
just need to double check if libjpeg-dev exists on debian/raspbian
U-20


jvonau:
both services respond get the splash pages under Ubuntu-20 aarch64

jvonau:
https://dl.gitea.io/gitea/1.11.3/ carries arm64, so that is an option also.

jvonau:
[lokole dependencies](https://github.com/ascoderu/opwen-webapp/blob/master/install.py#L482-L496)

PR #2302, opened 2020-03-26
Move 'rtc_id: ds3231' from vars/*.yml to vars/local_vars_*.yml

### LINKED ISSUE #2288: RTC routine 2 issues
The current code always installs support for an add-on RTC even if there is not one present, this is a result rtc_id being defined to a value other than 'none' in 0-init/defaults/main.yml. Feels like an unwise thing to do but so far has not caused issues sans for this warning in dmesg. 
> pi@box:~ $ dmesg | grep rtc
[    4.732719] rtc-ds1307: probe of 1-0068 failed with error -121

This is part of the 'dead wood' referred to in #2270. Might cause some issues should somebody make use of the gpio pins for a different purpose. Think rtc_id should really be defined in local_vars so the exact type of RTC added could be populated by the person who added the RTC so the model of the RTC is an exact match to what hardware was installed. 

The other issue would be the file to be modified is in a different location on Ubuntu, would be /boot/firmware/config.txt. Perhaps this RTC support can go away or should it get fixed?

holta:
And comment it out.

On @jvonau's request.  @georgejhunt agrees.

To resolve #2288

### DIFF
--- a/vars/default_vars.yml
+++ b/vars/default_vars.yml
@@ -43,6 +43,9 @@ download_timeout: 200
 default_language: en
 language_priority: en es fr
 
+# Real-time clock: set RTC chip family here.  Future auto-detection plausible?
+# rtc_id: ds3231
+
 # Set iiab_admin_user_install: False if you don't want iiab_admin_user & wheel
 # group auto-created in roles/iiab-admin/tasks/main.yml (hence disabling sudo-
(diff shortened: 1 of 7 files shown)


holta:
@Suyashdwivedi please review...and offer suggestions towards further refining!

PR #2305, opened 2020-03-26
python2 package name exists only on Ubuntu 20 for now.

jvonau:
'python' renamed to python2 upstream as python2.7 is going away at some point but unchanged on Ubuntu 18

### DIFF
--- a/roles/kalite/tasks/install.yml
+++ b/roles/kalite/tasks/install.yml
@@ -15,6 +15,7 @@
   package: 
     name: python2
     state: present
+  when: is_ubuntu_20 | bool
 
 - name: Use pip to install KA Lite static to {{ kalite_venv }}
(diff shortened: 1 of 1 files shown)


roles/kalite/tasks/install.yml
```diff
@@ -15,6 +15,7 @@
   package: 
     name: python2
     state: present
+  when: is_ubuntu_20 | bool
```
holta, line 18: We might be stuck with KA Lite a lot longer that we all imagined.  So can we please change:

```
when: is_ubuntu_20 | bool
```

...to something like:

```
when: is_ubuntu and not is_ubuntu_18 and not is_ubuntu_19
```

holta:
Alternative PR #2307 was merged.

PR #2309, opened 2020-03-30
NGINX: fancyindex instead of autoindex, for http://box/usb (for larger font on smartphones etc)

holta:
Proposed by @jvonau & tested by @shanti-bhardwa

### DIFF
--- a/roles/nginx/templates/iiab.conf.j2
+++ b/roles/nginx/templates/iiab.conf.j2
@@ -4,11 +4,11 @@ location / {
 
 location /usb {
     alias /library/www/html/local_content/;
-    autoindex on;
+    fancyindex on;    # autoindex on;
 }
 
 location /local_content/ {
-    autoindex on;
+    fancyindex on;    # autoindex on;
 }
 
 location /modules/ {


holta:
BACKGROUND:

1) `/etc/nginx/conf.d/iiab.conf` is no longer invoked by...
https://github.com/iiab/iiab/blob/master/roles/nginx/tasks/install.yml#L46-L50

   Now it's invoked a bit later in [Stage 4-server-options](https://github.com/iiab/iiab/tree/master/roles/4-server-options) by...
https://github.com/iiab/iiab/blob/master/roles/www_options/tasks/main.yml#L24

2) The best way to run this manually is probably `./runrole www_options` (to invoke [roles/nginx/tasks/homepage.yml](https://github.com/iiab/iiab/blob/master/roles/nginx/tasks/homepage.yml) thereby reinstalling `/etc/nginx/conf.d/iiab.conf`)

holta:
UX mechanics & suggestion thanks to @shanti-bhardwa:

> here is some background on what the the change is about. https://www.nginx.com/resources/wiki/modules/fancy_index/ and http://nginx.org/en/docs/http/ngx_http_autoindex_module.html so there is scope for more customisation

PR #2316, opened 2020-04-01
Ivan Savov's workaround [so Kolibri exercises work]

tim-moody:
temporary fix for  exercises


holta:
@tim-moody @ivanistheone any guesses as to how "temporary" this workaround should be...weeks...or months?

Anyway, fantastic news!

Background...upstream discussion is here: https://github.com/learningequality/kolibri-exercise-perseus-plugin/issues/213#issuecomment-604748222

holta:
FYI @tim-moody had already smoke tested and tested function, on an already-installed machine.

ivanistheone:
@holta I can't give a precise estimate on how long it will take, but probably O(months). It's a tricky issue since there is some dynamic loading of resources so it's not a trivial fix (MathJax components are lazily loaded after the initial load).

Issue #2312, opened 2020-03-30
KA Lite fails to install on Ubuntu 20.04 pre-release (arm64): "You are running Setuptools on Python 2, which is no longer supported" "ensure you are installing Setuptools using pip 9.x or later or pin to `setuptools<45`"


holta:
Thanks to @shanti-bhardwa & @jvonau for reporting this.

Issue appears to be: https://github.com/pypa/setuptools/issues/1458

Install was on Ubuntu 20.04 LTS (arm64 pre-release from http://cdimage.ubuntu.com/ubuntu-server/daily-preinstalled/current/) onto Raspberry Pi 4:

> TASK [kalite : Use pip to install KA Lite static to /usr/local/kalite/venv] **************************************************************************************************************************
[...]

PR #2321, opened 2020-04-02
wheel issue [for KA Lite, relating to python2 etc]

jvonau:


holta:
Hopefully this solves #2312 i.e. refining PR #2314, whether on a temporary or long-term basis!

holta:
Thanks to @jvonau who tested this on a very recent Ubuntu 20.04 pre-release, presumably on RPi 4.

holta:
http://minutes.iiab.io call consensus:

1) Let's not push this for a few days (48+ hrs) ...watching `apt update` for a new release of python-pip-whl

2) This could be reversed later in April if absolutely necssary.

holta:
This PR was tested to successfully/cleanly/freshly install KA Lite on Raspbian Lite on RPi 4.

Should it now be merged...or is other/specific testing wise?

jvonau:
> This PR was tested to successfully/cleanly/freshly install KA Lite on Raspbian Lite on RPi 4.
> 
Which OSs? the fix is intended for U-20 but should do no harm on other OSs. 

> Should it now be merged...or is other/specific testing wise?

I would prefer not to carry the wheel file in git but rather retrieve the file from a know web url using ansible's get_url

holta:
FWIW the file in question is now posted in both places:

- http://d.iiab.io/packages/ipaddress-0.0.0-py2.py3-none-any.whl
- https://github.com/jvonau/iiab/blob/wheel_issue/roles/kalite/files/ipaddress-0.0.0-py2.py3-none-any.whl -> click "Download"

holta:
Context:

> Why not lock in python-setuptools to its current version number, if that's the underlying goal?

> because setuptools greater than 44 will never exist on python2 that is where the devs choose to break backwards compatibility.
>
> and if a new version of python-setuptools 45 comes out that might supply the missing wheel as that is where the dependence lies..
>
> the goal is to be a supplier of last resort..
>
> we know what the fix is but upstream has to catchup fixing it, which apt package that carries the fix is unknown at this time, or continue to break python2..

PR #2322, opened 2020-04-04
pastebinit bugs: try dpaste.com instead (for iiab-diagnostics command)

holta:
Tested on Raspbian Lite on RPi 4.

If pastebinit fails to work with http://dpaste.com in future, remember to try `pastebinit -l` to list other possible pastebin site URLs &mdash; many of which no longer work &mdash; but that's the place to start!

PR #2330, opened 2020-04-08
Enhance iiab-diagnostics command

holta:
For now this adds the last 100 lines of:
- /var/log/messages
- /var/log/syslog

And provides the complete output (I think!) of Linux's `dmesg` command.

Builds on PR #2322


holta:
Many improvements added after discussion with others on our IIAB weekly call (http://minutes.iiab.io).

PR #2331, opened 2020-04-08
add iiab-admin to netdev group so that gui can control networking

georgejhunt:
Found to be necessary in raspberry desktop, when clicking on network->Wireless and "Wired Network Settings
Smoke tested

### DIFF
--- a/roles/iiab-admin/tasks/admin-user.yml
+++ b/roles/iiab-admin/tasks/admin-user.yml
@@ -19,7 +19,7 @@
 - name: 'Add user {{ iiab_admin_user }} to groups: wheel, sudo'
   user:
     name: "{{ iiab_admin_user }}"
-    groups: wheel,sudo
+    groups: wheel,sudo,netdev
 
 - name: Edit the sudoers file -- first make it editable
   file:


holta:
@jvonau take a look?

jvonau:
I take it you are logging into the desktop as the iiab-admin user and not the pi user?

georgejhunt:
Somewhere I've seen (and I think I used) a prep script which makes
iiab-admin the default user for the gui. Do you think that's a bad idea?

holta:
> Somewhere I've seen (and I think I used) a prep script which makes iiab-admin the default user for the gui. Do you think that's a bad idea?

I don't think that's a good idea as Raspbian Desktop software is sometimes hardcoded to user 'pi'.

Also users who later apply a screensaver/lock screen expect to be able to log back in as user 'pi'.

In both cases, fighting the existing Raspbian culture seems like...a losing battle that we cannot possibly win.

georgejhunt:
Ok I see your point. There are just so many issues cropping up as I try to
automate the production of images to use with rpi-imager.

georgejhunt:
Initially we had been adding developer keys to iiab-admin.  But more
recently, I think Tim and I have just been putting them in root.

Do you have an opinion on this?

holta:
> Initially we had been adding developer keys to iiab-admin. But more recently, I think Tim and I have just been putting them in root. Do you have an opinion on this?

'iiab-admin' has the potential to be safer on some OS's, in future possibly?

Though at the moment 'root' seems like an acceptable compromise, given iiab-admin already has some root privileges, in order to alert operators/implementers who are endangering themselves by not changing published passwords.

tim-moody:
I believe putting in root is safer and will continue to do so.

georgejhunt:
I'll put developer keys in root

PR #2334, opened 2020-04-10
remove any possible file before creating regions symlink [OSM]

tim-moody:

Issue #2208, opened 2020-01-27
Proposed OSM Vector Maps Data Structures


tim-moody:
###	External, global catalog
* Describes download sources
* Contains a list of all map sets available for download
* Key is map set id
* Question: map sets have language as a prefix to their names, such as en-osm-omt_spanish_speaking_regions. Is there also an es-osm-omt_spanish_speaking_regions or other set? Most of the maps are multi-lingual.

### Locally installed catalog
* Same format, but serves as list of installed map sets
* Has attributes for map sets even that may no longer be in global catalog. 
* Key is map set id

###	Version index
* Maps generic name to version instance  
* Key is perma_ref.
* Question: is there more than one perma_ref per region or is perma_ref actually region

###	Map set meta data
* It would be nice to have meta data for a map set version stored in that set’s directory. This would allow regeneration of Locally installed catalog and Version index. 
* Same format as one item in Locally installed catalog.

tim-moody:
## Examples from Current Data

### External, global catalog
"en-osm-omt_spanish_speaking_regions_2017-07-03_v0.23" : {
      "region" : "spanish_speaking_regions",
      "center_lat": "-8.0",
      "center_lon": "-70.0",
      "date": "2017-07-03",
      "east": "-18.0",
      "north": "33.00",
      "osm_size": "7888969728",
      "perma_ref": "en-osm-omt_spanish_speaking_regions",
      "sat_is_regional": "False",
      "sat_size": "976416768",
      "sat_url": "https://archive.org/download/satellite_z0-z9_v3.mbtiles/satellite_z0-z9_v3.mbtiles",
      "seq": 25,
      "size": "8699784000",
      "south": "-84.0",
      "title": "Spanish-speaking Regions",
      "url": "https://archive.org/download/en-osm-omt_spanish_speaking_regions_2017-07-03_v0.23/en-osm-omt_spanish_speaking_regions_2017-07-03_v0.23.zip",
      "version": "v0.23",
      "west": "-179.00",
      "zoom": "3"
    }
### Version Index

{
  "en-osm-omt_world": {
    "file_name": "en-osm-omt_world_2017-07-03_v0.23",
    "menu_item": "en-osm-omt_world",
    "size": "55752568832",
    "date": "2017-07-03",
    "region": "world",
    "language": "en"
  }

holta:
Added to OSM Ideas Checklist @ #877 so we keep tabs on this!

holta:
[...]

PR #2336, opened 2020-04-11
copy won't try to overwrite another file if present [for OSM install]

jvonau:
### Fixes Bug
issue of replacing a later installed file with an earlier file. #2197
use copy in place of file, removes bogus fix, allowing later installed file to prevail.
unbuntu-20

### DIFF
--- a/roles/osm-vector-maps/tasks/install.yml
+++ b/roles/osm-vector-maps/tasks/install.yml
@@ -11,17 +11,11 @@
     url: "{{ iiab_map_url }}/assets/regions.json"    # http://download.iiab.io/content/OSM/vector-tiles/maplist/hidden
     dest: "{{ vector_map_path }}/maplist/assets/"    # /library/www/osm-vector-maps
 
-# handle case where a dummy file is already here and would break the symlink step that follows
-- name: Remove {{ doc_root }}/common/assets/regions.json
-  file:
-    dest: "{{ doc_root }}/common/assets/regions.json"    # /library/www/html
-    state: absent
-
-- name: Symlink catalog {{ doc_root }}/common/assets/regions.json -> {{ vector_map_path }}/maplist/assets/regions.json
-  file:
+- name: Copy catalog {{ vector_map_path }}/maplist/assets/regions.json -> {{ doc_root }}/common/assets/regions.json
+  copy:
     src: "{{ vector_map_path }}/maplist/assets/regions.json"    # /library/www/osm-vector-maps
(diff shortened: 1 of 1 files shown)


holta:
Clarif...that this is a possible alternative to the recently merged:

PR #2334 "remove any possible file before creating regions symlink"

holta:
Tangentially related:

#2208 "Proposed OSM Vector Maps Data Structures"

tim-moody:
what problem does this solve?

georgejhunt:
I think we identified on the call this morning that there was a
problem on the admin console wherein the label "Spanish Speaking
Region" caused the layout engine to put the remaining regions to the
bottom (when the window was not wide enough). I agreed to change the
label to "Spanish Speakers" to fix the problem.

tim-moody:
other than pertaining to osm, what do these two issues have to do with one another? or put another way, how does #2336 solve the issue of the Spanish region having too long a name for the adm console layout?

georgejhunt:
I cannot explain why Adam/Jerry  took issue with this #2336. But when I
tried to pin down what observation led them to create this issue, the
observation/description of the problem could be created on my end by making
the window narrower.

holta:
@georgejhunt & All: what's the best course of action for this PR #2336 as we seek to finalize a Release Candidate for [IIAB 7.1](https://github.com/iiab/iiab/wiki/IIAB-7.1-Release-Notes) in coming days?
- merge this PR?
- close this PR?
- defer this PR to IIAB 7.2?

georgejhunt:
close it. or defer it to 7.2 since I do not understand what problem it is
trying to solve, but I don't want to piss people off.

tim-moody:
How about test this PR? First, articulate the problem that this PR solves. Second, test the branch behind this PR. If the problem is solved merge it.
This should not be a popularity contest or vaguely trying to decide if one developer is better than another.

jvonau:
> other than pertaining to osm, what do these two issues have to do with one another? or put another way, how does #2336 solve the issue of the Spanish region having too long a name for the adm console layout?

#2334 solves a problem introduced in admin-console where regions.json is file(without the 'Spanish region problem) and not a symlink. Blindly deleting the regions.json file and replacing it with the web copy and symlink just re-introduces the "Spanish region having too long a name" requiring a re-install of the admin-console to apply the neutered copy of regions.json to omit the problem map title.

This PR would skip the download and install should regions,json already exists.

holta:
@jvonau confirms this PR was superseded by PR #2350.

Issue #2318, opened 2020-04-01
Is Kolibri shutting down cleanly when IIAB powers off?


holta:
Or is Kolibri possibly/painfully delaying IIAB shutdown?

Thanks @tim-moody for noticing this issue...being discussed in both places here:
- http://community.rachelfriends.org/t/new-image-shutdown-period/1687
- https://community.learningequality.org/t/stop-job-on-raspberry-pi-when-shutting-down/1482

jvonau:
I saw the discussion also, we went with the upstream timeouts which are under review. I suspect TimeoutStopSec should be removed/altered, the stock timeout is 90 seconds, if anything it should be increased to allow a clean shutdown of the db backend.

jvonau:
Might be fixed on new installs now given the change in upstream change, if the PPA was used then it would be a apt upgrade fix but with the deb location fixed in the install.yml file ./runrole --reinstall kolibri is required.

holta:
Related:

#1832 "Tracking the best systemd unit file(s) for Kolibri?"

holta:
Recommendation from our http://minutes.iiab.io call today &mdash; thanks to @jvonau:

- Revert (or comment out) IIAB's customized timeouts for Kolibri
- Go with systemd's default of 90 seconds
- In future we can also consider more than 90 seconds if large Kolibri Channels require that

holta:
Can this be closed now, now that PR #2343 is merged?

Whereas the other major Kolibri issue remaining, as IIAB 7.1's release is finalized in coming days (this last issue is ongoing for 15 months) might have to be deferred, to the IIAB 7.2 release cycle:

#1532 "http://box/kolibri causes browser to infinitely loop, preventing login (works initially...server reboot does not help...happens with all client OS's & all server OS's)"

jvonau:
think we are good, haven't seen any long delays (re)booting.

Issue #1832, opened 2019-07-07
Tracking the best systemd unit file(s) for Kolibri?


holta:
@benjaoming wrote @ https://github.com/iiab/iiab/pull/1828#issuecomment-509012372:
> In light of this issue and the fact that IIAB is shipping its own systemd service, I would recommend that you open an issue to have Kolibri installed through the Debian package instead? Interactive with dpkg,

Just so I understand, are you suggesting something like either of the following?
https://docs.ansible.com/ansible/latest/modules/apt_module.html
https://docs.ansible.com/ansible/latest/modules/dpkg_selections_module.html

IIAB has had endless trouble installing certain .deb packages directly in the case of Calibre, and likewise the lastest PPA's for Ansible, but perhaps the process is becoming more sane in recent years ??  Not that pip/npm/yarn are perfect by any stretch of the imagination, when normal `apt` installs are not possible :-)

> whether as an uninstall, reinstall or fresh install will manage the systemd job, starting, stopping and reloading it when necessary.

If https://github.com/iiab/iiab/blob/master/roles/kolibri/templates/kolibri.service.j2 is not the latest we all definitely need to know yes!

Conversely where should we find the latest/recommended systemd unit file Kolibri that you recommend in general?  Certainly IIAB has to manage diverse such system unit files, which sometimes need delicate modifications on slow-booting / older machine etc, especially when many service are running, that @jvonau and others have greatly helped with~

Refs:
#1591 "Better Kolibri performance with static file serve and wsgi"
#1646 "Migrate IIAB to kolibri-server for RPi performance"

holta:
Quick Background on the narrow issue of systemd unit files: even "official" unit files e.g. for major packages like dnsmasq very often contain showstopper bugs, so IIAB has learned that we have to be extremely judicious in choosing/refining the best strategies here!

(Accepting system unit files from upstream without extensive testing+review is definitely not an effective strategy in the end, despite our best wishes that has NOT worked out ;-)

jvonau:
A [quick search of kolibri](https://github.com/learningequality/kolibri/search?q=kolibri.service&unscoped_q=kolibri.service) turns up no unit file, so we use ours. Closest [upstream kolibri-server unit file](https://github.com/learningequality/kolibri-server/blob/master/debian/kolibri-server.service) unit file. We do not currently use kolibri-server as the webserver is not apache but nginx.

benjaoming:
[...]

PR #2343, opened 2020-04-18
Comment out both kolibri.service timeout settings

holta:
Implements @jvonau's suggestion at https://github.com/iiab/iiab/issues/2318#issuecomment-614727305 "Is Kolibri shutting down cleanly when IIAB powers off?"

Related:

#1832 "Tracking the best systemd unit file(s) for Kolibri?"


holta:
I'll merge this in coming hours if there are no objections.

Issue #2347, opened 2020-04-19
php7.4-stem support for Ubuntu 20.04 LTS?  On which CPU architectures if so?


holta:
With PHP 7.4 becoming mainline as part of Ubuntu 20.04 starting this coming Thursday, should we consider adding support for some subset of {aarch64, rpi, x64} ?

( e.g. to enable WikiHow search in Spanish, via http://d.iiab.io/packages/ )

Background:
- https://github.com/iiab/iiab/blob/master/roles/www_base/tasks/main.yml#L9-L10
- https://github.com/iiab/iiab/blob/master/roles/www_base/tasks/php-stem.yml

Most recently @tim-moody built these at http://d.iiab.io/packages/ :
- php7.3-stem.aarch64.tar
- php7.3-stem.rpi.tar
- php7.3-stem.x64.tar

Refs: #2123, PR #2198, PR #2269, PR #2294, #2345

tim-moody:
After #2345 this is a lower priority, but I think to start with php7.4-stem.aarch64.tar and then php7.4-stem.x64.tar once 20.04 is released.

tim-moody:
php7.4-stem.aarch64.tar  and php7.4-stem.x64.tar uploaded to d.iiab.io

needs testing

holta:
Good enough for now, with PR #2349 now merged.

PR #2345, opened 2020-04-18
Med vars & php-stem error handling

tim-moody:

PR #2198, opened 2020-01-25
Php stem

tim-moody:
#2123

Issue #2123, opened 2020-01-09
Some IIAB installs contain a partial/old version of PHP at /etc/php/7.0 etc [on Raspbian only...due to WikiHow's php-stem?]


holta:
Thanks @jvonau for finding this:

```
root@box:~# cat /etc/issue.net
Raspbian GNU/Linux 10

root@box:~# apt -a list php
Listing... Done
php/stable 2:7.3+69 all

root@box:~# locate php | grep 7\\.0
/etc/php/7.0
/etc/php/7.0/apache2
/etc/php/7.0/mods-available
/etc/php/7.0/apache2/conf.d
/etc/php/7.0/apache2/conf.d/20-stem.ini
/etc/php/7.0/mods-available/stem.ini

root@box:~# du -hsc /etc/php/*
20K     /etc/php/7.0
468K    /etc/php/7.3
488K    total
```

jvonau:
suspect is the contents of [php-stem.rpi.tar](https://github.com/iiab/iiab/blob/master/roles/httpd/tasks/php-stem.yml)

holta:
> suspect is the contents of [php-stem.rpi.tar](https://github.com/iiab/iiab/blob/master/roles/httpd/tasks/php-stem.yml)

If so, that would be caused by...
https://github.com/iiab/iiab/blob/master/roles/httpd/tasks/main.yml which calls...
https://github.com/iiab/iiab/blob/master/roles/httpd/tasks/php-stem.yml
http://download.iiab.io/packages/php-stem.rpi.tar
http://download.iiab.io/packages/php-stem.x64.tar <- "no longer relevant, as this one's only installed for Debian 9"

jvonau:
also creates /usr/lib/php/20151012/stem.so

holta:
@tim-moody is your above php-stem patch to fix WikiHow module's search still working?  (Or is it obsolete?)

holta:
June/July 2018 refs:

#829 "Is Spanish WikiHow's search fixable? (works on RACHEL site!)"
PR #925 "php-stem fixes search @ http://box/modules/es-wikihow"

holta:
Call proposed to comment out line 8 here for now:
https://github.com/iiab/iiab/blob/master/roles/httpd/tasks/main.yml#L8

As no modern OS's are using PHP 7.0...

(Even if stem.so might/maybe work in Raspbian new dir /usr/lib/php/20180731 which replaces the old /usr/lib/php/20151012 ?)

[...]

PR #2349, opened 2020-04-19
Php 386 [& php-stem with PHP 7.4+ prep?]

jvonau:
### Fixes Bug
avoids using ignore_errors: yes
better grouping, moves the creation of the stem.ini link to be with the installation of the tar file 
lightly tested on U-20

### DIFF
--- a/roles/nginx/tasks/install.yml
+++ b/roles/nginx/tasks/install.yml
@@ -49,18 +49,6 @@
     msg: roles/nginx/tasks/homepage.yml will run LATER (invoked by roles/www_options/tasks/main.yml) SO THAT NGINX CAN REDIRECT http://box TO http://box{{ iiab_home_url }} (based on var iiab_home_url)
 # - include_tasks: roles/nginx/tasks/homepage.yml
 
-# php-stem extension installed by roles/www_base/tasks/php-stem.yml
-# here it is linked to php-fpm
-# we skip any errors which probably due to the lack of the .so for the current platform
-- name: Symlink /etc/php/{{ php_version }}/fpm/conf.d/20-stem.ini -> /etc/php/{{ php_version }}/mods-available/stem.ini
-  file:
-    src: "/etc/php/{{ php_version }}/mods-available/stem.ini"
-    path: "/etc/php/{{ php_version }}/fpm/conf.d/20-stem.ini"
-    state: link
-  ignore_errors: yes
-  #when: php_version != 7.4
-  #when: nginx_enabled | bool
-
 - name: Restart php{{ php_version }}-fpm systemd service
   systemd:
     name: "php{{ php_version }}-fpm"
--- a/roles/www_base/tasks/main.yml
+++ b/roles/www_base/tasks/main.yml
@@ -3,11 +3,12 @@
 #
 # (Don't take either name too literally!)
 
+- name: Using html.yml
+  include_tasks: html.yml
 
-- include_tasks: html.yml
-
-- include_tasks: php-stem.yml
-  when: php_version < 7.4
+- name: Using php-stem.yml
+  include_tasks: php-stem.yml
+  when: php_version < 7.4 and not ansible_architecture == "i386"
 
 - name: Create dir {{ doc_root }}/home -- if you customized var iiab_home_url e.g. in /etc/iiab/local_vars.yml, that dir is created later -- by www_options/tasks/main.yml
   file:
--- a/roles/www_base/tasks/main.yml
+++ b/roles/www_base/tasks/main.yml
@@ -8,7 +8,6 @@
 
 - name: Using php-stem.yml
   include_tasks: php-stem.yml
-  when: php_version < 7.4 and not ansible_architecture == "i386"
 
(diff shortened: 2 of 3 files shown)


tim-moody:
ignore_errors has been used 20 to 30 times, so not sure why this one is offensive. I like moving the link creation to the same place where the .so is put in place, but the current logic of trying to have a when that catches all exceptions is less flexible and more fragile. It would not work for the case where we had .so for php-7.4 only for one arch.

jvonau:
Once the .so files are available for 7.4 the version test I trust would be removed in main.yml and in it's place suitable stanzas to exclude arches that are not supported would be inserted in php-stem.yml to suppress the creation of the symlink.

What we gain is current users who are now excluded because 7.4 would only need to git pull runrole www_base to add the .so file and create the needed link as opposed to having to iiab-install --reinstall with having the split locations of unpacking the tar file in php-stem and creating the link in nginx. 

I'll rework this PR to a final layout for right now and future availablility.

tim-moody:
> Once the .so files are available for 7.4 the version test I trust would be removed in main.yml 

no. it will be changed to <7.5
but this will still not handle the case where there is an so for x64 and not for aarch64 or some other platform that someone decides to try.

> What we gain is current users who are now excluded because 7.4 would only need to git pull runrole www_base to add the .so file and create the needed link as opposed to having to iiab-install --reinstall with having the split locations of unpacking the tar file in php-stem and creating the link in nginx.

yes. I agreed to putting the link with the ini and so.

jvonau:
> > Once the .so files are available for 7.4 the version test I trust would be removed in main.yml
> 
> no. it will be changed to <7.5
> but this will still not handle the case where there is an so for x64 and not for aarch64 or some other platform that someone decides to try.

See the proposed  "Uncomment and alter to suit current supported arches"

tim-moody:
there are still 7.2s out there

jvonau:
the 3 supported arches?

tim-moody:
only x64 and armv7l I think

jvonau:
I went with x64 only as ubuntu-18.yml lists 7.2 and is the only 7.2 available on d.iiab.io

holta:
@tim-moody what's your summary recommendation, towards merging this PR or something like it, for [IIAB 7.1](https://github.com/iiab/iiab/wiki/IIAB-7.1-Release-Notes) in coming days?

Related:

#829 "Is Spanish WikiHow's search fixable? (works on RACHEL site!)"
#2123 "Some IIAB installs contain a partial/old version of PHP at /etc/php/7.0 etc [on Raspbian only...due to WikiHow's php-stem?]"
#2198 "Php stem"
#2345 "Med vars & php-stem error handling"
#2347 "php7.4-stem support for Ubuntu 20.04 LTS? On which CPU architectures if so?"

holta:
@tim-moody can you review this in the coming 24h if possible?

tim-moody:
with the current code there is no need to make changes if we get another version of phpx.x-stem, but in order for someone who has an earlier version to upgrade two roles will have to be run. with this PR only one role will have to be run to upgrade, but it will be necessary to change the code each time there is a new version of phpx.x-stem. it's pretty much personal preference. my views were expressed above and I no longer wish to discuss this. please do as you see fit.

holta:
@georgejhunt got time to look at this quickly today?

There's never a magical answer to code maintainability, but let's pick a way forward here together if possible~

holta:
See https://github.com/iiab/iiab/issues/2347#issuecomment-619431189 thanks to @tim-moody:

> php7.4-stem.aarch64.tar and php7.4-stem.x64.tar uploaded to d.iiab.io
>
> needs testing

holta:
We're wanting http://d.iiab.io/packages/php7.4-stem.aarch64.tar installed on Ubuntu 20.04 e.g. on RPi (that contains PHP 7.4).

However current master does not work:

> 2020-04-28 16:06:14,388 p=12847 u=root n=ansible | TASK [nginx : Symlink /etc/php/7.4/fpm/conf.d/20-stem.ini -> /etc/php/7.4/mods-available/stem.ini] *********************************************************************************************************************************************
2020-04-28 16:06:15,050 p=12847 u=root n=ansible | fatal: [127.0.0.1]: FAILED! => {"changed": false, "msg": "src file does not exist, use \"force=yes\" if you really want to create the link: /etc/php/7.4/mods-available/stem.ini", "path": "/etc/php/7.4/fpm/conf.d/20-stem.ini", "src": "/etc/php/7.4/mods-available/stem.ini"}

I'd really like to solve this today so we can move on and release [IIAB 7.1](https://github.com/iiab/iiab/wiki/IIAB-7.1-Release-Notes) with search working for WikiHow across OS's if possible, by merging this PR or another compromise PR that moves us forward.

@georgejhunt & All can you possibly help?

tim-moody:
so merge this and see if you have better luck.
#2345 fixed it in roles/nginx/tasks/install.yml but not in  roles/www_base/tasks/main.yml

holta:
Thanks everyone for the hard work + patience over these past 9 days!

PR #2350, opened 2020-04-19
centralize map catalog for use by osm and admin console

tim-moody:
This is what we should have done in the first place, centralize the maps catalog, as was done for kiwix and rachel, and roles that use it create their own symlinks. (adm cons will need to create one as well or rather change the source for the one it already creates.)

### DIFF
--- a/roles/osm-vector-maps/tasks/install.yml
+++ b/roles/osm-vector-maps/tasks/install.yml
@@ -6,21 +6,15 @@
     group: "{{ apache_user }}"
     mode: '0755'
 
-- name: Download map catalog {{ iiab_map_url }}/assets/regions.json to {{ vector_map_path }}/maplist/assets/
+- name: Download map catalog {{ iiab_map_url }}/assets/regions.json to {{ iiab_etc_path }}
   get_url:
     url: "{{ iiab_map_url }}/assets/regions.json"    # http://download.iiab.io/content/OSM/vector-tiles/maplist/hidden
-    dest: "{{ vector_map_path }}/maplist/assets/"    # /library/www/osm-vector-maps
+    dest: "{{ iiab_etc_path }}"
 
-# handle case where a dummy file is already here and would break the symlink step that follows
-- name: Remove {{ doc_root }}/common/assets/regions.json
+- name: Symlink map catalog {{ vector_map_path }}/maplist/assets/regions.json -> {{ iiab_etc_path }}
   file:
-    dest: "{{ doc_root }}/common/assets/regions.json"    # /library/www/html
-    state: absent
-
-- name: Symlink catalog {{ doc_root }}/common/assets/regions.json -> {{ vector_map_path }}/maplist/assets/regions.json
-  file:
-    src: "{{ vector_map_path }}/maplist/assets/regions.json"    # /library/www/osm-vector-maps
-    dest: "{{ doc_root }}/common/assets/regions.json"    # /library/www/html
+    src: "{{ iiab_etc_path }}/regions.json"
+    dest: "{{ vector_map_path }}/maplist/assets/regions.json"
     state: link
 
 - name: Download the JavaScript bundle with OpenLayers (main.js) into {{ vector_map_path }}/maplist/, for test page http://box/maps/maplist


tim-moody:
works on vm

holta:
@georgejhunt @jvonau please review today/tmrw if you can!

tim-moody:
@georgejhunt I think this is the direction the map reorg is taking.

jvonau:
in 301 Create a link to osm catalog in /common/assets is done unconditionally perhaps that could be  in "iiab" addition to the osm link? both reference /etc/iiab/

tim-moody:
I don't think osm uses that link, only adm cons.

georgejhunt:
Yes let's put regions.json in /etc/iiab. Should have been there from the beginning. And maps.catalog will be there

holta:
Should this be merged now?

tim-moody:
should osm copy a blank catalog to /etc/iiab if not enabled, so that adm cons doesn't have to be run to disable?

jvonau:
Wouldn't turning off OSM just entail not having nginx display the webpage? The contents of the webpage should not really matter.

jvonau:
FWIW I sooner see vector-map-idx.json installed here, it's nothing more that a file with {} present. Create the docroot symlink here also, that would leave admin-console to only call /usr/bin/iiab-update-map when enabled.

tim-moody:
> Wouldn't turning off OSM just entail not having nginx display the webpage? The contents of the webpage should not really matter.

At the moment adm cons still has the menu link even if disabled as a sort of a promotion, but I'd like the list to be empty.

tim-moody:
>FWIW I sooner see vector-map-idx.json installed here

that file is owned by admin console. it has no use in osm

holta:
Other Open OSM Tkts...to help us look over the horizon to IIAB 7.2 & 7.3:

- #2197 OSM Install Fails and gives the following Errors: "refusing to convert from file to symlink for /library/www/html/common/assets/regions.json" & "/usr/bin/iiab-update-map: not found"
- #2208 Proposed OSM Vector Maps Data Structures
- PR #2336 Copy won't try to overwrite another file if present [for OSM install]

jvonau:
At least /etc/iiab/regions.json is treated as a file here and in 301, with the links pointing to the one master location. Work flow as I understand it.
iiab
*_install True would populate /etc/iiab/regions.json 
*_enabled True creates nginx's in the config file
runrole --reinstall osm.... would re-create the web version same as now.

admin-console ./install
*_enabled True  None of the dummy files would be installed as was before 301 no change there
*_enabled False  dummy files would be installed overwrites /etc/iiab/regions.json  no change there
Create a link to osm catalog in /common/assets
*_enabled True run iiab-update-map

Lets give it a go if that is the work flow wanted.

holta:
> Lets give it a go if that is the work flow wanted.

@georgejhunt can you weigh in?

Then I'd like everyone's help quickly revising our public-facing "IIAB Maps" doc (https://github.com/iiab/iiab/wiki/IIAB-Maps) for IIAB 7.1 in the coming 10 days!

georgejhunt:
I think this PR can go forward, as I think it just moves the authoritative
file from /library/www/osm-vector-maps/maplist/assets/ to /etc/iiab/.

But I'm not sure of the case where admin-console is installed but not
osm_vector_maps. Is that a valuable option?

I think it would be easier to move forward if osm_vector_maps was always
installed. Or there may be a simple solution that I don't see.

tim-moody:
> But I'm not sure of the case where admin-console is installed but not
> osm_vector_maps. Is that a valuable option?

yes, because all of the medical images are currently like that. could change, but needs to be possible.

georgejhunt:
On Mon, Apr 20, 2020 at 9:03 AM A Holt <notifications@github.com> wrote:
>
> Lets give it a go if that is the work flow wanted.
>
> @georgejhunt can you weigh in?
>
> Then I'd like everyone's help quickly revising our public-facing "IIAB Maps" doc (https://github.com/iiab/iiab/wiki/IIAB-Maps) for IIAB 7.1 in the coming 10 days!
>
I'd like to remove the whole section relating to upgrades of maps in
https://github.com/iiab/iiab/wiki/IIAB-Maps. Looking towards the
future, 7.2, the partition between program and data will change. I
have not tested whether 7.1 and 7.2 maps can exist along side each
other (and the operation of http://box/maps seems problematic). And it
does not make sense to duplicate the data.

holta:
Please continue to test after this is merged now!

holta:
FYI this PR supersedes PR #2336 "copy won't try to overwrite another file if present [for OSM install]" which @jvonau is now closing.

holta:
Let's test this now that https://github.com/iiab/iiab-admin-console/pull/301 ("move regions.json to /etc/iiab") is merged, Thanks @tim-moody !

PR #2355, opened 2020-04-22
drop kolibri_url_without_slash

jvonau:
less confusing to use one variable that would set the user facing uri


holta:
On the one hand I sympathize, but on the other hand the norm across most all IIAB variables is for {{ APP_url }} to include the trailing slash, so this make it more risky/confusing...for me :/

tim-moody:
in my view including the trailing slash is a mistake in all contexts. it will often mean that /kolibri/ works, but /kolibri doesn't. the apache solution was to redirect /xxx to /xxx/. nginx has directives that always try both xxx and xxx/, so again I think the standard should be /xxx not /xxx/. Of course this is all purely theoretical.

PR #2354, opened 2020-04-22
nginx requires trailing slash [for Kolibri URL] to function properly

jvonau:
### Fixes Bug
/kolibri not working and looping login

PR #2356, opened 2020-04-22
Softcoding kolibri-nginx.conf

holta:
Builds on PR #2354, softcoding kolibri-nginx.conf.j2

This is an alternative to PR #2355 which (unintentionally!) breaks the norm used across most every IIAB playbook.

Background: variables of the form {{ APP_url }} contain the trailing slash according to the current norm to avoid confusion.

If that norm needs change in future, we should change it across all IIAB playbooks.


holta:
Recap: a PR like #2355 is fine, if it (or something like it) is done consistently across all (or most all) IIAB playbooks.

But inconsistently implementing {{ APP_url }} across playbooks would lead to a lot of confusion among new devs especially :-(

benjaoming:
Noooo, I was about to make this change earlier today :) Would have been my first IIAB contrib. Alas!

Issue #2367, opened 2020-04-27
vnStat playbook cannot be rerun on a VM w/o LAN? Is iiab-from-console.yml part of the issue?


holta:
Thanks @jvonau who is looking into this:

On VM 191-u20-srv-BIG-0423 I clicked Admin Console (http://10.8.0.26/admin) > Configure > Install Configured Options and 2.5 min later Admin Console > Utilities > Display Job Status showed:

> TASK [vnstat : Create database for LAN to collect vnStat data if not appliance config] ***
fatal: [127.0.0.1]: FAILED! => {"changed": true, "cmd": "/usr/bin/vnstat -i none", "delta": "0:00:00.005928", "end": "2020-04-27 17:48:08.988125", "msg": "non-zero return code", "rc": 1, "start": "2020-04-27 17:48:08.982197", "stderr": "", "stderr_lines": [], "stdout": "Error: Interface \"none\" not found in database.", "stdout_lines": ["Error: Interface \"none\" not found in database."]}
>
> PLAY RECAP *********************************************************************
127.0.0.1 : ok=239 changed=57 unreachable=0 failed=1 skipped=106 rescued=0 ignored=0

(FYI this is repeatable, and both /opt/iiab/iiab and /opt/iiab/iiab-admin-console had [nearly] the latest code from their master branches.)

iiab-diagnostics output: http://dpaste.com//35AQEAJ

However the original install 4 days ago worked, according to /opt/iiab/iiab/iiab-install.log :

> 2020-04-23 21:38:04,084 p=8538 u=root n=ansible | TASK [vnstat : Create database for LAN to collect vnStat data if not appliance config] ***
> 2020-04-23 21:38:04,115 p=8538 u=root n=ansible | skipping: [127.0.0.1]

This line https://github.com/iiab/iiab/blob/master/roles/vnstat/tasks/main.yml#L19 :

```
when: iiab_lan_iface is defined
```

...could be revised to:

```
when: iiab_lan_iface is defined and iiab_lan_iface != "none"
```

...but that might be papering over a deeper problem?

_Is it possible [iiab-from-console.yml](https://github.com/iiab/iiab/blob/master/iiab-from-console.yml) is setting vars differently from the original [iiab-stages.yml](https://github.com/iiab/iiab/blob/master/iiab-stages.yml) ??_

FYI Admin Console runs this command when I click Configure > Install Configured Options :

```
[...]

PR #2368, opened 2020-04-27
move pylib.yml to stage 4

jvonau:
enhances #2362 by not requiring a --reinstall to activate.


holta:
Should all 3 of these...
- iiab_const.py.j2
- iiab_lib.py
- iiab_env.py.j2

...move from [roles/2-common/templates](https://github.com/iiab/iiab/tree/master/roles/2-common/templates) to [roles/4-server-options/templates](https://github.com/iiab/iiab/tree/master/roles/4-server-options/templates) ?

Tangentially related to:

#2367 "vnStat playbook cannot be rerun on a VM w/o LAN? Is iiab-from-console.yml part of the issue?"

tim-moody:
sorry commented on the wrong pr.

if these files are moved to stage 4 how does that make it easier to update?

holta:
> if these files are moved to stage 4 how does that make it easier to update?

When you click Admin Console (http://box.lan/admin) -> Configure -> ICO (Install Configured Options) Ansible starts at Stage 4...up to Stage 9...and the does the Network stage (equivalent to running ./iiab-network).

holta:
> Should all 3 of these...
> 
> * iiab_const.py.j2
> * iiab_lib.py
> * iiab_env.py.j2
> 
> ...move from [roles/2-common/templates](https://github.com/iiab/iiab/tree/master/roles/2-common/templates) to [roles/4-server-options/templates](https://github.com/iiab/iiab/tree/master/roles/4-server-options/templates) ?

@jvonau mentions:
- 4-server-options/templates can work, sure, but doesn't exist yet
- He'd prefer to place all this logic into its own roles (e.g. roles/pylibs) which would then be invoked by 4-server-options/tasks/main.yml

tim-moody:
> When you click Admin Console -> Configure -> ICO (Install Configured Options) it starts at Stage 4.

I guess that's true but how does it relate to my question, which was with respect to 

>enhances #2362 by not requiring a --reinstall to activate.

runroles 2-common does not require --reinstall and is probably faster

jvonau:
> runroles 2-common does not require --reinstall and is probably faster

Don't think it would be faster as there are other functions in 2-common and in doing so does reset the stage counter when completed.

tim-moody:
the only downside is if something needs one of these three files prior to stage 4

roles/pylibs/tasks/main.yml
```diff
@@ -0,0 +1,11 @@
...
+  template:
+    src: "{{ item.src }}"
+    dest: "{{ item.dest }}"
+    owner: root
+    group: root
+    mode: 0644
+  with_items:
+    - { src: 'roles/2-common/templates/iiab_const.py.j2', dest: '{{ py3_dist_path }}/iiab/iiab_const.py' }
+    - { src: 'roles/2-common/templates/iiab_lib.py', dest: '{{ py3_dist_path }}/iiab/iiab_lib.py' }
+    - { src: 'roles/2-common/templates/iiab_env.py.j2', dest: '{{ iiab_etc_path }}/iiab_env.py' }
```
holta:
```suggestion
    - { src: 'iiab_const.py.j2', dest: '{{ py3_dist_path }}/iiab/iiab_const.py' }
    - { src: 'iiab_lib.py', dest: '{{ py3_dist_path }}/iiab/iiab_lib.py' }
    - { src: 'iiab_env.py.j2', dest: '{{ iiab_etc_path }}/iiab_env.py' }
```

jvonau:
Nothing that would import iiab_lib.py or iiab_env.py.j2 is ever called in the iiab repo.

tim-moody:
>Nothing that would import iiab_lib.py or iiab_env.py.j2 is ever called in the iiab repo

if that were true I would not have created  iiab_lib.py and  iiab_const.py

https://github.com/iiab/iiab/blob/master/roles/kiwix/templates/iiab-make-kiwix-lib3.py

But I think iiab_env.py.j2 is obsolete

jvonau:
Now as role one can just use runrole pylibs to test local changes while being able to update existing installs, should master change at some point, with ease.

holta:
Smoke-test is now running on 191-u20-srv-PR2368-BIG-0428 = 10.8.0.26, currently installing Stage 7...

holta:
> Smoke-test is now running on 191-u20-srv-PR2368-BIG-0428 = 10.8.0.26, currently installing Stage 7...

Admin Console > Configure > "Save Configuration" and then "Install Configured Options" both worked unlike yesterdat, thanks also to PR #2349 which is now contained in this PR #2368.

FYI:

```
root@box:/etc/iiab# diff local_vars.yml.bkp local_vars.yml
374a375,404
>
>
> ############################################################################
> # IIAB -- following variables are first set by browser via the Admin Console
> #  They may be changed via text editor, or by the Admin Console.
>
> gui_desired_network_role: Appliance
> gui_static_wan: False
> gui_static_wan_ip: 192.168.0.191
> gui_static_wan_netmask: 255.255.255.0
> gui_static_wan_gateway: 192.168.0.1
> gui_static_wan_nameserver: 192.168.0.1
> use_cache: False
> adm_cons_force_ssl: False
> gw_squid_whitelist: False
> gw_block_https: False
> hostapd_enabled: True
> dokuwiki_enabled: False
> gui_desired_home_url: home
> xo_services_enabled: False
> ejabberd_enabled: False
> kalite_cron_enabled: False
> osm_enabled: False
> pathagar_enabled: False
> owncloud_enabled: False
> schooltool_enabled: False
> xovis_enabled: False
> xovis_chart_heading: undefined
> sugar_stats_enabled: False
> teamviewer_enabled: False

jvonau:
> > Nothing that would import iiab_lib.py or iiab_env.py.j2 is ever called in the iiab repo
> 
> if that were true I would not have created iiab_lib.py and iiab_const.py
> 
> https://github.com/iiab/iiab/blob/master/roles/kiwix/templates/iiab-make-kiwix-lib3.py
>
 
That is installed in kiwix's role but used by the cmdsrv and iiab-make-kiwix-lib is never run at any point from within iiab's repo.

> But I think iiab_env.py.j2 is obsolete

I'll look to you & George to make that call and remove if required.

tim-moody:
> That is installed in kiwix's role but used by the cmdsrv and iiab-make-kiwix-lib is never run at any point from within iiab's repo.

It allows kiwix to be used without admin console.

> Smoke-test is now running on 191-u20-srv-PR2368-BIG-0428 = 10.8.0.26, currently installing Stage 7...

had this been done previously it would equally have solved the problem without this PR.

jvonau:
> > That is installed in kiwix's role but used by the cmdsrv and iiab-make-kiwix-lib is never run at any point from within iiab's repo.
> 
> It allows kiwix to be used without admin console.
> 

Agreed no issue there. We agree that iiab-make-kiwix-lib is available post-install for cmdline use, we differ on if it is used during install, and I can't seem to find where iiab-make-kiwix-lib is called.

tim-moody:
I thought it used to be run during the kiwix install, but I see that it is not.

jvonau:
Would be nice if [Save Configuration](https://github.com/iiab/iiab-admin-console/commits/master/roles/console/files/htmlf/20-configure.html) could be cleaned up as not to add services that are no longer supported and are commented out in both local_vars and default_vars from being append to the user's local_vars file.

PR #2337, opened 2020-04-13
JV-Ap0+GH-wifi up down [simultaneous WiFi-as-hotspot AND WiFi-as-client]

jvonau:
for review

PR #2380, opened 2020-05-01
Ap0+wifi up down rebased [simultaneous WiFi-as-hotspot AND WiFi-as-client]

jvonau:
### Fixes Bug
Removes Dup's commits that really should not be pushed but otherwise identical to #2337
number of commit reduced from 177 to 86


holta:
I used WiFi to install this PR #2380 onto a regular PC laptop = 133-u20-desk-BIG-PR2380 = 10.8.0.26

1) After the install completed and the ensuing reboot, is "dual wifi" supposed to work on a regular laptop like this?

   `/etc/iiab/iiab.env` showed "HOSTAPD_ENABLED=True" and 'brctl show' showed "ap0" ...both as they should, but perhaps "dual wifi" is not expected to work due to firmware quirks?  In any case, when my phone attempted to connect to SSID "133-u20-desk-BIG-PR2380" it stalled for several minutes each time, showing "Obtaining IP address..." (meanwhile the laptop's WiFi connection to the Internet continued to work).

2) So I tried to change /etc/iiab/local_vars.yml to "wifi_up_down: False" and then run ./iiab-network which failed:

   > 2020-05-01 17:41:52,274 p=4957 u=root n=ansible | TASK [network : Create /etc/hostapd/hostapd.conf from template] ****************
   > 2020-05-01 17:41:52,396 p=4957 u=root n=ansible | fatal: [127.0.0.1]: FAILED! => {"changed": false, "msg": "AnsibleUndefinedVariable: 'iiab_wireless_lan_iface' is undefined"}

@jvonau thanks for helping or logging in directly to 10.8.0.26 to diagnose if you can!

jvonau:
1 9ee93d7 systemd-networkd was not being enabled on wifi gateway (no_net_restart) installs resulting in br0 not being active upon reboot.

2 for the wifi only issue f8a59b4 Didn't show up while the ethernet cable was inserted..

holta:
Warning: @jvonau & I discovered that changing from `wifi_up_down: False` to `wifi_up_down: True` [1] can get very confusing, e.g. on Ubuntu 20 Server after I put `/etc/netplan/02-iiab-config.yaml` [2] in place to connect to WiFi-to-Internet.

[1] Make this change within /etc/iiab/local_vars.yml -- which then requires you to run `cd /opt/iiab/iiab && ./iiab-network && reboot`

[2] Include your "building's" SSID & password at the end, as seen in the 02-iiab-config.yml template at https://github.com/iiab/iiab/wiki/IIAB-Platforms#operating-systems (indentation is very important, and tabs are not tolerated!)

Underlying Reason This Can Get Confusing: hostapd can run IIAB's hotspot on the wrong radio channel (`iw dev` shows frequencies/channels) if you don't (1) turn off IIAB's hotspot first, using `iiab-hotspot-off` (2) set `wifi_up_down: True` in /etc/iiab/local_vars.yml (3) run `cd /opt/iiab/iiab && ./iiab-network && reboot`

So if you omitted step (1) above, before making the transition to `wifi_up_down: True` in steps (2) and (3) -- it's possible you *might* still be able to recover (getting hostapd onto the correct channel/frequency, to match that of the "building's" upstream WiFi) by running...
- iiab-hotspot-off
- systemctl restart netplan-wpa-wlp2s0
- cd /opt/iiab/iiab && ./iiab-network
- reboot

Finally, IIAB's 'dual wifi' (simultaneous upstream & downstream WiFi) can be verified as _likely working_ if most/all these things are true...
- /etc/iiab/iiab.env contains `HOSTAPD_ENABLED=True`
- Running `brctl show` shows `ap0` in the rightmost column
- `systemctl status hostapd` looks good?
- And of course test that your IIAB's knowledge/learning hotspot is working, using a nearby phone or other WiFi-enabled device!

Issue #2327, opened 2020-04-07
MediaWiki: login broken link + broken image


holta:
0) I installed BIG-sized IIAB onto Raspbian Lite on RPi 4, using `curl d.iiab.io/install-test.txt | sudo bash`

1) When trying to log in to MediaWiki (http://10.8.0.62/wiki -> "Personal tools" -> "Log in") one is sent to broken link:

   http://10.8.0.62/mwlink/index.php?title=Special:UserLogin&returnto=Main+Page

   (Browser shows "File not found." and nothing else.)

2) This PNG image does not appear (it shows a broken image) on the main MediaWiki page after installing IIAB onto Raspbian Lite on RPi 4:

   http://10.8.0.62/mwlink/resources/assets/poweredby_mediawiki_88x31.png

tim-moody:
not sure it works over vpn. can you confirm locally

I was able to login on my local network (not sure what the Personal Tools part is)

tim-moody:
are you talking about http://192.168.3.96/wiki/Special:SpecialPages. it has a login link that took me to the login page (http://192.168.3.96/wiki/Special:UserLogin).

there should not be any mwlink pages unless this is cached as the link is now w (someone requested that change)

holta:
> not sure it works over vpn. can you confirm locally

The failures are identical, regardless whether visiting MediaWiki from 10.8.0.x (OpenVPN), 192.168.0.x (Ethernet side of IIAB) or 172.18.y.z (IIAB hotspot).
 
> I was able to login on my local network (not sure what the Personal Tools part is)

The only obvious place to login on the HTML page (http://box/wiki) appears as follows:

> Navigation menu
>
> Personal tools
> - Not logged in
> - Talk
> - Contributions
> - Create account
> - Log in

When clicking on "Log in" I'm sent to http://box/mwlink/index.php?title=Special:UserLogin&returnto=Main+Page which fails (as explained above) with message "File not found." and nothing else on the browser page.

holta:
This bug report was based on the erroneous belief that `curl d.iiab.io/install-test.txt | sudo bash` was equivalent to master + PR #2266 as we agreed to during last Thursday's IIAB call (http://minutes.iiab.io).

[...]

PR #2254, opened 2020-02-16
Pi gen [pi-gen]

jvonau:

PR #2381, opened 2020-05-01
Pi gen

jvonau:
Was #1937 #2254 was tested while part of http://download.iiab.io/7.1/iiab-test when that was pointed to jv-test 53bc50d was the result of that testing #2327


jvonau:
2254 was tagged with 7.1 so re-opening.

tim-moody:
I'm afraid I still don't understand the purpose of this PR. Perhaps someone can explain it. The name suggests that it builds raspbian as there is already a pi-gen project that does that, but I assume that is not the case. We build iiab with ansible, so I don't see how it does that either. There is an implied workflow here that I need to understand.

I do see that for some reason it introduces another local vars file. Local vars is a custom override of default vars so I don't know why you need a custom override of a custom override. Personally I use a variety of local vars fragments with, for example, test settings for host name and the js-menu repo, that I merge into local vars with https://github.com/iiab/iiab-factory/blob/master/box/generic/merge_local_vars.py. If this is the same concept, like presets, I would not duplicate the name. BTW, merge_vars.py also allows you to strip out the myriad comments and duplicates of default values so that you can actually see what you're getting.

jvonau:
> I'm afraid I still don't understand the purpose of this PR. Perhaps someone can explain it. 

Changes to allow ansible to run in a chroot build environment, the specific issue is the use of ansible's mysql modules.
>The name suggests that it builds raspbian as there is already a pi-gen project that does that, but I assume that is not the case. We build iiab with ansible, so I don't see how it does that either. There is an implied workflow here that I need to understand.

I have a set of [addons](https://github.com/jvonau/pi-gen/tree/iiab-rebase) to Raspbian's official image builder that allow the equivalent of 'iiab' to be preinstalled on the created image provided that the mysql modules don't run. The skipped mysql modules are run upon first boot of the image.

> 
> I do see that for some reason it introduces another local vars file.

Sorry I do not see that in this PR can you point that out?
> Local vars is a custom override of default vars so I don't know why you need a custom override of a custom override. 

Need some clarity on this one, what are you referring to?
>Personally I use a variety of local vars fragments with, for example, test settings for host name and the js-menu repo, that I merge into local vars with https://github.com/iiab/iiab-factory/blob/master/box/generic/merge_local_vars.py. If this is the same concept, like presets, I would not duplicate the name. BTW, merge_vars.py also allows you to strip out the myriad comments and duplicates of default values so that you can actually see what you're getting.

tim-moody:
>>I do see that for some reason it introduces another local vars file.

>Sorry I do not see that in this PR can you point that out?

Maybe I misunderstood, but I see feature - support custom local_vars via /boot/local_vars.yml
3b8f8c5

tim-moody:
I think our normal workflow is to start with an sd card with an os on it and use ansible to install iiab. An even simpler workflow is to get a prebuilt image with an os and curated iiab on it and then add content. I'm OK with alternate workflows, but I don't think the normal workflow should be encumbered with them. So I would see pi-gen addons as something external to the iiab repo. 

As I understand it iiab-provision is a service that on first boot installs mysql and maybe other things that can not be installed by ansible in an environment other than the target os, i.e. ansible has nothing analogous to a cross compile. If I want to install a service that uses mysql, let's say nextcloud for argument's sake, mysql gets installed at first boot (hopefully I still have an internet connection). Does the ansible configure script also run at first boot after mysql has been installed? How many cases are there besides mysql? How does this service know what to install and how to configure? Is it a one-shot that checks for work to do and then stops if none?

Rather than getting rpi foundation to build iiab as part of pi-gen, I'd rather get them to include a link to our image catalog for rpi-imager, as they do for Ubuntu and others, and then just build those images by whatever means we wish. Or are you thinking of something like what kiwix does with an image generator web site? Certainly some are attracted to that idea.

jvonau:
> > > I do see that for some reason it introduces another local vars file.
> 
> > Sorry I do not see that in this PR can you point that out?
> 
> Maybe I misunderstood, but I see feature - support custom local_vars via /boot/local_vars.yml
> [3b8f8c5](https://github.com/iiab/iiab/commit/3b8f8c5eb1f12c6bbab994cf810d1bb549711e4f)

Similar code to what Raspbian does with wpa_supplicant.conf when present on the first dos partition. The idea is allow setting of passwords and enabled services prior to the first boot of the install via the local_vars file that is moved into place in /etc/iiab and those settings would be applied during first boot.

jvonau:
> I think our normal workflow is to start with an sd card with an os on it and use ansible to install iiab. An even simpler workflow is to get a prebuilt image with an os and curated iiab on it and then add content. I'm OK with alternate workflows, but I don't think the normal workflow should be encumbered with them. So I would see pi-gen addons as something external to the iiab repo.

I'll point out the iiab-rpi-root-resize.service provided by 1-prep falls into this same catagory.

> 
> As I understand it iiab-provision is a service that on first boot installs mysql and maybe other things that can not be installed by ansible in an environment other than the target os, i.e. ansible has nothing analogous to a cross compile. 

Misconception number one, the 'install' from the point of view of apt|ansible has been completed, there is nothing 'new' to install, just a couple of commands to run in the brand new environment. 

>If I want to install a service that uses mysql, let's say nextcloud for argument's sake, mysql gets installed at first boot (hopefully I still have an internet connection). 

The mysql package was installed during image creation so no internet required.
>Does the ansible configure script also run at first boot after mysql has been installed? 

No need to install mysql is already present.
>How many cases are there besides mysql? 

Postgre would be the only other one that I'm aware of. 

>How does this service know what to install and how to configure? 

References local_vars like all other functions in iiab
>Is it a one-shot that checks for work to do and then stops if none?

One shot deal, self disables if it runs to completion. 
> 
> Rather than getting rpi foundation to build iiab as part of pi-gen, 

Don't think having to host this on our behalf is required, it's just part of the tool chain. 
>I'd rather get them to include a link to our image catalog for rpi-imager, as they do for Ubuntu and others, and then just build those images by whatever means we wish. 

Sure thinking the first step would be to build release-able images that are machine generated with as little human intervention required with the same tools as upstream uses, that is what pi-gen provides.

>Or are you thinking of something like what kiwix does with an image generator web site? Certainly some are attracted to that idea.

Now this has crossed my mind, website that has ansible loaded on it with a webpage that one could enter the ip address and ssh userid|pw, maybe os type for ansible use, and run ansible remotely from the webserver against the target ip address with the provided info.

tim-moody:
>Changes to allow ansible to run in a chroot build environment, the specific issue is the use of ansible's mysql modules.

led me astray. so you're saying the provision service doesn't need to install packages, only needs to configure (such as databases).

>thinking the first step would be to build release-able images that are machine generated with as little human intervention required with the same tools as upstream uses

curated how, small, medium, large? Who is the audience for this capability? Or is this a travis-like function?

>The idea is allow setting of passwords and enabled services prior to the first boot of the install via the local_vars file that is moved into place in /etc/iiab and those settings would be applied during first boot.

this then overrides the local vars that was the basis of the install that is being booted? who does the 'setting'? I edit this when I write it to the sd card.

jvonau:
> > Changes to allow ansible to run in a chroot build environment, the specific issue is the use of ansible's mysql modules.
> 
> led me astray. so you're saying the provision service doesn't need to install packages, only needs to configure (such as databases).
> 

Correct.
> > thinking the first step would be to build release-able images that are machine generated with as little human intervention required with the same tools as upstream uses
> 
> curated how, small, medium, large? 

Thinking of offering 'base' which would have ansible and all the 1-3 apt packages preinstalled but without any of the roles run. The second offering I was thinking of a 'medium' with all of 'iiab' completed(admin-console kalite's routines) sans for the DB passwords. This could be expanded to include a desktop version and/or 'big' in some combination only time required to complete the image 'set' would increase. Think the 2 images would be enough coverage for most, and one could add whatever one wishes providing internet is available for what is not already pre-installed.

>Who is the audience for this capability? Or is this a travis-like function?
> 

Would be of developer and advanced deployment interests, and yes this would be a travis-like function being part of the toolchain of image creation. 

> > The idea is allow setting of passwords and enabled services prior to the first boot of the install via the local_vars file that is moved into place in /etc/iiab and those settings would be applied during first boot.
> 
> this then overrides the local vars that was the basis of the install that is being booted? who does the 'setting'? I edit this when I write it to the sd card.

Yes, it is intended to replace the copy in /etc/iiab. The only partition that is viewable after writing any image on windows machine is the first dos partition. Seems like the right place to source a user editable file for first boot, like how wpa-supplicant.conf is handled.  Service runs once and is disabled, if the file is not found it does nothing and disables.

tim-moody:
>The only partition that is viewable after writing any image on windows machine is the first dos partition. 

Yes, boot is a handy spot because of its file system; I used to copy scripts there when writing the os image. Lately, because I write the os on linux I run the following which copies setup scripts to the root directory on the linux partition. The scripts do things similar the oneliner but some other things I want as well.

```
    echo "writing to $SDCARD root dir"
    cp rpi-setup/* $ROOTFS_PART_MNT/root
    cp /opt/iiab/iiab-factory/box/generic/merge_local_vars.py $ROOTFS_PART_MNT/root
    chmod +x $ROOTFS_PART_MNT/root/rpi-*

    echo "export LS_OPTIONS='--color=auto'" >>$ROOTFS_PART_MNT/root/.bashrc
    echo "alias ll='ls \$LS_OPTIONS -lav'" >>$ROOTFS_PART_MNT/root/.bashrc

    touch $BOOTFS_PART_MNT/ssh
    df
    umount $BOOTFS_PART_MNT
    umount $ROOTFS_PART_MNT
```

tim-moody:
much of this comes down to personal judgments about what is practical and where there needs to be flexibility. my judgment is I prefer to get a content ready image into pi imager as a starting point. I have been thinking that it is time to go back to one-size-fits-all now that sd cards are larger. by this I mean an image in which 'everything' is installed but only a few things are enabled. Once that is on your sd card you have a lot of flexibility with respect to services and content.

tim-moody:
forgot to ask. in pi-gen are there control variables that let me do the entire install and config up front rather than at boot time if I build on the target platform?

jvonau:
https://github.com/jvonau/pi-gen/tree/iiab-rebase is the branch of upstream pi-gen, with build_vars.yml controlling what is installed and stage_iiab* being added to achieve the end results. The bulk of the work is done in https://github.com/jvonau/pi-gen/blob/iiab-rebase/stage_iiab4/iiab/00-run.sh. Build is started with ./build-iiab.sh, need to update the splash page stock upstream version. This branch has been run on Raspbian on a rpi4 and runs to completion, the docker-compose is completly unexplored at this point.

tim-moody:
>I'll point out the iiab-rpi-root-resize.service provided by 1-prep falls into this same catagory.

I've been thinking that maybe this should be moved to factory as it's only needed if min-sd has been used to prepare the image.

PR #2383, opened 2020-05-03
2382 [hostapd / restart / dnsmasq / systemd-networkd refinement]

jvonau:

### DIFF
--- a/roles/network/tasks/hostapd.yml
+++ b/roles/network/tasks/hostapd.yml
@@ -4,7 +4,7 @@
     enabled: no
     masked: no
 
-- name: Disable hostapd when not using ap0 and wifi gateway present or no WiFi hardware present
+- name: Disable hostapd when not using ap0 and wifi gateway present, or no WiFi hardware present
   set_fact:
     hostapd_enabled: False
   when: (not wifi_up_down and discovered_wireless_iface == iiab_wan_iface) or discovered_wireless_iface == "none"
--- a/roles/network/tasks/hostapd.yml
+++ b/roles/network/tasks/hostapd.yml
@@ -4,10 +4,10 @@
     enabled: no
     masked: no
 
-- name: Disable hostapd when not using ap0 and wifi gateway present
+- name: Disable hostapd when not using ap0 and wifi gateway present or no WiFi hardware present
   set_fact:
     hostapd_enabled: False
-  when: not wifi_up_down and discovered_wireless_iface == iiab_wan_iface
+  when: (not wifi_up_down and discovered_wireless_iface == iiab_wan_iface) or discovered_wireless_iface == "none"
 
 - name: Detect current Wifi channel
   shell: iw {{ discovered_wireless_iface }} info | grep channel | cut -d' ' -f2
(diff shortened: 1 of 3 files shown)


roles/network/tasks/restart.yml
```diff
@@ -81,10 +81,13 @@
     state: restarted
   when: hostapd_enabled and wifi_slave.stdout is defined and wifi_slave.stdout == 0
 
+#both interfaces.d and systemd-networkd should have br0 available and Appliance lacks br0
 - name: User choice of dnsmasq or dhcpd - restarting {{ dhcp_service2 }}
   systemd:
     name: "{{ dhcp_service2 }}"
     state: restarted
-  when: not no_net_restart and discovered_wireless_iface == "none"
+  when: (not no_net_restart and not wifi_up_down) or wifi_up_down
```
holta: Redundant Logic...can be more concise:
```suggestion
  when: (not no_net_restart) or wifi_up_down
```

roles/network/tasks/hostapd.yml
```diff
@@ -4,10 +4,10 @@
     enabled: no
     masked: no
 
-- name: Disable hostapd when not using ap0 and wifi gateway present
+- name: Disable hostapd when not using ap0 and wifi gateway present or no WiFi hardware present
```
holta:
```suggestion
- name: Disable hostapd when not using ap0 and wifi gateway present, or no WiFi hardware present
```

holta:
@jvonau tested this.

For the heck of it, I'm also going to run a fresh install smoke-test on 182-rpi4-2g-u20-wifi-MIN (master) that will appear on VPN soon in coming minutes.

holta:
Clarification that the install did indeed complete (182-rpi4-2g-u20-wifi-MIN).

But after reboot, WiFi devices could not connect reliably to the RPi's SSID.

@jvonau made some adjustments (on the above test machine) and has a possible solution.

PR #2384 seems far larger (?) but might also include this fix?

jvonau:
The fix used was simply adding a 3 second delay to the ip link call in /etc/networkd-dispatcher/routable.d/netd-disp. The current netd-disp template is used in other networkd-dispatcher carrier/no-carrier locations, so it is not just a case of adding the sleep into the master template. The later version of ap0wifi_up_down2 has d149bed0829437f6d5613f9a87798f07731fc7c2 and refined with 9aef6750b8943b39b012f67a1bfeafa4e20c8df1 where splitting of routable.d is done to accommodate potential changes in the upstream wifi channel and attempt to have hostapd line up with the new channel.

jvonau:
9142cde is a functional change and not just a wording edit, rendering any forward testing of code useless as I have to re-validate this "harmless" change.

roles/network/tasks/restart.yml
jvonau: If you think so, should this causes any issues with dnsmanq I will step back and let you discover the reason why if failed and not answer any issue raised about the failure.

holta:
> 9142cde is a functional change

Check your logic please.

In fact, 9142cde is not a functional change.

PR #2395, opened 2020-05-09
install and start dnsmasq early - add option to set upstream dns server

jvonau:
allows manual setting of a upstream dns server

### DIFF
--- a/roles/1-prep/tasks/main.yml
+++ b/roles/1-prep/tasks/main.yml
@@ -3,6 +3,9 @@
 - name: ...IS BEGINNING ============================================
   command: echo
 
+- name: dnsmasq (install now, configure LATER in 'network', after Stage 9)
+  include_tasks: roles/network/tasks/dnsmasq.yml
+
 - name: Install uuid-runtime package (debuntu)
   package:
     name: 
--- a/roles/1-prep/tasks/main.yml
+++ b/roles/1-prep/tasks/main.yml
@@ -5,6 +5,7 @@
 
 - name: dnsmasq (install now, configure LATER in 'network', after Stage 9)
   include_tasks: roles/network/tasks/dnsmasq.yml
+  #when: dnsmasq_install | bool
 
 - name: Install uuid-runtime package (debuntu)
   package:
--- a/roles/3-base-server/tasks/main.yml
+++ b/roles/3-base-server/tasks/main.yml
@@ -21,12 +21,6 @@
 - name: WWW_BASE (WWW_OPTIONS should be installed later)
   include_role:
     name: www_base
-  #when: www_base_install | bool
-  #when: apache_install or nginx_install
-
-- name: dnsmasq (install now, configure LATER in 'network', after Stage 9)
-  include_tasks: roles/network/tasks/dnsmasq.yml
-  when: dnsmasq_install | bool
 
 - name: Recording STAGE 3 HAS COMPLETED =====================
   lineinfile:
--- a/roles/network/tasks/dnsmasq.yml
+++ b/roles/network/tasks/dnsmasq.yml
@@ -14,7 +14,9 @@
    - { src: 'roles/network/templates/network/dnsmasq.service.u18', dest: '/etc/systemd/system/iiab-dnsmasq.service', mode: '0644' }
    - { src: 'roles/network/templates/network/dnsmasq-iiab', dest: '/etc/dnsmasq.d/dnsmasq-iiab', mode: '644' }
 
-- name: Don't use stock dnsmasq systemd unit file during boot
+- name: Don't use stock dnsmasq systemd unit file during boot but start now
   systemd:
     name: dnsmasq
+    daemon_reload: yes
     enabled: no
+    state: restarted
--- a/roles/network/tasks/enable_services.yml
+++ b/roles/network/tasks/enable_services.yml
@@ -64,7 +64,7 @@
   template:
     src: network/dnsmasq-iiab
     dest: /etc/dnsmasq.d/dnsmasq-iiab
-  when: dnsmasq_install
+  when: dnsmasq_install    # 2020-05-10: Are all these dnsmasq_install conditions really still necessary ?
 
 ## Another way to skin the cat
 ##- name: Check if systemd service networkd-dispatcher is enabled
--- a/roles/network/tasks/enable_services.yml
+++ b/roles/network/tasks/enable_services.yml
@@ -60,6 +60,12 @@
     dest: /etc/hosts.dnsmasq
   when: dnsmasq_install and dnsmasq_enabled and (iiab_network_mode != "Appliance")
 
+- name: Update /etc/dnsmasq.d/dnsmasq-iiab for custom dns setting
+  template:
+    src: network/dnsmasq-iiab
+    dest: /etc/dnsmasq.d/dnsmasq-iiab
+  when: dnsmasq_install
+
 ## Another way to skin the cat
 ##- name: Check if systemd service networkd-dispatcher is enabled
 ##  systemd:
--- a/roles/network/templates/network/dnsmasq-iiab
+++ b/roles/network/templates/network/dnsmasq-iiab
@@ -1,1 +1,7 @@
+#IIAB
 bind-interfaces
+# Wan nameserver if manually set
+{% if wan_nameserver != "" %}
+no-resolv
+server={{ wan_nameserver }}
+{% endif %}
--- a/roles/network/templates/network/dnsmasq.conf.j2
+++ b/roles/network/templates/network/dnsmasq.conf.j2
@@ -1,7 +1,5 @@
 # Never forward addresses in the non-routed address spaces.
 bogus-priv
-#　Add other name servers here, with domain specs if they are for　non-public domains.
-#server=/{{ iiab_domain }}/{{ iiab_hostname }}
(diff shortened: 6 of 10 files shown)


jvonau:
pretesting - without IIAB installed on a pristine image check when apt installs dnsmasq what the contents of /etc/resolv.conf is after apt install (should be 127.0.0.1) and if not then run `sudo systemctl restart dnsmasq` and recheck.

holta:
On an Ubuntu 20.04 Server VM, `/etc/resolv.conf` does not change after `apt install dnsmasq` :

```
# This file is managed by man:systemd-resolved(8). Do not edit.
#
# This is a dynamic resolv.conf file for connecting local clients to the
# internal DNS stub resolver of systemd-resolved. This file lists all
# configured search domains.
#
# Run "resolvectl status" to see details about the uplink DNS servers
# currently in use.
#
# Third party programs must not access this file directly, but only through the
# symlink at /etc/resolv.conf. To manage man:resolv.conf(5) in a different way,
# replace this symlink by a static file or a different symlink.
#
# See man:systemd-resolved.service(8) for details about the supported modes of
# operation for /etc/resolv.conf.

nameserver 127.0.0.53
options edns0
```

Note that dnsmasq cannot start:

```
root@u20-srv:~# systemctl status dnsmasq
● dnsmasq.service - dnsmasq - A lightweight DHCP and caching DNS server
     Loaded: loaded (/lib/systemd/system/dnsmasq.service; enabled; vendor preset: enabled)
     Active: failed (Result: exit-code) since Sat 2020-05-09 16:43:22 UTC; 2min 24s ago

May 09 16:43:22 u20-srv systemd[1]: Starting dnsmasq - A lightweight DHCP and caching DNS server...
May 09 16:43:22 u20-srv dnsmasq[3773]: dnsmasq: syntax check OK.
May 09 16:43:22 u20-srv dnsmasq[3778]: dnsmasq: failed to create listening socket for port 53: Address already in use
May 09 16:43:22 u20-srv dnsmasq[3778]: failed to create listening socket for port 53: Address already in use
May 09 16:43:22 u20-srv dnsmasq[3778]: FAILED to start up
May 09 16:43:22 u20-srv systemd[1]: dnsmasq.service: Control process exited, code=exited, status=2/INVALIDARGUMENT
May 09 16:43:22 u20-srv systemd[1]: dnsmasq.service: Failed with result 'exit-code'.
May 09 16:43:22 u20-srv systemd[1]: Failed to start dnsmasq - A lightweight DHCP and caching DNS server.
```

holta:
On fresh OS Raspbian Lite on RPi 4, `/etc/resolv.conf` appears as follows: (prior to running `apt install dnsmasq`)

```
root@raspberrypi:~# more /etc/resolv.conf
# Generated by resolvconf
nameserver 192.168.0.1
```

Then a few seconds after running `apt install dnsmasq` (not right away it would appear??) it changed to:
```
root@raspberrypi:~# more /etc/resolv.conf
# Generated by resolvconf
nameserver 127.0.0.1
```

As an aside, here's the output from `systemctl status dnsmasq` :

```
root@raspberrypi:~# systemctl status dnsmasq
● dnsmasq.service - dnsmasq - A lightweight DHCP and caching DNS server
   Loaded: loaded (/lib/systemd/system/dnsmasq.service; enabled; vendor preset: enabled)
   Active: active (running) since Sat 2020-05-09 13:35:14 EDT; 1min 27s ago
 Main PID: 877 (dnsmasq)
    Tasks: 1 (limit: 1599)
   Memory: 1.5M
   CGroup: /system.slice/dnsmasq.service
           └─877 /usr/sbin/dnsmasq -x /run/dnsmasq/dnsmasq.pid -u dnsmasq -r /run/dnsmasq/resolv.conf -7 /etc/dnsmasq.d,.dpkg-dist,.dpkg-old,.dpkg-new --local-service -

May 09 13:35:14 raspberrypi systemd[1]: Starting dnsmasq - A lightweight DHCP and caching DNS server...
May 09 13:35:14 raspberrypi dnsmasq[869]: dnsmasq: syntax check OK.
May 09 13:35:14 raspberrypi dnsmasq[877]: started, version 2.80 cachesize 150
May 09 13:35:14 raspberrypi dnsmasq[877]: DNS service limited to local subnets
May 09 13:35:14 raspberrypi dnsmasq[877]: compile time options: IPv6 GNU-getopt DBus i18n IDN DHCP DHCPv6 no-Lua TFTP conntrack ipset auth DNSSEC loop-detect inotify du
May 09 13:35:14 raspberrypi dnsmasq[877]: reading /run/dnsmasq/resolv.conf
May 09 13:35:14 raspberrypi dnsmasq[877]: using nameserver 192.168.0.1#53
May 09 13:35:14 raspberrypi dnsmasq[877]: read /etc/hosts - 5 addresses
May 09 13:35:14 raspberrypi dnsmasq[878]: Too few arguments.
May 09 13:35:14 raspberrypi systemd[1]: Started dnsmasq - A lightweight DHCP and caching DNS server.
```

jvonau:
Both are expected results, bind-interfaces takes care of the issue on Ubuntu.
Do an install with this PR applied and with wan_nameserver: set in local_vars.yml to some external dns server and post iiab-diagnostics pre-reboot.

holta:
I misunderstood this PR, thinking it would add a 3rd DNS server (as specified by Ansible var `wan_nameserver`) for reliability (on top of the ~2 DNS servers that most ISP's normally provide).

@jvonau clarifies that this is not the case.  Rather, this PR supplants (replaces) the ISP's DNS servers, and uses only that 1 single DNS server specified by variable `wan_nameserver`.

This single point of failure concerns me.  But this PR (if it's a good idea to move forward with it?) could still be used for debugging possibly &mdash; e.g. the try to better understand the near-daily failures to resolve X.raspberrypi.org and Y.ubuntu.com in recent weeks &mdash; which is clearly a systemic failure of some kind, blocking a majority of IIAB installs in my experience :-(

holta:
@jvonau clarifies:

>  `journalctl -b 0 -l --no-pager -u dnsmasq` should provide the upstream dns servers used.

> but that is not quite true on ubuntu as /etc/resolv.conf is not set to 127.0.0.1 so dnsmasq is not referenced when doing dnslookups

holta:
Related:

#2398 "while testing 2395 - handle no wifi"

holta:
> @jvonau clarifies:
> 
> > `journalctl -b 0 -l --no-pager -u dnsmasq` should provide the upstream dns servers used.
> 
> > but that is not quite true on ubuntu as /etc/resolv.conf is not set to 127.0.0.1 so dnsmasq is not referenced when doing dnslookups

> try `resolvectl status` to see what DNS server is in use?

```
root@box:~# resolvectl status
...
  Current DNS Server: 192.168.0.1
         DNS Servers: 192.168.0.1
```

> ...would seem to indicate it's ignoring the 'wan_nameserver: 1.1.1.1' that I specified in /etc/iiab/local_vars.yml

> /etc/resolv.conf dictates which backend is used for the machine itself, 127.0.0.1 you get dnsmasq (so it works right away on raspbian) while 127.0.0.53 gives you systemd-networkd (so ubuntu is NOT using the dnsmasq setting for lookups but clients using the LAN would use the upstream DNS)

holta:
@jvonau this PR removes the line `when: dnsmasq_install | bool` when installing dnsmasq.

Does this mean dnsmasq is now mandatory on all IIAB installations?

If so let's document and/or de-emphasize any variables that are effectively constants @ https://github.com/jvonau/iiab/blob/early-dns/vars/default_vars.yml#L139-L141 here:

```
# dnsmasq - handles DHCP and DNS
dnsmasq_install: True
dnsmasq_enabled: True

jvonau:
For the same effect wondering out loud if netplan can use, in the same block, dhcp4: yes and  
>      nameservers:
>        addresses: [8.8.8.8, 192.168.0.1]

While nameservers: is usually with static ip addressing but wondering if that key is valid when dhcp4 is used.

jvonau:
Think that might work along with an alternate method
https://askubuntu.com/questions/1120998/error-in-network-definition-unknown-key-dhcp4-overrides

holta:
There are a lot of great and evolving ideas in this PR (#2395) as https://netplan.io keeps evolving etc!

Which of these should be considered and/or now finalized as part of IIAB 7.1?

holta:
I moved this to [IIAB 7.2 Milestone](https://github.com/iiab/iiab/milestone/7) in consideration of @jvonau's recommendations:

> wait a bit after 7.1, there is a pretty big matrix to cover on the U-20 side... On the other hand it would be effective on raspbian.... I view it as a code change that doesn't need to be doc'd, with a 'future feature' that will be doc'd when refined..  It's kind of nice to go 'oh yea it works just turn it on for some users', as opposed to having to do a git pull for the feature...

So hopefully we can merge this (with netplan & cloud-init additions possibly) soon after [IIAB 7.1](https://github.com/iiab/iiab/wiki/IIAB-7.1-Release-Notes)'s release?

jvonau:
Given holta is the only consistant reproducer of this dns timeout issue, we could push as is and gather more data on if there is a positive effect or none on the dns timeout issue. Even without using an explicit external dns server dnsmasq should be acting as a caching nameserver so the dns lookups after the first request is received shouldn't contact the upstream dns server a second time for the same dns name on raspbian.

holta:
Ok this appears well-tested on 3 different OS's yesterday, do let's give it a shot now, and monitor it in coming weeks.

PR #2401, opened 2020-05-11
dnsmasq patch for PR #2395

holta:
Thanks @jvonau, Nzola & @shanti-bhardwa

Example error:

> [127.0.0.1] TASK [1-prep : Install uuid-runtime package (debuntu)] ***********************************************************************************************************************************************
[...]

PR #2402, opened 2020-05-11
yaml set the variable wan_nameserver as 'null' without quotes 

jvonau:
then jinja converts that 'none'
### Fixes Bug
skype


holta:
This is an alternative to PR #2401, to patch up a glitch resulting from PR #2395, that occurs when installing IIAB without specifying a value for var `wan_nameserver`

Issue #2265, opened 2020-02-26
Raspbian blocks hostapd when IIAB microSD is built in RPi 4 and then moved to RPi 3 or 3 B+ (or Zero W?) [until you run "rfkill unblock wifi"]


holta:
The workaround (thanks to @jvonau) is to run `rfkill unblock wifi` and then `systemctl restart hostapd`

But first, run `rfkill list` to confirm the problem appears as follows:

```
root@box:~# rfkill list
0: phy0: Wireless LAN
        Soft blocked: yes
        Hard blocked: no
1: hci0: Bluetooth
        Soft blocked: no
        Hard blocked: no
```

Background error messages:

```
root@box:~# systemctl status hostapd
● hostapd.service - Hostapd IEEE 802.11 AP, IEEE 802.1X/WPA/WPA2/EAP/RADIUS Authenticator
   Loaded: loaded (/etc/systemd/system/hostapd.service; enabled; vendor preset: enabled)
   Active: inactive (dead) since Fri 2020-02-21 23:33:05 GMT; 2min 20s ago
  Process: 311 ExecStart=/usr/sbin/hostapd -P /run/hostapd.pid /etc/hostapd/hostapd.conf (code=exited, status=0/SUCCESS)
 Main PID: 311 (code=exited, status=0/SUCCESS)

Feb 21 23:33:05 box.lan hostapd[311]: wlan0: Could not connect to kernel driver
Feb 21 23:33:05 box.lan hostapd[311]: Interface initialization failed
Feb 21 23:33:05 box.lan hostapd[311]: wlan0: interface state COUNTRY_UPDATE->DISABLED
Feb 21 23:33:05 box.lan hostapd[311]: wlan0: AP-DISABLED
Feb 21 23:33:05 box.lan hostapd[311]: wlan0: interface state DISABLED->DISABLED
Feb 21 23:33:05 box.lan hostapd[311]: wlan0: AP-DISABLED
Feb 21 23:33:05 box.lan hostapd[311]: wlan0: CTRL-EVENT-TERMINATING
Feb 21 23:33:05 box.lan hostapd[311]: hostapd_free_hapd_data: Interface wlan0 wasn't started
Feb 21 23:33:05 box.lan hostapd[311]: nl80211: deinit ifname=wlan0 disabled_11b_rates=0
Feb 21 23:33:05 box.lan systemd[1]: hostapd.service: Succeeded.
```
```
root@box:~# more /etc/hostapd/hostapd.conf
[...]

PR #2403, opened 2020-05-12
selective rfkill unblock for raspbian

jvonau:
### Fixes Bug
#2265 
add rfkill unblock for Raspbian.
Hand toggling rfkill block wifi and rebooting with current master and wifi_up_down True

This change will not auto unblock with wifi_up_down: False as this service would be disabled.

### DIFF
--- a/roles/network/templates/hostapd/test-wifi
+++ b/roles/network/templates/hostapd/test-wifi
@@ -15,6 +15,10 @@ fi
 # covers raspbian
 if [ -f /etc/wpa_supplicant/wpa_supplicant.conf ]; then
     RASPBIAN=1
+    if /usr/sbin/rfkill list wifi | grep -q "Soft blocked: yes" ; then
+        echo "unblocking WiFi"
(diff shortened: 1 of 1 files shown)


holta:
If an IIAB microSD is created in RPi 4, and then inserted into RPi 3 or 3 B+, is a manual step needed to engage this and make it work?

(Or does something like roles/network/templates/hostapd/test-wifi somehow run automatically ?)

jvonau:
roles/network/templates/hostapd/test-wifi is the script that is called by wifi-test.service that is installed and enabled if there is wifi detected and using wifi_up_down True.

holta:
Confirmed that this /sbin/test-wifi fix works!

i.e. to solve #2265 on a RPi 3 B+ originally built in a RPi 4, as verified by monitored `rfkill` and `systemctl status hostapd` &mdash; and then testing the RPi 3 B+'s internal WiFi hotspot using my phone.

Issue #2262, opened 2020-02-18
Nextcloud product intro video does not play [/etc/nginx/conf.d/nextcloud-nginx.conf needs fixing?]


holta:
@m-anish am giving this issue its own ticket so it doesn't get lost!

Details @ https://github.com/iiab/iiab/pull/2255#issuecomment-586789380

(Also mentioned within https://github.com/iiab/iiab/wiki/IIAB-7.1-Release-Notes#known-issues)

holta:
@m-anish

I and @jvonau both suspect the Nextcloud intro video (and videos in general?!) are not playing because of a glitch in [nextcloud-nginx.conf](https://github.com/iiab/iiab/blob/master/roles/nextcloud/templates/nextcloud-nginx.conf.j2) that you provided.

Can you possibly take a look to see if that's true?!

holta:
@floydianslips @m-anish

It appears Nextcloud's intro video http://box/nextcloud/apps/firstrunwizard/img/Nextcloud.mp4 is being blocked from playing (in the giant blue popup, when you first log in to http://box/nextcloud) because this .mp4 URL is (effectively?) being redirected to http://box/nextcloud/apps/files/?dir=/&fileid=2

However the underlying vid itself `/library/www/nextcloud/apps/firstrunwizard/img/Nextcloud.mp4` looks fine.

Does this mean that `/etc/nginx/conf.d/nextcloud-nginx.conf` needs fixing?

e.g. can you see anything that needs changing in https://github.com/iiab/iiab/blob/master/roles/nextcloud/templates/nextcloud-nginx.conf.j2 that would allow this vid to play?

holta:
@darkenvy can you make Nextcloud's video play, with help from @floydianslips ?

holta:
Huge thanks to @floydianslips who solved this...anyway good enough for [IIAB 7.1](https://github.com/iiab/iiab/wiki/IIAB-7.1-Release-Notes) Release Candidate 2 that we should release imminently!

PR #2404, opened 2020-05-12
added video MIME support for Nextcloud/Nginx per  #2262

floydianslips:
### Fixes Bug
 #2262

adds video MIME headers to Nextcloud

Ubuntu 20 RPi 4

### DIFF
--- a/roles/nextcloud/templates/nextcloud-nginx.conf.j2
+++ b/roles/nextcloud/templates/nextcloud-nginx.conf.j2
@@ -135,7 +135,7 @@ location ^~ {{ nextcloud_url }} {
         access_log off;
     }
 
-    location ~ ^\/nextcloud\/.+[^\/]\.(?:png|html|ttf|ico|jpg|jpeg|bcmap)$ {
+    location ~ ^\/nextcloud\/.+[^\/]\.(?:png|html|ttf|ico|jpg|jpeg|bcmap|mp4|webm)$ {
         try_files $uri /nextcloud/index.php$request_uri;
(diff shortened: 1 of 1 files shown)


floydianslips:
> This is incredible news...a critical fix to a longstanding bug: @m-anish take a look!
> 
> @tim-moody, do you understand why **svg|gif** etc are separate on Line 109 here? https://github.com/floydianslips/iiab/blob/master/roles/nextcloud/templates/nextcloud-nginx.conf.j2#L109
> 
> ...whereas **png|html|ttf|ico|jpg|jpeg|bcmap** (and now Josh Dennis' **mp4|webm** fix) are separate on Line 138 further below here?
> https://github.com/floydianslips/iiab/blob/master/roles/nextcloud/templates/nextcloud-nginx.conf.j2#L138
> 
> (Is this in fact the way it should be?)

This is actually the way that the Nextcloud Nginx config file is set up on their site. I think the version we are currently using is from an older version of Nextcloud which didn't have mp4|webm in the config file.
https://docs.nextcloud.com/server/latest/admin_manual/installation/nginx.html

holta:
> This is actually the way that the Nextcloud Nginx config file is set up on their site. I think the version we are currently using is from an older version of Nextcloud which didn't have mp4|webm in the config file.
> https://docs.nextcloud.com/server/latest/admin_manual/installation/nginx.html

Good enough for [IIAB 7.1](https://github.com/iiab/iiab/wiki/IIAB-7.1-Release-Notes) ?

Do you have general recommendations for IIAB 7.2 ?!

PR #2407, opened 2020-05-13
missing apache enable

jvonau:
### Fixes Bug
TK's report  - handles runrole, iiab-install, and iiab-configure
add 'enabled: yes' to the apache restart calls.

### DIFF
--- a/roles/0-init/tasks/main.yml
+++ b/roles/0-init/tasks/main.yml
@@ -44,6 +44,7 @@
   include_tasks: validate_vars.yml
 
 # SEE: https://github.com/iiab/iiab/blob/master/roles/nginx/README.md
+# apache is set to install in default_vars - each role will set apache service to enabled
 - name: "apache_install is currently '{{ apache_install }}' and apache_enabled is currently '{{ apache_enabled }}'.  Now let's set 'apache_install: True' and 'apache_enabled: True' if NGINX is set not to be enabled OR if any of {Elgg, Lokole, Moodle, Node-RED} are set to install."
   set_fact:
     apache_install: True
--- a/roles/elgg/tasks/apache.yml
+++ b/roles/elgg/tasks/apache.yml
@@ -9,4 +9,5 @@
 - name: (Re)Start '{{ apache_service }}' systemd service
   systemd:
     name: "{{ apache_service }}"    # apache2 or httpd, as set in /opt/iiab/iiab/vars/<OS>.yml
+    enabled: yes
     state: restarted
--- a/roles/lokole/tasks/apache.yml
+++ b/roles/lokole/tasks/apache.yml
@@ -9,4 +9,5 @@
 - name: (Re)Start '{{ apache_service }}' systemd service
   systemd:
     name: "{{ apache_service }}"    # apache2 or httpd, as set in /opt/iiab/iiab/vars/<OS>.yml
+    enabled: yes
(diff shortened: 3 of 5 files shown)


holta:
`apache_enabled` is always set True here, whenever `apache_install` is set to True:
https://github.com/iiab/iiab/blob/master/roles/0-init/tasks/main.yml#L50-L51

Both are acted upon during IIAB installs' Stage 3, which invokes:
https://github.com/iiab/iiab/blob/master/roles/httpd/tasks/main.yml#L38

In the end the above might be enough — or if not, what situations is this PR seeking to improve?

jvonau:
> `apache_enabled` is always set True here, whenever `apache_install` is set to True:
> https://github.com/iiab/iiab/blob/master/roles/0-init/tasks/main.yml#L50-L51
> 
> Both are acted upon during IIAB installs' Stage 3, which invokes:
> https://github.com/iiab/iiab/blob/master/roles/httpd/tasks/main.yml#L38

Now consider an initial install where none of the 4 trigger apps were installed, short of --reinstall there is nothing to ensure apache is enabled should it become enabled sometime post install. Given www_options does a restart, could be another place to enable the apache service, that would fix iiab-configure, but would not cover runrole where www_options is not included.

jvonau:
This does not address the inaccurate recorded state of apache in iiab.ini that would be #2408

holta:
While refactoring Apache for all sorts of reasons is possible, I'm not (seeing) this as changing much for TK.

i.e. as it stands anyway, the full set of actions in roles/httpd/tasks/install.yml should always run prior to enabling Apache's systemd service.

(We're not trying to support people who happen to type `apt install apache2` at the prompt, and hope that's equivalent to roles/httpd/tasks/install.yml !)

So if this and PR #2408 are more refactoring than helping unblock people immediately (insofar as I can tell) these look to be IIAB 7.2 material most likely.

jvonau:
> While refactoring Apache for all sorts of reasons is possible, I'm not (seeing) this as changing much for TK.
> 

How so? TK's complaint is iiab-configure needs to be run after boot(runs restart apache2) to have elgg working, otherwise nginx reports bad gateway pointing to the apache backend being unavailable because it was never told to start on boot.

> i.e. as it stands anyway, the full set of actions in roles/httpd/tasks/install.yml should always run prior to enabling Apache's systemd service.
> 

What is the purpose of using the entry in the iiab_state file then? Apache is just another service with install and enabled as two distinct operations, and is a dependency of 4 other roles that could be installed|enabled at any point in time. Are you suggesting having to run --reinstall because you now enabled an Apache dependent app after installing min_vars and made a change? 

> (We're not trying to support people who happen to type `apt install apache2` at the prompt, and hope that's equivalent to roles/httpd/tasks/install.yml !)
> 

just wow...

> So if this and PR #2408 are more refactoring than helping unblock people immediately (insofar as I can tell) these look to be IIAB 7.2 material most likely.

I can't prove how TK's machine got into that situation, but this is a cheap fix for him and ensures runrole(s) will result in a working role.

jvonau:
Have to ensure that the secondary webserver is enabled in all code paths not just iiab-install where the secondary webserver is enabled in 3-base-server, due to nothing in the role ensuring that apache is enabled but relying on a setting that is only run during the initial install(3-base-server). It's a bug brought on by the nginx|apache rewrite.

PR #2408, opened 2020-05-13
call httpd role to restart service and record apache state

jvonau:
### Fixes Bug
for discussion untested


holta:
Should be considered with PR #2407 for IIAB 7.2?

holta:
I probably mistated this earlier, but this looks like a possible substitute for (enhancement over?) PR #2407.

PR #2440, opened 2020-06-06
scripts/ansible patch for RPi: avoid multiple entries in /etc/apt/sources.list

holta:
Refines PR #2436 to avoid multiple entries.

### DIFF
--- a/scripts/ansible
+++ b/scripts/ansible
@@ -81,8 +81,12 @@ elif [ -f /etc/debian_version ]; then    # Includes Debian, Ubuntu & Raspbian
     fi
     if [ -f /etc/rpi-issue ]; then
         # 2020-06-06: https://www.raspberrypi.org/forums/viewtopic.php?t=276340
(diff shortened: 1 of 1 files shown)


jvonau:
> if [ -f /etc/rpi-issue ]; then

Relying on this has probability broken 64bit installs as the added apt urls are for 32bit.

holta:
- IIAB 7.1 (released overnight, announcement to come) is not intended to support the ["early beta" 64-bit RaspiOS](https://www.raspberrypi.org/forums/viewtopic.php?t=275370) that appeared ~10 days ago.  If however 64-bit RaspiOS proves increasingly more reliable/popular and legit-supported in coming months, IIAB 7.2 should also support it (and if appropriate safeguard folks from RaspiOS's usual flaky mirrors!?)
- Either way, 64-bit Ubuntu 20.04 continues to work reasonably well on RPi with IIAB (as reconfirmed with yesterday's testing, and as expected since that OS does not contain the file `/etc/rpi-issue` !)

Issue #2461, opened 2020-07-06
Maps for OSM for IIAB7.1 do not appear to be loading correctly


BagOfCoins:
### Expected behavior and actual behavior:
This is my first time posting and I don't know if this is the correct forum for raising this issue but here it goes.
It appears to me that with IIAB 7.1 that OpenStreetMap(OSM) is not working, unlike in IIAB 7.0. In IIAB 7.1, a map can be selected to download from the IIAB Admin section and the Job Status indicates that the three tasks associated with downloading and installing a map are marked "Succeeded".  But when OSM is opened, however, the OSM software appears to be working but there is no map image.  The Lat/Long moves with the mouse pointer, the zoom level changes when zoomed, but there is no map image.  There is also no response when a city is typed into the search box (i.e. no cities come up when searched).  In other words, all of the OSM functionality appears present, but there does not appear to be a loaded map.  This appears to be happening with both North America and Central America maps though other maps have not been tried.  Simple solutions like rebooting/reinstalling do not apear to correct the issue.  Other IIAB software appears to be functioning normally.  See the screenshot.
![OsmMaps from IIAB 7_1](https://user-images.githubusercontent.com/67908303/86597177-65495f00-bf69-11ea-9e2c-09c4110b02a8.png)
This was tried using the one-line-installer to load IIAB 7.1 on a fresh installation of Raspbian and it was tried using the image "iiab-7.1-200612-min-content-ready-g17253d4.img" and the results were the same.  
The maps loaded without any problem when using OSM in IIAB 7.0 in May 2020 (prior to the release of 7.1) when installed with the one-line-installer.  The maps also seem to load normally now when installed in IIAB 7.0 with the line "curl d.iiab.io/7.0/install.txt | sudo bash".
Maybe I'm doing something wrong but I've tried installing the OSM maps several times now using several different methods and this is my experience.  I am not super-knowledgeable about computer stuff but I can respond to simple questions about this issue.
Since you are still reading, I'd like to commend everyone working on this project as it is truly fantastic.  The improvements that you all have been making add alot of utility to the project.  IIAB really has the potential to make the world a better place and you all are making it happen.

### Steps to reproduce the problem:
Install the current (6 July 2020) version of Raspbian on a Raspberry Pi 3B+, then use the 1-line-installer to install IIAB 7.1.  Run IIAB and go to the Admin section and add the OSM map for North America.  Look at the map after installation.
OR
[...]

PR #2463, opened 2020-07-07
unconditionally install sqlite3

georgejhunt:
### Fixes bug:
Tim's image is fixed by installing php7.3-sqlite3

### DIFF
--- a/roles/httpd/tasks/install.yml
+++ b/roles/httpd/tasks/install.yml
@@ -33,6 +33,7 @@
 - name: Install php{{ php_version }}-sqlite3 (raspbian-9+ or debian-9+ or ubuntu-18+)
   package:
     name: "php{{ php_version }}-sqlite3"
+    state: present
   #when: is_raspbian_9 or is_debian_9 or is_ubuntu_18
   #when: is_debuntu and (not is_debian_8) and (not is_ubuntu_16)
   #when: (is_debian and ansible_distribution_major_version == "9") or is_ubuntu_18
--- a/roles/httpd/tasks/install.yml
+++ b/roles/httpd/tasks/install.yml
@@ -34,7 +34,7 @@
   package:
     name: "php{{ php_version }}-sqlite3"
   #when: is_raspbian_9 or is_debian_9 or is_ubuntu_18
-  when: is_debuntu and (not is_debian_8) and (not is_ubuntu_16)
+  #when: is_debuntu and (not is_debian_8) and (not is_ubuntu_16)
   #when: (is_debian and ansible_distribution_major_version == "9") or is_ubuntu_18
 
 - name: 'Install 4 packages: httpd, mod_authnz_external, php, php-curl (redhat)'
--- a/roles/osm-vector-maps/tasks/install.yml
+++ b/roles/osm-vector-maps/tasks/install.yml
@@ -1,3 +1,8 @@
+- name: Install php-sqlite3
+  package: 
+    name: "php{{ php_version }}-sqlite3"
+    state: present
+
 - name: Ensure directory {{ vector_map_path }}/maplist/assets exists
   file:
     path: "{{ vector_map_path }}/maplist/assets"    # /library/www/osm-vector-maps


holta:
Presumably this resolves #2461 ??

("Maps for OSM for IIAB7.1 do not appear to be loading correctly")

holta:
`when: is_debuntu and (not is_debian_8) and (not is_ubuntu_16)` covers all 3 of our currently supported OS's {RaspiOS 10, Ubuntu 20, Debian 10} so I'm confused how commenting this out helps &mdash; @georgejhunt can you explain more?

holta:
> state is required for package module

Good catch, thanks @georgejhunt 

Does this fix #2461 ?

tim-moody:
when logged in as root and with an internet connection for IIAB I did

```
apt update
apt install php7.3-sqlite3

```
The initial map still has no tiles even though it says zoom level 5, but zooming in and out got a map although with many missing tiles.

```
systemctl restart nginx
systemctl restart php7.3-fpm.service
```

helped, but some tiles still fail to appear, mostly satellite I think.

On reboot the initial map was populated, but still with some missing tiles.

tim-moody:
I still have some of these errors

```
Uncaught Error: Unimplemented type: 3
    at Pbf.skip (index.js:166)
    at Pbf.readFields (index.js:16)
    at MVT.readFeatures (MVT.js:225)
    at VectorTile.<anonymous> (featureloader.js:61)
```

georgejhunt:
#2463 not smoke tested

georgejhunt:
Yes I had the missing tile problem also. In the past, it has come and gone, and what seemed to fix it was to define data-root in tileserver.php (with our new viewer--program separated from data approach).
But doing that will require redoing the 12 map zip files. It may be that php-frm does not do the same as apache php.
I think it much easier to use apache for vector-maps than to redo 12 zipped bundles that are going away anyway.

holta:
Great if OSM is moving towards defining its own dependencies (e.g. sqlite3 in this case).

As such, shouldn't the installation of `php{{ php_version }}-sqlite3` be removed from roles/httpd/tasks/install.yml if it's no longer serving a purpose there?

(Or does Apache somehow truly require sqlite3 for some reason?)

georgejhunt:
I don't think that we have agreement on which direction we are "moving toward".

holta:
> I don't think that we have agreement on which direction we are "moving toward".

As will always be the case in any democracy.  e.g. it could take weeks if reverting to Apache needs to be investigated and carefully tested.

Along the way we should also answer the original Yes/No question: does Apache somehow truly require sqlite3 ?
(I'll assume the answer is No, as we fix master in the coming hours so that people can use OSM as part of regular IIAB 7.2 pre-releases.  But if I'm wrong, and Apache also needs sqlite3 for other reasons, please LMK.)

tim-moody:
first, is there possibly a timeout or file size that can be changed to fix the couple of tile blocks that don't load. feels to me like timeout as they load more readily as soon as things get cached by the server.

Is it viable to move forward to the viewer code rather than backwards to apache?
I agree we don't want to rebuild the zipped bundles.
What happens if you take the viewer code and after downloading a 7.1 bundle you move the mbtiles to /library/www/osm-vector-maps/viewer/tiles? the 3rd stage of the admin cons tile download could easily be modified to do this. 
What happens if you modify regions.json to point to the new mbtiles packages? my bet is that adm cons will be none the wiser and just download them with no change to the user interface.

tim-moody:
> Along the way we should also answer the original Yes/No question: does Apache somehow truly require sqlite3 

apache2 has no dependency on php or sqlite3

georgejhunt:
I don't think reverting to apache is a very good idea. The code base has
changed so much, and I cannot figure it out quickly.
I like the idea of exploring exactly why php-fpm is getting errors.
It seems very repeatable for me, and some tiles never seem to render
correctly. I remember that the satellite database initially had many
missing tiles. So I wrote a python program to verify and fetch the missing
ones, and they will be added at the end. So I'm sure access times for some
tiles will be longer, going through the index tree.

georgejhunt:
The defect seems to be in the satellite sqlite database. I substituted
satellite_z0-z9_v3.mbtiles. Cleared local cache-->didn't fix. Then rebooted
the server ->fixed.
So Tim's suggestion of changing the script in admin-console which loads the
zip file can work.
In each zip package there is a symbolic link satellite.mbtiles
-> satellite_z0-z9.mbtiles which needs changing to satellite.mbtiles
-> satellite_z0-z9_v3.mbtiles
The new satellite file is up at
https://archive.org/download/satellite_z0-z9_v3.mbtiles/satellite_z0-z9_v3.mbtiles

georgejhunt:
I wonder if  the "san_jose" zip file might not be the only one we need to
update.  I cannot imagine that 7.0 had the problem we have seen with
satellite tiles without a lot of complaints.
I think we just removed san-jose from regions -- but for 7.1 we renamed it
to "World to zoom 10", and started using it as our test platform because it
was smaller.
Redoing the "world to zoom 10" zip file might be another option.

georgejhunt:
Or just remove the "world to zoom 10" from regions.json

georgejhunt:
I can confirm that the central_america zip package has the _v3 version of
satellite.mbtiles

tim-moody:
Isn't World to zoom 10" just the name in the regions.json file?

georgejhunt:
yes

georgejhunt:
no actually "World to zoom 10" is the content of the title field in the
san_jose region.

holta:
Should no longer be necessary, as a result of PR #2464 which is now merged.

PR #2468, opened 2020-07-15
7.2 kiwix apk

tim-moody:
passes smoke test on rpi
in functional test works to linux, but on android something turns the zim download into zim.zip


tim-moody:
downloads as .zim on win10, so why not on android. I guess file associations.

tim-moody:
once I fixed the file associations in android this works there as well.

PR #2469, opened 2020-07-21
Stage1.5 [OSM]

georgejhunt:
Installs on May version of raspiOS, and ubuntu18.04.

Known problems:
1. Layer switcher can be unpredictable with more that one region installed
1. Attribution is missing
1. Zooming in transition from world to region at zoom 10 is problematic (OSM goes away for one zoom level).
1. Untested recently -- the right click adding points and images -- I expect problems it this area -- code is complex.


georgejhunt:
Still finding too much that it not working the way I want

PR #2470, opened 2020-07-24
No /etc/network/interfaces.d/iiab [fixes RPi 3 B+, allowing WiFi firmware & hostapd to launch reliably on boot]

jvonau:
### Fixes bug:
br0 race base on the down and dirty move test, but may impact bluetooth if AP is turned off so there would be no br0 but bluetooth might create it anyway.

### DIFF
--- a/roles/network/tasks/rpi_debian.yml
+++ b/roles/network/tasks/rpi_debian.yml
@@ -41,6 +41,12 @@
     dest: /etc/network/interfaces.d/iiab
     src: network/rpi.j2
     #src: network/iiab.j2    2019-02-05: caused ./iiab-network to fail repeatedly in recent days on one particular RPi 3, even after reboot ("Unable to start service networking: Job for networking.service failed..." at "Restart the networking service if appropriate") ...leaving dnsmasq off (#1452)
+  when: iiab_wired_lan_iface is defined
+
+- name: Use bind-dynamic for dnsmasq
+  template:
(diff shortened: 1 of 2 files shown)


holta:
CLARIF:

Manually removing /etc/network/interfaces.d/iiab appears to cause RPi 3 B+ to boot reliably, in 24 out of 24 attempts (WiFi firmware & hostapd launched properly!)

(Prior to this, almost 25% of boots failed to launch WiFi/driver and hostapd.)

holta:
Are any tests or smoke-tests needed -- e.g. to make sure this PR doesn't cause problems in other situations?

(Yesterday I only tested removing `/etc/network/interfaces.d/iiab` on RaspiOS Lite on RPi 3 B+)

holta:
@georgejhunt agrees we should move forward with this PR, especially as this PR introduces "lazy binding" of dnsmasq to any defined interfaces e.g. if br0 starts later on.

holta:
Merged experimentally after discussion with @georgejhunt & @jvonau call on IIAB weekly call (http://minutes.iiab.io).

Yes this needs more testing after merge -- thanks all for assisting in coming days, where possible!

holta:
Smoke-test ongoing as part of a standard MIN-sized IIAB install (http://d.iiab.io) onto 10.8.0.6 = 183-rpi4-2G-MIN

PR #2471, opened 2020-08-01
RPiOS - use dhcpcd hooks to restart hostapd when br0 appears [for Ubuntu too]

jvonau:
### Fixes bug:
on skype - if you stop and later start hostapd by hand, dnsmasq should be restarted once the bridge re-appears with an ip address. This removes the need to manually restart dnsmasq and should be better at dealing with bluetooth as the only device under the bridge.

### DIFF
--- a/roles/network/tasks/enable_services.yml
+++ b/roles/network/tasks/enable_services.yml
@@ -85,23 +85,22 @@
 #- debug:
 #    var: nd_enabled
 
-#- name: Check if /etc/networkd-dispatcher/routable.d exists
-#  stat:
-#    path: /etc/networkd-dispatcher/routable.d
-#  register: nd_dir
+- name: Check if /etc/networkd-dispatcher/routable.d exists
+  stat:
+    path: /etc/networkd-dispatcher/routable.d
+  register: nd_dir
 
 #- debug:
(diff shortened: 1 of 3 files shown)


holta:
Wouldn't this be an improvement on other OS's on RPi too?

(And/or other non-RPi hardware too?)

jvonau:
lol.. now it is..

jvonau:
> Wouldn't this be an improvement on other OS's on RPi too?
> 
The dhcpcd hook is specific to the RPiOS where dhcpcd is used. 


> (And/or other non-RPi hardware too?)

Re-enabling the network-dispatcher hook moves the dnsmasq startup to be more event driven as opposed to only using a time delay workaround to guess if the bridge is available.

holta:
Tangentially Related:

PR #2472 RFC Firmware downgrading [WiFi hotspot capacity workaround, for Raspberry Pi]

jvonau:
This really is an extension of #2470

holta:
fyi this PR #2471 (and PR #2472) are being tested together with PR #2474, on an RPi 4 @ 10.8.0.14

holta:
Applies to all hardware, not just RPi's.

Issue #823, opened 2018-05-30
Raspberry Pi 3 used to support 32 WiFi connections but is now limited to 10.  RPi 3 B+ and RPi 4 are now limited to ~4 [replacing Raspbian with Ubuntu 19.10.1 increases this to 32; what does Ubuntu 20.04 offer?]


holta:
Currently being discussed on xsce-devel mailing list...

Unlike in 2016, 32 simultaneous WiFi connections to RPi 3 (running the latest Raspbian Lite from 2018-04-18) no longer seem possible :/

holta:
Tangentially Related: TK Kang is observing total WiFi failure about an hour after boot (of IIAB 6.6 on Raspbian 2018-06-27) with symptoms similar to: https://github.com/raspberrypi/linux/issues/2453

Hopefully he and @jvonau can write up a more proper bug report ticket, so others can verify whether this is-or-is-not related to upstream firmware issues with Raspberry Pi's Wi-Fi?

jvonau:
Perhaps disabling the promiscuous mode hack from #638 might help or hurt but lets take that out of the picture as the firmware/kernel have been updated since the introduction of that 'fix'

holta:
As TK has reconfirmed the pattern in comprehensive tests on both Raspbian Lite and Raspbian Desktop... 

> With Internet access from Ethernet cable and connection to the the RPi3 Wi-Fi, will fail after an hour. Sometimes can't connect to the Wi-Fi AP and if it is connected it is not able to access http://box, etc.
> 
> If I boot with Ethernet cable unplugged  (i.e. no Internet), the wifi access and http://box, etc will remain stable for hours.

...I've published this as a "Known Issue" at the bottom of the Release Notes:

https://github.com/iiab/iiab/wiki/IIAB-6.6-Release-Notes#known-issues

holta:
Raspberry Pi 4's tested in classrooms in Oaxaca State, Mexico (earlier this month, December 2019) did not allow more than 16 simultaneous WiFi connections to the RPi 4's internal hotspot.

Whereas this is not true with Raspberry Pi 3's &mdash; which allow up to 32 simultaneous WiFi connections &mdash; even with the very same microSD card/image as used in the above RPi 4's.

_The Raspberry Pi 3 B+ should be verified, to reconfirm whether this one too still supports 32?_

[...]

PR #2472, opened 2020-08-02
RFC Firmware downgrading [WiFi hotspot capacity workaround, for Raspberry Pi]

jvonau:
only forces downgrade with no provisions to undo the rolling back of the firmware.

### DIFF
--- a/roles/firmware/tasks/main.yml
+++ b/roles/firmware/tasks/main.yml
@@ -8,7 +8,7 @@
     dest: "{{ item.dest }}"
     mode: "{{ item.mode }}"
   with_items:
-    - { src: 'fw-warn.sh', dest: '/usr/sbin/', mode: '0644' }
+    - { src: 'fw_warn.sh', dest: '/etc/profile.d/', mode: '0644' }
     - { src: 'check-firmware.service', dest: '/etc/systemd/system/', mode: '0644' }
     - { src: 'check-firmware.sh', dest: '/usr/sbin/', mode: '0755' }
 
--- a/roles/firmware/tasks/main.yml
+++ b/roles/firmware/tasks/main.yml
@@ -1,0 +1,20 @@
+- name: Include older RPi Firmware
+  include_tasks: install.yml
+  when: RPi_firmware_retrieved is undefined
+
+- name: Install check-firmware service files
+  template:
+    src: "{{ item.src }}"
+    dest: "{{ item.dest }}"
+    mode: "{{ item.mode }}"
+  with_items:
+    - { src: 'fw-warn.sh', dest: '/usr/sbin/' mode: '0644' }
+    - { src: 'check-firmware.service', dest: '/etc/systemd/system/' mode: '0644'}
+    - { src: 'check-firmware.sh', dest: '/usr/sbin/' mode: '0755' }
+
+- name: Enable and Start check-firmware.service
+  systemd:
+    name: check-firmware.service
+    state: reloaded    
+    enabled: yes
+    status: started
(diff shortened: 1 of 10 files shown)


roles/firmware/tasks/main.yml
```diff
@@ -0,0 +1,20 @@
...
+  include_tasks: install.yml
+  when: RPi_firmware_retrieved is undefined
+
+- name: Install check-firmware service files
+  template:
+    src: "{{ item.src }}"
+    dest: "{{ item.dest }}"
+    mode: "{{ item.mode }}"
+  with_items:
+    - { src: 'fw-warn.sh', dest: '/usr/sbin/' mode: '0644' }
```
holta: fw-warn.sh is a better name than fw_warn.sh below, if this is needed.
```suggestion
    - { src: 'fw-warn.sh', dest: '/usr/sbin/', mode: '0644' }
```

roles/firmware/tasks/main.yml
```diff
@@ -0,0 +1,20 @@
...
+    - { src: 'check-firmware.service', dest: '/etc/systemd/system/' mode: '0644'}
```
holta:
```suggestion
    - { src: 'check-firmware.service', dest: '/etc/systemd/system/', mode: '0644' }
```

roles/firmware/tasks/main.yml
```diff
@@ -0,0 +1,20 @@
...
+    - { src: 'check-firmware.sh', dest: '/usr/sbin/' mode: '0755' }
```
holta:
```suggestion
    - { src: 'check-firmware.sh', dest: '/usr/sbin/', mode: '0755' }
```

holta (commented):
Just a quick review (inlined comments) noticing a few small bugs.

Separately, I personally I lean towards:

1.  Including a backup every time firmware file(s) are overwritten (so the operator is protected no matter what, always able to roll back firmware if truly needed).  Similar to suggestions at: https://github.com/iiab/iiab/issues/823#issuecomment-667563551
2. Preserving the timestamp whenever firmware files are copied or backed up, making all these files (and a bit of their history) easier to understand, e.g. using `cp -p` or `rsync -t`

Tangentially Related:

PR #2471 "RPiOS - use dhcpcd hooks to restart hostapd when br0 appears [for Ubuntu too]"

roles/firmware/templates/fw_warn.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash
```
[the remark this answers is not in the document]
jvonau: latest commit

holta (commented):
Two (in the end three) missing commas.

jvonau:
There are 3 diffs being testing the 3 files, remember the stock 43430 files doesn't provide a clm_blob, only the replacement bin requires it. What is your issue?

holta:
1) I see that 3rd test now, understood.

2) If either .clm_blob changes, WiFi (generally) becomes very unreliable, in a classroom situation of many WiFi devices especially, as my July tests confirmed.

So testing for future changes to both .clm_blob files is probably wise (adding a 4th test).

That 4th file being http://d.iiab.io/packages/brcmfmac43455-sdio.clm_blob_2018-02-26_rpi (currently shipping with RaspiOS & Ubuntu-for-RPi both, but that could change in future).

holta:
While testing this, I see "RPi_firmware_retrieved: True" in /etc/iiab/iiab_state.yml

If possible can we keep this lowercase, and matching the exact name of the Ansible role, to help everyone keep track going forward?  (Possibly the role should be renamed to rpi_firmware ?)

(In any case keeping these names self-consistent wherever they appear, helping avoid many ambiguities, e.g. when running ./runrole etc.)

holta:
fyi this PR #2472 (and PR #2471) are being tested together with PR #2474, on an RPi 4 @ 10.8.0.14

holta:
Tested on RPi 4 2GB running RaspiOS Lite.

PR #2473, opened 2020-08-05
Stage1.5 [OSM]

georgejhunt:
This soft codes the repo name and branch from which the map code is pulled. We need the soft coding as we work towards integrating wikidata map information with the other map code.


roles/osm-vector-maps/tasks/main.yml
```diff
@@ -23,11 +15,9 @@
   include_tasks: install.yml
   when: osm_vector_maps_installed is undefined
 
-
-- name: Enable/Disable/Reload NGINX for OSM, if nginx_enabled
-  include_tasks: nginx.yml
-  when: nginx_enabled | bool
-
+- name: Install OSM Vector Maps if 'osm_vector_maps_installed' not defined, e.g. in {{ iiab_state_file }}    # /etc/iiab/iiab_state.yml
+  include_tasks: install.yml
```
holta: FYI this is duplicate code -- Lines 14-16 (just above) runs install.yml with the standard `when:` clause (used by all other roles) -- as invoked by https://github.com/iiab/iiab/blob/master/roles/7-edu-apps/tasks/main.yml#L26-L29

holta:
Not to interrupt @georgejhunt's hard work ongoing here!  i.e. George mentioned on today's call (http://minutes.iiab.io) that he will be testing/refining this to ensure the new IIAB Map Packs (that he's uploaded to Archive.org) install onto fresh IIAB installs going forward for [IIAB 7.2](https://github.com/iiab/iiab/wiki/IIAB-7.2-Release-Notes).

At the same time thanks to @jvonau who notes we should eventually: 

1) Preserve & use standard file [osm-vector-maps/tasks/nginx.yml](https://github.com/iiab/iiab/blob/master/roles/osm-vector-maps/tasks/nginx.yml)

   ...as invoked by [these 3 lines](https://github.com/iiab/iiab/blob/master/roles/osm-vector-maps/tasks/main.yml#L27-L29) of osm-vector-maps/main.yml :
   ```
   - name: Enable/Disable/Reload NGINX for OSM, if nginx_enabled
     include_tasks: nginx.yml
     when: nginx_enabled | bool
   ```

2) Preserve & use these standard lines from the [bottom of osm-vector-maps/install.yml](https://github.com/iiab/iiab/blob/master/roles/osm-vector-maps/tasks/install.yml#L75-L85) so that `/etc/iiab/iiab_state.yml` continues to record `osm_vector_maps_installed: True` using:

   ```
   # RECORD OSM Vector Maps AS INSTALLED

   - name: "Set 'osm_vector_maps_installed: True'"
     set_fact:
       osm_vector_maps_installed: True

   - name: "Add 'osm_vector_maps_installed: True' to {{ iiab_state_file }}"
     lineinfile:
       path: "{{ iiab_state_file }}"    # /etc/iiab/iiab_state.yml
       regexp: '^osm_vector_maps_installed'
       line: 'osm_vector_maps_installed: True'
   ```

holta:
1) @georgejhunt Lines 176-193 of `osm-vector-maps/tasks/install.yml` can be deleted now, right?

   (As nginx.yml takes care of that, after install.yml — as soon as the role is run with `osm_vector_maps_enabled: True`)

2) Is `osm_vector_maps_enabled: False` temporary during testing, in default_vars.yml and the 3 local_vars_{min, medium, bug}.yml files?

   Or is there some reason you're changing IIAB's current default?

georgejhunt:
See https://github.com/georgejhunt/iiab/tree/stage1.5/roles/osm-vector-maps
Install takes about 30MB, enabled creates a functioning map to zoom 10. All
regions install the world to zoom 10 as a base.
There's a discussion ahead whether small, medium or large should have
enabled on by default. But in 7.1 enabled did not install a functioning
map, so I was just wanting a discussion before I changed anything. My
inclination is to enable even for min, like we do for kiwix and kalite. But
I can also see restricting enabled to medium and big.

roles/osm-vector-maps/tasks/main.yml
```diff
@@ -15,9 +15,9 @@
   include_tasks: install.yml
   when: osm_vector_maps_installed is undefined
 
-- name: Install OSM Vector Maps if 'osm_vector_maps_installed' not defined, e.g. in {{ iiab_state_file }}    # /etc/iiab/iiab_state.yml
-  include_tasks: install.yml
-  when: osm_vector_maps_install | bool and not osm_vector_maps_installed is defined
+- name: Enable/Disable/Reload NGINX for OSM, if nginx_enabled
+  include_tasks: nginx.yml
+  when: nginx_enabled | bool
```
jvonau, line 20: The when clause is undesirable, nginx.yml should deal with both enabled and disabled as needed.

holta:
> I can also see restricting enabled to medium and big.

30MB is nothing these days (no less with 1TB microSD cards etc ;)

@georgejhunt personally I'd be in favor of keeping OSM enabled, as part of the 3 most common {MIN, MEDIUM, BIG} IIAB install paths.

Particularly as offline maps are IIAB's original/distinguishing feature.

roles/osm-vector-maps/tasks/main.yml
holta: > The when clause is undesirable, nginx.yml should deal with both enabled and disabled as needed.

@jvonau this `when:` clause looks ok to me, as it's conditioned on `nginx_enabled` (as in most other Ansible roles/playbooks) rather than `osm_vector_maps_enabled`.

(Or am I missing something?)

georgejhunt:
Answering Tims questions about whether this PR disables the old maps -- it does not disable, or prevent the admin console from installing the 7.1 map packages. I did the following test:
1. Install 7.1 master, and the test package (world to zoom 10) using admin console.
2. Install this PR. The installed world to zoom 10 still works.
3. Use the 7.2 installer to install  central-america--which installs and functions.
4. The new 7.2 installer does not update the home page. So the user needs to browse to http:box.lan/osm-vector-maps/viewer or just http://box/maps

Issue #2422, opened 2020-05-28
Test IIAB 7.2+ on 64-bit Raspbian (renamed to Raspberry Pi OS / RaspiOS) as this becomes more established? [UH-OH: WiFi firmware 7.45.18.0 not quite compatible; will a new kernel help?] [64-bit RasPiOS Bullseye allows for Calibre 5.12]


holta:
@jvonau & All: the early beta of Raspberry Pi Foundation's new OS is here:
https://www.raspberrypi.org/forums/viewtopic.php?f=117&t=275370

Latest 32-bit versions are here:
https://www.raspberrypi.org/downloads/raspbian/
http://downloads.raspberrypi.org/raspios_armhf/release_notes.txt

holta:
Thanks @floydianslips who observes:
- He had to do `apt install vlc-data` before the IIAB install would begin
- He could not log into Admin Console (http://box.lan/admin)

SEE ALSO @georgejhunt report:

#2424 "raspbian-64bit-dnsmasq does not start because br0 is not up [RaspiOS]"

holta:
@floydianslips @georgejhunt did you also see this error below on 64-bit RaspiOS on RPi 4?

> TASK [2-common : Install 24 common packages: acpid, bridge-utils, bzip2, curl, gawk, hostapd, htop, i2c-tools, logrotate, make, mlocate, netmask, net-tools, ntfs-3g, pandoc, pastebinit, rsync, sqlite3, sudo, tar, unzip, usbmount, usbutils, wget] ***
fatal: [127.0.0.1]: FAILED! => {"changed": false, "msg": "No package matching 'usbmount' is available"}

holta:
Even when I comment out pkg usbmount from https://github.com/iiab/iiab/blob/master/roles/2-common/tasks/packages.yml#L63 it still fails later on in Stage 4, predictably:

> TASK [usb_lib : Enable exFAT and NTFS in /etc/usbmount/usbmount.conf] **********
fatal: [127.0.0.1]: FAILED! => {"changed": false, "msg": "Destination /etc/usbmount/usbmount.conf does not exist !", "rc": 257}

holta:
Until the above issue is solved, I've temporarily (also) set the following in [/etc/iiab/local_vars.yml](http://wiki.laptop.org/go/IIAB/FAQ#What_is_local_vars.yml_and_how_do_I_customize_it.3F) :

```
usb_lib_install: False
usb_lib_enabled: False

[...]

PR #2490, opened 2020-08-25
scripts/ansible: revert to RaspiOS's own /etc/apt/sources.list for 64-bit RaspiOS etc

holta:
Testing ongoing.


holta:
Tested on 64-bit RaspiOS Lite as mentioned @ #2422

Issue #1849, opened 2019-07-08
@m-anish & @tim-moody ask: can systemd now replace Monit? [Debian 10 Buster no longer includes Monit]


holta:
Please weigh in as to whether [Monit](https://github.com/iiab/iiab/tree/master/roles/monit) should be deprecated in the IIAB 7.1 release cycle, much as Debian 10 "Buster" has now done:

#1842 "Ugly workaround so that IIAB installs on Debian 10 Buster (which suddenly removed monit)"

Thanks you to @m-anish & @tim-moody for raising this question during the IIAB community/team call earlier this morning (http://minutes.iiab.io).

ASIDE: the Monit project is not dead, as you can see here: https://mmonit.com/monit/changes/

As such if someone can find out why a community-centric project like Debian decided to drop Monit in very recent weeks, that might help us understand the situation?

holta:
@m-anish wrote an email to the Debian maintainer of `monit` asking if there are more details...explaining options going forward?

e.g. if https://packages.debian.org/buster/mon (or otherwise) might be an alternative to Monit going forward, even if `mon` is not a drop-in replacement to be added into https://github.com/iiab/iiab/tree/master/roles/monit/tasks apparently?

holta:
@m-anish's email to Debian's monit maintainer Sergey B Kirpichev, who responded that backports of Monit might (?) later be possible:
>> We don't see monit in the debian 10 package repos and wonder if there are any plans to
>> make it available on it?
>
> Maybe backports.  See release notes.

also this -- https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=887350#26

https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=927775

https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=930637

holta:
FYI the 'monit' package would (appear) to be returning to Debian 11 / Bullseye when that is released (around) mid-2021:

https://packages.debian.org/bullseye/monit
https://packages.debian.org/source/bullseye/monit

A reminder that the 'monit' package was suddenly removed from Debian 10 / Buster during the very final days prior to its mid-2019 release, as confirmed by the sudden disappearance of these 2 pages:

https://packages.debian.org/buster/monit
https://packages.debian.org/source/buster/monit

Please also see this file, which will be modified very shortly in coming days:

https://github.com/iiab/iiab/blob/master/roles/monit/tasks/main.yml

Ref: #2536

holta:
[...]

PR #2492, opened 2020-08-25
local_vars_big.yml: set monit_install & monit_enabled to False

holta:
I don't know of a single group using Monit.

And BIG-sized installs have issues on Debian 10 & 64-bit RaspiOS where the package is not available.

So let's remove Monit from BIG-sized IIAB install for now, if there are no objections?

Refs: #1849, #2422


holta:
As discussed on "IIAB Chat/Calls" channel.  No objections so far so let's go ahead.

PR #2493, opened 2020-08-25
RFC Mongo2 [reinforce MongoDB on most distros, e.g. for Sugarizer]

jvonau:
### Fixes bug:
mongodb installs on Debian-arm64
### Description of changes proposed in this pull request:
use upstream mongodb repo that provides the latest 4.4 version .
### Smoke-tested on which OS or OS's:
none as of yet

PR #2494, opened 2020-08-27
fix x86_64 [MongoDB]

jvonau:
### Fixes bug:
testing .38

### DIFF
--- a/roles/mongodb/tasks/install.yml
+++ b/roles/mongodb/tasks/install.yml
@@ -59,7 +59,7 @@
 
 
   # end block
-  when: not (ansible_architecture == "amd64" or ansible_architecture == "aarch64")
+  when: not (ansible_architecture == "x86_64" or ansible_architecture == "aarch64")
 # 32 bit OS's get caught above should handle aarch32 including 32-bit ubuntu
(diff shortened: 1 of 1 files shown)


roles/mongodb/tasks/install.yml
```diff
@@ -91,7 +91,7 @@
       repo: deb [ arch=amd64,arm64 ] https://repo.mongodb.org/apt/ubuntu focal/mongodb-org/4.4 multiverse
       state: present
       filename: mongodb-org
-    when: is_ubuntu_20 and (ansible_architecture == "aarch64" or ansible_architecture == "amd64")
+    when: is_ubuntu_20
```
holta, line 94: Later we might test for Ubuntu > 19 (or similar) &mdash; so that things like is_ubuntu_21 (i.e. Ubuntu 21.04 and 21.10) have a better shot when 2021 creeps up...

holta:
Builds on PR #2493 "RFC Mongo2 [reinforce MongoDB on most distros, e.g. for Sugarizer]"

roles/mongodb/tasks/install.yml
jvonau: addressed in #2495

Issue #2497, opened 2020-08-28
Concerning 1 line install on Raspberry pi 3b+ [kolibri.service failed to start due to Ansible 2.9.12; 'ansible_userspace_architecture' is undefined; WiFi flaky due to RaspiOS kernel 5.4.51-v7+ #1333 and/or 3rd party 'hat' doing USB boot?]


pumpkinpi85:
Architecture: raspberry pi 3b+ > 1tb expansion board (no micro sd > booting from usb)

What am I dealing with here:

curl d.iiab.io/install.txt | sudo bash 

(see image)

![Screen Shot 2020-08-27 at 8 37 39 PM](https://user-images.githubusercontent.com/59450868/91515337-2e773180-e8a6-11ea-85bc-4788cb3d71ee.png)

holta:
@pumpkinpi85 can you please provide the URL that results from running the `iiab-diagnostics` command?

(Run it as root if possible, Thanks!)

pumpkinpi85:
Thanks @holta. Is this what you're looking for? 

COMPLETE! Your diagnostics file (, 0 lines) is:

   /etc/iiab/diag/7.2_raspbian-10_2020-08-27_22:56:52_MDT_sudo iiab-diagnostics

Publish it to a web pastebin? [Y/n] y

PUBLISHING TO URL... /usr/bin/iiab-diagnostics: line 238: $outfile: ambiguous redirect

holta:
Hi @pumpkinpi85, let's try to understand more:

- What exact version of Raspberry Pi OS are you running?
  - 32-bit or the new/experimental 64-bit?
  - Lite or "with desktop" ?
  - Is your OS from 2020-08-20 at https://www.raspberrypi.org/downloads/raspberry-pi-os/ _(or an earlier version?)_
- What's the full output of `systemctl status kolibri` please, with a screenshot/photo if that's easiest?
- Did your Raspberry Pi 3 B+ run out of memory (RAM) ?  That's one possibility anyway, as Kolibri (part of BIG-sized IIAB installs) can sometimes be very resource-intensive.
  - Any chance you might have access to a Raspberry Pi 4 with 2GB or more of memory, to try installing there?

Thanks !

jvonau:
> /etc/iiab/diag/7.2_raspbian-10_2020-08-27_22:56:52_MDT_sudo iiab-diagnostics

I reproduced the iiab-diagnostics issue locally.
> Can you provide a short public nickname: (no spaces!)

[...]

Issue #2479, opened 2020-08-08
Validate Ansible 2.10 expected 2020-08-13 [with simplified auto-installation of "Collections" ETA 2020-09-22 ?]


holta:
1) ETA is ~5 days away according to: https://docs.ansible.com/ansible/devel/roadmap/ROADMAP_2_10.html

2) Should we ponder how long Ubuntu will be locked into Ansible 2.9.6, due to its incomplete transition to Python 3 ?  And if so what this means for IIAB's Ansible strategy, currently summarized at:

   https://github.com/iiab/iiab/blob/master/scripts/ansible.md

holta:
Working on it:

- #2481 [permissions bug, 666 becomes 600]
- PR #2482 [temporarily revert to Ansible 2.9.6, as workaround]
- Beyond...

(Hopefully Ansible 2.10.1 will be released in coming weeks and get us out of this mess!)

holta:
Ansible 2.10.x installation will be simplified around 2020-09-22:

'For users that prefer a batteries-included model, the "ansible" package in most distribution channels will be updated in the coming weeks to a curated snapshot of collections that depends on the ansible-base package. This will provide a one-stop installation experience, similar to all previous versions of Ansible'

https://groups.google.com/g/ansible-devel/c/T6sZnzAPDZA
https://docs.ansible.com/ansible/devel/roadmap/COLLECTIONS_2_10.html#release-schedule

holta:
Ansible ETA's Announced:

- 31 August 2020:
  ansible-base 2.10.1rc1
  ansible 2.9.13
  ansible 2.8.15

- 7 September 2020:
  ansible-base 2.10.1

- 28 September 2020:
  ansible-base 2.10.2rc1
  ansible 2.9.14rc1
  ansible 2.8.16rc1

- 5 October 2020:
  ansible-base 2.10.2
  ansible 2.9.14
  ansible 2.8.16

https://groups.google.com/forum/m/#!topic/ansible-announce/cFA8XBYSFSs

Yes we should clean up [scripts/ansible](https://github.com/iiab/iiab/blob/master/scripts/ansible) in the coming 10 days (for Ansible 2.9.13 and/or 2.10.1) — but things will likely change again October ~5th (when Ansible 2.10.2 likely introduces its more unified/simplified install process, including essential "collections").

holta:
Ansible 2.9.13 is now available FWIW:
https://launchpad.net/~ansible/+archive/ubuntu/ansible
[...]

PR #2500, opened 2020-09-02
Infastructure for proper versioning.

jvonau:
Add the recording of IIAB_REVISION to stage 9 to allow incriminating the value to reflect major change in a role that would require a re-install where such roles would be recorded in upgrade_roles, works with https://github.com/iiab/iiab-factory/pull/134

Change the value for any roles' source version that would require a re-install, that gets noted in upgrade_roles and iiab_revision gets bumped. The big mongodb change is a perfect test subject.

### DIFF
--- a/roles/0-init/defaults/main.yml
+++ b/roles/0-init/defaults/main.yml
@@ -8,6 +8,8 @@
 first_run: False
 rpi_model: none
 xo_model: none
+do_upgrade: False
+do_reinstall: False
 gw_active: False
 internet_available: False
 discovered_wan_iface: none
--- a/roles/0-init/tasks/main.yml
+++ b/roles/0-init/tasks/main.yml
@@ -1,20 +1,21 @@
 # Initialize
 - name: ...IS BEGINNING ============================================
-  stat:
-    path: "{{ iiab_env_file }}"
-  register: NewInstall
-
-- name: Set first_run flag
-  set_fact:
-    first_run: True
-  when: not NewInstall.stat.exists
-
-- name: Set top-level variables from local_facts for convenience
+#- name: Re-read local_facts.facts from /etc/ansible/facts.d
+#  setup:
+#    filter: ansible_local
+#- name: Set top-level variables from local_facts for convenience
   set_fact:
     rpi_model: "{{ ansible_local.local_facts.rpi_model }}"
     xo_model: "{{ ansible_local.local_facts.xo_model }}"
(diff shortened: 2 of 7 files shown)


jvonau:
You see the release workflow going on here right?
make a change that requires a reinstall
bump iiab_revision 
add iiab_revision number and role to upgrade_roles

134 has the logic in bash to remove the role based on upgrade_roles from the iiab_state.yml file just like runrole --reinstall does for a single role. Could be done in other places but lets start with this in place to have something to test against?

jvonau:
This PR was tagged with "question" a week ago, with no question asked. What is the deal here?

holta:
Let's surface exactly these design/architectural questions and needs for this PR and others (PR #2511 and others) during our Thursday 10AM NYC IIAB Calls (http://minutes.iiab.io) so everybody can understand where changes like this are critically important, how they serve specific implementer/field communities's needs, medium-term versus long-term etc.

All hugely important questions.

Keeping in mind that [IIAB 7.2](https://github.com/iiab/iiab/wiki/IIAB-7.2-Release-Notes) should ship this month with very few changes, if possible (i.e. most all structural/design changes should really target IIAB 8.0 at this point).  No rule is completely hard & fast -- but as stated at http://minutes.iiab.io -- we should try to keep those remaining IIAB 7.2 pre-release changes bounded in these 3 rough categories: (if at all possible!!)
- Ansible 2.10 support #2479
- Friendlier install of new OSM Map Packs + Satellite Photos (with Admin Console integration as nec?)
- WiFi hotspot reliability issues if/as needed by #2497 ?

jvonau:
The general idea is to use iiab_revision in conjunction with tagging|branching to denote an easy a point in time reference of the code-base without resorting to git hashes.

jvonau:
This goes into 7.2 or the freebies stop.

tim-moody:
is the proposal to track revisions by role or by entire install?

jvonau:
By the role

jvonau:
@holta Get off your high horse you power freak.

jvonau:
Despite the childish attempt to block testing progress by reverting this commit /iiab/iiab-factory#134 proved to be a success http://sprunge.us/YlHp09 note the order of entries in /etc/iiab/iiab_state.yml the 3 roles were automatically upgraded as the new entries are below calibreweb_installed, the last role to possibly be installed on a machine that was first installed on July 20.

PR #2488, opened 2020-08-23
MariaDB [rework security, eliminate {{ mysql_root_password }}]

jvonau:
### Fixes bug:
proposed fix for #2475
### Description of changes proposed in this pull request:
don't set password for root using 'localhost', add second admin account(mariaDBadmin)
### Smoke-tested on which OS or OS's:
pending
### Mention a team member @username e.g. to help with code review:
@holta

PR #2503, opened 2020-09-02
RFC Provision was pi-gen

jvonau:
Updated #2381, with #2488 in place this is much easier to deal with. Introduces *_provisioned to the setup.yml files for those role for tracking of the state of advanced configuration. Allows for a more virgin image where none of the database passwords have been set yet, so those are available to be changed on the next run of ansible.
provision_install: True to access the role
provision_active: True to suppress the setup.yml actions
provision_enable: True to enable firstboot services


tim-moody:
what does 'provision' mean in this context? is it different than configure?

jvonau:
just going to fork the code

PR #2505, opened 2020-09-04
Lockout installs and upgrades when no internet present

jvonau:
just stuff

### DIFF
--- a/roles/0-init/tasks/main.yml
+++ b/roles/0-init/tasks/main.yml
@@ -169,5 +169,13 @@
       value: "yes"
   when: is_VM is defined
 
+- name: 'DISALLOW "XYZ_install: True" WHEN "XYZ_installed is undefined" when no internet access'
+  assert:
+    that: "not {{ item }}_install or {{ item }}_installed is defined"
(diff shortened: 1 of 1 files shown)


tim-moody:
is configuration of a role possible when there is no internet? (e.g. packages installed on a previous run but database or other configuration has not completed.)

holta:
Good question.  Also there are so many different ways to reinstall vs. upgrade vs. reconfigure vs. change data(base) content vs. wipe/safeguard personal data — that this can get painful/confusing as hell :/

jvonau:
> is configuration of a role possible when there is no internet? (e.g. packages installed on a previous run but database or other configuration has not completed.)

Why does iiab-from-console.yml exist if one can't configure and enable|disable already installed web apps? Should I put in a PR to remove  iiab-from-console.yml as nobody does configuration after the initial iiab-install?

 > Good question. Also there are so many different ways to reinstall vs. upgrade vs. reconfigure vs. change data(base) content vs. wipe/safeguard personal data — that this can get painful/confusing as hell :/

Please keep the noise applicable to the respective PRs.. This PR is about early halting and warning the user that internet is required for the requested operation.

jvonau:
forking

PR #2477, opened 2020-08-08
Dual WiFi WARNING: 5 GHz connection from internal WiFi to Internet...ends up sabotaging legacy WiFi client devices that require 2.4GHz !

holta:
Thanks @jvonau for reviewing this verbiage/warning if poss.

Background: Willem in South Africa got stuck...when his phone could connect to IIAB's internal hotspot but his older laptop could not.  A big thank you to @jvonau for figuring out why this was happening!

PR #2509, opened 2020-09-08
Warning about 5 GHz sabotaging IIAB's internal WiFi was not being read. So raise it within local_vars.yml & MAKE IT LOUDER

holta:
Builds on PR #2477

Ref: #2497


jvonau:
How does moving the warning away from the section that the warning is related to help people notice the warning when setting the ssid/password?

PR #2511, opened 2020-09-08
Fieldback -- client connect time and byte count in graphical form [hourly/daily/weekly/monthly usage charts based on hostapd; hashes WiFi MAC addresses, to increase privacy]

georgejhunt:
Based upon hostapd statistics.
A working demo has accumulated over a week of data at http://10.8.0.34/gh
Hover over bars provides limited drill down.
See https://github.com/georgejhunt/iiab/blob/fback/roles/usage/README.md


holta:
Huge thanks @georgejhunt for presenting this Thur 10AM NYC Time if you can, during our IIAB Call off of http://minutes.iiab.io !

holta:
FYI as outlined at http://minutes.iiab.io this and several other late-breaking PR's might need to be moved to the [IIAB 8.0 Milestone](https://github.com/iiab/iiab/milestone/8), so that [IIAB 7.2](https://github.com/iiab/iiab/wiki/IIAB-7.2-Release-Notes) can hopefully be released as planned & on-time before the end of September, rapidly approaching in a few short weeks.

Nothing is carved in stone...but yes fyi it's time to bring IIAB 7.2's change rate way down during final testing now and in coming weeks!

holta:
Most of the guts of this initiative are contained within @georgejhunt's associated PR https://github.com/iiab/iiab-factory/pull/140

roles/usage/tasks/install.yml
```diff
@@ -0,0 +1,46 @@
...
+      mode: 0755
+  with_items:
+      - iiab-wifidata.py
+
+
+# RECORD OSM Vector Maps AS INSTALLED
+
+- name: "Set 'osm_vector_maps_installed: True'"
+  set_fact:
+    osm_vector_maps_installed: True
```
holta: `osm_vector_maps_installed` should be `usage_installed` on Lines 38 & 40.

holta:
Some clarifications and questions arising from the call earlier today: (http://minutes.iiab.io) 

1) @georgejhunt has posted a (prototype) of these charts here:
http://10.8.0.22/gh

2) @georgejhunt explained that a CSV-like JSON file (prototype for now) is also available here:
http://10.8.0.22/gh/fetch.php

3) In a production environment such sites would likely require a password (?) and be viewable off of:
http://box/admin/usage

4) @jvonau has questions about several problems that are very likely to arise as a result of restarting the `hostapd` systemd service every 2 minutes.

georgejhunt:
On Thu, Sep 10, 2020 at 11:52 AM A Holt <notifications@github.com> wrote:

> Some clarifications and questions arising from the call earlier today: (
> http://minutes.iiab.io)
>
>    1.
>
>    @georgejhunt <https://github.com/georgejhunt> put some prototype
>    charts (prototype) here:
>    http://10.8.0.22/gh
>    2.
>
>    @georgejhunt <https://github.com/georgejhunt> explained that a
>    CSV-like JSON file (prototype for now) is available here:
>    http://10.8.0.22/gh/fetch.php
>    3.
>
>    In a production environment, these would likely require a password to
>    be viewed off of:
>    http://box/admin/usage
>    4.
>
>    @jvonau <https://github.com/jvonau> has questions about several
>    problems that are very likely to arise as a result of the restarting of the
>    hostapd systemd service every 2 minutes.
>
> Well in fact, from /etc/crontab, it is actually the hostapd_cli that is
> run every 2 minutes. The accumulator buckets are reinitialized via a
> hostapd restart every hour. Whether this stops a whole classrom, and might
> be preblematic, will need to be tested.
>
/etc/crontab:

tim-moody:
Why is this installed in /opt/admin/console? It is not part of admin console. BTW, box/admin is no longer password protected; only cmdsrv is.

georgejhunt:
When I tried putting fieldback graphs in /opt/admin/console, I got a
request for authentication, which I thought might be a good idea. If you
don't want it there, or if the password authentication has changed, I'm
sure we can put it somewhere else.

tim-moody:
I'm very surprised that you got a prompt for authentication unless you are using an old version of admin console. In the current version only calls to cmdsrv should cause authentication. If you want that then it should be part of iiab-admin-console, otherwise I would prefer that it use another url than /admin.

georgejhunt:
We can put it in /library/www/html/. I'm not sure that usage data needs to
be protected by password.

jvonau:
195 commits and 90 files is a bit much to look at, better [diff](https://github.com/iiab/iiab/compare/master...georgejhunt:fback)

roles/8-mgmt-tools/tasks/main.yml
```diff
@@ -23,6 +23,11 @@
     name: munin
   when: munin_install | bool
 
+- name: USAGE
+  Include_role:
```
jvonau: typo Include_role: s/b lower case 'i' use include_role: 

On a side note the use of '| bool' has changed again... see #2576

PR #2512, opened 2020-09-09
revamp ansible for the future

jvonau:
### Fixes bug:
support ansible 2.9.7 or later(2.10.X) on Ubuntu
Switch Ubuntu to use ansible's ppa
lightly tested on RaspOS

### DIFF
--- a/scripts/ansible
+++ b/scripts/ansible
@@ -8,14 +8,24 @@ GOOD_VER="2.9.6"    # For XO laptops (pip install) & CentOS (yum install rpm)
 
 export DEBIAN_FRONTEND=noninteractive
 source /etc/os-release
-if [ $VERSION_CODENAME = "buster" ]; then
-    release=bionic
-elif [ $VERSION_CODENAME = "focal" ]; then
-    release=eoan
-else
-    echo -e "\nEXITING: Could not detect your OS (unsupported?)\n"
-    exit 1
-fi
+case $VERSION_CODENAME in
+    buster)
+        release=bionic
+    ;;
+    focal)
+        release=eoan
+    ;;
+    eoan)
+        release=eoan
+    ;;
+    bionic)
+        release=bionic
+    ;;
+   *)
+        echo -e "\nEXITING: Could not detect your OS (unsupported?)\n"
+        exit 1
+    ;;
+esac
 
 echo -e "\n\nYOU ARE RUNNING: /opt/iiab/iiab/scripts/ansible (TO INSTALL ANSIBLE)"
 #echo -e 'Alternative:     /opt/iiab/iiab/scripts/ansible-2.8.x ("Slow Food")\n'
--- a/scripts/ansible
+++ b/scripts/ansible
@@ -37,85 +37,38 @@ else
     echo -e "(Internet-in-a-Box requests Ansible $GOOD_VER or higher)\n"
 fi
 
-if [ -f /etc/olpc-release ]; then
-    yum -y install ca-certificates nss
-    yum -y install git bzip2 file findutils gzip hg svn sudo tar which unzip xz zip libselinux-python
-    yum -y install python-pip python-setuptools python-wheel patch
-    # Can above 3 lines be merged into 1 line?
-    pip install --upgrade pip setuptools wheel #EOL just do it
-    pip install ansible==$GOOD_VER --disable-pip-version-check
-elif [ -f /etc/centos-release ]; then
-    yum -y install ansible
-# 2018-09-07: the next 4 lines aren't needed according to https://docs.ansible.com/ansible/latest/installation_guide/intro_installation.html#latest-release-via-dnf-or-yum
-#    yum -y install ca-certificates nss epel-release
-#    yum -y install git bzip2 file findutils gzip hg svn sudo tar which unzip xz zip libselinux-python
-#    yum -y install python-pip python-setuptools python-wheel patch
-#    yum -y install https://releases.ansible.com/ansible/rpm/release/epel-7-x86_64/ansible-$GOOD_VER-1.el7.ans.noarch.rpm
-#elif [ -f /etc/fedora-release ]; then
-#    CURR_VER=`grep VERSION_ID /etc/*elease | cut -d= -f2`
-#    URL=https://github.com/jvonau/iiab/blob/ansible/vars/fedora-$CURR_VER.yml
-#    dnf -y install ansible git bzip2 file findutils gzip hg svn sudo tar which unzip xz zip libselinux-python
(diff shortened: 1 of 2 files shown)


holta:
Ansible 2.10.1 seems to be delayed by a few days, but when it's released (in the coming week likely!) this is the kind of thing #2479 needs.

scripts/ansible
```diff
@@ -25,97 +33,31 @@ if [ $(command -v ansible-playbook) ]; then    # "command -v" is POSIX compliant
...
-    #    echo -e "# Uncomment line below then 'apt-get update' to enable 'apt-get source'" >> /etc/apt/sources.list
-    #    echo -e "#deb-src http://raspbian.raspberrypi.org/raspbian/ buster main contrib non-free rpi" >> /etc/apt/sources.list
-    #fi
-    echo -e "\napt update; apt install ansible and python3 dependencies explained at:"
-    echo -e "https://github.com/iiab/iiab/tree/master/scripts/ansible.md\n"
-    $APT_PATH/apt update
-    $APT_PATH/apt -y --allow-downgrades install ansible python3-pymysql python3-psycopg2 \
-        python3-passlib python3-pip python3-setuptools python3-venv virtualenv
+echo -e "\napt update; install dirmngr; PPA to /etc/apt/sources.list.d/iiab-ansible.list\n"
+echo "deb http://ppa.launchpad.net/ansible/ansible/ubuntu "$release" main" \
```
holta, line 52: A more automated option (not sure this is desirable?!) might be:

```
apt install software-properties-common
apt-add-repository ppa:user/distro/repository
```

Where distro would presumably be `$release` as follows?

```
apt-add-repository ppa:ansible/$release/ansible
```

As documented by `apt-add-repository -h` (if we ever want to consider this more automated approach?)

If so, `apt-add-repository -r` (same as `--remove`) supports "remove repository from sources.list.d directory" if we ever need that.

jvonau: Yes, that would work but why, this method has worked well for a while? Don't think I like the dependencies like unattended-upgrades, video&audio support that would be dragged in on debian|raspOS.

>pi@box:/opt/iiab/iiab $ sudo apt install software-properties-common
Reading package lists... Done
Building dependency tree       
Reading state information... Done
The following additional packages will be installed:
  gir1.2-packagekitglib-1.0 libappstream4 libglib2.0-bin libgstreamer1.0-0
  libpackagekit-glib2-18 libstemmer0d packagekit packagekit-tools python3-distro-info
  python3-pycurl python3-software-properties unattended-upgrades
The following NEW packages will be installed:
  gir1.2-packagekitglib-1.0 libappstream4 libglib2.0-bin libgstreamer1.0-0
  libpackagekit-glib2-18 libstemmer0d packagekit packagekit-tools python3-distro-info
  python3-pycurl python3-software-properties software-properties-common
  unattended-upgrades

jvonau:
I found it a bit funny that we use ansible's ubuntu packages from OS's that were deemed unsupported. With 18.04 being LTS it's a bit of a heavy hammer to turn way users of this OS, not saying we can fully test the support but rather be listed as 'use at your own risk with feedback welcome'. As for 19.10, it's a short lived OS with an easy upgrade path to 20.4 which users should be aware of, it can be removed as it is EOL'd but I left it in for now.

scripts/ansible
holta: Indeed `apt-add-repository` is not necessary today.

It's just an option if an automated solution delivers more reliability in future.

Similar to Debian & RaspiOS both now moving to `apparmor` (https://github.com/raspberrypi/linux/pull/1698#issuecomment-684989048) and likewise `unattended-upgrades` will almost certainly arrive as default OS policy in coming years, e.g. to manage the flood of published security holes in RaspiOS etc (when "makers" don't do security audits, all too often selfishly creating chaos for others).

holta:
> I found it a bit funny that we use ansible's ubuntu packages from OS's that were deemed unsupported. With 18.04 being LTS it's a bit of a heavy hammer to turn way users of this OS, not saying we can fully test the support but rather be listed as 'use at your own risk with feedback welcome'. As for 19.10, it's a short lived OS with an easy upgrade path to 20.4 which users should be aware of, it can be removed as it is EOL'd but I left it in for now.

Good points.

That said, as discussed on last week's IIAB call, @georgejhunt & I strongly prefer that IIAB look to the future not the past &mdash; supporting far less OS's rather than far more OS's.  Especially when Ubuntu 19.10 is no longer supported even by Canonical.

Things like [Ubuntu 20.10 (apparently with official Desktop support for RPi?)](https://www.omgubuntu.co.uk/2020/06/ubuntu-desktop-support-for-raspberry-pi-4) appear to be a far higher priority for IIAB's educational/field community, if we absolutely must support Yet Another OS/Distro (at great cost in terms of our limited testing resources).  In any case, these decisions should be driven by Field Communities' Actual Needs far more than Developers' Dreamlines.

scripts/ansible
jvonau: The trouble with unattended-upgrades and calling `apt install foo` is the first touch point is in "iiab" with the mass os update, the issue needs to be addressed there. Checking for a lock file and disabling unattended-upgrades while the OS is being altered can be done but lets address that later.

jvonau:
Just looking at http://ppa.launchpad.net/ansible/ansible/ubuntu/pool/main/a/ and drilling down shows the code name ''eoan" is only available for 2.10.X, making this PR suitable for only the 2.10 series of ansible.

holta:
Numerically it's counterintuitive that [ansible-2.10.0](https://docs.ansible.com/ansible/devel/roadmap/COLLECTIONS_2_10.html) is being release AFTER [ansible-base-2.10.1](https://docs.ansible.com/ansible/devel/porting_guides/porting_guide_2.10.html#modules) but FYI here are the _projected new release dates_ for both:

> * 14-Sep-2020: ansible-base-2.10.1 final
> * 15-Sep-2020: ansible-2.10.0rc1 (moved from 10-Sep)
...
> * 22-Sep-2020: If no blockers are found, ansible-2.10.0 final will
release as previously scheduled.

https://groups.google.com/forum/#!topic/ansible-devel/618bSHJ7K64

(Presumably ansible-2.10.0 refers to the "batteries included" larger package with classic/easy install, i.e. presumably it will include BOTH ansible-base-2.10.1 AND common "Ansible Collections" ?  Even despite the numbering anomaly that "parent" package 2.10.0 will include "child" package 2.10.1 ;)

holta:
PR #2518 "Begin transition to Ansible 2.10 (now that ansible-base 2.10.1 is released & appears solid!)" merged, building (in part) on this work.

PR #2515, opened 2020-09-12
detect secondary gateway on wifi interface for no_net_restart

jvonau:
### Fixes bug:
It's not a bug but rather the use of `ansible_default_ipv4` as the normal usage to denote a default gateway that is present on the machine despite other default routes present in the routing table. Now this sort of issue can crop up with 2 wired interfaces with gateways, but shows up in other ways, that layout is not really supported. The use of both wired and wireless as gateway devices show this a weakness against the long standing assumption that only one interface has a default route to the internet. Now with both wired and wireless devices possessing a valid default route, `iiab_wan_iface` which is based on `ansible_default_ipv4.alias` just doesn't cut it. Need to treat the presence an active WiFi gateway the same whether only WiFi has a gateway or if both wired and wireless devices do.
preemptive patch #2497 should it be necessary. 
Update the WiFi detection to include default gateway presence to determine the need for `no_net_restart`
RaspOS wifi only with ap0 active
Reminder that active WiFi to the internet require a reboot it enable the internal hotspot with `no_net_restart`, this is done to ensure install actions that occur post-network configuration will not be disrupted. Same reboot requirement would hold true for the use of iiab-network with an active WiFi connection to the internet.

### DIFF
--- a/roles/network/tasks/detected_network.yml
+++ b/roles/network/tasks/detected_network.yml
@@ -145,6 +145,15 @@
   with_items:
     - "{{ lan_list_result.stdout_lines }}"
 
+- name: Detect wifi gateway active
+  shell: ip r | grep default | grep {{ discovered_wireless_iface }} | wc -l
+  when: discovered_wireless_iface != "none"
+  register: wifi_gateway_found
+
+- name: Set has_wifi_gateway for {{ discovered_wireless_iface }} if gateway is detected
+  set_fact:
+    has_wifi_gateway: True
+  when: wifi_gateway_found is defined and (wifi_gateway_found.stdout|int > 0)
 
 - name: Set iiab_wireless_lan_iface to {{ discovered_wireless_iface }} if not none
   set_fact:
--- a/roles/network/tasks/main.yml
+++ b/roles/network/tasks/main.yml
@@ -12,7 +12,7 @@
     # 0-init/defaults/main.yml, network/tasks/main.yml, debian.yml,
     # detected_network.yml, down-debian.yml, NM-debian.yml, restart.yml,
     # sysd-netd-debian.yml, computed_services.yml, rpi_debian.yml
-  when: discovered_wireless_iface == iiab_wan_iface
+  when: has_wifi_gateway is defined
 
 - name: computed_network
   include_tasks: computed_network.yml
--- a/roles/network/tasks/main.yml
+++ b/roles/network/tasks/main.yml
@@ -6,12 +6,13 @@
 - name: detected_network
   include_tasks: detected_network.yml
 
-- name: "Set 'no_net_restart: True' if discovered_wireless_iface == iiab_wan_iface"
+#- name: "Set 'no_net_restart: True' if discovered_wireless_iface == iiab_wan_iface"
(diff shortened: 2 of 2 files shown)


holta:
@pumpkinpi85 if you can help us further test @jvonau's work here, let us know and we can accelerate merging this into IIAB's mainline/master branch (whether or not it applies to your immediate question at #2497 ?)

Thanks!

tim-moody:
what is no_net_restart? It sounds like no network, but then there would not be a gateway.

jvonau:
> what is no_net_restart? It sounds like no network, but then there would not be a gateway.

Is that a review or a question? Has been in the Code since Nov 2017,.

holta:
I'd like to merge this if there are no objections/improvements in coming hours?

Recap: @jvonau has smoke-tested this, and this PR is very compact, helping safeguard against dual gateways e.g. resulting from dual routes to the Internet (not a theoretical risk, as this caused me and others problems in recent years!)

holta:
Also smoke-tested on Ubuntu 20.04.1 on RPi 4 (10.8.0.62) using `./iiab-network`

holta:
> Also smoke-tested on Ubuntu 20.04.1 on RPi 4 (10.8.0.62) using `./iiab-network`

/usr/bin/iiab-diagnostics output: http://sprunge.us/U5mg11?en

holta:
> what is no_net_restart?

I don't know its precise purpose, but FYI it's used in these 11 files:

```
root@box:/opt/iiab/iiab/roles# grep -rl no_net_restart | sort -f

0-init/defaults/main.yml
network/tasks/computed_services.yml
network/tasks/debian.yml
network/tasks/detected_network.yml
network/tasks/down-debian.yml
network/tasks/main.yml
network/tasks/netplan.yml
network/tasks/NM-debian.yml
network/tasks/restart.yml
network/tasks/rpi_debian.yml
network/tasks/sysd-netd-debian.yml
```

holta:
FYI @jvonau is investigating an issue that appears caused by this PR, on an Ubuntu Server 20.04.1 VM (fresh install of PR #2518) where all 9 stages completed successfully, then it failed in the network stage here:

> TASK [network : Set has_wifi_gateway for none if gateway is detected] **********
fatal: [127.0.0.1]: FAILED! => {"msg": "The conditional check 'wifi_gateway_found is defined and (wifi_gateway_found.stdout|int > 0)' failed. The error was: error while evaluating conditional (wifi_gateway_found is defined and (wifi_gateway_found.stdout|int > 0)): 'dict object' has no attribute 'stdout'\n\nThe error appears to be in '/opt/iiab/iiab/roles/network/tasks/detected_network.yml': line 153, column 3, but may\nbe elsewhere in the file depending on the exact syntax problem.\n\nThe offending line appears to be:\n\n\n- name: Set has_wifi_gateway for {{ discovered_wireless_iface }} if gateway is detected\n  ^ here\nWe could be wrong, but this one looks like it might be an issue with\nmissing quotes. Always quote template expression brackets when they\nstart a value. For instance:\n\n    with_items:\n      - {{ foo }}\n\nShould be written as:\n\n    with_items:\n      - \"{{ foo }}\"\n"}

PR #2521, opened 2020-09-15
revamp ansible to better match python3 versions

jvonau:
The finer point of #2512 was better matching of the python3 versions, bionic uses python3.7 while eoan uses python3.8


jvonau:
Should result in a faster install on Ubuntu 20 as less packages would need to be compiled during the apt install.

holta:
1) There is no 'eoan' .deb package for ansible-base 2.10.1 so I don't (think) this idea works.

2) If in future such a PR was needed, and its testing coverage burdens accounted for (rigorous testing takes longer than coding, generally far longer, which almost everyone forgets!) so we would also want to _verify_ any speed gains are truly worth all the many infra maintenance costs over the years.

   Repeated/broad testing cycles are Not something that volunteer-driven initiatives can organically/scalably sustain ~ so we need to be extraordinarily judicious together ~ about not biting off more than our testing capacity can chew . . .

3) Moving the 'SUCCESS' message to the very end is a good idea at this point.  In future it really should also spell out how to verify (or at least list?) that all necessary Ansible Collections are installed.

ASIDE: This current PR accidentally breaks more than a few explanation flows and line breaks...that were very carefully designed to communicate with and educate relative newcomers (output & comments both, for IIAB newcomers especially).  Again this was not at all intentional.  But FYI several of these are very real usability regressions.

jvonau:
- [Yup not yet,](http://ppa.launchpad.net/ansible/ansible/ubuntu/pool/main/a/ansible-base/?C=M;O=D) so is ansible dropping support of ubuntu in the long run given **all (sans focal)** prior releases are present for 2.10.0-1? I can foresee pip installs in the future...

jvonau:
> ASIDE: This current PR accidentally breaks more than a few explanation flows and line breaks...that were very carefully designed to communicate with and educate relative newcomers (output & comments both, for IIAB newcomers especially). Again this was not at all intentional. But FYI several of these are very real usability regressions.

Lets see moving the OS test and exit to the top is the right thing to do, why say your going to install then exit for non-supported os, that should be right up front. That leaves $GOOD_VER moving a couple of lines, only meaningful if upgrading,
'YOU ARE RUNNING'  is a dup'd line 16, I'll fix that.. Given this change is post #2520 don't see this big "usability regressions" that you are speaking of... the exit moved and releases= was used for substitution.

PR #2519, opened 2020-09-15
2515 fix [for non-WiFi situations]

jvonau:

PR #2524, opened 2020-09-15
2515 fix [another "secondary gateway" precaution, e.g. wired gateways especially]

jvonau:
Should address the rest of #1394 - 2 wired WAN gateways, with or without any LAN devices.
Tested here on RaspOS running ap0 with wifi gateway, both with and without a wired WAN connection.
I even stayed connected over ssh to br0 while running ./iiab-network!!

### DIFF
--- a/roles/network/defaults/main.yml
+++ b/roles/network/defaults/main.yml
@@ -59,7 +59,7 @@ virtual_network_devices: "-e ap0 -e lo -e br0 -e tun -e br- -e docker -e bridge0
 # Set defaults for discovery process as strings
 wifi1: "not found-1"
 wifi2: "not found-2"
-exclude_device: "none"
+exclude_devices: "none"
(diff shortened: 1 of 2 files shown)


holta:
Appears to build on PR #2519

Looks good:
What OS / network topologies should this be tested against?

Related: #2515

jvonau:
This specifically targets 2 wired devices with gateways present but also works when a wifi gateway is in play. Just don't ask for more that 2 wired or one wired plus one wireless gateway active at the same time. That should cover 'in home testing' a bit better. The refactoring in 2519 allowed the 2 wired layout to just drop into place, it was a requested feature at one point and just easy enough to do. This about as far as I would like to take the 'out of box' support, the use of the reserved_device is available for further tinkering.

holta:
What is the intention of `reserved_device` ?

PR #2530, opened 2020-09-16
Hostapd ctl

jvonau:
e0936a656ec35ace8633776ccc293db7345ff1cc from #2511
/iiab/iiab-factory/pull/144

Both want to do the same thing so lets just do it.

### DIFF
--- a/roles/network/templates/hostapd/hostapd.conf.j2
+++ b/roles/network/templates/hostapd/hostapd.conf.j2
@@ -30,3 +30,5 @@ wpa_passphrase={{ hostapd_password }}
 # Use AES, instead of TKIP
 rsn_pairwise=CCMP
 {% endif %}
+ctrl_interface=/var/run/hostapd
+ctrl_interface_group=0
--- a/roles/network/templates/hostapd/hostapd.conf.j2
+++ b/roles/network/templates/hostapd/hostapd.conf.j2
@@ -1,6 +1,8 @@
 # Basic configuration
 
 interface={{ iiab_wireless_lan_iface }}
+ctrl_interface=/var/run/hostapd
+ctrl_interface_group=0
 
 ssid={{ host_ssid }}
 channel={{ host_channel }}
@@ -30,5 +32,3 @@ wpa_passphrase={{ hostapd_password }}
 # Use AES, instead of TKIP
 rsn_pairwise=CCMP
 {% endif %}
-ctrl_interface=/var/run/hostapd
-ctrl_interface_group=0


holta:
Why is this not hostapd's default, does anyone know?

Are there any associated log bloat or privacy risks...with enacting this `/var/run/hostapd` change?

jvonau:
Maybe it is, we have always be just replaced the stock file. On the next fresh install pause to hand install hostapd and inspect the stock file. Doesn't really matter as the masses want it.

holta:
FYI `/etc/hostapd/hostapd.conf` does not exist after running `apt install hostapd` on Ubuntu 20.04.1 (desktop, VM).

holta:
As hostapd_cli commands don't work until we add those 2 lines to /etc/hostapd/hostapd.conf ...it would seem that the hostapd package's example `hostapd.conf` below...contains ideas-to-try more than hostapd defaults:

> root@u20-desk:~# pastebinit /usr/share/doc/hostapd/examples/hostapd.conf
https://paste.ubuntu.com/p/25krJbyPmV/

> root@u20-desk:~# hostapd -v
hostapd v2.9
User space daemon for IEEE 802.11 AP management,
IEEE 802.1X/WPA/WPA2/EAP/RADIUS Authenticator
Copyright (c) 2002-2019, Jouni Malinen <j@w1.fi> and contributors

jvonau:
More or less in the same order of appearance going down from the top.

holta:
`ctrl_interface_group=0` restricts access to root, so let's go ahead and take the risk here (keeping an eye on things!)

holta:
Recap from `hostapd` apt package's pastebin above:

```
# Interface for separate control program. If this is specified, hostapd
# will create this directory and a UNIX domain socket for listening to requests
# from external programs (CLI/GUI, etc.) for status information and
# configuration. The socket file will be named based on the interface name, so
# multiple hostapd processes/interfaces can be run at the same time if more
# than one interface is used.
# /var/run/hostapd is the recommended directory for sockets and by default,
# hostapd_cli will use it when trying to connect with hostapd.
ctrl_interface=/var/run/hostapd

# Access control for the control interface can be configured by setting the
# directory to allow only members of a group to use sockets. This way, it is
# possible to run hostapd as root (since it needs to change network
# configuration and open raw sockets) and still allow GUI/CLI components to be
# run as non-root users. However, since the control interface can be used to
# change the network configuration, this access needs to be protected in many
# cases. By default, hostapd is configured to use gid 0 (root). If you
# want to allow non-root users to use the contron interface, add a new group
# and change this value to match with that group. Add users that should have
# control interface access to this group.
#
# This variable can be a group name or gid.
#ctrl_interface_group=wheel
ctrl_interface_group=0

jvonau:
If you are truly concerned about security I would be addressing services that run as root.

holta:
> If you are truly concerned about security I would be addressing services that run as root.

Understanding unintentional privacy leaks (arising from any hostapd-related monitoring) is the primary question here.  And if there are any, protecting people.

OS security is secondary, a universal and important concern indeed, but not one that any single person can solve in 2020..

Finally log bloat (or more positively, community ownership of & meaningful visibility into their own usage analytics) is a GIANT question staring everyone in the face, where I believe we can make more headway...

Issue #2475, opened 2020-08-05
MySQL ERROR 1045 (28000): Access denied for user 'root'@'localhost' ...FATAL ERROR: Upgrade failed


holta:
Is this very ugly series of errors in `/var/log/syslog` fixable?  Even if some suggest it's possibly inconsequential?

```
Aug  5 16:18:39 box /etc/mysql/debian-start[1064]: Upgrading MySQL tables if necessary.
Aug  5 16:18:39 box /etc/mysql/debian-start[1070]: /usr/bin/mysql_upgrade: the '--basedir' option is always ignored
Aug  5 16:18:40 box /etc/mysql/debian-start[1070]: Looking for 'mysql' as: /usr/bin/mysql
Aug  5 16:18:40 box /etc/mysql/debian-start[1070]: Looking for 'mysqlcheck' as: /usr/bin/mysqlcheck
Aug  5 16:18:40 box /etc/mysql/debian-start[1070]: Version check failed. Got the following error when calling the 'mysql' command line client
Aug  5 16:18:40 box /etc/mysql/debian-start[1070]: ERROR 1045 (28000): Access denied for user 'root'@'localhost' (using password: [REDACTED]
Aug  5 16:18:40 box /etc/mysql/debian-start[1070]: FATAL ERROR: Upgrade failed
Aug  5 16:18:40 box /etc/mysql/debian-start[1095]: Checking for insecure root accounts.
```

Excerpted from short-term pastebin https://paste.ubuntu.com/p/J7f7SG3vdW/

(But this seemingly affects all installs onto Raspberry Pi OS...and possibly other/all installs onto Debian-related distros ...as suggested by debian-start in the error messages ??)

Ref:

#1714 'MySQL database password-changing failure during fresh install of IIAB [using "sudo iiab"] on Ubuntu 19.04 [and now 18.04 too...whereas 19.10 & Debian 10.x work!] -- Ansible Python 3 needs python3-distutils'

holta:
While oddly not catastrophic to IIAB's general functioning, this is a longstanding issue @jvonau has seen for many months.

He & @tim-moody suspect the password is being stored in the wrong file / wrong place (i.e. in a different place from where the service is expecting it).

Likely solvable by copying the password/credential into the right place?!

Likely Related:

- /root/.my.cnf
- /etc/mysql/my.cnf -> /etc/alternatives/my.cnf
- /etc/mysql/my.cnf.fallback
- /etc/mysql/mariadb.cnf

[...]

PR #2533, opened 2020-09-18
replacement for 2500

jvonau:
Replacement for the rudely reverted #2500 with new ability to back fit the fix for #2475 to recent installs

### DIFF
--- a/roles/0-init/defaults/main.yml
+++ b/roles/0-init/defaults/main.yml
@@ -8,6 +8,8 @@
 first_run: False
 rpi_model: none
 xo_model: none
+do_upgrade: False
+do_reinstall: False
 gw_active: False
 internet_available: False
 discovered_wan_iface: none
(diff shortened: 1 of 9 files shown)


jvonau:
more disregard for pending PRs

tim-moody:
`# use 'Update' in the name line when the role is not optional`

how does 'update' mean 'required'?

jvonau:
For the most part "install" is conditional upon local_vars being set to True for the role, with install.yml being present in the role, pylib's doesn't have an install.yml file and no entries in local_vars.yml. 'Update' is used as a visual reminder that the role is not optional. Would "Mandatory Update" be more descriptive?

tim-moody:
I like Mandatory. looking at packages.yml in 2-common, I see they all say install

jvonau:
forking

Issue #2375, opened 2020-04-29
RFQ: Adaptive Network Service


tim-moody:
Here are some ideas I've been thinking about for awhile. I notice that jvonau:ap0+wifi_up_down2 is starting in this direction so I'd like to suggest going even further:

To configure networking what about a different approach? Instead of tweaking a bunch of configuration files and trying to make third party software do the right thing using commands at install time, just let the network come up in its base configuration at boot, which includes at least one way of reaching the server over the network. When its done, the adaptive-network service runs, examines the environment, looks at the variables that define the desired network, and reconfigures it with cli on the fly. It does each step in the order needed and only for hardware present. 

If there is a problem, nothing is carved in stone. If adaptive-network fails it could turn itself off either with disable or a flag somewhere. Failure means there is no way to reach the server. To change a running server, you just change the variables and restart the service. There are probably different versions of this service for different platforms and the appropriate one is installed with ansible.

jvonau:
Thank for noticing ap0+wifi_up_down2, it's a working prototype on U-20. The issue at hand due to firmware limitations is one of keeping the channels in sync between wlan0 for upstream use and ap0 for AP use. Currently on raspbian the dhcpcd hook does the editing of hostapd.conf on the fly but doesn't restart the service as a restart deletes ap0. We must have the device cloned before starting wpa_ supplicant or nothing works, a reboot aligns the channels up. There is no hook like that currently for ubuntu because I was thinking with the cloning split apart from starting hostapd we might be able to (re)start hostapd once the channel was adjusted in hostapd.conf on the fly in dhcpcd's hooks or one written for ubuntu just haven't explored that far yet. The test-wifi routine and related service is intended for boot time staging so the services come up in a sane order. Come to think about it that would be the place to slip the netplan country= workaround in. This all works ok with the upstream channel not changing, needs testing for when the wifi channel does change.

tim-moody:
no meaningful comments. closing

PR #2537, opened 2020-09-21
Samba role standardized/modularized, for runrole etc

### LINKED ISSUE #2536: Design decisions made by committee, later warnings ignored. Is the release cycle open again?
### Expected behavior and actual behavior:
Not have roles install when the local_vars.yml is set to False
### Steps to reproduce the problem:
grep samba /etc/iiab/local_vars.yml 
samba_install: False
samba_enabled: False
sudo ./runrole samba or sudo ./runroles samba

>TASK [Set 'samba_installed: True'] ******************************************************
ok: [127.0.0.1]

>TASK [Add 'samba' variable values to /etc/iiab/iiab.ini] ********************************
changed: [127.0.0.1] => (item={'option': 'name', 'value': 'Samba'})
changed: [127.0.0.1] => (item={'option': 'description', 'value': '"Samba is a Microsoft-compatible network file system that re-implements SMB/CIFS (Common Internet File System)."'})
changed: [127.0.0.1] => (item={'option': 'enabled', 'value': False})

Now this has set the user up to fail at the next run of ansible when validate.yml catches the local_vars|iiab_state mismatch of samba_install{ed) in this case. Breaks the basic presumption of False is false.

layout of meta agreed to by committee  https://github.com/iiab/iiab/pull/2052
lockout relation change by committee https://github.com/iiab/iiab/pull/2126
highjack of PR removing installed check from main,yml https://github.com/iiab/iiab/pull/2134
moving of installed from main.yml  ***warning***  https://github.com/iiab/iiab/pull/2142
moving of installed to meta https://github.com/iiab/iiab/pull/2190/files

holta:
Resolves (Partially!) #2536

In recent years, IIAB's Samba role was unmaintained as several of us were thinking of deprecating it.  However this created sloppy usage with runrole etc.  Thanks @jvonau for identifying this within #2536.

To avoid those risks, this PR modernizes IIAB's Samba role, so it checks variables `samba_install` and `samba_enabled`, as other roles have been doing for many months now.

Tested on Ubuntu Desktop 20.04.1

(At the same time let's keep open the option of deprecating / de-supporting Samba completely in future, as it's very rarely used and might be more trouble than it's worth!)

### DIFF
--- a/roles/1-prep/tasks/main.yml
+++ b/roles/1-prep/tasks/main.yml
@@ -78,12 +78,12 @@
 - name: SSHD
   include_role:
     name: sshd
-  # has no "when: XXXXX_install" flag
+  #when: sshd_install | bool    # Flag might be created in future?
 
 - name: IIAB-ADMIN
   include_role:
     name: iiab-admin
-  # has no "when: XXXXX_install" flag
+  #when: iiab-admin_install | bool    # Flag might be created in future?
 
 - name: OPENVPN
   include_role:
--- a/roles/3-base-server/tasks/main.yml
+++ b/roles/3-base-server/tasks/main.yml
@@ -27,6 +27,7 @@
 - name: WWW_BASE (WWW_OPTIONS should be installed later)
   include_role:
     name: www_base
+  #when: www_base_install | bool    # Flag might be created in future?
 
 - name: Recording STAGE 3 HAS COMPLETED =====================
   lineinfile:
--- a/roles/4-server-options/tasks/main.yml
+++ b/roles/4-server-options/tasks/main.yml
@@ -5,7 +5,10 @@
     path: "{{ iiab_state_file }}"    # /etc/iiab/iiab_state.yml
   register: iiab_state_file_check
 
-# STRICT CHECK in case {iiab-stages.yml, run-one-role.yml, etc} omit this:
+# 2020-09-21: Why is checking for iiab_state.yml in Stage 4 out of curiosity?
+# Possibly move this to Stage 0?  Either way...this is a STRICT CHECK in case
+# {iiab-stages.yml, run-one-role.yml, etc} omit the file from "vars_files:"
+
 - name: ENFORCE precondition that {{ iiab_state_file }} MUST exist
   assert:
     that: iiab_state_file_check.stat.exists
--- a/roles/4-server-options/tasks/main.yml
+++ b/roles/4-server-options/tasks/main.yml
@@ -1,19 +1,25 @@
 # Server Options
 
 - name: ...IS BEGINNING ==================================
-  file:
-    path: "{{ iiab_state_file }}"
-    state: touch
+  stat:
+    path: "{{ iiab_state_file }}"    # /etc/iiab/iiab_state.yml
+  register: iiab_state_file_check
 
-- name: Install IIAB's python libs
+# STRICT CHECK in case {iiab-stages.yml, run-one-role.yml, etc} omit this:
+- name: ENFORCE precondition that {{ iiab_state_file }} MUST exist
+  assert:
+    that: iiab_state_file_check.stat.exists
+
+- name: Install pylibs (IIAB's python libs)
   include_role:
     name: pylibs
+  #when: pylibs_install | bool    # Flag might be created in future?
 
 - name: Install named / BIND
   include_tasks: roles/network/tasks/named.yml
   when: named_install | bool
 
-- name: Installing dhcpd
+- name: Install dhcpd
   include_tasks: roles/network/tasks/dhcpd.yml
   when: dhcpd_install | bool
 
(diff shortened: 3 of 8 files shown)


holta:
FYI these 4 roles fall in the same bucket as Samba, and could/should also be refined:

- cups
- minetest
- phpmyadmin
- transmission

The asterisks near the bottom of this [README.md](https://github.com/iiab/iiab/blob/master/roles/nginx/README.md) help keep track of these roles...that need improvement (or need to deprecated, as the case may be!)

i.e. we can-and-should deprecate things that incur ongoing maintenance/security costs, and have no known community of users &mdash; e.g. cups, phpmyadmin, etc ?

holta:
> FYI these 4 roles fall in the same bucket as Samba, and could/should also be refined:

See also @jvonau's longer list at https://github.com/iiab/iiab/issues/2536#issuecomment-696200743

(Of course several of these should be deprecated, rather that generating yet more code debt, especially when communities of users simply don't exist for that functionality.)

jvonau:
I still find it rather interesting and amusing that the test for a proper boolean is after the test for "is true". Makes more sense to ensure that the variable is a proper boolean before checking if the boolean variable is set to 'True'.

holta:
> I still find it rather interesting and amusing that the test for a proper boolean is after the test for "is true". Makes more sense to ensure that the variable is a proper boolean before checking if the boolean variable is set to 'True'.

A role's x_**install** is tested to be boolean _while_ it's being tested "is sameas true" on Line 11: (as we mandate that it must be both!)

```
  assert:
    that: samba_install is sameas true
```

Whereas a role's x_**enabled** can have either value (e.g. True or False) so is only tested to be boolean on Line 17:

```
 assert:
    that: samba_enabled | type_debug == 'bool'

jvonau:
> Whereas a role's x_**enabled** can have either value (e.g. True or False) so is only tested to be boolean on Line 17:
> 
> ```
>  assert:
>     that: samba_enabled | type_debug == 'bool'
> ```

What is the reasoning behind this? It's a duplicate of the code in validate.yml that is suppose to catch this error much earlier.
I can see this in place while working up a brand new role but once the role is merged the new role needs to be added to {{ vars_checklist }} in validate.yml anyway. Might want to check for missing roles. 

>- name: Assert that {{ vars_checklist | length }} "XYZ_enabled" vars are all... type boolean (NOT type string, which can invert logic!)
  assert:
    that: "{{ item }}_enabled | type_debug == 'bool'"

Fastest code is code that doesn't run...

holta:
> What is the reasoning behind this? It's a duplicate of the code in validate.yml

Safe software enforces its own preconditions, at different levels of modularity/granularity.

i.e. this is duplicate code very much on purpose, to quickly surface errors no matter how code is reused in future.

> Might want to check for missing roles.

Indeed: [validate_vars.yml](https://github.com/iiab/iiab/blob/master/roles/0-init/tasks/validate_vars.yml) needs regular maintenance, as vars come and go.

Roles enforcing their own preconditions...helps here (to a degree) though doesn't surface errors near the very beginning of iiab-install, iiab-configure etc &mdash; as validate_vars.yml seeks to do.

holta:
FYI this PR annotates AzuraCast as "UNMAINTAINED" as development (IIAB-AzuraCast integration) is not moving forward at the moment.

Please contact @m-anish (Anish Mangal) if you can help change this!

holta:
That's enough for 1 PR.  To be continued in another PR.

Not that [all ~15 roles](https://github.com/iiab/iiab/issues/2536#issuecomment-696200743) should be carried forward (as there's much dead weight here, not serving anybody in 2020, that needs to be eliminated completely!)

But yes a goodly number of them should be cleaned up (as was done for Samba in this PR).

jvonau:
> > Might want to check for missing roles.
> 
> Indeed: [validate_vars.yml](https://github.com/iiab/iiab/blob/master/roles/0-init/tasks/validate_vars.yml) needs regular maintenance, as vars come and go.

That is your baby.

> Roles enforcing their own preconditions...helps here (to a degree) though doesn't surface errors near the very beginning of iiab-install, iiab-configure etc — as validate_vars.yml seeks to do.

You should then have no issue with #2539, stop the user if sudo is not used right up front and add 'sudo' to the user facing feedback. Move validate to the very first ansible role to lead the parade. With runrole being more of a developer|power user tool targeting a single role, lets keep the noise to a minimum. Now iiab-network should really validate a small subset of user settings but we have been lucky so far. I really don't want to slow down the runtime needlessly with an eye towards something like #2375 with a smaller subset of network running during|post boot via systemd. Heck I have even done a full install or iiab-configure using a unit file and one systemd command, but you have to watch syslog or journalctl to see it running.

jvonau:
1d3627510573f343fa140b96a166669bf47e87d0 is interesting, most the top level yaml files use iiab_state.yml as a reference and would blowup without it. The only routine that might bump into this is 'runroles' when 4-server-options is used in the list of roles.

PR #2539, opened 2020-09-21
Basic Programming

jvonau:
Move validate.yml to it's own role and call before 0-init, where required.
iiab-network  & runrole deal with a single role where checking 50+ unrelated entries is just a waste of time.
Add check-vars utility for pre-validating local_vars.yml
Feedback should be aimed at the average user, add sudo to what the user sees as suggested commands
Add sudo check to scripts that need root privileges.


jvonau:
more total disregard for pending PR's

jvonau:
forking

PR #2527, opened 2020-09-15
Communicate Ansible 2.10 options, to implementers & maintainers

holta:
Building on #2479, #2481 and PR's #2512, #2518, #2520, #2521, #2523.

Issue #2538, opened 2020-09-21
http://pantry.learningequality.org is offline for 24+ hours, preventing IIAB installs that need KA Lite's en.zip [and multilingual .torrent files to install content]


holta:
@benjaoming is there someone at Learning Equality that we should ask about this?

The main file in question is:

http://pantry.learningequality.org/downloads/ka-lite/0.17/content/contentpacks/en.zip

Thanks @pumpkinpi85 for reporting this!

Tangentially related: #871

holta:
@ivanistheone do you know if access to this file (en.zip) can somehow be restored?

Or might you know who we might talk to at Learning Equality, to get some kind of access to it?

(Thanks if possible !!)

holta:
Just FYI, KA Lite's content torrent files in ~7 languages are also currently offline:
https://ka-lite.readthedocs.io/en/latest/usermanual/userman_admin.html#downloading-videos-in-bulk

As these live here &mdash; on this same host/site (panty.learningequality.org) that appears offline for about 48 hours now:
http://pantry.learningequality.org/downloads/ka-lite/0.17/content/

jvonau:
I did an install last night into this morning had no issues.
>Install KA Lite's mandatory 0.9 GB English Pack... (en.zip)
--2020-09-22 05:52:15--  http://pantry.learningequality.org/downloads/ka-lite/0.17/content/contentpacks/en.zip
Resolving pantry.learningequality.org (pantry.learningequality.org)... 104.18.9.53, 104.18.8.53, 2606:4700::6812:935, ...
Connecting to pantry.learningequality.org (pantry.learningequality.org)|104.18.9.53|:80... connected.
HTTP request sent, awaiting response... 200 OK
Length: 929916955 (887M) [application/zip]
Saving to: 'en.zip'
en.zip                 100%[=========================>] 886.84M  4.40MB/s    in 4m 5s   
2020-09-22 05:56:21 (3.61 MB/s) - 'en.zip' saved [929916955/929916955]

holta:
> I did an install last night into this morning had no issues.

Ok, Strange!  I've tried different ISP's over 3 days with no luck, i.e. personally I haven't been able to access http://pantry.learningequality.org Sunday, Monday and now Tuesday.  Similar to @pumpkin85, who has a different ISP (in a different country).

As confirmed each day by https://downforeveryoneorjustme.com/pantry.learningequality.org

[...]

PR #2541, opened 2020-09-22
Standardize 9 core roles (basic var checks, modularize) + 3 lesser roles now record to iiab_state.yml [firmware role beefed up with better explanations, this works w/o local_vars.yml]

holta:
This is a good start towards cleaning up 12 + 3 more imperfect roles, as alluded to in #2536.

Similar to PR #2537 (Samba fixup), these 9 core roles...also now have essential var checks added...and are more modularized:

- bluetooth (Stage 4)
- cups (Stage 4)
- azuracast (Stage 6, UNMAINTAINED)
- mosquitto (Stage 6)
- monit (Stage 8)
- phpmyadmin (Stage 8)
- transmission (Stage 8)
- vnstat (Stage 8)
- minetest (Stage 9)

In addition to these 3 lesser roles...which newly record their X_installed to /etc/iiab/iiab_state.yml:

- iiab-admin (Stage 1)
- pylibs (Stage 4)
- firmware (Network Stage)

Similar to these 3 lesser roles...that already recorded their Y_installed to /etc/iiab/iiab_state.yml:

- sshd (Stage 1)
- www_base (Stage 3)
- www_options (Stage 4)

This PR was extensively functionality tested on Ubuntu Desktop 20.04.1 LTS.
I'll also run a BIG-sized test from scratch to increase coverage.

Tangentially also related to:

- #1437 "[Debian 10] MongoDB not installable due to missing packages [for Sugarizer]"
- #1849 "@m-anish & @tim-moody ask: can systemd now replace Monit? [Debian 10 Buster no longer includes Monit]"
- PR #2527 "Communicate Ansible 2.10 options, to implementers & maintainers"


holta:
Smoke-tested successfully on 193-deb10-srv-BIG-PR2541 = 10.8.0.70

ASIDE: KA Lite was not included, as a result of their ongoing hosting site failure: #2538

holta:
For good measure, another BIG-sized smoke test is now running on 32-bit RaspiOS Lite @ 183-rpi4-2g-lite-BIG-PR2541 = 10.8.0.54

jvonau:
Think pylibs & firmware should not have the 'installed' treatment, the recording of the completion is fine. We are able to push bug-fixes without needing to 'reinstall' the role for what are support libraries or functions, given pylibs randomly gets updated. I would rather not have to override the installed flag to update a couple of files that admin-console and friends depend on.

jvonau:
Changing iiab-admin to record is nice but the actual user editable variables in play that should be tested are different due to the sub task within the role see: https://github.com/iiab/iiab/pull/2540/commits/a003b637a12826776a5f3c4b186062a975a6d02b

jvonau:
> Think pylibs & firmware should not have the 'installed' treatment, the recording of the completion is fine. We are able to push bug-fixes without needing to 'reinstall' the role for what are support libraries or functions, given pylibs randomly gets updated. I would rather not have to override the installed flag to update a couple of files that admin-console and friends depend on.

reviewed up to 8279f64, no issues that I can spot easily.

holta:
What was /usr/sbin/check-firmware.sh has become `/usr/sbin/iiab-check-firmware` FYI, to make it more findable and preserve IIAB naming norms.

holta:
This PR was very heavily tested yesterday and today, so I'm merging now.

If further bugs arise, we'll deal with those alongside the mountain of IIAB 7.2 release communications work that is ongoing.

holta:
> Think pylibs & firmware should not have the 'installed' treatment, the recording of the completion is fine. We are able to push bug-fixes without needing to 'reinstall' the role for what are support libraries or functions, given pylibs randomly gets updated. I would rather not have to override the installed flag to update a couple of files that admin-console and friends depend on.

This thinking makes sense, with a twist...that being that these 6+ (formerly 3+) XYZ_install flags do not block reinstalls (or block anything) but instead serve as a very helpful rapid/visual reminder in `/etc/iiab/iiab_state.yml` as to what is and is not installed:

- sshd (Stage 1)
- iiab-admin (Stage 1)
- www_base (Stage 3)
- pylibs (Stage 4)
- www_options (Stage 4)
- firmware (Network Stage)

Yes the [firmware role](https://github.com/iiab/iiab/tree/master/roles/firmware) is a bit different than the others, including an unconventional flag, so I've begun carefully documenting that within its code i.e. see the 3 files within `/opt/iiab/iiab/roles/firmware/tasks`

jvonau:
1a3f386 Now other than shifting the contents of install.yml to download.yml and the contents of main.yml to install.yml what has this accomplished? Both firmware_downloaded and the newer firmware_installed flags are present as informational only in the iiab_state.yml file as roles/tasks/main.yml runs firmware when rpi hardware is present. Both flags denote a one time action called from network, that is the reason why I went with the unique flag to begin with, to denote an out of the ordinary calling of a role from network as every thing else is done in stages 1 through 9. That aspect of the flag was made clear when you and I were working up #2472 so I'm a bit at a loss for why this was required.

You should really learn how to use `git mv` in place of copy & pasting to rename files (iiab-check-firmware) so the git history can track the progression of the files and who was involved with the files in the past, unless of course your goal **is** to erase the history to create the illusion that you are the original author.

PR #2544, opened 2020-09-24
iiab-diagnostics: collect last 100 lines from iiab-configure.log (generated by ./iiab-configure)

holta:
`/usr/bin/iiab-diagnostics` pastebin snapshots are getting a bit overweight (around 2000 lines) but `/opt/iiab/iiab/iiab-configure.log` seems important, as generated by the `./iiab-configure` command, as is newly recommended here...

http://FAQ.IIAB.IO > "24 What is local_vars.yml and how do I customize it?"

Builds on PRs #2439 #2476 #2485


holta:
If files and dirs need to be added/removed from the list here...
https://github.com/iiab/iiab/blob/master/scripts/iiab-diagnostics#L106-L219

...let's definitely stay current (and make those adjustments).

So remote debugging of IIAB's is as efficient as...is realistically possible ;-)

PR #2557, opened 2020-10-03
Document that admin_console_install and _enabled now work

holta:
Per https://github.com/iiab/iiab-factory/pull/151


holta:
For supplemental testing coverage, I will retest this and iiab/iiab-factory#151 on another OS (RaspiOS) in the coming hour.

In addition to Debian 10.6 where that was already tested.

Ref PRs: #2558 #2559

PR #2558, opened 2020-10-05
Mistaken Assumption that Released images of Medical didn't install admin-console

jvonau:
Alters local_vars_medical.yml to match released images.
Complements #2557 and https://github.com/iiab/iiab-factory/pull/151

### DIFF
--- a/vars/local_vars_medical.yml
+++ b/vars/local_vars_medical.yml
@@ -16,3 +16,5 @@ openvpn_handle: "Medical IIAB - Put Your Name Here"
 # By default
 # kiwix
 # awstats
+admin_console_install: False
+admin_console_enabled: False


tim-moody:
>Alters local_vars_medical.yml to match released images.

what released image?

wouldn't a better title be modified local_vars_medical.yml causes admin console not to install?

holta:
> > Alters local_vars_medical.yml to match released images.
> 
> what released image?

I presume @jvonau is talking about medical IIAB images published to Archive.org

Do those all IIAB images not contain Admin Console?

@tim-moody would know.

tim-moody:
>Do those all IIAB images not contain Admin Console?

They always have and still do, so that users of these content ready images can add content.

tim-moody:
Use of admin_console_install and enabled by install scripts is not a good idea. It will lead to contention as anyone subsequently running iiab-admin-console/install will get a different result. This should be a flag to iiab-install.

holta:
> > Do those all IIAB images not contain Admin Console?
> 
> They always have and still do, so that users of these content ready images can add content.

Ok, @jvonau must have assumed otherwise.

Innocent mistake, no worries.

As several IIAB images have been produced without Admin Console overs the years, by @tim-moody and others.

Prob we should close this PR now.

tim-moody:
>As several IIAB images have been produced without Admin Console overs the years, by @tim-moody 

I don't recall any; certainly none that were published.

jvonau:
I was thinking more of the curated images available from https://mdwiki.org/wiki/WikiProjectMed:Internet-in-a-Box, didn't think one would ship a single purpose image with the the admin-console enabled as one of the slideshows used in an IIAB [presentation](https://upload.wikimedia.org/wikipedia/commons/0/0a/Internet_in_a_Box_for_Education.pdf) which mentions only kiwix is used for medical, but that was at a time when 'menus' was not part of the admin-console install but from iiab-menu matching what appears in local_vars_medical.yml now. Where are the scrapings stored or are they regenerated on demand from the admin-console? I'll assume the scraped sites are loaded outside of admin-console due to the size of the data involved.

tim-moody:
>I was thinking more of the curated images available from https://mdwiki.org/wiki/WikiProjectMed:Internet-in-a-Box, didn't think one would ship a single purpose image with the the admin-console enabled 

I can't say for sure if cmdsrv was disabled on the 16G Nigeria image. I know I thought about it, but I doubt admin console was not installed as it was used by the curator to create those images. I'll grant that a couple of years ago before this code was written it might have been useful, but not in the last two years or so.

>I'll assume the scraped sites are loaded outside of admin-console due to the size of the data involved.

By scraped I assume you mean cdc and wikem and maybe HealthPhone. All are added to the oer2go catalog by admin console so they can be installed like any other module. Zims are installed by curators with the admin console. As you mention menus are now maintained with admin console and have been for some time now.

jvonau:
Understood, needs to be enabled to gather the content, once gathered and the menu tailored to suit, may be optionally disabled to conserve resources by the end user.

jvonau:
Let take this conversation over to https://github.com/iiab/iiab-factory/pull/151

PR #2564, opened 2020-10-07
util to return state of var

tim-moody:
takes var name as argument and returns value favoring local vars over default vars

### DIFF
--- a/roles/0-init/tasks/main.yml
+++ b/roles/0-init/tasks/main.yml
@@ -30,6 +30,12 @@
     dest: /usr/bin/
     mode: '0755'
 
+- name: Copy /opt/iiab/iiab/scripts/iiab-vars-util.py to /usr/bin/iiab-vars-util.py
+  copy:
+    src: "{{ iiab_dir }}/scripts/iiab-vars-util.py"
+    dest: /usr/bin/
+    mode: '0755'
+
 - name: Create globally-writable directory /etc/iiab/diag so non-root users can run iiab-diagnostics
   file:
     state: directory
--- a/roles/pylibs/templates/iiab_lib.py
+++ b/roles/pylibs/templates/iiab_lib.py
@@ -6,6 +6,7 @@ import os
 import json
 import subprocess
 import shlex
+import yaml
 import xml.etree.ElementTree as ET
 import iiab.iiab_const as CONST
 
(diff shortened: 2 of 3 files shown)


holta:
Quick questions:

- Does this provide a way to disambiguate between `null` (undefined Ansible variables) and Ansible variables set to the 4-letter string "null" ?
- Maybe not important today, but perhaps in future this should provide a way to verify the type of Ansible variables? (which oftentimes in the past have been non-strings by accident!)
- Can the stanza proposed for roles/0-init/tasks/main.yml move up or down?  Currently it is breaking the flow of `iiab-diagnostics` & `/etc/iiab/diag` (2 stanzas that work together).

holta:
This can be deferred to IIAB 8.0 if more discussion and testing is necessary.

(Certainly if other assorted code by many different people will be relying on this for many years to come &mdash; we should battle-test and clearly document these code safety techniques for all &mdash; either now or in future.)

holta:
If this PR proves useful/fast/safe/popular/etc[*] it could/should probably use templates to incorporate our longstanding path vars like:

- {{ iiab_etc_path }}/local_vars.yml == {{ iiab_local_vars_file }}
- {{ iiab_dir }}/vars/default_vars.yml

...auto-sourced from https://github.com/iiab/iiab/blob/master/vars/default_vars.yml#L19-L30

[*] A standardized approach would certainly be wise long-term, to help developers avoid falling in all the same old Ansible-inspired traps, time and again.

holta:
This PR (or things like it) are moving us in the right direction.

Towards IIAB configuration variables that new-and-veteran devs/ops/implementers can understand and depend on (reliably, efficiently...and without drowning in complexity).

i.e. (1) code safety _and_ (2) code clarity — _both_.

Not at all easy :/  But We're Getting There.

<BR>
ASIDE: hopefully Ansible too, after almost a decade now.

holta:
Discussed at length during our IIAB call a couple hours ago (http://minutes.iiab.io).

In Short: something like this PR is a really good idea.

Not so much for the current 4-5 instances of grep checking local_vars.yml etc (/usr/sbin/iiab 1-line installer, iiab-check-firmware, iiab-support, the new iiab-install-map-region and possible commands like iiab-hotspot-on?) which work very well today[*] _but more for the future..._

i.e. so IIAB devs (no matter their age & experience levels) stop falling in all the usual Ansible traps &mdash; writing clean code, no matter the language &mdash; _without_ their time being wasted on the soup-of-subtleties around Ansible variables, be they string-type or non-string-type, etc...

[*] these 4-5 can be retrofitted in the coming year as nec, to promote hygienic code re-use etc!

PR #2566, opened 2020-10-09
Fix for Graphical and TTY Logins [clean up auto-checking for iiab-admin's published password]

holta:
Resolves #2561 and its very serious `sudo` problem.

By greatly simplifying `/etc/profile.d/sshpwd-profile-iiab.sh` since the prior version was quite broken, for those with graphical desktops especially, as TK & @shanti-bhardwa confirmed.

Needs more testing on more OS's than just Ubuntu Desktop 20.04.1 (where it was smoke-tested & functionality tested) so this important fix can hopefully be merged in short order.

Refs: #1252, PR #1253, #1537, PR #1551, #2432, PR #2434, PR #2571

PR #2570, opened 2020-10-12
Respect iiab_admin_user_install better [and iiab-admin design discussion]

jvonau:
Install {{ iiab_admin_user }} password checks only when asked to create user.
Augments #2566 by allowing enduser to bypass iiab-admin creation should in field testing show
stange behavior with unknown/untested desktops and/or OSs.

### DIFF
--- a/roles/iiab-admin/tasks/admin-user.yml
+++ b/roles/iiab-admin/tasks/admin-user.yml
@@ -51,3 +51,35 @@
   file:
     path: /etc/sudoers
     mode: 0440
+
+- name: Install /etc/profile.d/sshpwd-profile-iiab.sh from template, to issue warnings (during shell/ssh logins) if iiab-admin password is the default
+  template:
+    src: sshpwd-profile-iiab.sh
+    dest: /etc/profile.d/
+    mode: '0644'
+
+- name: Is this LXDE-pi?
+  stat:
+    path: /etc/xdg/lxsession/LXDE-pi
+  register: lx
+
+- name: "Likewise for Raspbian, installing: /etc/xdg/lxsession/LXDE-pi/sshpwd-lxde-iiab.sh"
+  template:
+    src: sshpwd-lxde-iiab.sh
+    dest: /etc/xdg/lxsession/LXDE-pi/
+    mode: '0755'
+  when: lx.stat.isdir is defined and lx.stat.isdir and is_raspbian and is_debuntu
+
+# 2019-03-07: This popup (/etc/xdg/lxsession/LXDE-pi/sshpwd-lxde-iiab.sh) does
+# not actually appear when triggered by /etc/xdg/autostart/pprompt-iiab.desktop
+# (or pprompt.desktop as Raspbian has working since 2018-11-13!)  Too bad as it
+# would be really nice to standardize this popup across Ubermix & all distros..
+# Is this a permissions/security issue presumably?  Official autostart spec is:
+# https://specifications.freedesktop.org/autostart-spec/autostart-spec-latest.html
+# Raspbian's 2016-2018 evolution here: https://github.com/iiab/iiab/issues/1537
+
+- name: Put line in /etc/xdg/lxsession/LXDE-pi/autostart to run the above (raspbian)
+  lineinfile:
+    path: /etc/xdg/lxsession/LXDE-pi/autostart
+    line: "@/etc/xdg/lxsession/LXDE-pi/sshpwd-lxde-iiab.sh"
+  when: lx.stat.isdir is defined and lx.stat.isdir and is_raspbian and is_debuntu
--- a/roles/iiab-admin/tasks/main.yml
+++ b/roles/iiab-admin/tasks/main.yml
@@ -3,38 +3,6 @@
 
 - include_tasks: access.yml
 
-- name: Install /etc/profile.d/sshpwd-profile-iiab.sh from template, to issue warnings (during shell/ssh logins) if iiab-admin password is the default
-  template:
-    src: sshpwd-profile-iiab.sh
-    dest: /etc/profile.d/
-    mode: '0644'
-
-- name: Is this LXDE-pi?
-  stat:
-    path: /etc/xdg/lxsession/LXDE-pi
-  register: lx
-
-- name: "Likewise for Raspbian, installing: /etc/xdg/lxsession/LXDE-pi/sshpwd-lxde-iiab.sh"
-  template:
-    src: sshpwd-lxde-iiab.sh
-    dest: /etc/xdg/lxsession/LXDE-pi/
-    mode: '0755'
-  when: lx.stat.isdir is defined and lx.stat.isdir and is_raspbian and is_debuntu
-
-# 2019-03-07: This popup (/etc/xdg/lxsession/LXDE-pi/sshpwd-lxde-iiab.sh) does
-# not actually appear when triggered by /etc/xdg/autostart/pprompt-iiab.desktop
-# (or pprompt.desktop as Raspbian has working since 2018-11-13!)  Too bad as it
-# would be really nice to standardize this popup across Ubermix & all distros..
-# Is this a permissions/security issue presumably?  Official autostart spec is:
-# https://specifications.freedesktop.org/autostart-spec/autostart-spec-latest.html
-# Raspbian's 2016-2018 evolution here: https://github.com/iiab/iiab/issues/1537
-
-- name: Put line in /etc/xdg/lxsession/LXDE-pi/autostart to run the above (raspbian)
-  lineinfile:
-    path: /etc/xdg/lxsession/LXDE-pi/autostart
-    line: "@/etc/xdg/lxsession/LXDE-pi/sshpwd-lxde-iiab.sh"
-  when: lx.stat.isdir is defined and lx.stat.isdir and is_raspbian and is_debuntu
-
 
 # RECORD iiab-admin AS INSTALLED
 


holta:
I considered exactly this idea earlier in the week (or was it last week?)

My conclusion back then was that nagging about published passwords is a very necessary evil &mdash; in case `iiab-admin` or similar ({{ iiab_admin_user }}) is added after the fact &mdash; with IIAB's published password (g0adm1n) just as dangerous later on.

(I'm personally torn.  I feel like it's an open question that could go either way?  Nobody likes security nags -- but then again such security nagware is really needed for exactly that human reason -- if the nagware's interventionist UX genuinely helps ppl help themselves ;)

jvonau:
I suppose one could use runrole after the install to add a different named user and different default password for that user, the warning should work ok but {{ iiab_admin_published_pwd }} would need tinkering. The test is more applicable to iiab_user.yml as {{ iiab_admin_published_pwd }} is directly related to the default value of {{ iiab_admin_user }}. If {{ iiab_admin_user }} is changed from the default of iiab-admin, {{ iiab_admin_published_pwd }} would never match the replacement value used for {{ iiab_admin_user }}. Both default users 'pi' and 'ubuntu' have routines on their respective OSs in places to deal with stock passwords, let not clutter up the filesystem where we know in advance a match would be an impossibility. The object is to pass the user account {{ iiab_admin_user }} to admin-console to know which account to offer password changing capabilities for.

tim-moody:
I'm wondering if it makes sense to separate the sudo stuff from the user creation, like even a sudo role. I've also been thinking to create an iiab-admin group and take iiab_admin_user out of sudo. I don't think there's any reason for admin to be a sudoer, though it would require iiab/iiab-admin-console#345.

holta:
1. Lots To Think About Here (nagware needs to be effective, otherwise it's completely ignored and useless) but I personally I find this particular nagware can be extremely useful on Debian/Ubuntu/Mint when iiab-admin appears after the fact &mdash; certainly it's very useful to me &mdash; on IIAB systems where `iiab-admin` or `pi` password was later set to `g0adm1n` for a colleague to access etc...

2. Unusual that nagware can help both newcomers and veterans (whose interests rarely align otherwise!) but this seems to be one of those rare cases, so I'd rather not eliminate a cautionary mechanism that would seem to still be useful at keeping Linux know-it-all's (plenty of them ;) from shooting themselves in the foot too often?  But yes it's a usability/insurance question weighing a whole bunch of different pro's and con's we should all ponder carefully, whatever improvements are possible here.

3. Along the way (or when we get to it!) I'd like to understand better how/when Admin Console (and others, not just Admin Console) needs sudo access.  (That's a much longer conversation for another day, for sure, but also an important one that many others too would like more insight into...)

tim-moody:
> I'd like to understand better how/when Admin Console (and others, not just Admin Console) needs sudo access

the only reason iiab-admin needs to be in the sudo group is that I chose it as a convenience. I just changed authentication to the iiab-admin group and iiab-admin was able to log into adm cons with no problem and change the iiab-admin password.

If the iiab_admin_user has no role outside of admin console, then it should be created in the iiab-admin-console ansible run.

holta:
Is there any risk that some of Admin Console's many functionalities — might silently (or overtly) fail if iiab-admin no longer has sudoer privileges?

tim-moody:
>Is there any risk that some of Admin Console's many functionalities — might silently (or overtly) fail if iiab-admin no longer has sudoer privileges?

Do people ever make mistakes? Yes. But the cmdsrv runs as root, not iiab-admin, and if I search for the word sudo on the repo, none of the results involves iiab-admin.

holta:
Why does iiab-admin need to be part of any group out of curiosity?

FWIW I'm presuming Admin Console's enforcing group membership (sudo for now) when it shows that red "Invalid Login" message...when you try to log in to http://box/admin but lack group membership?

tim-moody:
>I'm presuming Admin Console's enforcing group membership 

because that is more flexible and lets other users be admin if they are in the right group(s)

holta:
> because that is more flexible and lets other users be admin if they are in the right group(s)

Good to know.  Would it be possible for the "Invalid Login" error message to say that in future?  Possibly something along the lines of...

```
Invalid Login: you need to a member of the 'iiab-admin' Linux group
```

holta:
See merged PR iiab/iiab-admin-console#346

tim-moody:
There are a number of reasons that the login can be invalid, and I think it is best practices not to tell the user why their authentication failed. However, for troubleshooting the reason is in the uwsgi log.

jvonau:
> Why does iiab-admin need to be part of any group out of curiosity?

By default most distros create a group with the same name as the user. 

If you don't create the 'iiab-admin user' a 'iiab-admin group' can still be created with select users added at will. The new 'iiab-admin group' should basically mirror the privileges of the sudo group, not sure that is worth the hassle.

holta:
> The new 'iiab-admin group' should basically mirror the privileges of the sudo group

Isn't @tim-moody trying to somewhat reduce security exposure?

i.e. by removing iiab-admin user(s) from the sudo group to eliminate some risks?

(Granted iiab-admin group members will still have some root privileges as a result of their logging in to Admin Console.)

tim-moody:
>By default most distros create a group with the same name as the user

any idea of the ansible module does this? I was assuming not, but hard to tell since iiab-admin is created in the iiab script.

>The new 'iiab-admin group' should basically mirror the privileges of the sudo group, not sure that is worth the hassle.

Not at all. The only privilege someone in the iiab-admin group has is to have cmdsrv run one of its commands, which are a pretty short list compared to what sudo can do and are mostly sanitized (e.g. if there is a copy command you don't get to supply the directory and it knows about ../../) A sudoer can do basically anything, which is way more than someone in iiab-admin can do.

A sudoer can become root. iiab-admin can only ask root to do a few things for it.

holta:
Please see the latest https://github.com/iiab/iiab/blob/master/roles/iiab-admin/README.rst

It's work in progress (isn't everything ;) but moves us all forward &mdash; with the imminent IIAB 7.2 RC3 that will be broadly and publicly announced if it passes various overnight smoke tests.

jvonau:
> > By default most distros create a group with the same name as the user
> 
> any idea of the ansible module does this? I was assuming not, but hard to tell since iiab-admin is created in the iiab script.
> 

Think that would depend on how each OS has /etc/login.defs configured and what groups and options are passed while using ansible's [user module.](https://docs.ansible.com/ansible/latest/collections/ansible/builtin/user_module.html). Can't recall the features of the user module when first introduced in [2014](https://github.com/XSCE/xsce/pull/164/files) 

> > The new 'iiab-admin group' should basically mirror the privileges of the sudo group, not sure that is worth the hassle.
> 
> Not at all. The only privilege someone in the iiab-admin group has is to have cmdsrv run one of its commands, which are a pretty short list compared to what sudo can do and are mostly sanitized (e.g. if there is a copy command you don't get to supply the directory and it knows about ../../) A sudoer can do basically anything, which is way more than someone in iiab-admin can do.
> 
> A sudoer can become root. iiab-admin can only ask root to do a few things for it.

Ok,better vision of the intended logic, just to make sure I understand, iiab-admin group would have a reduced set of privileges and/or just be able to use admin-console? What about Adam's use of the login user iiab-admin and most of the cmdline iiab-*commands currently requiring sudo, guess those would need access by the iiab-group also? Personally I don't login into my machines as iiab-admin, I wouldn't miss the user but I like having it for support issues. The console now accepts iiab-admin or sudo group membership as criteria for execution, perhaps admin-console could use the logged in user's id for the password changing utility in place of {{ iiab_admin_user }}?

tim-moody:
>just to make sure I understand, iiab-admin group would have a reduced set of privileges and/or just be able to use admin-console?

Not sure I understand the question. Someone authenticated by admin console can use it, which means cause cmdsrv to do anything it is capable of. Besides root and non-root, what other privileges are we talking about, file permissions? File permission and the privilege to run any other scripts are outside of the scope of admin console.

I'm also not saying that iiab-admin cannot be a sudoer, only that it with this change doesn't have to be to run admin console. I authenticate off a linux user and group for convenience. Otherwise I have to maintain passwords and do resets and the whole rest of it, which would be impossible for an offline install.

>perhaps admin-console could use the logged in user's id for the password changing utility in place of {{ iiab_admin_user }}?

I considered this, but it needs more though to make sure there cannot be a privilege escalation if this is allowed. There is security in that you have to supply the current password in order to change it, but I was still nervous about opening this up wider. 

Personally I always log in as root. I actually believe root with a key is stronger security than sudo. In fact I have contemplated generating a key for the admin user, but not much point unless they want to login and have root permission.

jvonau:
> > just to make sure I understand, iiab-admin group would have a reduced set of privileges and/or just be able to use admin-console?
> 
> Not sure I understand the question. Someone authenticated by admin console can use it, which means cause cmdsrv to do anything it is capable of. Besides root and non-root, what other privileges are we talking about, file permissions? File permission and the privilege to run any other scripts are outside of the scope of admin console.

That is inside the scope of iiab-admin being used as a **login user** via ssh or cmdline to run basic command like 'iiab' and friends or other system level activities such as journctl, et al.

> 
> I'm also not saying that iiab-admin cannot be a sudoer, only that it with this change doesn't have to be to run admin console. I authenticate off a linux user and group for convenience. Otherwise I have to maintain passwords and do resets and the whole rest of it, which would be impossible for an offline install.

Sure the recent change now allows for non-sudo users, that opens up many possibilities in the future.   
> > perhaps admin-console could use the logged in user's id for the password changing utility in place of {{ iiab_admin_user }}?
> 
> I considered this, but it needs more though to make sure there cannot be a privilege escalation if this is allowed. There is security in that you have to supply the current password in order to change it, but I was still nervous about opening this up wider.
> 
> Personally I always log in as root. I actually believe root with a key is stronger security than sudo. In fact I have contemplated generating a key for the admin user, but not much point unless they want to login and have root permission.

To each their own, still comes down to the user's choice of "I don't want a second account" with *_install False. The debate is then on the "any related support code being installed for what is in effect a non-existent user" when False is employed by an enduser or builder in the current use of the code.

jvonau:
Closing as the mass updates are meant to entice others not to contribute,

Issue #2553, opened 2020-09-30
Extremely slow downloads of IIAB Maps from archive.org (often 1-2 Mbit/s) - would torrents solve this - or is there a better way? [OSM]


holta:
The extreme slowness of downloads from archive.org (still as slow as 1-2 mbit/sec almost a year later) means that folks are unable to take advantage of IIAB Maps, when even a medium-sized Map Pack tediously can take all-day-and-night-and... to download.

(Or many days to download in the case of the 80GB "World" Map Pack!)

Is there a pragmatic solution here?

Would torrents (that Internet Archive already seeds) be one of those, e.g. could this every become a genuinely hassle-free solution, that idoesn't drive everyone crazy?  i.e. without endless maintenance/onboarding and other hidden costs?

(Are there other options we haven't considered?)

holta:
fyi the "World" Map Pack (80.48 GB) takes 3.7 days **(more than half a week)** to download at Archive.org's usual speed (2Mbit/s = 0.9 GByte/hour).

That's a best case scenario without network drops/interruptions (!)

i.e. I don't know if /usr/bin/[iiab-install-map-region](https://github.com/iiab/maps/blob/master/osm-source/pages/viewer/scripts/iiab-install-map-region) and /usr/bin/[iiab-extend-sat.py](https://github.com/iiab/maps/blob/master/osm-source/pages/viewer/scripts/iiab-extend-sat.py) recover gracefully after network outages, but possibly @georgejhunt might know: https://github.com/iiab/iiab/pull/2551#issuecomment-700342446

holta:
Of course things could be Even Worse (:

i.e. in comparison, `iiab-extend-sat.py` sustains about 0.33 Mbit/s as it downloads many small files (containing satellite photos) if we measure incremental data (MB) based on the growth of its `/library/www/osm-vector-maps/viewer/tiles/satellite_z0-z9_v3.mbtiles` database file.

That number appears to be pretty consistently between 0.32 and 0.35Mbit/s when downloading 3 different Hi-Res Satellite Photo Regions (2 squares of 100x100 km, and 1 square of 300x300 km).

[...]

PR #2572, opened 2020-10-15
Bittorrent

georgejhunt:
Installed on an clean rpi 4 with current master. Initially it did not work. But changed to a venv, and it seemed to work. I'm assuming that there were some packages that archive.org wanted that were the wrong version. 

The defaults seem to be that local cache gets shared, which is not what we want without explicit setting in local_vars. But this is in response to the observation that there was not anything  visible about this effort.

A lot of my testing has been interactive, though that's not what is needed long term. There is an interactive --all option which makes the local machine into a seeder.

I'm sure there is much that I haven't thought of.


holta:
As background to @georgejhunt's initiative here, please see the design discussion ongoing here:

#2553 "Extremely slow downloads of IIAB Maps from archive.org (often 1-2 Mbit/s) - would torrents solve this - or is there a better way? [OSM]"

(Thanks Much to @sptankard and @mitra42 !)

Of course Learning Content Peering applies online AND in-the-field (we call these grassroots civic networks IIAB's "Sneakernet-of-Alexandria" somewhat similar to Cuba's https://en.wikipedia.org/wiki/El_Paquete_Semanal etc).

So hopefully both workflows + their HCI/UX affordances _(one more online and one more offline)_ will continue to inform each other in support of teachers & students everywhere !!

e.g. SHAREit, Xender, Zapya file transfer apps for Android — are quite ubiquitous apps in the developing world — making **user interfaces far more humane** for Bluetooth & WiFi Direct (& USB On-The-Go and other cables, microSD, etc).

As this is all about much more than Map Packs + Khan Academy + Wikipedia "mass media consumption" in the end ~ this also must support indigenous cultural exchanges at the very root of the human right to self-determination ~

georgejhunt:
Adam was seeing 2Mb (250kB/sec). Using transmission-daemon, I saw 5-10MB download speeds (my download pipe is only 8MB). So I don't have a way to test, once we get a few other seeders working-- because archive.org can fill my downstream bandwidth without any other seeders we might put online. 

I'm not sure I know all that will be necessary to get IIAB devices to seed if they are behind a NAT router. I may need a tool, or learn to use the tools we already have, to verify that machines we put online are discovered, and used.

Tim, are you willing to use iiab.me as a seeder? How much of your upstream are you willing to make available? I googled "transmission-daemon security". There was a vulnerability discovered in 2018, that was foxed in our version 2.92. See https://www.cvedetails.com/vulnerability-list/vendor_id-9749/product_id-17422/Transmissionbt-Transmission.html
https://majenko.co.uk/blog/securing-transmission-daemon

tim-moody:
@georgejhunt please contact me directly to discuss security. My first take is that CVE-2018-5702 does apply to the Ubuntu 18.04 version 2.92, but not to Ubuntu 20.04 version 2.94. I'm also not sure it is relevant to seeding.

Issue #2561, opened 2020-10-07
TK: Installing IIAB blocks most logins, except for sudoers with NOPASSWD: set [EXCEPTIONS: (1) RaspiOS graphical logins work (2) sudoers can survive non-graphical logins by typing in their passwords twice]


holta:
@jvonau is monitoring TK Kang's progress in understanding the problem, which has been reproduced ~3 times now on Ubuntu 20.04 VM's and PC's.

jvonau:
#2562 for the wired slave detection as somebody actually used that code path, thanks TK as I don't have a U-20 amd64 reproducer handy.

> which has been reproduced ~3 times now on Ubuntu 20.04 VM's and PC's.

At least 2 different people should be tripped up not being able to login before panicking. Think this would be showing up on Mint 20 as it's close cousin to Ubuntu. From Chat

>To narrow this down I would do a stock desktop install on hardware or VM and note if you are forced to login at the select user dialog box after rebooting several times. Once the stock behavior is established, run just the part 'iiab' to the point where the iiab-admin user is created and `ctrl c`after the password is set. Reboot and note if the select user login box is present or not.

jvonau:
More context
>On Tue, Oct 6, 2020 at 11:54 PM T.K. Kang wrote:
            OK... if I login with iiab-admin the desktop will reappear. Hence the issues are account that are created prior to IIAB or after IIAB install.
            Cheers
    On Wed, Oct 7, 2020 at 1:30 PM Jerry Vonau wrote:
        Is this a case of having 2 possible logins and now the display manager can't or refuses to auto login?
On Wed, Oct 7, 2020 at 6:47 PM T.K. wrote:
    Only the iiab-admin account can login in via local desktop to get the ubuntu interface. My admin account will just display a blank page. However can ssh from another terminal.

holta:
More context from TK Kang:
> > Can you clarify the exact URL of the Ubuntu 20.04 installer you are using?
> >
> > e.g. if https://ubuntu.com/download/desktop (or any other!)

> ubuntu-20.04-desktop-amd64
>
> Will try the latest ubuntu-20.04.1-desktop-amd64 for another install again

shanti-bhardwa:
did 7.2 build on Ubuntu-20.04.1-desktop-amd64.iso . Medium build, cannot loging using my original userid, but can do so using iiab-admin. see http://spruge.us/JBOgXV

looks like my profile as original user is being changed by iiab-admin

holta:
[...]

Issue #2560, opened 2020-10-06
references to admin console


tim-moody:
I don't understand 4-server-options/tasks/main.yml 59ff:
```
# 2020-02-17: what was roles/homepage lives in roles/www_options.  Eventually
# softcoding of iiab_home_url should happen everywhere (incl Admin Console) for
# more field options, e.g. changing /library/www/html/home even when offline...
```
Is this still needed, nginx/tasks/install.yml 19ff?

```
- name: Add user '{{ apache_user }}' to shadow group, so it can authenticate Admin Console
  user:
    name: "{{ apache_user }}"    # www-data or apache, per /opt/iiab/iiab/vars/<OS>.yml
    groups: shadow

```
captiveportal/tasks/enable-or-disable.yml and elsewhere. If desired admin console can install uwsgi: it already installs the config files for admin console. (btw it only uses 1 child process, though with 4 threads)

```
# 2020-01-22: Package 'uwsgi' is installed by nginx/tasks/install.yml for
# Captive Portal (responds to browsers trying to detect a Captive Portal) and
# also for Admin Console.

# IIAB makes the 'uwsgi' service mandatory for now (see Line 45) but this could
# perhaps be auto-disabled in future (or never installed?)  e.g. if Captive
# Portal + Admin Console are both unneeded...and impact of uWSGI is too high ?

- name: Stop 'uwsgi' systemd service (may include 3 child processes for Captive Portal, and 3 child processes for Admin Console)
  systemd:
    name: uwsgi
    state: stopped

```

holta:
@jvonau helped clarify the 3 questions above with @tim-moody during our call today (http://minutes.iiab.io)

1. Comment should be freshened up (I'll work on that).
2. Likely that stanza should be removed, @jvonau & @tim-moody agree.
3. @tim-moody wonders if Admin Console should install uWSGI, just to be safe.  @jvonau is not sure that necessary, but perhaps no harm.

holta:
> 1. Comment should be freshened up (I'll work on that).

Done.  This is part of PR #2565:

https://github.com/holta/iiab/commit/fff2829fd83c1b4d2ebdf81141f126fb7d3fbbe0

holta:
Appears resolved.  @tim-moody please re-open as nec.

PR #2573, opened 2020-10-15
Let's consider variable "iiab_admin_can_sudo: True" & support "admin_console_group: iiab-admin" and explain them for all in roles/iiab-admin/README.rst

holta:
Thanks to input from @jvonau

Tested on Ubuntu Desktop 20.04.1

Might need a few more revisions.

Most important, I'll work on explaining how this idea can work well with Admin Console in coming hours, as I revise https://github.com/holta/iiab/blob/sudoers_anonymous/roles/iiab-admin/README.rst

FYI in order to make IIAB's Security more coherent and understandable, this PR proposes deprecating Ansible's highly bureaucratic, highly unused and highly questionable approach to setting our iiab-admin's password ({{ iiab_admin_pwd_hash }}) &mdash; in favor of auto-generating its salt and hash using {{ iiab_admin_published_pwd }}.

Refs: #2560, #2561, PR #2566, PR #2570, PR #2571, PR iiab/iiab-factory#153, iiab/iiab-admin-console#345, PR iiab/iiab-admin-console#346, PR iiab/iiab-admin-console#347, PR iiab/iiab-admin-console#348


holta:
Please read over the new iiab-admin README:

https://github.com/holta/iiab/blob/sudoers_anonymous/roles/iiab-admin/README.rst

The quite separate question of how Admin Console should expand access to multiple users (and change the password of those multiple users, etc!) is distinct from this PR, but is an important ongoing inquiry (and onramping challenge. i.e. how to keep it dead simple + understandable for everyone!) very much worth mentioning nonetheless.

For example if Linux group(s) are indeed the way forward for Admin Console authentication, here are 2 things to keep in mind:

- The actual (not assumed) primary group (as specified by the GID in the 4th column of /etc/passwd) associated with {{ iiab_admin_user }} might be useful, if only as a security check that `USERGROUPS_ENAB yes` in `/etc/login.defs` is being respected?  See the output of Linux command `id -gn iiab-admin` unlike the (also useful!) output of the command `id -Gn iiab-admin` similar to `groups iiab-admin`

- I agree with @jvonau that another option might be to recognize more explicitly that this `iiab-admin` user serves 2 quite distinct purposes (Admin Console & community support command-line things like /usr/bin/iiab-diagnostics).  And as such a distinct variable {{ admin_console_group }} might make the most sense, e.g. calling a spade a spade?  Regardless whether a variable like {{ admin_console_group }} inherits directly from {{ iiab_admin_user }} or from `id -gn {{ iiab_admin_user }}` or any other group now or in future, this might help us all articulate what's really needed here?

Implementation Asides: (more Ansible war stories, hopefully leading us forward not backward ;)

1. Currently Admin Console fails to allow iiab-admin to log in unless it joins the iiab-admin group twice (its primary group is in /etc/passwd, and then currently/weirdly it must also become a member of its own group in /etc/group as well?)  This looks like it will trip up a lot of people.  It should probably be fixed, to avoid widespread confusion, given `USERGROUPS_ENAB yes` is standard in `/etc/login.defs` across all modern Linux OS's.  _Of course Ansible doesn't make it at all easy here..._ with separate parameters `group` and `groups` for https://docs.ansible.com/ansible/latest/collections/ansible/builtin/user_module.html

2. Aside/Similarly Ansible makes it nearly impossible to remove a user from a single group...so this PR implements that (to remove iiab-admin from the sudo group) using shell command `gpasswd` instead: https://github.com/holta/iiab/blob/sudoers_anonymous/roles/iiab-admin/tasks/admin-user.yml#L37

holta:
Fresh install (BIG-sized) of this PR is currently being tested on Debian 10.6 together with:

- PR #2576 'Remove "| bool" 1.5yrs later, as Ansible 2.10 changed conditional_bare_variables'

Both will be merged today for [IIAB 7.2](https://github.com/iiab/iiab/wiki/IIAB-7.2-Release-Notes) RC3 if testing continues to look good.

PR #2575, opened 2020-10-15
phpMyAdmin 5.0.3 -> 5.0.4 [and change download site from d.iiab.io/packages to phpmyadmin.net/files/*]

holta:
Hot fix will be merged later today, for: https://github.com/phpmyadmin/phpmyadmin/releases/tag/RELEASE_5_0_4

More details: https://www.phpmyadmin.net/files/5.0.4/

Ref: PR #2568


holta:
phpMyAdmin 5.0.3 appears to have been very broken (at around 12MB instead of 14MB as with all other 5.x releases).

So this (5.0.4) is very timely!

See: https://www.phpmyadmin.net/files/

PS for now this is also available at:
http://d.iiab.io/packages/phpMyAdmin-5.0.4-all-languages.zip

holta:
Re-hosting a copy of each file like phpMyAdmin-5.0.4-all-languages.zip to http://d.iiab.io/packages every few days/months was not a good use of my time.

So I've changed the download source to https://www.phpmyadmin.net/files/ (which should be reliable enough for BIG-sized IIAB installs, and similar) to reduce all of our upkeep/administration time going forward.

Tested on Ubuntu 20.04.1

PR #2147, opened 2020-01-14
(1) Clean Munin (2) rm excess '| bool' (casts to boolean) (3) Experimentally comment out Ansible tags (4) Fix 3 /etc/iiab/iiab.ini descriptions

holta:
WIP in preparation for @jvonau's PR #2141 etc.

Ref: #1632

PR #2576, opened 2020-10-16
Remove "| bool" 1.5yrs later, as Ansible 2.10 changed conditional_bare_variables default

### LINKED ISSUE #1632: Ansible 2.8+ ADVISORY: avoid warnings by using "when: var | bool" for top-level BARE vars (in case they're strings, instead of boolean)

holta:
Resolves #1632.

The ` | bool` is no longer useful on the right side of `when:` conditions, especially now that Ansible 2.10 makes this more official:

- https://docs.ansible.com/ansible/latest/porting_guides/porting_guide_2.8.html#bare-variables-in-conditionals
- https://github.com/ansible/ansible/issues/70682

So I ran `grep -r '\s*|\s\+bool' roles | xargs sed -i 's/\s*|\s\+bool//'` in /opt/iiab/iiab while careful not to modify these 3 files:

```
root@box:/opt/iiab/iiab# grep -rl '\s*|\s*bool'

.git/objects/pack/pack-9660aafb9d44d7a4dce8dbf2f22394f53730bbe0.pack
roles/www_base/files/html/js/bootstrap4.min.js
roles/0-init/tasks/validate_vars.yml
```

PS this PR also converts 5 files from DOS to UNIX endings using `find -type f -exec dos2unix {} \;` so that grep will work coherently going forward:

```
roles/kiwix/files/test_zim/Notes.txt
roles/kiwix/tasks/kiwix-apk.yml
roles/kiwix/templates/iiab-make-kiwix-lib
roles/nginx/templates/mime.types.j2
roles/www_base/files/html/js/Markdown.Converter.js
```

Otherwise simple things like `grep pattern$ file` fail to show lines ending in that pattern, and hackier things like `grep 'pattern[[:cntrl:]]$' file` and `grep pattern$'\r'$ file` [*] very confusingly show blank lines, instead of the actual matching lines.

[*] `$''` for C-style escape (in this case `$'\r'` for carriage return) so no need to force single character `^M` to appear using `Ctrl-v` `Ctrl-m` &mdash; but if you do force that single control character to appear at the command-line (`^M` for carriage return) then `grep pattern^M$ file` also works, as above.

ASIDE: `cat -v file` is a quick way to visually confirm a DOS file, seeing the `^M` characters (carriage return, first part of CRLF) inside it &mdash; this is also useful for characters like `^@` (NUL character, i.e. what's often used to [terminate strings](https://en.wikipedia.org/wiki/Null-terminated_string)).  And `file testfile` tell you at a high-level what kind of file it is, summarized in a single line like...

```
root@box:~# file textfile1
textfile1: ASCII text, with CRLF line terminators
```

Above being a DOS file.  Whereas a Linux text files appears as...

```
root@box:~# file textfile2
textfile2: ASCII text
```

And a JPEG file might appear as...

```
root@box:~# file /library/www/html/js-menu/menu-files/images/en-afristory.jpg
/library/www/html/js-menu/menu-files/images/en-afristory.jpg: JPEG image data, Exif standard: [TIFF image data, little-endian, direntries=0], baseline, precision 8, 700x520, components 3
```

Refs: #1632, PR #1696, PR #2147


jvonau:
Stop right there. Obviously there is no real code freeze and this BS spouting about no changes late in the release cycle is just that, The errors then(1632) was a bug and the knee jerk reaction that went into place was unwarranted. I find it extremely rude to be jerking around with milestones 7.2 -> 8.0 -> 7.2 and file changes en mass that results in having to re-base **any and all** pending PR's **again**. Push this to 8.0 or look at pushing PRs first then make this mass change.

holta:
I humbly disagree.

Release is in fact the most important time for code readability improvements and documentation at very level.

PR #2548, opened 2020-09-26
Lint for release: osm-vector-maps indentation/spacing/quotes

holta:
Not a functional change.

Testing nonetheless, in case of Ansible snafus.

Ref: #1605 'New OSM Vector Maps need clean UX + clean instructions ("pick a continent or region")'

PR #2579, opened 2020-10-19
Lint for release: scripts/local_facts.fact & remove legacy "tags:" from 6 remaining files (.yml's)

holta:
Somewhat similar: PR #2548


holta:
With the obvious caveat that [Duct Tape Programmers](https://www.joelonsoftware.com/2009/09/23/the-duct-tape-programmer/) absolutely despise lint and code readability...because it certainly compiled when _they_ ran it ;)

holta:
Lightly tested with a fresh install on 'RaspiOS with desktop' on RPi 4, but I'll do some functional tests too to be sure.

holta:
Installs tests on RaspiOS & Ubuntu 20.04.1 look perfect, but I'll do a couple more functional tests to be sure.

tim-moody:
What problem does this solve that is required before the 7.2 release? This requires a great deal of retesting including a variety of scripts and functionality in the admin console. Also, what does it have to do with local facts? @holta your position was that issues should not mix topics. Equally PRs should only address a single problem, not a stream of consciousness.

holta:
> great deal of retesting including a variety of scripts and functionality in the admin console

I don't believe that's true.  I certainly hope that's not true.

As `scripts/local_facts.fact` is tested to (1) produce the correct output across its 12 variables, (2) more safely and (3) more approachably.

Regarding Ansible tags they've been phased out by IIAB over the last couple years, and if these last 6 files have any consequences in this regard, I don't believe that's the case after the multi-year deprecation glide path around our former use of Ansible tags.

Thankfully nobody is perfect but this is the time for testing and code clarity and documenting high-level public-facing IIAB infra especially, that makes our work an approachable and purposeful community project.

And if there are aspects of Admin Console that need additional testing please share those so that I and many others can help you and all with any test plan here.

jvonau:
[jerry@box iiab]$ scripts/local_facts.fact
>cat: /proc/device-tree/mfg-data/MN: No such file or directory
cat: /proc/device-tree/model: No such file or directory
Failed to get unit file state for dhcpcd.service: No such file or directory
Failed to get unit file state for network-manager.service: No such file or directory

I can see the original check being used just omitting the else clause as the default is already assigned. ie:

`if [ -f /proc/device-tree/mfg-data/MN ]; then
    XO_VERSION=`cat /proc/device-tree/mfg-data/MN`
fi`

Just because local_facts.facts is run under ansible I don't think the misuse of && is the right thing to do here.

holta:
I don't know of any misuse of `&&`.

@jvonau are you saying you don't like the short form syntax / syntactic sugar below?

```
CMD1 &&
    CMD2
```

i.e. that you visually prefer this long-form syntax?

```
if CMD1; then
    CMD2
fi
```

jvonau:
'cat: /proc/device-tree/mfg-data/MN: No such file or directory'  <<-- now present
if [ -f /proc/device-tree/mfg-data/MN ]; then
..XO_VERSION=cat /proc/device-tree/mfg-data/MN
fi
Yes personally long form, the needless feedback is harmless as is but was not being generated prior to this change. 
Isn't the point of defensive programming not to do things that you already should know the result of by the absence of the file (the declared default) or am I missing something?

holta:
> needless feedback

On my end I find the feedback important at several levels.

Both for new people who choose to run `scripts/local_facts.fact` to build an understanding of IIAB at the command-line.

And also for veterans evolving and testing this code going forward, surfacing more-than-hiding flareups &mdash; towards keeping the landscape and risks clear.

> defensive programming

Aligning each var's CMD with the condition (using the UNIX norm of returning 0/zero on success, and non-zero on failure) traps all errors instead of just some of the errors.  Wherever possible.

i.e. mentioning each var's CMD-and-possible-path-etc just once (instead of twice) reduces duplication and risk, as others maintain this code in future (we don't know who that will be).

In any case, if the list of 12 variables (and their associated commands) ever needs to change in the future, a standard form greatly helps across many/most of these ~12 vars, for code readability and maintainability, for new and veteran devs both.

>Yes personally long form

Ok.  I'll definitely keep that in mind and avoid the short-form as much as possible in future.

Code style guidelines should not be too strict, but this makes sense as a very reasonable suggestion towards embracing newcomers who might not be as fluent.

PR #2582, opened 2020-10-21
Overseas request for immediate testing of IIAB on Debian 11 Bullseye + similarly clean out long-deprecated apache_config_dir from high-level <OS>.yml vars files [ASIDE: dnsmasq & dhclient issues? host_country_code MUST match PC/HW, or hostapd won't start]

holta:
While Debian 11 (Bullseye) is an unreleased OS that's obviously not supported, vars are put in place here to encourage general testing in anticipation of 2021's Debian 11, Raspbian 11 a.k.a. RaspiOS 11 a.k.a. Raspberry Pi OS 11 &mdash; and other Debian derivatives like Qubes OS, etc.

Aside: a reminder that server/headless versions of Debian don't immediately jump on Wi-Fi (and don't permit bringing up your Wi-Fi network interface, even if the driver's installed!) unless you first run something like `allow-hotplug wlp2s0` as documented at https://wiki.debian.org/WiFi/HowToUse#Command_Line

cc: @sptankard, @deldesir, @arky and other Debian implementers (:

Tangentially related: iiab/iiab-factory#154

### DIFF
--- a/scripts/local_facts.fact
+++ b/scripts/local_facts.fact
@@ -34,8 +34,12 @@ NM_PATH=`which NetworkManager`
     #"raspbian-8"   | \
     #"raspbian-9"   | \
 
+# 2020-10-21: Debian 11 (Bullseye) not yet supported but adding this line to
+# its /etc/os-release can help testing this unreleased OS: VERSION_ID="11"
+
 case $OS_VER in
     "debian-10"    | \
+    "debian-11"    | \
     "ubuntu-20"    | \
     "linuxmint-20" | \
     "raspbian-10")
--- a/vars/centos-7.yml
+++ b/vars/centos-7.yml
@@ -12,7 +12,6 @@ proxy_user: squid
 apache_service: httpd
 apache_user: apache
 apache_conf_dir: httpd/conf.d
-apache_config_dir: "{{ apache_conf_dir }}"    # for iiab-admin-console/roles/console/tasks/main.yml Line 150
 apache_log_dir: /var/log/httpd
 smb_service: smb
 nmb_service: nmb
--- a/vars/debian-10.yml
+++ b/vars/debian-10.yml
@@ -12,7 +12,6 @@ proxy: squid
 proxy_user: proxy
 apache_service: apache2
 apache_conf_dir: apache2/sites-available
-apache_config_dir: "{{ apache_conf_dir }}"    # for iiab-admin-console/roles/console/tasks/main.yml Line 150
 apache_user: www-data
 apache_log_dir: /var/log/apache2
 smb_service: smbd
--- a/vars/debian-11.yml
+++ b/vars/debian-11.yml
@@ -1,0 +1,27 @@
+is_debuntu: True
+is_debian: True
+is_debian_11: True
+
+# 2019-01-31: These apply if-only-if named_install and/or dhcpd_install are True
+# (This is quite rare now that vars/default_vars.yml sets dnsmasq_install: True)
+dns_service: bind9
+dhcp_service: isc-dhcp-server
+dns_user: bind
+
+proxy: squid
+proxy_user: proxy
+apache_service: apache2
+apache_conf_dir: apache2/sites-available
+apache_config_dir: "{{ apache_conf_dir }}"    # for iiab-admin-console/roles/console/tasks/main.yml Line 150
+apache_user: www-data
+apache_log_dir: /var/log/apache2
+smb_service: smbd
+nmb_service: nmbd
+systemctl_program: /bin/systemctl
+mysql_service: mariadb
+apache_log: /var/log/apache2/access.log
+sshd_package: openssh-server
+sshd_service: ssh
+php_version: 7.4
+postgresql_version: 13
+systemd_location: /lib/systemd/system
--- a/vars/debian-11.yml
+++ b/vars/debian-11.yml
@@ -12,7 +12,6 @@ proxy: squid
 proxy_user: proxy
 apache_service: apache2
 apache_conf_dir: apache2/sites-available
-apache_config_dir: "{{ apache_conf_dir }}"    # for iiab-admin-console/roles/console/tasks/main.yml Line 150
 apache_user: www-data
 apache_log_dir: /var/log/apache2
 smb_service: smbd
--- a/vars/debian-8.yml
+++ b/vars/debian-8.yml
@@ -11,7 +11,6 @@ proxy: squid3
 proxy_user: proxy
 apache_service: apache2
 apache_conf_dir: apache2/sites-available
-apache_config_dir: "{{ apache_conf_dir }}"    # for iiab-admin-console/roles/console/tasks/main.yml Line 150
 apache_user: www-data
 apache_log_dir: /var/log/apache2
 smb_service: smbd
--- a/vars/debian-9.yml
+++ b/vars/debian-9.yml
@@ -12,7 +12,6 @@ proxy: squid
 proxy_user: proxy
 apache_service: apache2
 apache_conf_dir: apache2/sites-available
-apache_config_dir: "{{ apache_conf_dir }}"    # for iiab-admin-console/roles/console/tasks/main.yml Line 150
 apache_user: www-data
 apache_log_dir: /var/log/apache2
 smb_service: smbd
--- a/vars/default_vars.yml
+++ b/vars/default_vars.yml
@@ -698,6 +698,7 @@ is_linuxmint: False    # Subset of is_ubuntu
 is_linuxmint_20: False
 
 is_debian: False    # Covers both: Debian, Raspberry Pi OS (Raspbian)
+is_debian_11: False
 is_debian_10: False
 is_debian_9: False
 is_debian_8: False
--- a/vars/fedora-18.yml
+++ b/vars/fedora-18.yml
@@ -12,7 +12,6 @@ proxy_user: squid
 apache_service: httpd
 apache_user: apache
 apache_conf_dir: httpd/conf.d
-apache_config_dir: "{{ apache_conf_dir }}"    # for iiab-admin-console/roles/console/tasks/main.yml Line 150
 apache_log_dir: /var/log/httpd
 smb_service: smb
 nmb_service: nmb
--- a/vars/fedora-22.yml
+++ b/vars/fedora-22.yml
@@ -12,7 +12,6 @@ proxy_user: squid
 apache_service: httpd
 apache_user: apache
 apache_conf_dir: httpd/conf.d
-apache_config_dir: "{{ apache_conf_dir }}"    # for iiab-admin-console/roles/console/tasks/main.yml Line 150
 apache_log_dir: /var/log/httpd
 smb_service: smb
 nmb_service: nmb
--- a/vars/linuxmint-20.yml
+++ b/vars/linuxmint-20.yml
@@ -15,7 +15,6 @@ proxy_user: proxy
 apache_service: apache2
 apache_user: www-data
 apache_conf_dir: apache2/sites-available
-apache_config_dir: "{{ apache_conf_dir }}"    # for iiab-admin-console/roles/console/tasks/main.yml Line 150
 apache_log_dir: /var/log/apache2
 smb_service: smbd
 nmb_service: nmbd
--- a/vars/raspbian-10.yml
+++ b/vars/raspbian-10.yml
@@ -14,7 +14,6 @@ proxy: squid
 proxy_user: proxy
 apache_service: apache2
 apache_conf_dir: apache2/sites-available
-apache_config_dir: "{{ apache_conf_dir }}"    # for iiab-admin-console/roles/console/tasks/main.yml Line 150
 apache_user: www-data
 apache_log_dir: /var/log/apache2
 smb_service: smbd
--- a/vars/raspbian-8.yml
+++ b/vars/raspbian-8.yml
@@ -13,7 +13,6 @@ proxy: squid3
 proxy_user: proxy
 apache_service: apache2
 apache_conf_dir: apache2/sites-available
-apache_config_dir: "{{ apache_conf_dir }}"    # for iiab-admin-console/roles/console/tasks/main.yml Line 150
 apache_user: www-data
 apache_log_dir: /var/log/apache2
 smb_service: smbd
--- a/vars/raspbian-9.yml
+++ b/vars/raspbian-9.yml
@@ -14,7 +14,6 @@ proxy: squid
 proxy_user: proxy
 apache_service: apache2
 apache_conf_dir: apache2/sites-available
-apache_config_dir: "{{ apache_conf_dir }}"    # for iiab-admin-console/roles/console/tasks/main.yml Line 150
 apache_user: www-data
 apache_log_dir: /var/log/apache2
 smb_service: smbd
--- a/vars/ubuntu-16.yml
+++ b/vars/ubuntu-16.yml
@@ -13,7 +13,6 @@ proxy_user: proxy
 apache_service: apache2
 apache_user: www-data
 apache_conf_dir: apache2/sites-available
-apache_config_dir: "{{ apache_conf_dir }}"    # for iiab-admin-console/roles/console/tasks/main.yml Line 150
 apache_log_dir: /var/log/apache2
 smb_service: smbd
 nmb_service: nmbd
--- a/vars/ubuntu-17.yml
+++ b/vars/ubuntu-17.yml
@@ -13,7 +13,6 @@ proxy_user: proxy
 apache_service: apache2
 apache_user: www-data
 apache_conf_dir: apache2/sites-available
-apache_config_dir: "{{ apache_conf_dir }}"    # for iiab-admin-console/roles/console/tasks/main.yml Line 150
 apache_log_dir: /var/log/apache2
 smb_service: smbd
 nmb_service: nmbd
--- a/vars/ubuntu-18.yml
+++ b/vars/ubuntu-18.yml
@@ -13,7 +13,6 @@ proxy_user: proxy
 apache_service: apache2
 apache_user: www-data
 apache_conf_dir: apache2/sites-available
-apache_config_dir: "{{ apache_conf_dir }}"    # for iiab-admin-console/roles/console/tasks/main.yml Line 150
 apache_log_dir: /var/log/apache2
 smb_service: smbd
 nmb_service: nmbd
--- a/vars/ubuntu-19.yml
+++ b/vars/ubuntu-19.yml
@@ -13,7 +13,6 @@ proxy_user: proxy
 apache_service: apache2
 apache_user: www-data
 apache_conf_dir: apache2/sites-available
-apache_config_dir: "{{ apache_conf_dir }}"    # for iiab-admin-console/roles/console/tasks/main.yml Line 150
 apache_log_dir: /var/log/apache2
 smb_service: smbd
 nmb_service: nmbd
--- a/vars/ubuntu-20.yml
+++ b/vars/ubuntu-20.yml
@@ -13,7 +13,6 @@ proxy_user: proxy
 apache_service: apache2
 apache_user: www-data
 apache_conf_dir: apache2/sites-available
-apache_config_dir: "{{ apache_conf_dir }}"    # for iiab-admin-console/roles/console/tasks/main.yml Line 150
 apache_log_dir: /var/log/apache2
 smb_service: smbd
 nmb_service: nmbd


holta:
CLARIF: the focus is on the `testing` branch of Debian 11 Bullseye.

Where `/etc/os-release` currently appears as follows:

```
PRETTY_NAME="Debian GNU/Linux bullseye/sid"
NAME="Debian GNU/Linux"
ID=debian
HOME_URL="https://www.debian.org/"
SUPPORT_URL="https://www.debian.org/support"
BUG_REPORT_URL="https://bugs.debian.org/"
```

ASIDE: I don't believe anybody has recently installed IIAB onto the `unstable` brranch of Debian (that's known as [sid](https://www.debian.org/releases/sid/)), but if I'm wrong please someone in the Debian community let us all know!

holta:
Just as background, Debian 10 Buster's `/etc/os-release` appears as follows:

```
PRETTY_NAME="Debian GNU/Linux 10 (buster)"
NAME="Debian GNU/Linux"
VERSION_ID="10"
VERSION="10 (buster)"
VERSION_CODENAME=buster
ID=debian
HOME_URL="https://www.debian.org/"
SUPPORT_URL="https://www.debian.org/support"
BUG_REPORT_URL="https://bugs.debian.org/"

**Comment by deldesir** [comment]
That explains it all for a good start :)

> 
> 
> While Debian 11 (Bullseye) is an unreleased OS that's obviously not supported, vars are put in place here to encourage general testing in anticipation of 2021's Debian 11, Raspbian 11 a.k.a. RaspiOS 11 a.k.a. Raspberry Pi OS 11 — and other Debian derivatives like Qubes OS, etc.
> 
> Aside: a reminder that server/headless versions of Debian don't immediately jump on Wi-Fi (and don't permit bringing up your Wi-Fi network interface, even if the driver's installed!) unless you first run something like `allow-hotplug wlp2s0` as documented at https://wiki.debian.org/WiFi/HowToUse#Command_Line
> 
> cc: @sptankard, @deldesir, @arky and other Debian implementers (:
> 
> Tangentially related: [iiab/iiab-factory#154](https://github.com/iiab/iiab-factory/issues/154)

**Comment by holta** [comment]
If anybody wants MongoDB (primarily for Sugarizer) to work on Debian 11 Bullseye, see @deldesir's [interim workaround](https://github.com/iiab/iiab/pull/2493#issuecomment-713866120) on Line 77 of [/opt/iiab/iiab/roles/mongodb/tasks/install.yml](https://github.com/iiab/iiab/blob/master/roles/mongodb/tasks/install.yml#L77) changing `{{ ansible_distribution_release }}` to `buster` (so the variable doesn't evaluate to `bullseye`).

(A quick hack during testing, that's all!)

**Comment by jvonau** [comment]
> While Debian 11 (Bullseye) is an unreleased OS that's obviously not supported, vars are put in place here to encourage general testing in anticipation of 2021's Debian 11, Raspbian 11 a.k.a. RaspiOS 11 a.k.a. Raspberry Pi OS 11 — and other Debian derivatives like Qubes OS, etc.
> 
> Aside: a reminder that server/headless versions of Debian don't immediately jump on Wi-Fi (and don't permit bringing up your Wi-Fi network interface, even if the driver's installed!) unless you first run something like `allow-hotplug wlp2s0` as documented at https://wiki.debian.org/WiFi/HowToUse#Command_Line

Can' t say I recall seeing an iiab-diagnoistics from a Debian-10 install with WiFi and/or a related issue filed, this is a difference in behavior from Debian-10?

> If anybody wants MongoDB (primarily for Sugarizer) to work on Debian 11 Bullseye, see @deldesir's [interim workaround](https://github.com/iiab/iiab/pull/2493#issuecomment-713866120) on Line 77 of [/opt/iiab/iiab/roles/mongodb/tasks/install.yml](https://github.com/iiab/iiab/blob/master/roles/mongodb/tasks/install.yml#L77) changing `{{ ansible_distribution_release }}` to `buster` (so the variable doesn't evaluate to `bullseye`).
> 
> (A quick hack during testing, that's all!)

Nodesource has a similar installer and policy, most likely run into the same kind issue there also.

**Comment by deldesir** [comment]
Thanks @jvonau for mentioning nodesource. I'll know what to do then when I'll be at nodejs.

What is gui_desired_network_role? Installation stopped saying it's undefined.

**Comment by holta** [comment]
> Nodesource has a similar installer and policy, most likely run into the same kind issue there also.

Interestingly IIAB installs Node.js 12.9.0 onto Debian 11 Bullseye without a hitch, as shown by the output of `node -v` &mdash; which makes Debian pre-release testing easier than Ubuntu pre-release testing.

The reason that the install is so easy is because Nodesource.com supports Debian's `testing` branch (Bullseye) &mdash; and interestingly they also support Debian's `unstable` branch (sid) &mdash; as they document here:

https://github.com/nodesource/distributions#debian-and-ubuntu-based-distributions

**Comment by deldesir** [comment]
I remembered @jvonau solved this 2y ago on Ubuntu 18.04 (https://github.com/iiab/iiab/issues/1306). Now I get the same thing here. Unable to start service dnsmasq.

**Comment by holta** [comment]
> I remembered @jvonau solved this 2y ago on Ubuntu 18.04 (#1306). Now I get the same thing here. Unable to start service dnsmasq.

I'm not seeing that problem after installing IIAB (MIN-sized + Node-RED) onto Debian 11 Bullseye (server not desktop, latest daily net-install ISO) on a VM.  Certainly not yet, after a few reboots!

However earlier (prior to installing IIAB) the Debian 11 VM failed to get an IP address over Ethernet (Bridged Adapter from VirtualBox's perspective) during one of its many boots.  So that one time when the networking failed I had to manually run `dhclient enp0s3` which solved the problem.

@deldesir what hardware are you using for Debian 11 Bullseye testing...an older laptop?

**Comment by holta** [comment]
FYI this PR was tested by 2 different people, but discussion of Debian 11 Bullseye can continue here (@deldesir's dnsmasq issue or whatever!) even after it's now been merged, if this is a good enough place for such topics.

**Comment by deldesir** [comment]
I am using HP Stream Notebook 13 - c002dx

**Comment by jvonau** [comment]
> What is gui_desired_network_role? Installation stopped saying it's undefined.

Hum, that is a rarely seen part of the code, but as a doctor I must order up an x-ray as my remote viewing crystal ball is in for repairs, mind supplying the url generated from iiab-diagnostics and posting the resulting url here in this issue or send it to me via email.

> However earlier (prior to installing IIAB) the Debian 11 VM failed to get an IP address over Ethernet (Bridged Adapter from VirtualBox's perspective) during one of its many boots. So that one time when the networking failed I had to manually run `dhclient enp0s3` which solved the problem.

Can you expand on the 'one time'? as somewhere between 2nd <-> 10th reboot? Looking for a general trend and as you mentioned this was noted to be pre-iiab install so what ever ills Debian has, is preexisting.

**Comment by holta** [comment]
@deldesir please run: [sudo iiab-diagnostics](https://github.com/iiab/iiab/blob/master/scripts/iiab-diagnostics.README.md)

Then post the resulting URL here.

( This URL should contain **sprunge.us** )

> > However earlier (prior to installing IIAB) the Debian 11 VM failed to get an IP address over Ethernet (Bridged Adapter from VirtualBox's perspective) during one of its many boots. So that one time when the networking failed I had to manually run `dhclient enp0s3` which solved the problem.
> 
> Can you expand on the 'one time'? as somewhere between 2nd <-> 10th reboot? Looking for a general trend and as you mentioned this was noted to be pre-iiab install so what ever ills Debian has, is preexisting.

It just happened once.  It was probably the 3rd or 4th boot of the VM (a very minimal install of Debian 11 Bullseye).  I probably booted the VM about 6-8 times total (prior to installing IIAB onto the VM).

**Comment by jvonau** [comment]
> I'm not seeing that problem after installing IIAB (MIN-sized + Node-RED) onto Debian 11 Bullseye (server not desktop, latest daily net-install ISO) on a VM. Certainly not yet, after a few reboots!

Like pulling hens teeth... Think we need to establish how Debian was installed in the ill behaving install, something is different between the 2 (note rarely seen part of the code above) and saying 'is works here' is not productive without context. Recall #2287 where I suggested successful installs could post a iiab-diagnostics to a tracking issue to support claiming a successful install on given OS/hardware combination as a progress indicator for a release? Didn't think so, and we have nothing to reference what a successful install looks like in this discussion. Over to you.

@deldesir how exactly did you install Debian? 

> It just happened once. It was probably the 3rd or 4th boot of the VM (a very minimal install of Debian 11 Bullseye). I probably booted the VM about 6-8 times total (prior to installing IIAB onto the VM).

Think that is your own local issue, slow dhcp has been noted by myself and others at your location in the past across various OSs.

**Comment by holta** [comment]
> installs could post a iiab-diagnostics to a tracking issue to support claiming a successful install on given OS/hardware combination

How could privacy best be strengthened in support of ethical community telemetry?

What are some ways that people could take pride in participating, in support of others (and meeting others!) with distros/hardware/content like their own?

This is especially important as most IIAB implementers are humanitarians seeking a more believable alternative to the 'Surveillance Capitalism' exploitive model that dominates many 'Silicon Valley' initiatives.

**Comment by holta** [comment]
Thanks @deldesir for your pastebin here, extracted manually from directory `/etc/iiab/diag` after your Debian networking unfortunately stopped working:

https://pastebin.ubuntu.com/p/M5y8hpzY3R/

**Comment by jvonau** [comment]
> > I'm not seeing that problem after installing IIAB (MIN-sized + Node-RED) onto Debian 11 Bullseye (server not desktop, latest daily net-install ISO) on a VM. Certainly not yet, after a few reboots!
> 
> Like pulling hens teeth... Think we need to establish how Debian was installed in the ill behaving install, something is different between the 2 (note rarely seen part of the code above) and saying 'is works here' is not productive without context. Recall #2287 where I suggested successful installs could post a iiab-diagnostics to a tracking issue to support claiming a successful install on given OS/hardware combination as a progress indicator for a release? Didn't think so, and we have nothing to reference what a successful install looks like in this discussion. Over to you.

Where is the information from the 'Debian 11 Bullseye (server not desktop, latest daily net-install ISO) on a VM' installation that I requested? I noted above was that systemd-networkd is disabled but without the prior data that was been requested in the past I can't come to a general conclusion on what is the 'stock out of the box layout' of config files and daemons that are commonly used within debian or if systemd-networkd was purposely turned off by the user.

I also noted that the above diagnostics contains both entries for SSIDs in /etc/wpa_supplicant/wpa_supplicant.conf and /etc/network/interfaces for the WiFi network device with different ssid/pw entries present, these two methods are mutually exclusive, one shouldn't use both methods at the same time. Points to a user configuration error and/or misunderstanding before installation of iiab ever begun.

With #2579 changing what was returned from a 3-way state of "" for not installed, 'disabled' for installed but not intended to run, and 'enabled' for actively running to using a default of 'none' for not installed or enabled for the network services kind of muddies the waters. How is this better or more readable now that the intended use has been altered? Now I have to ask if the services are in fact installed but disabled as I can no longer tell at a glance, how this is viewed 'progress' and not a regression is above my pay grade.

**Comment by jvonau** [comment]
> Instead of 'ifupdown' what is recommended these days?

I documented that in #2382 in the past, Should systemd-networkd be enabled the preferred method I would use (replacing IFACE with interface name of the WiFi device, keep the @ in place) is to create /etc/wpa_supplicant/wpa_supplicant-IFACE.conf with the same general contents as wpa_supplicant.conf and  

/etc/systemd/network/IFACE.network containing:

[Match]
Name=IFACE

[Link]
RequiredForOnline=no

[Network]
DHCP=yes

start the service to test sudo systemctl start wpa_supplicant@IFACE
enable the service for reboot sudo systemctl enable wpa_supplicant@IFACE
to test sudo systemctl restart systemd-networkd 
confirm network access, reboot to confirm if the settings still work.

**Comment by holta** [comment]
> Where is the information from the 'Debian 11 Bullseye (server not desktop, latest daily net-install ISO) on a VM' installation that I requested?

Sorry I didn't understand these details were being asked for.  Fortunately (VM was supposed to be deleted by now) I was able to recover that info:

http://sprunge.us/SZoMp0

**Comment by jvonau** [comment]
> > Where is the information from the 'Debian 11 Bullseye (server not desktop, latest daily net-install ISO) on a VM' installation that I requested?
> 
> Sorry I didn't understand these details were being asked for. Fortunately (VM was supposed to be deleted by now) I was able to recover that info:
> 
> http://sprunge.us/SZoMp0

I'm not a daily Debian user so without a bit of feedback I have no clue if the code is functioning as intended, with the test matrix as large as what network has to accommodate I have to rely on reported results to gauge functionality. It's very important to gather feedback from successful installs to gather a baseline of expected results for other installs going forwarded. Thank you, that clears things up a bunch as Debian is not the primary focus of supported OSs, it appears that Debian is the last remaining cling-on to ifupdown as deducted by noting
>systemd_networkd_active = False
wan_in_interfaces = True

With that cleared up #2586 was the first use of Debian with 2 network interfaces, WiFi with ap0 in this case that I recall. The other trigger would be second wired ethernet device, that should create br0 also, that should be tested as some point also.

The next point of would be how should the wifi be configured with ifupdown in play, answer in /etc/network/interfaces as per the original report that is just perfect and the network code makes no attempt to alter what is present. I'm a bit perplexed as to how /etc/wpa_supplicant/wpa_supplicant.conf became into existence.

**Comment by jvonau** [comment]
the plot thickens, 
>sudo: unable to open log file: /var/log/sudo.log: Read-only file system
When I do sudo
... and it goes to root:
root@box:/home/iiab-admin#
And this is now I do:
mount -o remount,rw /dev/mmcblk0p1
All this is possible because I commented UUIDs on /etc/fstab. Without this it won't boot outside recovery mode.
Send a message

**Comment by jvonau** [comment]
both are missing 
>; then

https://github.com/iiab/iiab/blob/master/roles/network/templates/network/iiab-hotspot-off#L31
https://github.com/iiab/iiab/blob/master/roles/network/templates/network/iiab-hotspot-on#L32

**Comment by holta** [comment]
FYI @jvonau & @deldesir did a ton of work today to track down these issues and others.

(The `iiab-hotspot-off` and `iiab-hotspot-on` fixes would appear to help many-if-not-all folk using graphical desktops, as many/most depend on NetworkManager ?)

**Comment by holta** [comment]
PS they discovered a few things that should be changed in iiab-diagnostics

e.g. how to quickly tell if a laptop/device is NOT capable of dual Wi-Fi.

**Comment by jvonau** [comment]
> (The `iiab-hotspot-off` and `iiab-hotspot-on` fixes would appear to help many-if-not-all folk using graphical desktops, as many/most depend on NetworkManager ?)

None of the above, the only way to hit the error is to intentionally change wifi_up_down  to False to disable ap0 and run in legacy mode prior to the install or prior to running iiab-network. Only then was the error exposed in iiab-hotspot-on|off which to date nobody it appears to has needed or wanted to disable ap0.

**Comment by jvonau** [comment]
`iw list` with the string "interface combinations are not supported" is the key to identifying WiFi devices that don't support the usage of ap0.

**Comment by holta** [comment]
Also, here's another password redaction `iiab-diagnostics` could use — see the last line:

```
=IIAB==========================================================================
-rw-r--r-- 1 root root 348 Oct 22 12:03 /etc/network/interfaces

# This file describes the network interfaces available on your system
# and how to activate them. For more information, see interfaces(5).

source /etc/network/interfaces.d/*

# The loopback network interface
auto lo
iface lo inet loopback

# My wifi
allow-hotplug wlp2s0
iface wlp2s0 inet dhcp
        wpa-ssid w1network1
        wpa-psk XXXXXXXX
```

This might also be desirable across the board in `/etc/network/interfaces.d/*` if that can be accomplished?

**Comment by jvonau** [comment]
/etc/netplan/* if not already covered. sample file in #2382

**Comment by jvonau** [comment]
The another thing to note is that in local_vars.yml host_country_code: **must** be set to the same regulatory domain as the machine was provisioned with or hostapd will not start. This information maybe available as REGDOMAIN= in /etc/default/crda, what is populating country= in /etc/wpa_supplicant/wpa_supplicant.conf. or reveled with `iw reg get`

**Comment by holta** [comment]
> The another thing to note is that in local_vars.yml host_country_code: **must** be set to the same regulatory domain as the machine was provisioned with or hostapd will not start. This information maybe available as REGDOMAIN= in /etc/default/crda, what is populating country= in /etc/wpa_supplicant/wpa_supplicant.conf. or reveled with `iw reg get`

Where would be the best place to enforce that programmatically?

And if so, presumably remove `host_country_code:` from MIN/MEDIUM/BIG-sized local_vars.yml (if not also from default_vars.yml) ?

**Comment by jvonau** [comment]
> > The another thing to note is that in local_vars.yml host_country_code: **must** be set to the same regulatory domain as the machine was provisioned with or hostapd will not start. This information maybe available as REGDOMAIN= in /etc/default/crda, what is populating country= in /etc/wpa_supplicant/wpa_supplicant.conf. or reveled with `iw reg get`
> 
> Where would be the best place to enforce that programmatically?
> 

This gets funky real fast, it's pretty ugly for a very basic setting across the various distros and display managers involved, none of which has a standardized way of setting the country code or referencing a global setting. Too many different knobs in a variety of places with no clear path to the light.

> And if so, presumably remove `host_country_code:` from MIN/MEDIUM/BIG-sized local_vars.yml (if not also from default_vars.yml) ?

Well that might be a thought but the setting was mainly geared towards RaspOS, where the setting is taken as being authoritative and the code deals with rfkill, adjusts wpa_supplicant.conf and hostapd.conf, with everything playing happily along. The other distros are more problematic with if the country code is set or used, and there is a fallback world domain to consider in any output. There should be a pile of data collected across the various distros on the subject before I will even look at cutting any code, but might consider guiding that effort. This is next release material if I even choose to look at the subject given my past research into it. I'd be looking at Release Note fodder nothing more.

**Comment by holta** [comment]
> across the various distros and display managers involved, none of which has a standardized way of setting the country code or referencing a global setting

This is the kind of thing Ansible should be good at...

One would assume (perhaps wrongly!!) that Ansible might already have a somewhat standard way of accessing the OS's country code?

Refs:

- #724 "Raspbian 2018-03-13 blocks WiFi til country set; complicates 1-line installs?"
- PR #784 "host_country_code: US moved to local_vars.yml; download_timeout: 100 -> 200 for Haiti"

**Comment by jvonau** [comment]
> > across the various distros and display managers involved, none of which has a standardized way of setting the country code or referencing a global setting
> 
> This is the kind of thing Ansible should be good at...
> 
> One would assume (perhaps wrongly!!) that Ansible might already have a somewhat standard way of accessing the OS's country code?

Your assignment if you so choose is to find that info and present a PR for review.

**Comment by tim-moody** [comment]
>another thing to note is that in local_vars.yml host_country_code: must be set to the same regulatory domain as the machine was provisioned with or hostapd will not start.

is this true across all OSes and all versions of wifi firmware? I would have guessed that it would not be true for the old firmware.

Also, this should be a runtime configuration, not install-time, so ansible is probably not the right place to attack it.

**Comment by jvonau** [comment]
> > another thing to note is that in local_vars.yml host_country_code: must be set to the same regulatory domain as the machine was provisioned with or hostapd will not start.
> 
> is this true across all OSes and all versions of wifi firmware? I would have guessed that it would not be true for the old firmware.
> 
Not firmware related, kernel driver related and has been that way since kernel 2.6.27 

http://www.linuxwireless.sipsolutions.net/en/developers/Regulatory/CRDA/index.html
https://unix.stackexchange.com/questions/171992/what-are-crda-regulatory-domains-and-why-does-my-wifi-card-keep-trying-to-set-t


> Also, this should be a runtime configuration, not install-time, so ansible is probably not the right place to attack it.

That is what is intended with the country= in wpa_supplicant.conf while NetworkManager relies on reception of country IE frames from the upstream access point. The thing with NM is because of power saving features being used it tends to look for and set the country code very frequently because if a laptop is coming out of one of the sleep modes it may not be in the same regulatory domain as it was when it went to sleep.

The most reliable means to set the regulatory domain is with direct editing of a wpa_supplicant.conf file with netplan being the worst to deal with. RaspOS has a user utility to handle manipulation of the wpa_supplicant.conf for use with dhcpcd, Now systemd-networkd requires hand editing of a wpa_supplicant-IFACE.conf file so that is pretty straight forward while netplan just does not deal with the issue at all. There is a hack in place for systemd-networkd under netplan in https://github.com/iiab/iiab/blob/master/roles/network/templates/hostapd/test-wifi that works changing the regulatory domain from 'world' to US(s/b softcoded up for discussion) that keeps Ubuntu happy on Rpi hardware with both wlan0 and ap0 active.

**Comment by tim-moody** [comment]
I'm trying to get a handle on where each scenario is used

https://iiab.zulipchat.com/#narrow/stream/231879-iiab-devel/topic/wifi.20country.20code/near/214489785

Should Admin Console allow setting of country code when setting upstream password on Control screen? (currently rpi only)

**Comment by jvonau** [comment]
> I'm trying to get a handle on where each scenario is used
> 
> https://iiab.zulipchat.com/#narrow/stream/231879-iiab-devel/topic/wifi.20country.20code/near/214489785
> 
> Should Admin Console allow setting of country code when setting upstream password on Control screen? (currently rpi only)

I see you have that hard coded also, can't hurt to make that settable but the country_code= value in hostapd.conf must match. 

>    except IOError: # for now patch missing file
        wpa_txt = 'ctrl_interface=DIR=/var/run/wpa_supplicant GROUP=netdev\nupdate_config=1\ncountry=US\n'

Netplan is hooped from the start as there was no provision for country
https://bugs.launchpad.net/ubuntu/+source/linux-firmware/+bug/1862760

**Comment by tim-moody** [comment]
>changing the regulatory domain from 'world' to US

is there a lowest common denominator regulatory domain that can be used? leaving aside 5GHz I can't imagine any domain that doesn't permit channel 6, our default, but maybe there are power of signal considerations as well.

**Comment by jvonau** [comment]
There is a default 'world'

rpi | RaspiOS | /etc/wpa_supplicant/wpa_supplicant.conf
Any | Debian (and friends) with ifupdown | /etc/wpa_supplicant/wpa_supplicant.conf or /etc/network/interfaces [1]
Any hardware  | Any distro using systemd-networkd natively (no netplan) | /etc/wpa_supplicant/wpa_supplicant-$IFACE.conf
Any hardware | Ubuntu using netplan with systemd-networkd backend | /run/netplan/wpa-$IFACE.conf
Any hardware | Ubuntu using netplan with NetworkManager backend | NM uses wpa_supplicant over dbus

Doe this help any? NM does a double connect, first one is to get the country from the router and sets it and the second is the real connection that persists. 

1.  I'm on the fence, user reports setting ssid/pw in /etc/network/interfaces while reviewing I saw /etc/wpa_supplicant/wpa_supplicant.conf populated also, the latter I think was populated by installer. User reports receiving an error that doesn't stop the network from loading. Think one of the set of ssid/pw entries are redundant. FWIW I would keep wpa_supplicant.conf and remove the ssid/psk lines but keep the other settings in /etc/network/interfaces.

**Comment by jvonau** [comment]
> https://iiab.zulipchat.com/#narrow/stream/231879-iiab-devel/topic/wifi.20country.20code/near/214489785
> 

Signed up, don't like the ability to send private messages for what is suppose to be an open project. Wondering why the 'glitters project' never received a public response? There is way too much compartmentalization going on though the documented 'contact us' process going on already I don't see any current benefits.

**Comment by jvonau** [comment]
Debian-11 might need the same treatment as Debian-10 [here](https://github.com/iiab/iiab/blob/master/roles/postgresql/tasks/install.yml)

**Comment by holta** [comment]
@deldesir are you able to run `apt install postgresql` on a relatively fresh/clean pre-release install of Debian 11?

If so please take a look at what systemd unit files it installs:

```
apt install mlocate
updatedb
locate postgresql | grep .service
```

(The main one should be `/usr/lib/systemd/system/postgresql.service` but there will likely be others!)

Context: https://github.com/iiab/iiab/blob/master/roles/postgresql/tasks/install.yml#L83-L96

holta:
Just FYI this tkt continues to help monitoring general testing coverage of IIAB on Debian 10/11...

...alongside https://github.com/iiab/iiab/issues/2585#issuecomment-718918905 for Ubuntu 20.10, also acting as an incredibly useful canary-in-the-coal-mine.

Issue #2585, opened 2020-10-22
Testing the new Ubuntu Desktop 20.10+ for RPi 4 (rumor confirmed!) [Bluetooth, MongoDB, NetworkManager issues resolved] [Ubuntu MATE 20.10 seems more polished on RPi 4]


holta:
Can someone help evaluate and test install http://download.iiab.io on this just-released hot-off-the-press new Ubuntu Desktop 20.10?

There's only 1 graphical desktop version available.  But the almost-7-month-old gossip (that Ubuntu would mainline a graphical desktop for Raspberry Pi !) turns out to be true in the end:

https://ubuntu.com/download/raspberry-pi

Noting that Ubuntu Server 20.10 is released in the usual array of 5 versions available for RPi 2, 3 and 4 (32-bit and 64-bit).  I don't expect any problems with the 64-bit server version for RPi 4 as its pre-releases have already been quite heavily tested in recent weeks.  But a final test here would also be very useful to sanity check https://github.com/iiab/iiab/wiki/IIAB-7.2-Release-Notes &mdash; and also to confirm which WiFi drivers they're officially shipping within `/lib/firmware/brcm` per https://github.com/iiab/iiab/issues/823#issuecomment-662285202

Ref:

- #2610 "Who can help test on Ubuntu 20.10 (Server &/or Desktop version!) on x86_64, VM or PC?"

shanti-bhardwa:
```
/boot/firmware$ cat config.txt
[pi4]
max_framebuffers=2

[all]
arm_64bit=1
kernel=vmlinuz
cmdline=cmdline.txt
initramfs initrd.img followkernel

# Enable the audio output, I2C and SPI interfaces on the GPIO header
dtparam=audio=on
dtparam=i2c_arm=on
dtparam=spi=on

# Enable the FKMS ("Fake" KMS) graphics overlay, enable the camera firmware
# and allocate 128Mb to the GPU memory
dtoverlay=vc4-fkms-v3d
gpu_mem=128
start_x=1

# Comment out the following line if the edges of the desktop appear outside
# the edges of your display
disable_overscan=1

# If you have issues with audio, you may try uncommenting the following line
# which forces the HDMI output into HDMI mode instead of DVI (which doesn't
# support audio output)
#hdmi_drive=2

# If you have a CM4, uncomment the following line to enable the USB2 outputs
# on the IO board (assuming your CM4 is plugged into such a board)
#dtoverlay=dwc2,dr_mode=host

[...]

PR #554, opened 2017-11-23
U17-2: refine network for Ubuntu 18.04 LTS daily builds

jvonau:
### Fixes Bug
post 6.5 release

PR #2592, opened 2020-10-27
U-20.10 changed the NM service name

jvonau:
### Fixes bug:
https://github.com/iiab/iiab/issues/2585#issuecomment-716798464

### DIFF
--- a/scripts/local_facts.fact
+++ b/scripts/local_facts.fact
@@ -106,6 +106,8 @@ tmp=$(systemctl is-enabled dhcpcd)
 
 # Debian family only, as is_redhat would use NetworkManager as the service name
 tmp=$(systemctl is-enabled network-manager)
+[[ $tmp == "" ]] &&
+tmp=$(systemctl is-enabled NetworkManager)
 [[ $tmp != "" ]] &&
     NETWORK_MANAGER=$tmp
 


holta:
The symlink (network-manager) that existed since Ubuntu 15.04 appears to have been removed:

> "The unit file for network-manager.service is a symlink to NetworkManager.service"

> "The name network-manager remains simply for compatibility, so people can use old style commands like sudo service network-manager restart"

https://askubuntu.com/questions/919609/whats-the-difference-between-network-manager-service-and-networkmanager-service/919614#919614

jvonau:
>The symlink (network-manager) that existed since Ubuntu 15.04 appears to have been removed:

Nothing new, the change came full circle to the name used in Fedora on the XOs where this started from, with local_facts.facts going back to the early days of ansible adoption in 2012-2013. The 'network-manager service' name is in part inherited from Debian on Ubuntu in the 14-ish - 16.04 XSCE time frame. The detection was introduced by myself in IIAB https://github.com/iiab/iiab/pull/554/files because some people run servers on desktops and it's a bit of a challenge for myself to pull off the automation. I happen like TK's testing, he is the main tester of the Desktop network code over the years that provided **useful** feedback on real hardware to get the Desktop routine to where it is today otherwise I would of filed the reports under ***not supported*** and not have to listen to whining about issues related to Desktops. I don't have to spend my time on any of this, I could of just sat back and waited to see if you figured it out on your own.

PR #2593, opened 2020-10-27
NetworkManager (NM) detection for Ubuntu 20.10&All + remove long-deprecated phplib_dir + Linux Mint /etc/os-release fix

holta:
Partial fix for #2585, aligning IIAB with Debian and Ubuntu's 2015 change from `network-manager.service` to `NetworkManager.service`

Thanks to @jvonau and @floydianslips for tracking this down.

Related: PR #2592

### DIFF
--- a/scripts/local_facts.fact
+++ b/scripts/local_facts.fact
@@ -99,7 +99,7 @@ tmp=$(systemctl is-enabled dhcpcd)
 
 # is_redhat uses "NetworkManager".  Debian 7 & Ubuntu 14.10 required
 # "network-manager" (prior to 2015/systemd).  Ubuntu 20.10 dropped the
-# legacy symlink from "network-manager" to "NetworkManager"
+# legacy symlink from "network-manager.service" to "NetworkManager.service"
 tmp=$(systemctl is-enabled NetworkManager)
 [[ $tmp != "" ]] &&
     NETWORK_MANAGER=$tmp
--- a/scripts/local_facts.fact
+++ b/scripts/local_facts.fact
@@ -1,6 +1,6 @@
 #!/bin/bash
 
-# 2020-10-19: Most of the 12 variables require a command[*] to be run to
+# 2020-10-27: Most of the 11 variables require a command[*] to be run to
 # establish the var's value.  WE DISPLAY ALL ERRORS / DIAGNOSTICS AND CONTINUE.
 #
 # [*] DOESN'T MATTER WHAT COMMAND: so long as it fails with Return Code != 0
@@ -11,7 +11,6 @@ OS="none"
 VERSION_ID="none"    # This var's combined with the above, before being output
 IIAB_BRANCH="none"
 IIAB_COMMIT="none"
-PHPLIB_DIR="none"
 XO_MODEL="none"
 RPI_MODEL="none"
 ANSIBLE_VERSION="none"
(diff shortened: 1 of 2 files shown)


holta:
Tested on 6 mainline OS's {Raspbian, Debian, Ubuntu, Mint} with and without NetworkManager.

jvonau:
The challenge, show me just one howto out on the internet from the 16.04 time frame where NetworkManager was used in place of network-manager in the instructions. Good luck, hell even current [Debian Docs](https://www.debian.org/doc/manuals/debian-reference/ch05.en.html) mentions 'network-manager'

holta:
Debian 8 (April 2015) is similar to Ubuntu 15.04 here:

> To disable Network Manager on Debian 8 or later:
$ sudo systemctl stop NetworkManager.service
$ sudo systemctl disable NetworkManager.service

https://www.myridia.com/dev_posts/view/2022

jvonau:
> Debian 8 (April 2015) is similar to Ubuntu 15.04 here:
> 
> > To disable Network Manager on Debian 8 or later:
> > $ sudo systemctl stop NetworkManager.service
> > $ sudo systemctl disable NetworkManager.service
> 
> https://www.myridia.com/dev_posts/view/2022

Written in 27.02.18, after the release dates of the distros mentioned, next.

holta:
> [Debian Docs](https://www.debian.org/doc/manuals/debian-reference/ch05.en.html)

Above mentions `NetworkManager.service` not `network-manager.service`

holta:
What Matters: IIAB does not need to support Debian 7 and Ubuntu 14.x from 2013/2014.

jvonau:
> Restart NM or Wicd by the following.
$ sudo /etc/init.d/network-manager restart
$ sudo /etc/init.d/wicd restart

> What Matters: IIAB does not need to support Debian 7 and Ubuntu 14.x from 2013/2014.

Good you can go back and remove the phplib_dir part of local_facts.facts code nothing uses that variable now, but xo_model is a bit trickier
https://github.com/iiab/iiab/search?q=phplib_dir
https://github.com/iiab/iiab/search?q=xo_model

jvonau:
> Tested on 6 mainline OS's {Raspbian, Debian, Ubuntu, Mint} with and without NetworkManager.

Your numbers don't compute, I see 4 OSs listed yet you mention 6. here is what I think is tested:

RaspOS-10 lite
Ubuntu-20.04 server amd64 and arm32/64
Debian-10|11 server amd64
RaspOS-10 desktop
That would be the OSs without network-manager (package name) installed

Ubuntu-20.04 desktop amd64 (pre-installed)
Ubuntu-20.04 arm32/64 being an add-on to server (pulls in NM)
Mint-20 desktop amd64 (only no arm) (pre-installed)
Ubuntu-20.10 desktop arm32/64 for #2585 (pre-installed)
Debian-10 desktop amd64 Can't recall ever seeing any feedback in recent memory assuming (pre-installed),
That would be the OSs with network-manager (package name) installed.

holta:
The long-deprecated `phplib_dir` should also be removed from Admin Console's https://github.com/iiab/iiab-admin-console/blob/master/vars/default_vars.yml#L14 if possible.

holta:
Smoke-tested on 4 OS's {Raspbian Lite, Debian 10.6 VM, Ubuntu Desktop 20.10 on RPi, Linux Mint 20 VM} the latter 2 with NetworkManager.

scripts/local_facts.fact
```diff
@@ -36,7 +35,9 @@ if [ -f /etc/rpi-issue ]; then
     OS="raspbian"
 fi
 
-if tmp=$(grep ^VERSION_ID= /etc/*elease); then
+# /etc/lsb-release could also be grep'd.  But /etc/upstream-release/lsb-release
+# on Linux Mint 20 caused grep of /etc/*elease to fail (on directory not file)
+if tmp=$(grep ^VERSION_ID= /etc/os-release); then
```
jvonau, line 40: Any reason not to have line 30 the same as the change in line 40? It's intended to be the same file that is referenced.

jvonau:
Might want to run something like shellcheck on the bash scripts, kind of enlightening. https://github.com/iiab/iiab/compare/master...jvonau:shellcheck?expand=1 is what found.

scripts/local_facts.fact
holta: Thanks for catching this.  It's serious and should be fixed by PR #2596

holta:
> Might want to run something like shellcheck on the bash scripts, kind of enlightening. https://github.com/iiab/iiab/compare/master...jvonau:shellcheck?expand=1 is what found.

PR #2597 "scripts/local_facts.fact: quoting vars is safer...on these 3 lines"

PR #2596, opened 2020-10-27
Fix $OS in scripts/local_facts.fact for Mint

holta:
Fixes PR #2593 and earlier.

Thanks @jvonau for catching this, as sourcing variables from `/etc/iiab/iiab.env` was masking the bug on a prior install.

Ref: #2585


holta:
Tested on Linux Mint 20 VM.

PR #2597, opened 2020-10-27
scripts/local_facts.fact: quoting vars is safer...on these 3 lines

holta:
As recommended by @jvonau at:

- https://github.com/iiab/iiab/pull/2593#issuecomment-717398600
- https://github.com/jvonau/iiab/blob/shellcheck/scripts/local_facts.fact

As lint-like programs go, [ShellCheck](https://www.shellcheck.net/) (shell script analysis tool) isn't perfect but it's far better than nothing to surface common errors!


holta:
Tested on Linux Mint 20.

PR #2602, opened 2020-10-28
MongoDB 4.4.x for Debian 11 "Bullseye" testing (and Debian Sid too?)

holta:
For broader testing of IIAB on [Debian 11 (Bullseye)](https://github.com/iiab/iiab/wiki/IIAB-Platforms#operating-systems) especially, whose daily builds are already incredibly solid 6-9 months before release.

Of course if you do choose to test IIAB on Debian 11, please remember to insert this line into your `/etc/os-release` :

```
VERSION_ID="11"
```

Somewhat related PRs: #2493, #2582, #2595, #2600

cc: @deldesir, @jvonau

PR #2595, opened 2020-10-27
MongoDB for Ubuntu 20.10 (Groovy Gorilla)

holta:
For #2585, building on @jvonau's major cleanup of MongoDB (PR #2493) in August 2020.

Tested by installing Sugarizer which successfully pulled in MongoDB 4.4.1 on Ubuntu Desktop 20.10 on RPi 4.

Tangentially related: PR #2593, PR #2594.

cc: @floydianslips, @deldesir

PR #2604, opened 2020-10-30
Clean MongoDB/Sugarizer for Debian 11/Sid & All OS's

holta:
Supersedes PR #2602

Allows MongoDB and Sugarizer to run on Debian 11 Bullseye and probably also Debian Sid.

But more importantly this makes Sugarizer's dependency (via systemd, not via Ansible) on MongoDB more readable and more maintainable.

Tested on Debian and Ubuntu.

cc: @jvonau

Refs:

- PR #2582 "Overseas request for immediate testing of IIAB on Debian 11 Bullseye + similarly clean out long-deprecated apache_config_dir from high-level <OS>.yml vars files [ASIDE: dnsmasq & dhclient issues? host_country_code MUST match PC/HW, or hostapd won't start]"
- PR #2595 "MongoDB for Ubuntu 20.10 (Groovy Gorilla)"
- PR #2606 "Node.js 12 -> 14 LTS for Node-RED, Sugarizer and Internet Archive"


holta:
Aside: Sugarizer was tested on Node.js 14 LTS (v14.15.0) and it works fine, as @llaske indicated.

jvonau:
Shouldn't really mix the 0-init changes in with a different theme of changes, but it is nice you are using bits of my PRs. If you remove  the apache.yml you should also remove the site config template install.yml also, I see that apache.yml & nginx.yml replaced with enable-or-disable.yml is a deviance from all the other web-apps, Is this going to be the new normal layout across all web apps? Should everybody expect this change across all roles?

holta:
FYI this PR was tested on 3 different OS's including 64-bit RaspiOS Lite from 2020-08-20.

> I see that apache.yml & nginx.yml replaced with enable-or-disable.yml is a deviance from all the other web-apps, Is this going to be the new normal layout across all web apps? Should everybody expect this change across all roles?

Those are good questions.  I want code to be more approachable to wider community newcomers.  So this is definitely an option if NGINX replaces Apache in a more comprehensive way going forward.

> If you remove the apache.yml you should also remove the site config template install.yml also

Which site config template install.yml ?

roles/sugarizer/tasks/install.yml
```diff
@@ -153,6 +154,10 @@
     src: sugarizer.service
     dest: /etc/systemd/system
 
+- name: systemctl daemon_reload, so systemd (re)reads sugarizer.service
+  systemd:
+    daemon_reload: yes
+
```
jvonau, line 160: Re: apache template stanza in install.yml starts at line lines 161 below.

holta:
FYI MongoDB doc on port-changing, thanks to @jvonau:

https://docs.mongodb.com/manual/reference/default-mongodb-port/

roles/sugarizer/tasks/main.yml
```diff
@@ -19,25 +19,22 @@
...
 
-# If we got here sugarizer_install is True
 - name: "Set 'mongodb_install: True'"
   set_fact:
     mongodb_install: True
 
-# MongoDB auto-started by sugarizer.service - let's set 'mongodb_enabled: True'
-# regardless - anticipating that MongoDB (should!) run on demand:
-- name: "Set 'mongodb_enabled: True' when sugarizer_enabled"
-  set_fact:
```
jvonau, line 33: This is longer true once you invoke the systemd module with "enabled: yes" in mongodb role.

jvonau:
The point is, with always having enabled no as the default runtime setting for mongodb one only needs to `sudo systemd disable sugarizer` and reboot to fully disable both sugarizer and mongodb as opposed to the future **requiring** `runrole sugarizer` or `sudo systemd disable mongodb` to fully turn off mongodb. Or it seems like you are chasing some sysV/init script like issue on Debian.

roles/sugarizer/tasks/main.yml
holta: This PR completely avoids enabling MongoDB.

As explained in several Ansible `name:` comments — the idea is to keep Ansible variable `mongodb_enabled` more properly in sync with the output of systemd's `systemctl is-enabled mongodb` — as the prior code left them very confusingly out of sync.

So while I'm not in full agreement that `mongodb.service` should remain disabled, my goal here was only to make the code more readable — while preserving its current mechanics as much as possible — i.e. respecting de facto status of systemd's `is-enabled` and bringing Ansible var `mongodb_enabled` into line with de facto current practice.

This allows MongoDB to be easily enabled in future (both on the Ansible side and the systemd side, keeping both in sync going forward to maintain sanity) — if enabling it is indeed thought to be preferable[*].

[*] e.g. for purposes of MongoDB reliability.

tim-moody:
It would be great one day to have a set of scripts like iiab-<role>on and off that hide the details of how to enable and disable an already installed role. @jvonau could probably write them all in an hour. Admin Console could call them. <role> should be the same as <role>_enabled.

holta:
> The point is, with always having enabled no as the default runtime setting for mongodb one only needs to `sudo systemd disable sugarizer` and reboot to fully disable both sugarizer and mongodb as opposed to the future **requiring** `runrole sugarizer` or `sudo systemd disable mongodb` to fully turn off mongodb.

MongoDB is _not_ being enabled, as explained in this PR's code and above: https://github.com/iiab/iiab/pull/2604#discussion_r515043928

Again: I don't fully agree with this, but the goal here was to honor de facto current practices as much as possible, adding far cleaner explanations, while also bringing Ansible var `mongodb_enabled` more properly into line with de facto systemd status (output of `systemctl is-enabled mongodb`).

roles/sugarizer/tasks/install.yml
holta: Done.  Thanks.

jvonau:
The "port usage" url above points out the port used by mongod can vary based on the cmd-line flag used to start mongod but haven't looked any farther. Just wondering about the port dance going on in https://github.com/iiab/iiab/blob/master/roles/sugarizer/tasks/install.yml#L195-L199, why, and if we are using the correct cmdline options to start the mongodb service. I'll point to the altering of the {{ mongodb_port }}(set to 27018). would only cover 32bit arm OS's via the mongod.conf, the 64bit os are hardcoded in mongodb's install.yml with a lineinfile statement, forgot to softcode that one,

roles/sugarizer/tasks/main.yml
holta: Aside: the reason we might want to enable MongoDB in future (on the Ansible side and systemd side, keeping both in sync as much as possible to maintain sanity!) is to make MongoDB more reliable.

e.g. for IIAB installations that are not constantly being rebooted, but rather are left on for many weeks like a traditional server.

Allowing Sugarizer (and potentially other apps using MongoDB) to become more reliable.

jvonau:
> It would be great one day to have a set of scripts like iiab-on and off that hide the details of how to enable and disable an already installed role. @jvonau could probably write them all in an hour. Admin Console could call them. should be the same as _enabled.

how about 15 minutes, but you dislike sed..

tim-moody:
apart from the fact that I dislike sed, I'm talking about scripts that would read the value of <role>_enabled and take the appropriate action for each role, such as stopping and disabling services, without having to run ansible.

jvonau:
Upon further review it appears that if {{ mongodb_port }} set to 27017 the need for https://github.com/iiab/iiab/blob/master/roles/sugarizer/tasks/install.yml#L195-L199 and https://github.com/iiab/iiab/blob/master/roles/mongodb/tasks/install.yml#L115-L120 goes away. Perhaps the requirement to use port 27018 is because 27017 is already occupied and now there might be 2 mongod processes running on different ports? Out for the day check back later.

jvonau:
> apart from the fact that I dislike sed, I'm talking about scripts that would read the value of _enabled and take the appropriate action for each role, such as stopping and disabling services, without having to run ansible.

Why, doesn't that throw the iiab.ini file out of sync with what is really enabled|disabled on the machine?

jvonau:
With the port altered to be one that is also used by the same process with slightly different invocation as noted in the "port usage" url, looking from a more experienced mongodb administrators point of view(and I don't claim to be one) can really lead to some confusion when seeing a port is bound but the expected features are absent. We go out of the way to alter sugarizer which uses the stock port, anybody know why a different port is used other than it has been like that forever?

roles/sugarizer/tasks/main.yml
jvonau: > This PR completely avoids enabling MongoDB.
> 
No, there is code that will, just currently unused.

> As explained in several Ansible `name:` comments — the idea is to keep Ansible variable `mongodb_enabled` more properly in sync with the output of systemd's `systemctl is-enabled mongodb` — as the prior code left them very confusingly out of sync.
> 

Why so? I was assuming the iiab.ini file was recording the "expected state of the service when booted", how the service is started is irrelevant for that purpose.

> So while I'm not in full agreement that `mongodb.service` should remain disabled, my goal here was only to make the code more readable — while preserving its current mechanics as much as possible — i.e. respecting de facto status of systemd's `is-enabled` and bringing Ansible var `mongodb_enabled` into line with de facto current practice.
> 

The pertinent information is "should the service be running when booted" not one of if systemd should automaticlly start the service standalone, as the service is started on demand by sugarizer.

> This allows MongoDB to be easily enabled in future (both on the Ansible side and the systemd side, keeping both in sync going forward to maintain sanity) — if enabling it is indeed thought to be preferable[*].
> 
> [*] e.g. for purposes of MongoDB reliability.

If you think so but whatever.

jvonau: > Aside: the reason we might want to enable MongoDB in future (on the Ansible side and systemd side, keeping both in sync as much as possible to maintain sanity!) is to make MongoDB more reliable.
>
More reliable in which way?
 
> e.g. for IIAB installations that are not constantly being rebooted, but rather are left on for many weeks like a traditional server.
> 
Now I'm confused, once the service starts the service is running, what are you referring to? 

> Allowing Sugarizer (and potentially other apps using MongoDB) to become more reliable.

If you want other apps to easily integrate then maybe look at port usage issue I pointed out.

jvonau:
> > I see that apache.yml & nginx.yml replaced with enable-or-disable.yml is a deviance from all the other web-apps, Is this going to be the new normal layout across all web apps? Should everybody expect this change across all roles?
> 
> Those are good questions. I want code to be more approachable to wider community newcomers. So this is definitely an option if NGINX replaces Apache in a more comprehensive way going forward.

You came up with the apache|nginx layout and now are switching back to the way it was 10 months ago makes it seem like there was a bunch of needless code churn at this point. All you really need to do is drop the apache.yml include to disable the apache routine now that the apache site template has been removed. Since the debate is what the value should read, just had a semi-brainwave why not just fix the message in iiab.ini to just record 'on_demand' to be accurate?
>- option: enabled
  value: "{{ mongodb_enabled }}"

  value: on_demand

tim-moody:
>Why, doesn't that throw the iiab.ini file out of sync with what is really enabled|disabled on the machine?

it would. I thought you were talking about local vars.

If you can write these scripts in 15 minutes it would be a good use of your time.

jvonau:
> > Why, doesn't that throw the iiab.ini file out of sync with what is really enabled|disabled on the machine?
> 
> it would. I thought you were talking about local vars.
>

I was inferring using sed to alter local_vars then call runrole thus keeping the system state, local_vars, and iiab.ini in sync.

> If you can write these scripts in 15 minutes it would be a good use of your time.

I'll disagree once the debate starts or the PR is just ignored as usual. For what use would be putting a wrapper around 'systemctl dis|enable' and/or 'start|stop'? To allow admin-console to move away from ansible and call the scripts in place of ansible? No thanks, outside what interests me as a volunteer so it would be best if you recruited Adam to write those scripts as only then the scripts will be _perfect_ and ready to merge.

tim-moody:
>No thanks,

OK. Sorry to trouble.

holta:
> > * option: enabled
> >   value: "{{ mongodb_enabled }}"
> 
> value: on_demand

Poetically elegant narrative for https://github.com/iiab/iiab/blob/master/roles/sugarizer/templates/sugarizer.service#L4 which is indeed a cute hack after all.  It this were 2018 and many other roles used this same hack I'd agree.  But in 2020 with the True/False code safety validation and intentionally simplified _install _enabled scheme that's stabilized after converging for half a decade, I personally don't think it's worth throwing all we've just built up into the air for the sake of (what is indeed) classy poetry.

The real question for schools that treat IIAB as a traditional server (leaving it on for months, rather than rebooting it every hour) would be if `mongodb.service` can/should be beefed up, to mitigate unpredictable weirdo crashes that we cannot anticipate today, and if so how we'd best go about that (including playing with MongoDB port[s] optimization as nec, if that's somehow related).

In any case this PR and it predecessor have been very intensively tested on 5 OS's over 5 days now, and design debates will surely continue long after this PR, as very well they should!

jvonau:
> > > * option: enabled
> > >   value: "{{ mongodb_enabled }}"
> > 
> > 
> > value: on_demand
> 
> Poetically elegant narrative for https://github.com/iiab/iiab/blob/master/roles/sugarizer/templates/sugarizer.service#L4 which is indeed a cute hack after all.

Can't call it a hack when upstream has documented the usage, and I didn't introduce the usage in the service file.
>It this were 2018 and many other roles used this same hack I'd agree. But in 2020 with the True/False code safety validation 
and intentionally simplified _install _enabled scheme that's stabilized after converging for half a decade, 
Given the non-user usage of the of the mongodb flags there is no need to over zealously validate this as a user setting at all.

>I personally don't think it's worth throwing all we've just built up this air for the sake of (what is indeed) classy poetry.
>
So much for the 'say as I do' maxim that bantered about. Nothing is tossed out, the proposed change to the recording in *iiab.ini* doesn't alter what is in local_vars.yml thus not impacting validation at all, this is just more FUD being spun. 

> The real question for schools that treat IIAB as a traditional server (leaving it on for months, rather than rebooting it every hour) would be if `mongodb.service` can/should be beefed up, to mitigate unpredictable weirdo crashes that we cannot anticipate today, and if so how we'd best go about that (including playing with MongoDB port[s] optimization as nec, if that's somehow related).
>

This statement is a bit of smokescreen, how the service is started is a very different issue from the service crashing later during use. Any crashes post-startup would not be an iiab issue but rather a bug in sugarizer or mongodb which need to be addressed at that level and has nothing to do with this discussion unless you are suggesting either of the services are not as robust as first thought requiring further mitigation efforts. Any hard evidence or even anecdotal to support that notion?

> In any case this PR and it predecessor have been very intensively tested on 5 OS's over 5 days now, and design debates will surely continue long after this PR, as very well they should!

You glossed over the port usage shows the lack of technical understanding of the issue so I will spell it out as plainly as possible. Both sugarizer and mongodb come out of the box using the stock port 27017 yet both are altered to use 27018 and the question is why? I mocked both services to use the stock port 27017 as a test, which works as expected. Why is the port being altered without having a technical reason to do? This non-stock port usage potentially effects other future mondodb projects, requiring the same unneeded port change that is required by sugarizer. The use of port 27018 can be confusing and can clash with another mongd process that is started with a different flag, set either in the config file or the cmdline that causes a process to bind with 27018. Closer to out of the box the easier for other new roles to drop into place without the hoop jumping that is currently required. Do i need to file an issue for greater visibility?

holta:
Again there's nothing wrong with the cute hack at https://github.com/iiab/iiab/blob/master/roles/sugarizer/templates/sugarizer.service#L4 and it should be preserved in all likelihood — especially if `mongodb.service` itself can deliver (1) respectful timing during boot (2) long-term stability over months for schools that want to use Sugarizer as an institutional tool rather than a silly toy.

i.e. I personally see no reason the above hack cannot work _alongside_ [mongodb.service](https://github.com/iiab/iiab/blob/master/roles/mongodb/templates/mongodb.service.j2) also being enabled by systemd, in service to schools that want things reliable.

@llaske may have recommendations here?

In any case, I would not want other apps that need MongoDB to have to jump through complex hoops.

Ansible variables `mongo_install` and `mongo_enabled` abide by this clean simple norm so that complexity is avoided, and so that things 'Just Work' going forward, as core infra ([validated](https://github.com/iiab/iiab/blob/master/roles/0-init/tasks/validate_vars.yml) ideally!) needs to abide by [KISS](https://en.wikipedia.org/wiki/KISS_principle) and [SSOT](https://en.wikipedia.org/wiki/Single_source_of_truth) to be broadly understandable in any public-participatory community project.

jvonau:
> Again there's nothing wrong with the cute hack at https://github.com/iiab/iiab/blob/master/roles/sugarizer/templates/sugarizer.service#L4 and it should be preserved in all likelihood — especially if `mongodb.service` itself can deliver (1) respectful timing during boot (2) long-term stability over months for schools that want to use Sugarizer as an institutional tool rather than a silly toy.
>
I would suggest a dedicated machine for that situation rather than the 'preview' with a bunch of other services enabled that iiab supplies to provide as much system resources for sugarizer and mongodb to use as possible to best showcase sugarizer's performance. 
 
> i.e. I personally see no reason the above hack cannot work _alongside_ [mongodb.service](https://github.com/iiab/iiab/blob/master/roles/mongodb/templates/mongodb.service.j2) also being enabled by systemd, in service to schools that want things reliable.
> 

Again this is not a hack, think of it as an auto dependency for an underlining process that needs to be present before this new process can start. There is no reliability issue that you or anybody else has ever put forth regarding the starting of mongodb, once the service file was corrected in #814, This actually provides better staging of the boot time events (see the /var/run/mongodb log issue in 814) with an auto fallback to mongodb not started on boot should suguarizer be disabled directly via systemctl without having to remember to disable mongodb with systemctl also.

> @llaske may have recommendations here?
>
I welcome the input.
 
> In any case, I would not want other apps that need MongoDB to have to jump through complex hoops.
> 
> Ansible variables `mongo_install` and `mongo_enabled` abide by this clean simple norm so that complexity is avoided, and so that things 'Just Work' going forward, as core infra ([validated](https://github.com/iiab/iiab/blob/master/roles/0-init/tasks/validate_vars.yml) ideally!) needs to abide by [KISS](https://en.wikipedia.org/wiki/KISS_principle) and [SSOT](https://en.wikipedia.org/wiki/Single_source_of_truth) to be broadly understandable in any public-participatory community project.

holta:
I agree that boot timing is very important, and the people who worked on PR #814 deserve the credit.

But a week later (and a month later) also matters, beyond the ribbon-cutting ceremony...

A common example is very poor countries that don't turn off their IIAB on weekends because public schools receive free electricity from the government, so they are lackadaisical about environmental/energy costs.

So the core question is about reliability of Sugarizer and MongoDB in multi-week uptime situations.

jvonau:
> I agree that boot timing is very important, and the people who worked on PR #814 deserve the credit.
> 
> But a week later (and a month later) also matters, beyond the ribbon-cutting ceremony...
> 
> A common example is very poor countries that don't turn off their IIAB on weekends because public schools receive free electricity from the government, so they are lackadaisical about environmental/energy costs.
> 
> So the core question is about reliability of Sugarizer and MongoDB in multi-week uptime situations.

That really is outside of the scope of this discussion on when/how the services start. If there is something crashing then there are other issues that need to be addressed, why did the crash occur? Bug in either in sugarizer, mongo or perhaps the machine is in a out of memory state, all of which is way beyond the end-user's ability to fix, with a reboot being the easiest method for the end-user to restore a functioning machine at least until the next crash. Do you have any data to support that something is crashing or is this just a case of conjecture?

holta:
It's a chicken+egg problem.

Schools won't adopt Sugarizer/MongoDB unless it's clearly resilient.

Especially when their DBA's look under the covers and see a systemd service that's oxymoronically[*] "permanently-on-yet-disabled" — hardly inspiring confidence that this is a professional solution (?)

[*] Or at least confoundingly, confusingly.

jvonau:
> It's a chicken+egg problem.
> 
> Schools won't adopt Sugarizer/MongoDB unless it's clearly resilient.
> 
Without feedback from the field one never knows so it more egg+chicken issue, other than whatever improvements might be gained from upstream progress. 

> Especially when their DBA's look under the covers and see a systemd service that's oxymoronically[*] "permanently-on-yet-disabled" — hardly inspiring confidence that this is a professional solution (?)
>

I can see it the other way, 'why is this db service disabled on boot yet running'. It's not upstream that provided any of the systemd files to use or be inspired from, this project had to create them from scratch ported from instructions at https://github.com/llaske/sugarizer-server/blob/master/docs/install.md to actual systemd service files and commands to handle those tasks and have done a pretty decent job.

> [*] Or at least confoundingly, confusingly.

A real linux DBA should know what to look for systemd wise as the use of Requires= is not uncommon and fully documented by upstream.

holta:
> https://github.com/llaske/sugarizer-server/blob/master/docs/install.md

@llaske do you have any systemd recommendations for Sugarizer and/or MongoDB?

> A real linux DBA should know what to look for systemd wise as the use of Requires= is not uncommon and fully documented by upstream.

Again I have no problem at all with this systemd hackery (as it clearly works at boot).  The question is about long-term operations, not boot-time, and as such [/etc/systemd/system/mongodb.service](https://github.com/iiab/iiab/blob/master/roles/mongodb/templates/mongodb.service.j2) possibly should _also_ be enabled.

But the world has heard enough from Jerry and me.  I'm more interested in others' opinions.

jvonau:
> > https://github.com/llaske/sugarizer-server/blob/master/docs/install.md
> 
> @llaske do you have any systemd recommendations for Sugarizer and/or MongoDB?
> 
> > A real linux DBA should know what to look for systemd wise as the use of Requires= is not uncommon and fully documented by upstream.
> 
> Again I have no problem at all with this systemd hackery (as it clearly works at boot). The question is about long-term operations, not boot-time, and as such [/etc/systemd/system/mongodb.service](https://github.com/iiab/iiab/blob/master/roles/mongodb/templates/mongodb.service.j2) possibly should _also_ be enabled.
> 
> But the world has heard enough from Jerry and me. I'm more interested in others' opinions.

Almost, the use of [Restart=](https://unix.stackexchange.com/questions/564443/what-does-restart-on-abort-mean-in-a-systemd-service) might be useful, but the down side from something like that is the logging can just full up with failed service restart messages if the condition doesn't self-correct with a restart. Putting on my system interrogator hat I would rather have the service fail and have the crash isolated in the logs than looking through 100s of lines of failed error messages that can fill up the logs and require just that much more disk space. Will not hear another peep out of me.

llaske:
> @llaske do you have any systemd recommendations for Sugarizer and/or MongoDB?

I'm not an expert on MongoDB, I have no recommendation here.

BTW if you think that your config is too small to support a running Sugarizer Server instance with MongoDB, you've got the option to use Sugarizer in "no server mode". See [here](https://github.com/llaske/sugarizer/issues/356) and [here](https://github.com/llaske/sugarizer/pull/871).

PR #2606, opened 2020-10-30
Node.js 12 -> 14 LTS for Node-RED, Sugarizer and Internet Archive

holta:
Node 14.x seems to work fine (e.g. with Sugarizer) but further testing is ongoing for Node-RED and https://github.com/iiab/iiab/tree/master/roles/internetarchive

cc: @mitra42

Related: PR #2604 "Clean MongoDB/Sugarizer for Debian 11/Sid & All OS's"

### DIFF
--- a/roles/internetarchive/tasks/install.yml
+++ b/roles/internetarchive/tasks/install.yml
@@ -9,10 +9,10 @@
   include_role:
     name: nodejs
 
-- name: FAIL (STOP INSTALLING) IF nodejs_version is not set to 10.x or 12.x
+- name: FAIL (STOP INSTALLING) IF nodejs_version is not set to 10.x, 12.x or 14.x
   fail:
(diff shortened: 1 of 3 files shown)


mitra42:
I wouldn't expect problems -

holta:
FYI light-testing of http://box/nodered indicates it also works fine with Node.js 14.x

holta:
> I wouldn't expect problems -

Confirmed!

(In any case, I quickly tested a BIG-sized IIAB install on Ubuntu Server 20.04)

jvonau:
[Other fallout](https://github.com/llaske/sugarizer-server/issues/272) and the [fix](https://github.com/llaske/sugarizer-server/commit/45d9cd6c2b68707b7627f8e346d4d0618ad65615) will not be picked up in sugarizer role's as the git clone is based on a tag and not a branch.

PR #2282, opened 2020-03-08
Usbmount fixes

jvonau:
### Fixes Bug
#2277
Even thou usbmount contains a un-mount routine it is now ignored, use the service file to clean up the mountpoint.

PR #2607, opened 2020-10-30
drop udev-reload service

jvonau:
Remove unnecessary workaround for #710 introduced in #345 that was solved with #722


holta:
@georgejhunt can you review this PR?  Thanks if possible!

A bit of background on systemd-udevd.service:
https://www.freedesktop.org/software/systemd/man/systemd-udevd.service.html

holta:
Has this PR been smoke-tested with a fresh install or similar?

georgejhunt:
I think this PR is trying to address the visibility of the root file system
in rpi's booted via external USB. Is that true?
Yes, it's ok with me,  If it is sufficiently tested. I wonder who will test
it.
It's removing something that we added, (udev-reload.service), and just
relying on a systemd unit file. So the change is towards the main stream,
which seems lower risk to me.

georgejhunt:
I'm just remembering that creating the links for external USB sticks was
happening before the sda2 had been remounted read-write (in early boot, it
is read-only). My experience has been that testing of timing issues may
take some time, if external hard disks, slow USB sticks, etc, are added to
the mix.

georgejhunt:
After verification that USB stick contents appear, and disappear, as they
should, I'd worry about whether contents continued to be visible after a
reboot. I remember that we had a failure mode where udev did the correct
thing on device change, but did not handle boot time correctly.

jvonau:
> I'm just remembering that creating the links for external USB sticks was happening before the sda2 had been remounted read-write (in early boot, it is read-only). My experience has been that testing of timing issues may take some time, if external hard disks, slow USB sticks, etc, are added to the mix.

Easy testing `sudo systemctl disable udev-reload` on any current install then reboot with the usb drive present, the basic idea  was tested in #2282 showed that disabling has no ill results. The fix for the read-only mounting issue was delaying the starting point of usbmount with After=rc-local.service, now this kluge can go away.

holta:
> After verification that USB stick contents appear, and disappear, as they should, I'd worry about whether contents continued to be visible after a reboot. I remember that we had a failure mode where udev did the correct thing on device change, but did not handle boot time correctly.

Great idea.  That seems to work (so far) with a few brief tests on 32-bit RaspiOS Lite on RPi 4 w/ 2GB RAM.

This PR looks just about ready to merge, to enable broader community testing.

holta:
> I think this PR is trying to address the visibility of the root file system in rpi's booted via external USB.

Is this true?

If so this should definitely be promoted!

PR #2611, opened 2020-11-01
mongo port - use the KISS method - revert from the mistaken use of 27018 to 27017

jvonau:
### Fixes bug:
non-stock port usage of mongoDB
return to the "factory defaults"
RaspOS 32bit
#2604

### DIFF
--- a/roles/mongodb/defaults/main.yml
+++ b/roles/mongodb/defaults/main.yml
@@ -15,7 +15,6 @@
 
 # mongodb_enabled: False
 
-# mongodb_port: 27017
 
 # All above are set in: github.com/iiab/iiab/blob/master/vars/default_vars.yml
 # If nec, change them by editing /etc/iiab/local_vars.yml prior to installing!
--- a/roles/mongodb/defaults/main.yml
+++ b/roles/mongodb/defaults/main.yml
@@ -15,7 +15,7 @@
 
 # mongodb_enabled: False
 
-# mongodb_port: 27018
+# mongodb_port: 27017
 
 # All above are set in: github.com/iiab/iiab/blob/master/vars/default_vars.yml
 # If nec, change them by editing /etc/iiab/local_vars.yml prior to installing!
--- a/roles/mongodb/tasks/install.yml
+++ b/roles/mongodb/tasks/install.yml
@@ -114,13 +114,6 @@
         - mongodb-org-server
       state: present
 
-#  - name: Change {{ mongodb_conf }} port to {{ mongodb_port }} -- takes effect on next (re)start of the service -- via enable-or-disable.yml or via sugarizer.service auto-starting MongoDB on demand
-#    lineinfile:
-#      path: "{{ mongodb_conf }}"
-#      regexp: "port: 27017"
-#      backrefs: yes
-#      line: "  port: {{ mongodb_port }}"    # 27018
-
   # end block
   when: (ansible_architecture == "aarch64") or (ansible_architecture == "x86_64")
 
--- a/roles/mongodb/tasks/install.yml
+++ b/roles/mongodb/tasks/install.yml
@@ -49,7 +49,7 @@
 
   - name: Install {{ mongodb_conf }} from template (aarch32)
     template:
-      src: mongod.conf.j2
+      src: mongod.conf
       dest: "{{ mongodb_conf }}"    # /etc/mongod.conf
       #owner: root
       #group: root
--- a/roles/mongodb/tasks/install.yml
+++ b/roles/mongodb/tasks/install.yml
@@ -114,12 +114,12 @@
         - mongodb-org-server
       state: present
 
-  - name: Change {{ mongodb_conf }} port to {{ mongodb_port }} -- takes effect on next (re)start of the service -- via enable-or-disable.yml or via sugarizer.service auto-starting MongoDB on demand
-    lineinfile:
-      path: "{{ mongodb_conf }}"
-      regexp: "port: 27017"
-      backrefs: yes
-      line: "  port: {{ mongodb_port }}"    # 27018
+#  - name: Change {{ mongodb_conf }} port to {{ mongodb_port }} -- takes effect on next (re)start of the service -- via enable-or-disable.yml or via sugarizer.service auto-starting MongoDB on demand
+#    lineinfile:
+#      path: "{{ mongodb_conf }}"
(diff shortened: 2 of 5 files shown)


holta:
@llaske does this seem reasonable to you?

Or is there any concrete reason folks stick with 27018?

Thanks if you have ideas/experiences to share!

jvonau:
The use of 27018 goes back to [ day one ](https://github.com/XSCE/xsce/pull/680) of the introduction of sugarizer, I don't go out of my way to second guess other devs efforts, never questioned the use. Maybe I should of spoken up then, but why bother, only to be ignored anyways.

holta:
I don't know anything about why 27018 was chosen, years ago.

Hopefully @llaske and @georgejhunt know a bit more.

Question: if IIAB changes its recommendation from 27018 to 27017, can we please preserve well-tested logic that sets `mongodb_port` so it remains meaningfully softcoded / auto-modifiable in `/opt/iiab/sugarizer-server/env/sugarizer.ini` ?

i.e. if I'm reading this correctly, these Lines 201-205 here should remain uncommented:
https://github.com/iiab/iiab/pull/2611/files#diff-9b71051e09215354666c64f8299d702e31d6a099f70e0814b8bb4943f4db9eea

jvonau:
> I don't know anything about why 27018 was chosen, years ago.
> 
> Hopefully @llaske and @georgejhunt know a bit more.
> 
> Question: if IIAB changes its recommendation from 27018 to 27017, can we please preserve well-tested logic that sets `mongodb_port` so it remains meaningfully softcoded / auto-modifiable in `/opt/iiab/sugarizer-server/env/sugarizer.ini` ?

Wow, just wow, sugarizer.ini come populated with 27017 [from upstream](https://github.com/llaske/sugarizer-server/blob/master/env/sugarizer.ini), not to mention that mongodb is listed in [IANA service name port number listing.](https://www.iana.org/assignments/service-names-port-numbers/service-names-port-numbers.txt) so I removed the soft-coding altogether.

holta:
27017 IANA recommendation might well be wise for IIAB to adopt too (seems likely, if others agree!)

But why should working code be removed for people that might want to customize ports for security/obfuscation reasons etc?

jvonau:
> 27017 IANA recommendation might well be wise for IIAB to adopt too (seems likely, if others agree!)
> 
> But why should working code be removed for people that might want to customize ports for security/obfuscation reasons etc?

Furious argument, the service is bound to 127.0.0.1, so it acts sort of like a socket and is not seen by the outside world as available.

tim-moody:
https://docs.mongodb.com/manual/reference/default-mongodb-port/

all of the above?

holta:
Aside: IIAB ports / recommendations are enumerated here:
https://github.com/iiab/iiab/wiki/IIAB-Networking#list-of-ports--services

(Whatever we end up with, this is subject to change, but hopefully not too often ;)

tim-moody:
https://github.com/llaske/sugarizer-server has the port as 27018

is there a reason to change it?

holta:
> https://github.com/llaske/sugarizer-server has the port as 27018
> 
> is there a reason to change it?

Personally I'm happy to change it to 27017 as @llaske appears to have done here:
https://github.com/llaske/sugarizer-server/blob/master/env/sugarizer.ini#L22

Following @llaske's recommendations whatever those may be?

But removing the port-softcoding logic definitely feels like overkill.  To me.

holta:
I guess Sugarizer (Sugarizer Server, to be exact) made that change from port 27018 to 27017 on 2018-10-14, thanks to @tarunsinghal92 who appears to be based in Toronto!

https://github.com/llaske/sugarizer-server/commit/b139e55b8cb89f523f5ecbe2318893bc52ff4405#diff-2af86d17090d1ed1caa8cd099561ebd56e029713c325d237bb02dff38ef3275f

tim-moody:
I see. The env file does not match the readme, which is all I looked at. Since we only use mongodb for sugarizer, I'd also follow their usage, but I'd keep the soft coding as the link I gave above lists 27017, 18, 19 under various circumstances.

jvonau:
> https://docs.mongodb.com/manual/reference/default-mongodb-port/

The above was referenced in 2604, 2 days ago after it was posted in chat by me as gentle persuasion. 
> 
> all of the above?

Yes it could be one of the three or a combination of ports depending on the intended use, with or without clustering, of mongodb on the machine.
>27017 | The default port for mongod and mongos instances. You can change this port with port or --port.
27018 | The default port for mongod when running with --shardsvr command-line option or the shardsvr value for the clusterRole setting in a configuration file.
27019 | The default port for mongod when running with --configsvr command-line option or the configsvr value for the clusterRole setting in a configuration file.

https://github.com/iiab/iiab/pull/2604#issuecomment-719524984

> I see. The env file does not match the readme, which is all I looked at. Since we only use mongodb for sugarizer, I'd also follow their usage, but I'd keep the soft coding as the link I gave above lists 27017, 18, 19 under various circumstances.

I could see the need to alter the port *sugarizer* tries to start the connection **_to_** [on a clustered mongodb install](https://github.com/llaske/sugarizer-server/issues/268) not the port that mongo gets bound with and _**listens on**_ as it appears that the port assignment might be automatic based on the above docs, however the mongodb role is not attempting to setup and use the advanced cluster features at all. I have not personally tried to setup up a _working_ cluster or replication with the above 'shardsvr' 'configsvr' options but the options appear to be ignored on the tar packaged version of mongodb or the port assignment is not auto adjusted as the above info suggests on the old 3.0.14 tar version that is installed on 32bit arm.
>pi@box:/opt/iiab/iiab $ mongod --version
db version v3.0.14
git version: 08352afcca24bfc145240a0fac9d28b978ab77f3

tim-moody:
> could see the need to alter the port sugarizer tries to start the connection to

I agree. In fact I stopped short of suggesting that the mongodb port is a given and it is only sugarizer that requires configuration to match the mongo setup.

There's nothing broken here, so I would wait for the next release.

llaske:
@holta Sugarizer Server can work with MongoDB on any port 27017, 27018 or any other port not used.
Historically Sugarizer Server was on port 27018 by default because... the port 27017 was already used on my development machine by another MongoDB instance 😕 
Because it could not be considered to be a good reason 😄 , we've decided to revert back to the "standard" 27017 port.
I think it's better to do the same for IIAB.

PR #2613, opened 2020-11-03
Change mongodb_port from 27018 to more standard 27017, following Sugarizer Server's lead

holta:
@jvonau gets the credit for doing the research.

Thanks @llaske and https://Sugarizer.org for bringing this all together over almost a decade now!

This PR is an alternative to PR #2611 to clean up PR #2604 for [IIAB 7.2](https://github.com/iiab/iiab/wiki/IIAB-7.2-Release-Notes)'s release.

Let's keep port flexibility.  Linux/FOSS folk cherish softcoding[*] and we never know where other folks are going to take our work.

[*] taking Liberation Theology to an excess...time and again softcoding too early in the game very often *creates* bugs by generalizing in counterproductive ways, before the software life cycle / clientele patterns are understood... e.g. often in ways the community & customers will never use!  But in this case the code is pretty well tested over the years, and sysadmins worldwide all have their quirky habits insisting on particular ports for weird in-house cultural reasons of one kind or another (!) so let's respect these oddball traditions, given we already have working code in place.


holta:
1) This PR was tested on 64-bit RaspiOS Lite by placing `mongodb_port: 27019` in [/etc/iiab/local_vars.yml](http://wiki.laptop.org/go/IIAB/FAQ#What_is_local_vars.yml_and_how_do_I_customize_it.3F) by running:

   ```
   sudo su -
   cd /opt/iiab/iiab
   time ./runrole --reinstall sugarizer    # 2min 51sec on RPi 4
   ```

   _Result: http://box/sugarizer worked and `/etc/mongod.conf` correctly contained 27019._

2) I also tested standard/default port 27017 by running the above with a stock `local_vars.yml` file (2min 54sec).

   _Result: http://box/sugarizer worked and `/etc/mongod.conf` correctly contained 27017._

<br>

Aside A: It would be really great if `./runrole --reinstall mongodb` worked to clear out `mongodb_installed: True` from `/etc/iiab/iiab_state.yml` but we can work on things like that another day.  Validation can be and perhaps should be restored at [/opt/iiab/iiab/roles/0-init/tasks/validate_vars.yml#L101](https://github.com/iiab/iiab/blob/master/roles/0-init/tasks/validate_vars.yml#L101) some day.  For now I do that 1-line removal manually, every time I test MongoDB, Sugarizer, etc.  Streamlining test community workflows is very important, but some other day!  (As of now: testers please use that manual step to force [/opt/iiab/iiab/roles/mongodb/tasks/install.yml](https://github.com/iiab/iiab/blob/master/roles/mongodb/tasks/install.yml) to run.)

Aside B: @llaske do you know why the password characters (animals to click on, instead of typing in letters) flash in color for a half-second when you login to Sugarizer?  Then after that fraction of a second, the animals revert to black+white.

roles/mongodb/tasks/install.yml
```diff
@@ -117,9 +117,9 @@
   - name: Change {{ mongodb_conf }} port to {{ mongodb_port }} -- takes effect on next (re)start of the service -- via enable-or-disable.yml or via sugarizer.service auto-starting MongoDB on demand
```
jvonau: The 'name' is no longer accurate, nothing is being 'changed' from the out of the box install, and without 'port' being defined at all the default port will be used.

holta: True (most of the time).

Ansible's idempotent claim (declarative language dream, just articulate your life dream and Ansible will take you there poof!) makes many hundreds of our Ansible `name: ` descriptions grammatically incorrect (or at least grammatically quite imperfect!)

I think part of this longstanding challenge here is that people (not just developers) fundamentally think in terms of action (let's get sheit done, i.e. let's proceed forward with procedural linguistics...)  Which is usually a much friendlier way of describing transformation...rather than a math geek/lawyer's declarative "it must be so" legalistics.

Still, pontificating has its place (i.e. declarative languages, poets, dreams) otherwise we'll be stuck in this pretty messy world forever...so a few Declarations of Independence ("electric cars MUST replace fossil fuel by 2030") are good things and occasionally nec (:

jvonau:
Why not just follow the lead of the other database roles that do not extend the flexibility of changing the working port as there is nothing really to be gained but complexity when it comes to the client apps that access the database.

holta:
Documentation https://github.com/iiab/iiab/wiki/IIAB-Networking#list-of-ports--services is now changed to...

| Protocol | Port | Service |
| - | - | - |
| TCP | 27017 (formerly 27018) | mongodb (if used by Sugarizer) |

jvonau:
> 
> Aside A: It would be really great if `./runrole --reinstall mongodb` worked to clear out `mongodb_installed: True` from `/etc/iiab/iiab_state.yml` but we can work on things like that another day. 

It does, you are misdiagnosing the problem. 

> Validation can be and perhaps should be restored at [/opt/iiab/iiab/roles/0-init/tasks/validate_vars.yml#L101](https://github.com/iiab/iiab/blob/master/roles/0-init/tasks/validate_vars.yml#L101) some day. For now I do that 1-line removal manually, every time I test MongoDB, Sugarizer, etc. Streamlining test community workflows is very important, but some other day! (As of now: testers please use that manual step to force [/opt/iiab/iiab/roles/mongodb/tasks/install.yml](https://github.com/iiab/iiab/blob/master/roles/mongodb/tasks/install.yml) to run.)
>

sudo ./runrole --reinstall mongodb
>TASK [Assert that "mongodb_install is sameas true" (boolean not string etc)] **************************
fatal: [127.0.0.1]: FAILED! => {"assertion": "mongodb_install is sameas true", "changed": false, "evaluated_to": false, "msg": "PLEASE SET 'mongodb_install: True' e.g. IN: /etc/iiab/local_vars.yml"}

It's because default_vars.yml has mongodb_install: False and the only way it is set to True is from within roles/sugarizer/tasks/main.yml before calling the mongodb role. Toggle default_vars.yml to True and drop the ' name: "Set 'mongodb_install: True'" ' stanza from roles/sugarizer/tasks/main.yml as the only way to get to mongodb's role is indirectly via sugarizer role or directly with runrole.

holta:
_Testers who want to force a "hard reinstall" of [Sugarizer](https://github.com/iiab/iiab/tree/master/roles/sugarizer) AND its dependencies ([Node.js](https://github.com/iiab/iiab/tree/master/roles/nodejs) and [MongoDB](https://github.com/iiab/iiab/tree/master/roles/mongodb)) should read this &mdash; as they have options:_

1. Let's not touch [/opt/iiab/iiab/vars/default_vars.yml](https://github.com/iiab/iiab/blob/master/vars/default_vars.yml) and instead edit [/etc/iiab/local_vars.yml](http://wiki.laptop.org/go/IIAB/FAQ#What_is_local_vars.yml_and_how_do_I_customize_it.3F) to set both of these:

   ```
   mongodb_install: True
   nodejs_install: True
   ```

   Now you can run these: _(no problem!)_

   ```
   ./runrole --reinstall mongodb      # 1min 21sec
   ./runrole --reinstall nodejs       # 1min 6sec
   ./runrole --reinstall sugarizer    # 2min 28sec
   ```

   Indeed we really should have read and actually listened to the earlier error message :-)

   ```
   PLEASE SET 'mongodb_install: True' e.g. IN: /etc/iiab/local_vars.yml
   ```

2. So the above routine's really great the 1st time (so testers build an understanding of dependencies, IIAB understructure, etc).

   But it's slow as molasses &mdash; and it's even worse to be interrupted repeatedly (instead of typing 1 command) after the layout's clear.  As every tester has higher-level tasks to dig into if they want to deliver on targetted-and-useful testing goals &mdash; rather than punching keys mindlessly & sitting around _(which rarely delivers useful testing coverage anyway!)_

   Hence manually removing the 2 dependencies' lines from `/etc/iiab/iiab_state.yml` and then running the following is probably best, as I did earlier: _(for more reasons than I earlier realized!)_

   ```
   ./runrole --reinstall sugarizer    # 2min 46sec
   ```

   **This is exactly as it should be, giving folks the flexibility to reinstall individual dependencies _(or not!)_**

3. Just FYI [./runroles](https://github.com/iiab/iiab/blob/master/runroles) offers no particular advantages over [./runrole](https://github.com/iiab/iiab/blob/master/runrole) in this case.  In fact it's (1) no faster, (2) you still need to manually remove all-3-not-just-2 lines from `/etc/iiab/iiab_state.yml`, and (3) you have more to type here too as well, to run the consolidated test:

   ```
   ./runroles mongodb nodejs sugarizer    # 2min 47sec
   ```

holta:
fyi this PR has been very heavily tested and will likely be merged soon.

jvonau:
The thing is the --reinstall functionally is as you have stated in the past not officially supported nor are upgrades, hence the way the code is. Even funnier is that you reverted the foundation #2500 for such progress and ignored it's replacement #2533 and companion in iiab/iiab-factory#134 that provides fully versioned upgrade capabilities and does the dirty work of 'reinstall' as required without user intervention. I see it as a moot point anyway why do I bother..

holta:
`./runrole --reinstall <SOME APP>` has really blossomed, demonstrably useful in so many different situations, and should now be supported (for CLI Kids anyway, and more of them keep showing up ;)

llaske:
> Aside B: @llaske do you know why the password characters (animals to click on, instead of typing in letters) flash in color for a half-second when you login to Sugarizer? Then after that fraction of a second, the animals revert to black+white.

It's related to the connection speed (time to load the font file). I've open an [issue](https://github.com/llaske/sugarizer/issues/897) to try to improve the visual effect.

PR #2546, opened 2020-09-25
Near-final Release Cleanup of validate_vars, default_vars, local_vars, Stages 0-4, SSHD, Moodle, MediaWiki 1.34.4

holta:
In preparation for [IIAB 7.2](https://github.com/iiab/iiab/wiki/IIAB-7.2-Release-Notes) final release:

- Inspired by @jvonau's PR #2540
- Trims [validate_vars.yml](https://github.com/iiab/iiab/blob/master/roles/0-init/tasks/validate_vars.yml) from 53 to 47 vars
- [SSHD role](https://github.com/iiab/iiab/tree/master/roles/sshd) standardized per IIAB norms
- Moodle fix to install PostgreSQL
- Emergency patch MediaWiki 1.34.3 -> [1.34.4](https://lists.wikimedia.org/pipermail/mediawiki-announce/2020-September/000262.html)
- Cleaner explanations in [default_vars.yml](https://github.com/iiab/iiab/blob/master/vars/default_vars.yml), local_vars.yml etc
- Tested on 32-bit RaspiOS Lite & 64-bit Ubuntu Desktop 20.04.1

Ref: PRs #2535 #2545

PR #2616, opened 2020-11-04
Use '*_install: True' with on demand roles for runrole(s)

jvonau:
remove set_facts as default_vars now has the *_install value set to True with the *_enabled part being ignored anyway with nodejs & yarn apt repos
### Fixes bug:
restores runrole --reinstall functionality for the on demand roles:  nodejs, yarn, postgresql, mongodb
toggle the default *_install and remove unnecessary set_facts from the roles. 
lightly tested with ./runrole

### DIFF
--- a/roles/internetarchive/tasks/install.yml
+++ b/roles/internetarchive/tasks/install.yml
@@ -1,10 +1,5 @@
 # 1. INSTALL 3 PREREQS
 
-- name: "Set 'nodejs_install: True' and 'nodejs_enabled: True'"
-  set_fact:
-    nodejs_install: True
-    nodejs_enabled: True
-
 - name: NODEJS - run 'nodejs' role (attempt to install & enable Node.js)
   include_role:
     name: nodejs
@@ -13,11 +8,6 @@
   fail:
     msg: "Internet Archive install cannot proceed, as it currently requires Node.js 10.x or 12.x or 14.x, and your nodejs_version is set to {{ nodejs_version }}.  Please check the value of nodejs_version in /opt/iiab/iiab/vars/default_vars.yml and possibly also /etc/iiab/local_vars.yml"
   when: (nodejs_version != "10.x") and (nodejs_version != "12.x") and (nodejs_version != "14.x")
-
-- name: "Set 'yarn_install: True' and 'yarn_enabled: True'"
-  set_fact:
-    yarn_install: True
-    yarn_enabled: True
 
 - name: YARN - run 'yarn' role (attempt to install & enable Yarn package manager)
   include_role:
--- a/roles/nodered/tasks/install.yml
+++ b/roles/nodered/tasks/install.yml
@@ -11,11 +11,6 @@
 # 2019-01-16: @jvonau's PR #1403 moved installation of Node.js (8.x for now) &
 # npm to roles/nodejs/tasks/main.yml
 
-- name: "Set 'nodejs_install: True' and 'nodejs_enabled: True'"
-  set_fact:
-    nodejs_install: True
-    nodejs_enabled: True
-
 # 2020-01-04 no longer triggered by roles/nodered/meta/main.yml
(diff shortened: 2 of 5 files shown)


jvonau:
Lockout of --reinstall introduced in #2546 
Glorifying comments of runrole in #2613

tim-moody:
I'm OK with a strategy that says just set install True prior to install rather than on the fly, but this should not be the default. Default should be False and min or medium or big that have things set to True that depend on these services should also set the services to be installed and enabled.

holta:
I have what looks to be a very clean solution.

I'll post in a bit.

jvonau:
> I'm OK with a strategy that says just set install True prior to install rather than on the fly, but this should not be the default. Default should be False and min or medium or big that have things set to True that depend on these services should also set the services to be installed and enabled.

That is a complete 180 turn in your thinking on default_vars.yml which is full of items that are set to 'True' for no other reason than to pass "validation" or because you feel the role is forms part of a 'standard' install.

tim-moody:
My rule of thumb is that anything used by all of min, medical, medium, and big are good candidates to default to True and other things not. Not would include nodejs, mongodb, postgresql. If there are things that are currently True but no one ever wants them, then I'd be happy to see them set to False. A number of things currently True were not set by me. I think kiwix is used universally, but kalite, osm vector maps, and a number of other things can be turned off if you like. I confess I probably wanted awstats as I think it is included in all versions.

What I don't like is doing an install with a previous local vars and getting a new result because someone decided to add something to default vars.

jvonau:
pi@box:/opt/iiab/iiab $ grep "install: True" vars/default_vars.yml
>\# installed with 'APP_XYZ_install: True' below, do NOT later change that.
iiab_admin_user_install: True    # If False, THE SETTING BELOW WILL BE IGNORED.
admin_console_install: True
js_menu_install: True
hostapd_install: True    # 2020-01-21: this var MIGHT be implemented in future.
dnsmasq_install: True
captiveportal_install: True
bluetooth_install: True
sshd_install: True    # Required by OpenVPN
openvpn_install: True
mysql_install: True
nginx_install: True
usb_lib_install: True
kalite_install: True
\# kiwix_install: True is REQUIRED, if you install IIAB's Admin Console
kiwix_install: True
osm_vector_maps_install: True
awstats_install: True
\# iiab_install: True

jvonau:
> A reminder that MIN-sized is functionally identical to default_vars.yml, but with low-level stuff removed to make its local_vars.yml friendlier, more customizable, more understandable.
> 
> (About 2 years ago MEDIUM-sized was functionally identical to default_vars.yml)

This is not my first go around suggesting all *optional* roles, as opposed to mandatory or dependency, should be "False" in default_vars and whatever the enduser wishes be set to "True" in local_vars because opt-in is less totalitarian than forcing an opt-out choice as local_vars is viewed authoritative. In an ideal world a blank local_vars file should result in stages 1-3 being completed with states 4-9 run through resulting in nothing installed from those stages, that is a 'base' install with min, medium, med, or big being layered over top and those optional roles applied.

jvonau:
> (About 2 years ago MEDIUM-sized was functionally identical to default_vars.yml)

Once upon a time local_vars was copied from default_vars then modified by the end-user, however default_vars was never updated to reflect the new reality of local_vars being split into the levels in use today.

holta:
While we don't have agreement on these issues, we had a very good discussion on IIAB's weekly call earlier today (http://minutes.iiab.io) as to the overall trickyness of handling dependencies.

e.g. we could take this in many directions including `/runrole --hardinstall sugarizer` to force the reinstall of all dependencies (child roles as well as parent roles) but many of these are probably not quite necessary, certainly in 2020 (:

Related:

- PR #2617 "runrole: auto-prompt to pass 'APP_install: True' to Ansible + properly validate 5 dependencies"

PR #2617, opened 2020-11-04
runrole: auto-prompt to pass 'APP_install: True' to Ansible + properly validate 5 dependencies

holta:
Alternative to @jvonau's PR #2616 with interactive explanations so testers know exactly what's happening.

This streamlines testing by cleaning up ad hoc installation of IIAB's 5 auto-installed dependencies {apache, nodejs, postgresql, mongodb, yarn}, allowing `./runrole <APP>` and `./runrole --reinstall <APP>` to work gracefully for most all of IIAB's Ansible [roles](https://github.com/iiab/iiab/tree/master/roles), in most all situations.

It does this by automatically prompting implementers/testers/you e.g. if you forgot to set APP_install: True in [/etc/iiab/local_vars.yml](http://wiki.laptop.org/go/IIAB/FAQ#What_is_local_vars.yml_and_how_do_I_customize_it.3F) &mdash; allowing you to quickly (A) pass variable `APP_install: True` directly to Ansible, by hitting [Enter] at the prompt &mdash; or (B) exit to the command-line to edit `/etc/iiab/local_vars.yml` if you prefer that instead.

Also var validation of IIAB's 5 core dependencies is fixed.  With blue-on-green Ansible output when [0-init](https://github.com/iiab/iiab/tree/master/roles/0-init) runs, making clear which validations the 5 dependencies are bypassing.

Background: PR https://github.com/iiab/iiab/pull/2613#issuecomment-721329979

This PR has been heavily tested on Ubuntu Server 20.04.1

PR #2622, opened 2020-11-06
runrole: explain unknown flag (--reinstall) with more context

holta:
Refined guidance for new users of `./runrole <APP>`

Displays to implicit valued of 'APP_enabled: True-or-False' as derived from [defaults_vars.yml](https://github.com/iiab/iiab/blob/master/vars/default_vars.yml) and [local_vars.yml](http://wiki.laptop.org/go/IIAB/FAQ#What_is_local_vars.yml_and_how_do_I_customize_it.3F) if Ansible if the app is already installed...and enforcing that (i.e. enable or disable the app/service) appears to be the operator's goal.

Ref: https://github.com/iiab/iiab/pull/2617#issuecomment-723104684


holta:
Tested on 64-bit RaspiOS Lite on RPi 4.

Issue #2626, opened 2020-11-08
Copying zims manually without running iiab-make-kiwix-lib will cause Admin Console installation to fail (and kiwix not to work)


holta:
@tim-moody did this Admin Console fresh install fail possibly because I pre-positioned a few ZIM files into `/library/zims/content` ?

(Is pre-positioning ZIM files not allowed?)

FYI I'm trying to install IIAB (http://download.iiab.io) onto 32-bit 'RaspiOS with desktop' on RPi 4 with 4GB RAM:

> TASK [js-menu : Add any IIAB roles that are installed to home/menu.js] *********
[...]

PR #2627, opened 2020-11-08
Kiwix [role improvements for pre-positioning of ZIM files into /library/zims/content, etc]

jvonau:
### Fixes bug:
#2626
Call iiab-make-kiwix-lib in enable-or-disable

### DIFF
--- a/roles/kiwix/tasks/enable-or-disable.yml
+++ b/roles/kiwix/tasks/enable-or-disable.yml
@@ -1,18 +1,20 @@
-- name: Enable & Restart 'kiwix-serve' systemd service
-  systemd:
-    name: kiwix-serve
-    daemon_reload: yes
-    enabled: yes
-    state: restarted
-  when: kiwix_enabled
-
 - name: Disable & Stop 'kiwix-serve' systemd service
   systemd:
     name: kiwix-serve
     enabled: no
     state: stopped
-  when: not kiwix_enabled
 
+- name: Update '{{ kiwix_library_xml }}'
+    command: /usr/bin/iiab-make-kiwix-lib
+  when: kiwix_enabled
+# the above starts the service
+
+- name: Enable & Ensure 'kiwix-serve' systemd service is started
+  systemd:
+    name: kiwix-serve
+    enabled: yes
+    state: started
+  when: kiwix_enabled
 
 # TO DO: BOTH CRON ENTRIES BELOW *SHOULD* BE DELETED "when: not kiwix_enabled"
 
--- a/roles/kiwix/tasks/enable-or-disable.yml
+++ b/roles/kiwix/tasks/enable-or-disable.yml
@@ -4,8 +4,8 @@
     enabled: no
     state: stopped
 
-- name: Update '{{ kiwix_library_xml }}'
-    command: /usr/bin/iiab-make-kiwix-lib
+- name: Update {{ kiwix_library_xml }}
+  command: /usr/bin/iiab-make-kiwix-lib
   when: kiwix_enabled
 # the above starts the service
 
--- a/roles/kiwix/tasks/install.yml
+++ b/roles/kiwix/tasks/install.yml
@@ -51,7 +51,7 @@
 
 - name: Check for {{ iiab_zim_path }}/content/*
   stat:
-    path: "{{ iiab_zim_path }}/content/*"
+    path: "{{ iiab_zim_path }}/content/*.zim"
   register: kiwix_zims
 
 - name: Install {{ iiab_zim_path }}/content/test.zim
--- a/roles/kiwix/tasks/install.yml
+++ b/roles/kiwix/tasks/install.yml
@@ -54,13 +54,6 @@
     path: "{{ kiwix_library_xml }}"    # /library/zims/library.xml
   register: kiwix_xml
 
-- name: Install stub /library/zims/library.xml from template, if one doesn't exist
-  template:
-    src: library.xml
-    dest: "{{ kiwix_library_xml }}"
-    force: no
-  when: not kiwix_xml.stat.exists
-
 - name: Install {{ iiab_zim_path }}/content/test.zim
   copy:
     src: test.zim
--- a/roles/kiwix/tasks/install.yml
+++ b/roles/kiwix/tasks/install.yml
@@ -66,6 +66,7 @@
     src: test.zim
     dest: "{{ iiab_zim_path }}/content/test.zim"
     force: no
+  when: not kiwix_xml.stat.exists
 
 
 # 2. INSTALL KIWIX-TOOLS EXECUTABLES
--- a/roles/kiwix/tasks/install.yml
+++ b/roles/kiwix/tasks/install.yml
@@ -49,7 +49,7 @@
     - "{{ iiab_zim_path }}/content"
     - "{{ iiab_zim_path }}/index"
 
-- name: Check for {{ iiab_zim_path }}/content/*
+- name: Check for {{ iiab_zim_path }}/content/*.zim
   stat:
     path: "{{ iiab_zim_path }}/content/*.zim"
   register: kiwix_zims
--- a/roles/kiwix/tasks/install.yml
+++ b/roles/kiwix/tasks/install.yml
@@ -49,17 +49,17 @@
     - "{{ iiab_zim_path }}/content"
     - "{{ iiab_zim_path }}/index"
 
-- name: Check for {{ kiwix_library_xml }}
+- name: Check for {{ iiab_zim_path }}/content/*
   stat:
-    path: "{{ kiwix_library_xml }}"    # /library/zims/library.xml
-  register: kiwix_xml
+    path: "{{ iiab_zim_path }}/content/*"
+  register: kiwix_zims
 
 - name: Install {{ iiab_zim_path }}/content/test.zim
   copy:
     src: test.zim
     dest: "{{ iiab_zim_path }}/content/test.zim"
     force: no
-  when: not kiwix_xml.stat.exists
+  when: not kiwix_zims.stat.exists
 
 
 # 2. INSTALL KIWIX-TOOLS EXECUTABLES
--- a/roles/kiwix/tasks/install.yml
+++ b/roles/kiwix/tasks/install.yml
@@ -44,6 +44,7 @@
     path: "{{ item }}"
     state: directory
   with_items:
+    - "{{ kiwix_path }}/bin"    # /opt/iiab/kiwix
     - "{{ iiab_zim_path }}"    # /library/zims
     - "{{ iiab_zim_path }}/content"
     - "{{ iiab_zim_path }}/index"
@@ -65,11 +66,6 @@
     src: test.zim
     dest: "{{ iiab_zim_path }}/content/test.zim"
     force: no
-
-- name: Create {{ kiwix_path }}/bin directory
-  file:
-    path: "{{ kiwix_path }}/bin"    # /opt/iiab/kiwix
-    state: directory
 
 
 # 2. INSTALL KIWIX-TOOLS EXECUTABLES
--- a/roles/kiwix/tasks/install.yml
+++ b/roles/kiwix/tasks/install.yml
@@ -107,6 +107,10 @@
     - { src: 'iiab-make-kiwix-lib', dest: '/usr/bin/iiab-make-kiwix-lib', mode: '0755' }
     - { src: 'iiab-make-kiwix-lib3.py', dest: '/usr/bin/iiab-make-kiwix-lib.py', mode: '0755' }
 
+- name: Reload systemd for new kiwix-serve.service file
+  systemd:
+    daemon_reload: yes
+
 - name: "Install from template: Apache's kiwix.conf"
   template:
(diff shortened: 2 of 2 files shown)


tim-moody:
This will not fix #2626 

In all likelihood any manual installation of zims will occur after iiab installation. This is a documentation problem (actually it probably is already in write-only-memory). If you manually copy zims you must run iiab-make-kiwix-lib in order for kiwix and other things to work.

tim-moody:
also in order for adm cons to work there needs to be at least one zim and a corresponding library.xml. so if the library/zims/content exists and is empty test.zim will not get installed, so if you want to condition the copy it needs to be on the presence of a zim in the dir

jvonau:
The provided library.xml file contains only one entry for test.zim, the two are a matched set, and both would be installed when library.yml does not exist keeping admin-cosole happy. Granted running iiab-make-kiwix-lib when the default library.xml is deployed doesn't gain anything but one doesn't need library.xml present first to run iab-make-kiwix-lib, so perhaps the supplying of the library.xml can go away as well.
>pi@box:/opt/iiab/iiab $ sudo mv /library/zims/library.xml /library/zims/library.xml.old 
pi@box:/opt/iiab/iiab $ sudo iiab-make-kiwix-lib
Waiting to run iiab-make-kiwix-lib.py
Now running iiab-make-kiwix-lib.py
cp: cannot stat '/library/zims/library.xml': No such file or directory
Writing zim_versions_idx
rm: cannot remove '/library/zims/library.xml': No such file or directory
Finished making Kiwix library.xml
pi@box:/opt/iiab/iiab $ ls /library/zims/
content  index  library.xml  library.xml.old
pi@box:/opt/iiab/iiab $

holta:
Related:

PR iiab/iiab-admin-console#363 "run iiab-make-kiwix-lib in case zims have been added manually"

jvonau:
@tim-moody  Please explain why calling `iiab-make-kiwix-lib` in iiab-admin-console's install is fine while calling iiab-make-kiwix-lib in the kiwix's role in iiab is not.

tim-moody:
I now agree with this as I did not realize the case was zims installed prior to iiab install, but with the following provisions:
1. there must always be at least 1 zim file, test.zim or other
1. library.xml must contain all zims
2. /usr/bin/iiab-make-kiwix-lib I think requires kiwix-manage so watch the order of execution. 
3. It also requires iiab_lib.py.

tim-moody:
please note that iiab-make-kiwix-lib will still need to be called in adm console as it does work that can not yet be done, which is why I originally did not think it should be done here. but I now agree as I said above.

holta:
> 2. library.xml must contain all zims

My opinion's a bit different.

While it would be lovely if `iiab-make-kiwix-lib` has already been run, ZIM files are getting more popular in non-IIAB contexts and occasionally such content (new ZIM files) will be dropped into `/library/zims/content` prior to IIAB software installation &mdash; when it's impossible to run a program (`iiab-make-kiwix-lib`) that has not been installed yet.

Something within IIAB &/or Admin Console's installation process (or both if that makes us resilient?!) should really allow for non-techie content people to gracefully recover.

In my opinion!

holta:
See Also:

PR iiab/iiab-admin-console#364 "Give error message and exit if iiab-make-kiwix-lib needs to run"

jvonau:
> > 1. library.xml must contain all zims
> 
> My opinion's a bit different.
> 
> While it would be lovely if `iiab-make-kiwix-lib` has already been run, ZIM files are getting more popular in non-IIAB contexts and occasionally such content (new ZIM files) will be dropped into `/library/zims/content` prior to IIAB software installation — when it's impossible to run a program (`iiab-make-kiwix-lib`) that has not been installed yet.

iiab-make-kiwix-lib & companion iiab-make-kiwix-lib.py are installed in this role, available to be run.

> I now agree with this as I did not realize the case was zims installed prior to iiab install, but with the following provisions:
> 
>     1. there must always be at least 1 zim file, test.zim or other

Check, test.zim provided when library.xml does not exist 
 
>     2. library.xml must contain all zims

Check, running iiab-make-kiwix-lib populates library.xml
>     3. /usr/bin/iiab-make-kiwix-lib I think requires kiwix-manage so watch the order of execution.

pi@box:/library/zims $ sudo /opt/iiab/kiwix/bin/kiwix-manage --help
>/opt/iiab/kiwix/bin/kiwix-manage: unrecognized option '--help'
Usage:
	kiwix-manage LIBRARY_PATH add ZIM_PATH [OPTIONS]
	kiwix-manage LIBRARY_PATH remove ZIM_ID [ZIM_ID]...
	kiwix-manage LIBRARY_PATH show [ZIM_ID]...
Purpose:
	Manipulates the Kiwix library XML file
Arguments:
	LIBRARY_PATH	is the XML library file path.
	ACTION		is the pre-defined string to specify the action to run on the XML library file.
			Must be one of the following values:
			* add: add a ZIM file to the library
			* remove: remove a ZIM file from the library
			* show: show the content of the library
	ZIM_ID		ZIM file unique ID
	OPTIONS		Custom options for "add" action:
			--zimPathToSave=CUSTOM_ZIM_PATH to replace the current ZIM file path
			--url=HTTP_ZIM_URL to create an "url" attribute for the online version of the ZIM file
			Other options:
			-v, --version to print the software version
Examples:
	Add ZIM files to library: kiwix-manage my_library.xml add first.zim second.zim
	Remove ZIM files from library: kiwix-manage my_library.xml remove e5c2c003-b49e-2756-5176-5d9c86393dd9
	Show all library ZIM files: kiwix-manage my_library.xml show
Documentation:
	Source code	https://github.com/kiwix/kiwix-tools
	More info	https://wiki.kiwix.org/wiki/Kiwix-manage

Looks iiab-make-kiwix-lib is designed to replace kiwix-manage, bouncing the service after replacing the library.xml file. The path="content/SOMEZIM.zim" in the library.xml file is the value to note.
 
>     4. It also requires iiab_lib.py.

Check, installed in stage 4

tim-moody:
just wanted to make sure kiwix was installed before running make kiwix lib, but that is in main and this is enable, so I agree is fine. also wasn't sure where pylib was in the install in relation to kiwix, again seems fine.

>Looks iiab-make-kiwix-lib is designed to replace kiwix-manage, bouncing the service after replacing the library.xml file.

not sure I understand this, but doesn't matter.

roles/kiwix/tasks/enable-or-disable.yml
```diff
@@ -1,18 +1,20 @@
...
   systemd:
     name: kiwix-serve
     enabled: no
     state: stopped
-  when: not kiwix_enabled
 
+- name: Update '{{ kiwix_library_xml }}'
+    command: /usr/bin/iiab-make-kiwix-lib
+  when: kiwix_enabled
+# the above starts the service
```
holta, line 10: The above starts which service?

The same `kiwix-serve.service` being started just below, or a different one?

jvonau: What role is this found in? kiwix_enabled implies the use of kiwix-serve.service, You are being pretty petty for a note that could just be removed,

jvonau:
> > Looks iiab-make-kiwix-lib is designed to replace kiwix-manage, bouncing the service after replacing the library.xml file.
> 
> not sure I understand this, but doesn't matter.

Does this sound better?
In place of passing kiwix-manage each new or absent zimfile name iiab-make-kiwix-lib (re)creates library.xml based on the zims present in the content directory then restarts the service to pickup the revised library.xml file.

tim-moody:
>Does this sound better?
In place of passing kiwix-manage each new or absent zimfile name iiab-make-kiwix-lib (re)creates library.xml based on the zims present in the content directory then restarts the service to pickup the revised library.xml file.

This I understand, but it is not true. It was rewritten some time ago to handle deltas, both additions and deletions (after asking kiwix to fix their problem with deletions.)

roles/kiwix/tasks/enable-or-disable.yml
holta: I was asking a sincere question.

I guess it's harmless that it (kiwix-serve.service) is "started" above and below, with these 2 stanzas in a row.

jvonau:
Perhaps triggering the installation of test.zim could be the absence of any other zims being present in /library/zims/content/ in place of using the presence of library.xml, that would leave less user cleanup post install as the test.zim would be the first zim to be removed anyway.

tim-moody:
I would prefer that

holta:
Is the Test ZIM menu-def added across the board, to all fresh installs of IIAB / Admin Console at the moment?

(Or only when something like `/library/zims/content/test.zim` is detected?)

jvonau:
PR in final form tested with runrole --reinstall with/without content/test.zim in place, supplies test.zim in absence of any zims as expected, library.xml regenerated.

roles/kiwix/tasks/install.yml
```diff
@@ -44,32 +44,22 @@
     path: "{{ item }}"
     state: directory
   with_items:
+    - "{{ kiwix_path }}/bin"    # /opt/iiab/kiwix
     - "{{ iiab_zim_path }}"    # /library/zims
     - "{{ iiab_zim_path }}/content"
     - "{{ iiab_zim_path }}/index"
 
-- name: Check for {{ kiwix_library_xml }}
+- name: Check for {{ iiab_zim_path }}/content/*
```
holta: Line 52 and 54 should probably test for *.zim suffix to be a bit safer?

```suggestion
- name: Check for {{ iiab_zim_path }}/content/*.zim
```

roles/kiwix/tasks/install.yml
```diff
@@ -44,32 +44,22 @@
...
-    path: "{{ kiwix_library_xml }}"    # /library/zims/library.xml
-  register: kiwix_xml
-
-- name: Install stub /library/zims/library.xml from template, if one doesn't exist
-  template:
-    src: library.xml
-    dest: "{{ kiwix_library_xml }}"
-    force: no
-  when: not kiwix_xml.stat.exists
+    path: "{{ iiab_zim_path }}/content/*"
```
holta:
```suggestion
    path: "{{ iiab_zim_path }}/content/*.zim"
```

tim-moody:
>Is the Test ZIM menu-def added across the board

all existing menu defs are always installed, basically a clone of https://github.com/iiab-share/js-menu-files

roles/kiwix/tasks/install.yml
jvonau: done

roles/kiwix/tasks/install.yml
jvonau: done

holta:
> > Is the Test ZIM menu-def added across the board
> 
> all existing menu defs are always installed, basically a clone of https://github.com/iiab-share/js-menu-files

Does Test ZIM currently appear on IIAB main page http://box.lan after installing Admin Console -- even if `/library/zims/content/test.zim` doesn't exist?

tim-moody:
>Does Test ZIM currently appear on IIAB main page http://box.lan after installing Admin Console -- even if /library/zims/content/test.zim doesn't exist?

currently in the default menu.json. iiab-update=menus is supposed to add any new zims but not remove old ones. we may have made an exception for adding test.zim, otherwise it would add same as any other if present.

holta:
> currently in the default menu.json

So I guess that will lead to a broken link within that Test Zim widget on IIAB's main page after this PR is merged, in the case of fresh installs with pre-positioned ZIM files placed in `/library/zims/content` and hence lacking `/library/zims/content/test.zim` ?

Maybe not a serious issue, but still if this ugliness can be avoided we prob should.

holta:
@jvonau prefers this be labelled as a bug fix rather than an enhancement, so I changed it.

jvonau:
> @jvonau prefers this be labelled as a bug fix rather than an enhancement, so I changed it.

That is not what I asked in https://github.com/iiab/iiab/pull/2629#issuecomment-723978416, I want an explanation as to how this PR was an "enhancement" while the replacement is a "bugfix", with the same commits no less, and what criteria is used to make the distinction between the two labels because inquiring minds want to know.

holta:
Some companies label every single ticket a 'defect' to avoid exactly such never-ending philosophical debates...particularly as there's a pattern in some companies/orgs where developers will spend all day arguing over what's a bug and what isn't...so they don't have to do any real work ;)

As such it's not as terrible of an idea as it sounds...to label every single ticket a bug...as this gives the clientele / user base / community's frame of reference more of a say in the matter...without too many rhetorical soap operas around the original intent of the framers of the Constitution etc.

jvonau:
Nice useless response, I asked a direct question "what criteria is used to make the distinction between the two labels." Think the "bug" label is being misused, should be used to triage incoming *issues* while [bugfix|feature|enhancement], in uppercase if you want that more visible, in the PR's title line would be more useful.

holta:
> uppercase...in the PR's title line

I agree this can help.

The downside risk being that every label can be argued to the moon-and-back.

Whether "Legislating The Truth" actually leads us all forward-or-backwards is a bit less clear.

Certainly legalistic poets DO mean well.

Ref: #2630 "How should issues and PRs be labelled?"

Issue #2610, opened 2020-11-01
Who can help test on Ubuntu 20.10+ (Server &/or Desktop version!) on x86_64 / amd64, VM or PC?


holta:
Somewhat related:

- #1737 "Who can help test IIAB 7.2 on Ubuntu 20.04+, Debian 10.6+, Linux Mint 20 & Raspberry Pi ? [Kolibri 502 Bad Gateway?]"
- PR #2582 "Overseas request for immediate testing of IIAB on Debian 11 Bullseye + similarly clean out long-deprecated apache_config_dir from high-level <OS>.yml vars files [ASIDE: dnsmasq & dhclient issues? host_country_code MUST match PC/HW, or hostapd won't start]"
- #2585 "Testing the new Ubuntu Desktop 20.10 for RPi 4 (rumor confirmed!) [Bluetooth, MongoDB, NetworkManager]"
- PR iiab/iiab-factory#158 "1-line-installer: facilitate testing w/ instant install of CLI-specified PR's e.g. 'sudo iiab 361 2607'"

shanti-bhardwa:
installed 20.10 Ubuntu Groovy Gorilla desktop on PI 4 with 4GB of RAM. Desktop works fine.  This version requires installation of curl first before running the 1 line install script. for testing  IIAB 7.2.
Doing a small build gives this error first:

W: GPG error: http://ppa.launchpad.net/ansible/ansible/ubuntu bionic InRelease: The following signatures couldn't be verified because the public key is not available: NO_PUBKEY 93C4A3FD7BB9C367
E: The repository 'http://ppa.launchpad.net/ansible/ansible/ubuntu bionic InRelease' is not signed.

have to run

sudo apt-key adv --keyserver hkp://keyserver.ubuntu.com:80 --recv-keys 93C4A3FD7BB9C367"

then sudo iiab again to restart.

Install reboots the desktop and doing sudo iiab completes the install

shanti-bhardwa:
http://sprunge.us/rYBwZr

holta:
@shanti-bhardwa thanks!  Noting your report is really more appropriate for #2585 !  (This ticket [#2610] is about PC and VM uses of Ubuntu 20.10)

But since we're here (:

> have to run
> 
> sudo apt-key adv --keyserver hkp://keyserver.ubuntu.com:80 --recv-keys 93C4A3FD7BB9C367"

[...]

PR #2649, opened 2020-11-27
gitea/tasks/install.yml: timeout: 10 -> 200 (for get_url from dl.gitea.org)

holta:
Thanks @deldesir for surfacing this longstanding issue: https://github.com/iiab/iiab/issues/2610#issuecomment-734315053

When IIAB is installed in developing world countries, it's especially important that Ansible's download timeout default (10 seconds) be raised to a much higher number.

This PR brings that IIAB norm (200 seconds) to Gitea, in addition to the 23 other places where `timeout: "{{ download_timeout }}"` is already in use across IIAB code, thanks to https://github.com/iiab/iiab/blob/master/vars/default_vars.yml#L43-L44 &mdash; which is currently:

```
# Ansible's default timeout for "get_url:" downloads (10 seconds) often fails
download_timeout: 200
```

CC: @aidan-fitz

### DIFF
--- a/roles/gitea/tasks/install.yml
+++ b/roles/gitea/tasks/install.yml
@@ -48,6 +48,7 @@
     url: "{{ gitea_download_url }}"
     dest: "{{ gitea_install_path }}"
     mode: '0775'
+    timeout: "{{ download_timeout }}"
   when: internet_available
 
 - name: Download Gitea GPG signature


holta:
@deldesir were you aware that you could change `download_timeout: 200` at the top of every [/etc/iiab/local_vars.yml](http://wiki.laptop.org/go/IIAB/FAQ#What_is_local_vars.yml_and_how_do_I_customize_it.3F) ?

In any case, thanks for keeping an eye on these things!

As others have been unable to reproduce the download failures experienced by people using particular ISP's in Haiti and India :/

Issue #2650, opened 2020-11-27
How many of these 10 sysctl {kernel, IPv4, IPv6} tweaks are still needed ~5 years later?


holta:
PRs like #2647 (@jvonau) and #2648 weren't quite enough, as the [sysctl](https://docs.ansible.com/ansible/latest/collections/ansible/posix/sysctl_module.html) Ansible module is also suddenly moved into of Ansible's new [ansible.posix](https://galaxy.ansible.com/ansible/posix) collection.

(Same story as with Ansible module [selinux](https://docs.ansible.com/ansible/latest/collections/ansible/posix/selinux_module.html), that we just faced in the above PR's.)

As a test I've commented out the 10 sysctl commands at https://github.com/iiab/iiab/blob/master/roles/2-common/tasks/main.yml#L24-L38 which allowed BIG-sized IIAB install to complete on Debian 11 VM 10.8.0.26 = deb11-srv-BIG-1127 over the past hour.  Now that IIAB installed successfully onto this VM, can we use this to investigate the QUESTION in this ticket's subject line: _how many of these 10 `systctl` tweaks are truly still needed and necessary a half-decade later?_

```
- name: Use 'sysctl' to set 10 network/kernel settings, turning off IPv6 if possible
  sysctl:
    name: "{{ item.name }}"
    value: "{{ item.value }}"
  with_items:
    - { name: 'net.ipv4.ip_forward', value: '1' }
    - { name: 'net.ipv4.conf.default.rp_filter', value: '1' }
    - { name: 'net.ipv4.conf.default.accept_source_route', value: '0' }
    - { name: 'kernel.sysrq', value: '1' }
    - { name: 'kernel.core_uses_pid', value: '1' }
    - { name: 'net.ipv4.tcp_syncookies', value: '1' }
    - { name: 'kernel.shmmax', value: '268435456' }
    - { name: 'net.ipv6.conf.all.disable_ipv6', value: '1' }    # IPv6 disabled
    - { name: 'net.ipv6.conf.default.disable_ipv6', value: '1' }
    - { name: 'net.ipv6.conf.lo.disable_ipv6', value: '1' }
```

holta:
@jvonau @georgejhunt @tim-moody feel free to log into 10.8.0.26 = deb11-srv-BIG-1127 if it helps us all answer this question?

jvonau:
Not in front of a vpn machine replying from my phone, I'll login later today, but
[...]

PR #2651, opened 2020-11-27
Emergency/Interim Patch + Experiment to help surface which of the 10 sysctl tweaks are still needed + necessary

holta:
For #2650

### DIFF
--- a/roles/2-common/tasks/main.yml
+++ b/roles/2-common/tasks/main.yml
@@ -21,21 +21,21 @@
 - include_tasks: packages.yml
 - include_tasks: iptables.yml
 
-- name: Use 'sysctl' to set 10 network/kernel settings, turning off IPv6 if possible
-  sysctl:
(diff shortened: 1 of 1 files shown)


holta:
Tested on Debian 11 pre-release 10.8.0.26 -- as an interim patch, this PR allows IIAB installation to complete, given the recent changes to Ansible's Collections (moving Ansible modules `selinux` and `sysctl` into ansible.posix).

Until #2650 is answered more methodically.

holta:
While I'm not thrilled with this PR, it seems a decent way to start surfacing the real issues in coming days &mdash; as to which of these 10 sysctl commands are truly necessary vs. legacy cruft ariding from OS's no longer in use etc.

Possibly the 3 IPv6 tweaks are more urgent than the other 7?

Let's all try to figure this out in coming days...i.e. which of these 10 are truly priorities...and which are long-gone relics from another era/decade.

PR #2647, opened 2020-11-26
For discussion where should "Collections" reside? [Ansible]

jvonau:
idea..

PR #2518, opened 2020-09-15
Begin transition to Ansible 2.10 (now that ansible-base 2.10.1 is released & appears solid!)

holta:
Building on and evolving from:

- #2479 'Validate Ansible 2.10 expected 2020-08-13 [with simplified auto-installation of "Collections" ETA 2020-09-22 ?]'
- #2481 'Ansible promised not to change things, but this seems like a big one: file permissions like 666 become 600, in new Ansible 2.10.0 (and in Ansible 2.9.12 too?)'
- PR #2512 'Revamp ansible for the future'

PR #2652, opened 2020-11-28
ansible.posix collection for 2-common sysctl & js-menu synchronize (formerly needed by 1-prep's selinux)

holta:
Currently IIAB and Admin Console do not install properly.

This is a 1st stab cleaner resolution of issues arising from the sudden upstream changes in Ansible Collections &mdash; including things like PR #2647, PR #2648, #2650, PR #2651 &mdash; and also the install of Admin Console's js-menu which is currently (silently) failing at: https://github.com/iiab/iiab-admin-console/blob/master/roles/js-menu/tasks/main.yml#L77-L80

This is sure to evolve over time, but is tested on RaspiOS Lite & Ubuntu Server 20.04.1 to get us back on track.

Ref: PR #2518 "Begin transition to Ansible 2.10 (now that ansible-base 2.10.1 is released & appears solid!)"

### DIFF
--- a/roles/2-common/tasks/main.yml
+++ b/roles/2-common/tasks/main.yml
@@ -27,7 +27,7 @@
     name: "{{ item.name }}"
     value: "{{ item.value }}"
   with_items:
-    - { name: 'net.ipv4.ip_forward', value: '1' }
+    - { name: 'net.ipv4.ip_forward', value: '1' }  # Masquerading LAN->Internet
     - { name: 'net.ipv4.conf.default.rp_filter', value: '1' }
     - { name: 'net.ipv4.conf.default.accept_source_route', value: '0' }
     #- { name: 'kernel.sysrq', value: '1' }             # OS values differ, Ok?
--- a/roles/2-common/tasks/main.yml
+++ b/roles/2-common/tasks/main.yml
@@ -30,9 +30,9 @@
     - { name: 'net.ipv4.ip_forward', value: '1' }
     - { name: 'net.ipv4.conf.default.rp_filter', value: '1' }
     - { name: 'net.ipv4.conf.default.accept_source_route', value: '0' }
-    #- { name: 'kernel.sysrq', value: '1' }    # Already set by most Linux OS's
+    #- { name: 'kernel.sysrq', value: '1' }             # OS values differ, Ok?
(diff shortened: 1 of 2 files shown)


holta:
@jvonau & All: do you know if anybody's really using `kernel.core_uses_pid` ?

i.e. if nobody's using it, we should revert it from (1) to the default (0).

holta:
Each OS has its own default value for `kernel.sysrq` (Linux's magic SysRq key, to try to recover from system lockups etc).

This PR intentionally respects those OS defaults.

_But:_ if there are good reasons _not_ to respect OS defaults...please do explain/discuss a bit...i.e. who needs this most, for what situations _(e.g. how has this been used over the past decade?)_

holta:
Some quick lessons learned:

> Do we know yet whether it affects all OS's ?

Re: https://github.com/iiab/iiab/pull/2647#issuecomment-734628627

1) I've confirmed these mysteriously sudden upstream changes (several Ansible modules suddenly no longer available, and needing Ansible Collection [ansible.posix](https://galaxy.ansible.com/ansible/posix)) seem to affect all Linux OS's...having sampled 4 anyway, among the several [OS's IIAB supports/tolerates](http://wiki.laptop.org/go/IIAB/FAQ#What_OS_should_I_use.3F).

   Those who created [POSIX](https://en.wikipedia.org/wiki/POSIX) ~30 years ago to help standardize UNIX/Linux are likely crying in their sleep, but at least it finally/mostly works!

2) Curiously Ansible still doesn't have the ability to cleanly upgrade all collections, a year later...despite all of IBM / Red Hat's resources to try to elegantly modularize Ansible...they're still not quite there yet:
   - https://github.com/ansible/ansible/issues/65699
   - https://github.com/ansible/ansible/pull/71850

3) Until Ansible cleans this up, things like the following can help -- e.g. to force the upgrade of Ansible Collection 'community.general' from 1.2.0 to 1.3.0 -- and its dependency collections (that it's pulling in) too:

   ```
   ansible-galaxy collection install community.general --force-with-deps
   ```

   (Though I doubt we want to use the `--force-with-deps` flag every time, on initial install, etc ?!)

4) @jvonau (and ansible/ansible#65699 above) also mention that IIAB could use a `requirements.yml` files to help manage dependencies across Ansible Collections, if we really need to go down that road later...

Issue #2669, opened 2021-01-28
IIAB 7.2 installation stuck on "Add 'runtime' variable values" Rpi 4 4GB Raspberry Pi OS 32-bit


chjohn22:
### Expected behavior and actual behavior:
After running the one-liner install, iiab should install all components. Instead, it does abort the installation after some minutes. Details below.
### Steps to reproduce the problem:
execute the one-liner, do not use custom .yaml, press enter to aknowledge that you can continue with `sudo iiab`, wait for the install to stop

### IIAB version, OS version, hardware details:
7.2, Raspberry Pi OS 32-bit with desktop and recommended software, Raspberry Pi 4 4GB, 256GB microSD
### Please run the `iiab-diagnostics` command, to generate a URL summarizing your IIAB configuration, for volunteers seeking to help you: (Include that short URL here, Thanks!)

Uploading to web pastebin does not work: `PUBLISHING TO URL... /usr/bin/iiab-diagnostics: Zeile 242: pastebinit: Kommando nicht gefunden`

The installation does run after executing: ` curl d.iiab.io/install.txt | sudo bash` but does not continue after the task named
`TASK [0-init : Add 'runtime' variable values to /etc/iiab/iiab.ini] ************`

Log:

`TASK [0-init : Add 'runtime' variable values to /etc/iiab/iiab.ini] ************
ok: [127.0.0.1] => (item={'option': 'iiab_stage', 'value': '0'})
ok: [127.0.0.1] => (item={'option': 'iiab_base_ver', 'value': 7.2})
ok: [127.0.0.1] => (item={'option': 'iiab_revision', 'value': 0})
ok: [127.0.0.1] => (item={'option': 'runtime_branch', 'value': 'master'})
ok: [127.0.0.1] => (item={'option': 'runtime_commit', 'value': 'f46d469eb7055176b94d8149d1798a28f40d78b2'})
changed: [127.0.0.1] => (item={'option': 'runtime_date', 'value': '2021-01-28T21:53:53Z'})
ok: [127.0.0.1] => (item={'option': 'ansible_version', 'value': '2.10.5'})
ok: [127.0.0.1] => (item={'option': 'kernel', 'value': '5.4.83-v7l+'})
ok: [127.0.0.1] => (item={'option': 'memory_mb', 'value': 3827})
ok: [127.0.0.1] => (item={'option': 'swap_mb', 'value': 99})
[...]

PR #2670, opened 2021-01-28
Comment out non-existent var {{ ansible_product_uuid }} [not quite enough: {{ prior_gateway_device }} has a similar issue]

holta:
Proposed emergency fix for #2669

### DIFF
--- a/roles/0-init/tasks/main.yml
+++ b/roles/0-init/tasks/main.yml
@@ -137,7 +137,7 @@
       value: "{{ ansible_memtotal_mb }}"
     - option: swap_mb
       value: "{{ ansible_swaptotal_mb }}"
-    # 2021-01-28: Non-existent var, so fails with Ansible 2.10.5 (#2669)
+    # 2021-01-28: Non-existent var, so fails with ansible-base 2.10.5 (#2669)
     #- option: product_id
     #  value: "{{ ansible_product_uuid }}"
     - option: gw_active
--- a/roles/0-init/tasks/main.yml
+++ b/roles/0-init/tasks/main.yml
@@ -137,7 +137,7 @@
       value: "{{ ansible_memtotal_mb }}"
     - option: swap_mb
       value: "{{ ansible_swaptotal_mb }}"
-    # 2021-01-28: Non-existent, so fails with Ansible 2.10.5 (#2669)
+    # 2021-01-28: Non-existent var, so fails with Ansible 2.10.5 (#2669)
     #- option: product_id
     #  value: "{{ ansible_product_uuid }}"
     - option: gw_active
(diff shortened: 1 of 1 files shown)


holta:
This PR wasn't enough, as seen during this fresh IIAB install attempt on 193-deb10-BIG-0128 = 10.8.0.46:

> TASK [network : Add 'detected_network' variable values to /etc/iiab/iiab.ini] ***
...
> failed: [127.0.0.1] (item={'option': 'prior_gateway_device', 'value': ''}) => {"ansible_loop_var": "item", "changed": false, "item": {"option": "prior_gateway_device", "value": ""}, "msg": "Parameter 'value' must not be empty if state=present and allow_no_value=False"}

@jvonau and others may have ideas on the most thoughtful ways to clean up here, i.e. given Ansible's removal of `{{ ansible_product_uuid }}` and similar problems with undeclared vars ?

jvonau:
Fun, the version reported by apt is 2.10.5-1 but is referred to as "Ansible 2.10.6 contains Ansible-base version 2.10.5. This is a newer version than version 2.10.4 contained in the previous Ansible release", released on 2021-01-26 according to https://github.com/ansible-community/ansible-build-data/blob/main/2.10/CHANGELOG-v2.10.rst. 

The fix for above would be just to define a better default value than "" in roles/network/defaults/main.yml, I suggest "not-found" for what is used as a recording variable.

jvonau:
> ubuntu@box:~$ ansible -m setup localhost --connection=local | grep product
        "ansible_product_name": "",
        "ansible_product_serial": "",
        "ansible_product_uuid": "",
        "ansible_product_version": "",

" ansible_product_uuid  " error smells like a upstream bug somewhere, that is a built-in variable supplied by the setup module, that in fact does exist but has returned "", it's the stricter checking that caused this ansible supplied variable to puke.

holta:
Currently https://github.com/iiab/iiab/blob/master/roles/network/defaults/main.yml#L64 shows:

```
prior_gw_device: ""
```

Likely we should adopt & use an industry norm keyword like the following?

```
prior_gw_device: undefined
```

Possibly using the long-form variable name  `prior_gateway_device` ?

tim-moody:
>prior_gw_device: undefined

I would use 'none' like the other parameters. To me let x = 'undefined' is an oxymoron.

holta:
An initial value like "unknown" "undeclared" "undefined" "not-yet-found" is needed to (very intentionally) convey the ambiguity of initial state.  Similar to "unset" used here:

https://github.com/iiab/iiab/blob/master/roles/network/defaults/main.yml#L86-L87

Whereas wordings like "none" erroneously convey certainty.

We really should convey (accurate) ambiguity — wherever realistically possible anyway.  No solution is perfect given Ansible limitations etc.  But let's avoid Little Lies as much as possible...as these accidentally/inevitably become Big Lies (and bugs) down the road :/

holta:
@chjohn22

I'd like others to review it for possible improvements / future directions, but FYI this fix has been smoke-tested to work: PR #2672

PR #2672, opened 2021-01-29
Initialize vars more meaningfully, for #2669 and #2670, thx to ansible-base 2.10.5's more strict checking [roles/network/defaults/main.yml]

holta:
Possible emergency fix, given ansible-base 2.10.5's changed behavior.

See ongoing discussion at PR #2670.  See also #2669 and PR #2671.

### DIFF
--- a/roles/0-init/tasks/validate_vars.yml
+++ b/roles/0-init/tasks/validate_vars.yml
@@ -36,6 +36,11 @@
 # 2020-07-08 - Excellent analysis & summary by Jon Spriggs: "In Ansible,
 # determine the type of a value, and casting those values to other types"
 # https://jon.sprig.gs/blog/post/1801
+#
+# 2021-01-29 - ansible-base 2.10.5 (1) is more strict about empty string vars
+# (2) no longer supports "when: myvar is boolean", "is integer" & "is float"
+# (3) brings yet more "Ansible collection" dependency changes (undocumented!)
+# Details: https://github.com/iiab/iiab/pull/2672 (see also #2669)
 
 # 3. "How do i fail a task in Ansible if the variable contains a boolean value?
 # I want to perform input validation for Ansible playbooks"
--- a/roles/0-init/tasks/validate_vars.yml
+++ b/roles/0-init/tasks/validate_vars.yml
@@ -39,7 +39,7 @@
 #
 # 2021-01-29 - ansible-base 2.10.5 (1) is more strict about empty string vars
 # (2) no longer supports "when: myvar is boolean", "is integer" & "is float"
-# (3) brings yet more "Ansible collection" dependency changes (undocumented!)
+# (3) brings yet more "Ansible Collections" dependency changes (undocumented!)
 # Details: https://github.com/iiab/iiab/pull/2672 (see also #2669)
 
 # 3. "How do i fail a task in Ansible if the variable contains a boolean value?
--- a/roles/network/defaults/main.yml
+++ b/roles/network/defaults/main.yml
@@ -17,7 +17,7 @@
 
 # Wi-Fi - IF NECESSARY, CUSTOMIZE THESE 6 VARIABLES IN /etc/iiab/local_vars.yml
 # host_country_code: US
-# host_ssid: "Internet in a Box"
+# host_ssid: Internet in a Box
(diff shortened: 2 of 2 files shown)


holta:
@jvonau this was Smoke-Tested successfully on VM 193-deb10-BIG-0128 = 10.8.0.46

I've used your original lingo ("unset") for now.

Please review this PR if you have time?

jvonau:
Is it so hard to change a single variable (prior_gw_device) instead of this mass change? I have this suspicion that removing " " from around variables changes them from strings to booleans or perhaps being seen as **None** in the case of where **none** was used, VM's with a single network interface are useless to validate network code, doesn't exercise the full scope of the code. You have been warned.

holta:
> I have this suspicion that removing " " from around variables changes them from strings to booleans

As documented at https://github.com/iiab/iiab/blob/master/roles/0-init/tasks/validate_vars.yml#L19-L27 an Ansible var will be type "bool" if-and-only-if you set it to one of these ~18 values: (without using quotation marks, or vertical bar symbol!)

### yes|Yes|YES|no|No|NO
### |true|True|TRUE|false|False|FALSE
### |on|On|ON|off|Off|OFF

> or perhaps being seen as None in the case of where none was used

Not a risk.  The following 2 snippets are identical: (both create an Ansible string variable, i.e. of type "AnsibleUnicode")

```
myvar: none
```

```
myvar: "none"
```

jvonau:
@chjohn22

Personally given that the above test was preformed on Ubuntu and you are on RaspOS as the main author of the network code I would suggest that you try:

`cd /opt/iiab/iiab`
`sudo git checkout -b testing ` Creates new branch to keep change apart from the main code.
Now edit (must have sudo access or be root) roles/network/defaults/main.yml changing 
prior_gw_device: ""
to 
prior_gw_device: "not-found"
save the file
Check to see if the main install was completed
`sudo cat /etc/iiab/iiab.env` looking for what is present in the STAGE= variable
If STAGE= is less than 9 complete the install with `sudo ./iiab-install` otherwise use `sudo ./iiab-network` to configure the network.

Warning, there may be other errors that might appear with this new version of ansible and I would rather deal with any errors individually rather than risk upsetting the whole apple cart.

@holta 

Do as you wish, you asked, I provided feedback.

holta:
FYI this PR was also Smoke-Tested on RaspiOS Lite on RPi 4.

Thanks @chjohn22 if you can test any other environments as @jvonau has outlined above.

chjohn22:
@jvonau I only did run `sudo iiab` two hours ago and the install finished successfully.  I did not have to do anything else.

holta:
FYI here's an easy way to display the type of any Ansible variable: (in general)

```
- debug:
    msg: "myvar has type: {{ myvar | type_debug }}"
```

The output should always be...one of these 8 things:

1. bool
2. int
3. float
4. AnsibleUnicode [which means string]
5. list
6. dict
7. NoneType [if `myvar` doesn't yet have a type, e.g. `myvar:` was declared with no value]
8. AnsibleUndefined [if variable `myvar` has not yet been declared]

IIAB might in future use the above values for code safety.

Even snappier, one can test the type of a variable using these 4-5 (formerly 7-8) tests along the lines of `when: myvar is not string` and `when: myvar is number` etc:

1. <strike>is boolean</strike> (no longer supported by ansible-base 2.10.5)
2. <strike>is integer</strike> (no longer supported by ansible-base 2.10.5)
3. <strike>is float</strike> (no longer supported by ansible-base 2.10.5)
4. is number [int or float]
5. is string
6. is sequence [dict or list or string]
7. is mapping [dict or string]
8. is iterable [dict or list or string, apparently identical to `is sequence` above]

More details @ https://jon.sprig.gs/blog/post/1801 and https://jon.sprig.gs/blog/post/tag/ansible

jvonau:
> @jvonau I only did run `sudo iiab` two hours ago and the install finished successfully. I did not have to do anything else.

I doubt your network is fully configured, the noted error https://github.com/iiab/iiab/issues/2669#issuecomment-769681513 would not be exposed while using 'sudo iiab' a second time if the STAGE= counter is already at 9, the network configuration would be skipped, that is the reason for using ./iiab-network as I noted above.

PR #2686, opened 2021-02-23
kick ap0 [fix for WiFi hotspot after RaspiOS is apt updated from kernel 5.4 to 5.10]

jvonau:
From fix .66 git pull, ./iiab-network, and reboot to apply

### DIFF
--- a/roles/network/templates/hostapd/50-hostapd
+++ b/roles/network/templates/hostapd/50-hostapd
@@ -1,6 +1,6 @@
 if [ "$interface" = "br0" ] && [ $if_up = "true" ]; then
-    syslog info "50-iiab IF_UP br0 restarting dnsmasq"
-    sleep 2
+    syslog info "50-iiab IF_UP br0 restarting dnsmasq - kicking ap0"
+    ip link set ap0 up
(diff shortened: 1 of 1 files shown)


jvonau:
change from prior behavior

holta:
Tested on 10.8.0.66 = 180-rpi4-1g-lite-MIN (where all 48 RaspiOS apt updates had been applied since RaspiOS Lite 2021-01-11, causing the WiFi problem uncovered by @fisherbryan in South Africa).

Background:

1) This timing issue arose starting approx 2021-02-13 when RaspiOS pushed the new 5.10.11 kernel (for those who `apt update` etc) regardless of which WiFi firmware the RPi 4's internal hotspot was running, e.g. 7.45.206 or 7.45.229 ...or 7.45.18 which permits 32 client devices (#823).

2) For some people the problem only happened when removing the RPi 4's Ethernet cord to the Internet (or when there was no Ethernet cord plugged in to the RPi 4 during boot).

   For others, the problem happened regardless whether the RPi 4 had an Ethernet cord to the Internet or not.

Ref: #2610

Issue #2105, opened 2020-01-05
Update /opt/iiab/iiab/.travis.yml Python version? distro version? (integrate Ansible Lint?)


holta:
1) https://github.com/iiab/iiab/blob/master/.travis.yml#L3 currently shows:

   ```
   python: "2.7"
   ```

2) https://github.com/iiab/iiab/blob/master/.travis.yml#L6 currently shows:

   ```
   dist: trusty
   ```

3) Can [Ansible Lint](https://github.com/ansible/ansible-lint) ([docs](https://docs.ansible.com/ansible-lint/)) be plugged into the above file and if so how?

Ref:

#2096 NGINX & ./runrole need approachable docs for newbie/legacy implementers

holta:
Example output of `ansible-lint -p -R -v iiab-stages.yml` in the link below:

https://paste.ubuntu.com/p/9S9qWxNbB8/

(Thanks to @jvonau)

holta:
@jvonau suggests this should be "ported to bionic, use Python 3, and install the [Ansible] collections..."

holta:
This might become increasingly important if GitHub/Travis' testing environment's network connections fail more and more often?  As was happening earlier this morning, e.g. https://travis-ci.com/github/iiab/iiab/builds/218107893#L527 showed things like:

```
Err:77 http://archive.ubuntu.com/ubuntu trusty InRelease
  Could not connect to archive.ubuntu.com:80 (91.189.88.152), connection timed out [IP: 91.189.88.152 80]
```

jvonau:
https://config.travis-ci.com/explore Give this a shot.
>language: python
python: "3.8"
dist: focal
\# Install ansible
install:
\# Create ansible.cfg with correct roles_path and local_tmp
\# not sure is is needed now
\- "{ echo '[defaults]'; echo 'roles_path = ./roles/'; } >> ansible.cfg"
script:
\- /opt/iiab/iiab/scripts/ansible
\# now call the test
\- ansible-playbook tests/test.yml -i tests/inventory --syntax-check

holta:
> `script:`
> `- opt/iiab/iiab/scripts/ansible`

Very interesting idea.  Will this slow down Ansible installs during each CI syntax check?  (That ideally we'd like completed within ~2 minutes flat!)

holta:
FYI [scripts/ansible](https://github.com/iiab/iiab/blob/master/scripts/ansible) fails when part of a Travis job as follows:

```
[...]

PR #2689, opened 2021-02-24
Update & tighten up .travis.yml from Trusty (Ubuntu 14.04) + Python 2.7 to Focal (Ubuntu 20.04) + Python 3.8

holta:
Tested.

Bonus it runs substantially faster than in recent months, now that it avoids installing all ~80 Ansible Collections!

Thanks to @jvonau offering suggestions at #2105.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,13 +7,21 @@ addons:
     sources:
       - sourceline: ppa:ansible/ansible
     packages:
+#      - python-pip    # @arky had used this starting in 2018
       - ansible-base    # Install latest ansible-base e.g. 2.10.6+, similar to https://github.com/iiab/iiab/blob/master/scripts/ansible
-#      - python-pip
+#      - python3-pymysql    # These 7-or-8 packages are not needed during this very rapid --syntax-check
+#      - python3-psycopg2
+#      - python3-passlib
+#      - python3-pip
+#      - python3-setuptools
+      - python3-packaging    # To avoid warning "packaging Python module unavailable; unable to validate collection..."
+#      - python3-venv
+#      - virtualenv
 
 install:
 #  - scripts/ansible    # See #2105: fails to install latest Ansible (& ~4 Ansible Collections from collections.yml) due to Travis VM's disk layout/perms being different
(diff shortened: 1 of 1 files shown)


jvonau:
Really should enable the ansible ppa and install the same apt package set as used in scripts/ansible.

jvonau:
in place of `pip install ansible-base` numbers are spaces:
>addons:
2apt:
  4sources:
  6 - sourceline: 'ppa:ansible/ansible-2.10'
  4packages:
  6 - ansible-base
  6 - python3-pymysql
  6 -  python3-psycopg2
  6 -  python3-passlib
  6 - python3-pip
  6 - python3-setuptools 
  6 - python3-packaging 
  6 - python3-venv 
  6 - virtualenv

holta:
> in place of `pip install ansible-base` numbers are spaces:
> 
> > addons:
> > 2apt:
> > 4sources:
> > 6 - sourceline: 'ppa:ansible/ansible-2.10'
> > 4packages:
> > 6 - ansible-base
> > 6 - python3-pymysql
> > 6 -  python3-psycopg2
> > 6 -  python3-passlib
> > 6 - python3-pip
> > 6 - python3-setuptools
> > 6 - python3-packaging
> > 6 - python3-venv
> > 6 - virtualenv

Do we really need all those packages for a simple syntax check that takes about 1 second to run?

jvonau:
#2521

jvonau:
>Do we really need all those packages for a simple syntax check that takes about 1 second to run?

Then don't add all of them, just use the ppa for now, pip is not apt, with apt used in the field would it not be the best to have the exact same environment should 'linting' be added to the testing mix?

holta:
I have no idea why the apt install of ansible-base appears slower than the pip install of ansible-base (and certainly a whole lot wordier, in both .travis.yml and the Travis output!)

But it's good enough.  And we can always revert to a pip install of ansible-base later, if we want that bit of extra speed.

jvonau:
All the listed dependencies in the deb are installed, pip is bare bones and can't install system level dependencies not part of python.

PR #2690, opened 2021-02-25
Install bionic-or-focal version of ansible-base, depending on OS

holta:
Catches a few more OS/distro corner cases than PR #2521, tightens code, adds explanations/options for those wanting help uninstalling/installing Ansible (ansible-base) in other ways.

Refs: #2105, PR #2521, PR #2527, #2647, PR #2652, PR #2654, PR #2655, #2659, #2669, PR #2670, PR #2671, PR #2672, PR #2685, PR #2689.

PR #2691, opened 2021-02-25
scripts/ansible: Minor readability tweaks

holta:
Ref: PR #2690

### DIFF
--- a/scripts/ansible
+++ b/scripts/ansible
@@ -1,11 +1,11 @@
-#!/bin/bash -e
+#!/bin/bash -e   # Exit right away on error
 
 # PLZ SEE http://FAQ.IIAB.IO > "What is Ansible and what version should I use?"
 #         https://github.com/iiab/iiab/tree/master/scripts/ansible.md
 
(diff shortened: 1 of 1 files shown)


tim-moody:
`CURR_VER=undefined` 
is an oxymoron.

holta:
> `CURR_VER=undefined`
> is an oxymoron.

I think this is a legacy of @jvonau or @georgejhunt wanting vars described/declared near the top of the file to make the code generally readable.  Which is a very reasonable ask, even if declaring some vars like CURR_VER and GOOD_VER isn't adding a whole lot of value in this case, for sure.  As both are barely used at all (!)

holta:
FYI this PR was tested on Ubuntu 20.04

holta:
Conversely, I can strip out bash's `-e` flag completely if others (@jvonau ?) believe it's basically useless.

jvonau:
> Conversely, I can strip out bash's `-e` flag completely if others (@jvonau ?) believe it's basically useless.

Bad idea, you want the install to stop dead should any of the commands fail resulting in ansible not being full installed perhaps due to network issues at the user's location.

PR #2692, opened 2021-02-25
exit early with no /etc/debian_version present

jvonau:
before the grep


holta:
Needs some thought as inline and stdout documentation would need to be substantially reworked.

It works as is (even when a file like os-release or debian_version does not exist) but I agree if porting to non-Debuntu distros doesn't happen, this can be tightened up.

jvonau:
> It works as is (even when a file like os-release or debian_version does not exist) but I agree if porting to non-Debuntu distros doesn't happen, this can be tightened up.

If I don't do the work it won't get done, more smokescreen.

PR #2694, opened 2021-02-26
nathansalapat.com is offline [Minetest 5.1.1 installer for RPi]

tim-moody:


tim-moody:
install succeeded on latest raspios

Issue #2696, opened 2021-03-01
WiFi client machines failing to (re)connect to RPi 4 internal hotspot w/ passwords in South Africa [hostapd generates EAPOL errors; usually happens right after "ieee80211 phy0: brcmf_psm_watchdog_notify: PSM's watchdog has fired!" in dmesg output]


holta:
We need more logs from @fisherbryan (the output of `dmesg`, and the pastebin resulting from [iiab-diagnostics](https://github.com/iiab/iiab/blob/master/scripts/iiab-diagnostics.README.md)).

So far we have what appears to be `/var/log/syslog` :

> 1 Android v6.0
1 -Tablet,
Then Adding another up to 10 [tablets..]
http://sprunge.us/2DjFUm?en
>
> Result, some tablets are connecting and some [are] not...my desktop ubuntu machine also can't connect to wifi

We suspect these facts are not relevant to the problem:

- `host_country_code: ZA` was set in /etc/iiab/local_vars.yml
- `iiab_gateway_enabled: True` was set in /etc/iiab/local_vars.yml
- RaspiOS includes hostapd 2.7 &mdash; hostapd 2.9 could easily be tested using Ubuntu on RPi 4 is nec

These aspects however (might) be relevant:

- A wpa2 password is being used in `/etc/hostapd/hostapd.conf` — originally the password contained an '@' — but FYI even after removing that special character, the problems persisted
- RaspiOS recently apt updated from Linux kernel 5.4 LTS to 5.10 LTS
- RPi 4 IIAB is running WiFi firmware 7.45.18, as shown by `dmesg | grep brcm`
- The very large number tablets (typically 10-to-20 per school) are "Point of View (China) with Android 10"
- RaspiOS runs hostapd 2.7 from 2018-12-02 (conceivably hostapd 2.9 from 2019-08-07 might do better here?)

Last week's output might (or might not) be related:

- http://sprunge.us/yyVc1t?en (iiab-diagnostics shows possible WiFi firmware crash on Line 1422, 34m32s after boot?)
- http://sprunge.us/QAQyvk?en /opt/iiab/iiab/iiab-network.log (13,079 lines, after 6 successful runs of `./iiab-network`)

Refs: #823, #1737, #2610, PR #2686

holta:
A new batch of data from @fisherbryan &mdash; problems appeared when the 4th tablet tried to connect to the RPi 4 IIAB's WiFi hotspot:

[...]

PR #2697, opened 2021-03-03
WARN implementers that hotspot passwords may FAIL, especially if firmware customized with 'wifi_hotspot_capacity_rpi_fix: True' ?

holta:
We don't know that the situation is much better with 'wifi_hotspot_capacity_rpi_fix: False'

But the warning is left intentionally broad for now, to warn IIAB implementers about hotspot passwords in general.

See #2696 for many more and hopefully ongoing details, as we and others progressively learn more here.

### DIFF
--- a/vars/default_vars.yml
+++ b/vars/default_vars.yml
@@ -115,8 +115,8 @@ host_country_code: US
 host_ssid: Internet in a Box
 host_wifi_mode: g
 host_channel: 6
-hostapd_secure: False
-hostapd_password: changeme
+hostapd_secure: False # 2021-03-02 #2696 WiFi EAPOL fails if hotspot passwords,
+hostapd_password: changeme # eg if firmware wifi_hotspot_capacity_rpi_fix: True
 hostapd_install: True    # 2020-01-21: this var MIGHT be implemented in future.
 hostapd_enabled: True
 wifi_hotspot_capacity_rpi_fix: True    # Restores the ability of RPi internal
--- a/vars/local_vars_big.yml
+++ b/vars/local_vars_big.yml
@@ -63,8 +63,8 @@ host_country_code: US
 host_ssid: Internet in a Box
 host_wifi_mode: g
 host_channel: 6
-hostapd_secure: False
-hostapd_password: changeme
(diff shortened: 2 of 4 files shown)


holta:
We can further revise this warning as we learn more about #2696.

A few core patterns have now been confirmed — RPi internal hotspot passwords (`hostapd_secure: True`) cause WiFi meltdowns...

- with kernel 5.8 or 5.10 for sure — and (possibly) with earlier kernels like 5.4 too??
- with Ubuntu 20.10 or RaspiOS Lite 2021-01-11 (apt updates were applied in both cases)
- with hostapd 2.9 or 2.7
- with WiFi firmware 7.45.18 arising from `wifi_hotspot_capacity_rpi_fix: False` (#823)
- with `wifi_up_down: True`
- with `iiab_gateway_enabled: True`
- with `host_country_code: ZA`
- with RPi 4

Conclusion: this opens up many potential avenues to try to bisect, attempting to rapidly reproduce the failure with (earlier kernels, other firmwares e.g. 7.45.229, wifi_up_down: False, iiab_gateway_enabled: False, RPi 3 or RPi 400, etc...)

Testing is Extremely Rapid: `hostapd_cli -i ap0 list_sta | wc -l` shows all WiFi clients kicked off within less than 5 minutes every time, when (even just 1-or-2 of 10+ ?) WiFi client machines are actively browsing.

holta:
FYI personally I've been unable to make a 7.45.18 RPi 4 hotspot (with password) fail, after 32+ hours of testing with 20 WiFi-connected client machines.  The RPi 4 is running RaspiOS Lite 2021-01-11 apt updated to the 5.10 LTS kernel.

/etc/iiab/local_vars.yml contains `iiab_gateway_enabled: True` and Kolibri is installed, in an attempt to simulate #2696 conditions, as closely as possible.

In brief spurts, I've also tried aggressive browsing on a couple of the WiFi client devices, to see if that might cause the RPi 4's WiFi to crash, but it's all continued to work.

(In Short: blaming `wifi_hotspot_capacity_rpi_fix: True` and the 6-year-old 7.45.18 high-capacity WiFi firmware so far remains completely speculative, without any material or reproducible evidence to date.)

Issue #2713, opened 2021-03-19
Let's provide a less confusing way for Kolibri communities to save [export Channels] to USB drives


holta:
![IMG-20210319-WA0001](https://user-images.githubusercontent.com/2458907/111840497-da2faf80-88d2-11eb-9c00-a352b4e14e28.jpg)

@fisherbryan eventually got it to work at the Command-Line as root, using:

```
umount /media/usb0
mount -o rw,users,umask=000 /dev/sda1 /media/usb0
```

(I don't know if all those -o options are truly needed?  Possibly something as simple as `-o rw` might be enough?)

But FWIW Kolibri is now confirmed working for him — as a result of the above be is now able to save materials from Kolibri to his VFAT-formatted 8 GB USB stick.

As background he is running a very recent install of IIAB on Raspberry OS Lite on RPi 4's in South Africa, including Kolibri 0.14.6

Conclusion: IIAB should offer a hassle-free option that does something very similar to the above.  As there are at least 3 different organizations that have run into this same problem.  So let's investigate how to do that herein:

https://github.com/iiab/iiab/tree/master/roles/usb_lib

(IIAB's user-facing documentation for everyday use of USB sticks is here...)

http://FAQ.IIAB.IO > "4 Can teachers display their own content?"

holta:
Possibly related concern, unearthed by @jvonau: (is it possible Kolibri 0.14.x doesn't play well with the fuse apt packages like `exfat-fuse` and `exfat-utils` [*] e.g. possibly also affecting filesystems like VFAT / FAT32 and NTFS ?)

https://github.com/learningequality/kolibri/blame/release-v0.14.x/kolibri/core/discovery/utils/filesystem/posix.py

[*] as installed by https://github.com/iiab/iiab/blob/master/roles/2-common/tasks/packages.yml#L22-L23

holta:
FYI @jvonau suspects this entire ticket is possibly a non-issue &mdash; that would not arise at all (i.e. the inability to save from Kolibri) if the USB stick/drive had been cleanly unmounted[*] the last time it was used.

@fisherbryan please do confirm/deny that when you can (with multiple confirmed-cleanly-ejected USB sticks if possible!)

holta:
[...]

PR #2714, opened 2021-03-20
WIP: /etc/usbmount/usbmount.conf setting so that Kolibri exports work

holta:
Intended for @fisherbryan and the several others stuck with the inability to export Kolibri Channels to USB drives, as described in #2713.

In my testing of this PR (on RaspiOS Lite on RPi 4) Kolibri exports-to-USB and imports-from-USB both work fine.   (Imports probably work fine without this PR as well, though Bryan had some unspecified trouble yesterday, trying to import Kolibri Channels from his USB drive.)

### DIFF
--- a/roles/usb_lib/tasks/install.yml
+++ b/roles/usb_lib/tasks/install.yml
@@ -67,8 +67,8 @@
     line: 'FILESYSTEMS="vfat ext2 ext3 ext4 hfsplus exfat fuseblk ntfs"'
     path: /etc/usbmount/usbmount.conf
 
-# Setting 'umask=0000' for partic filesystems e.g. {fat, vfat, ntfs} appears to accomplish the exact same thing:
-# FS_MOUNTOPTIONS="-fstype=fat,umask=0000 -fstype=vfat,umask=0000 -fstype=ntfs,umask=0000"
+# Setting 'umask=0000' for partic filesystems e.g. {fat, vfat, ntfs, exfat} appears to accomplish the exact same thing:
+# FS_MOUNTOPTIONS="-fstype=fat,umask=0000 -fstype=vfat,umask=0000 -fstype=ntfs,umask=0000 -fstype=exfat,umask=0000"
 - name: "Add ',umask=0000' to MOUNTOPTIONS var in /etc/usbmount/usbmount.conf, so Kolibri exports work"
   lineinfile:
     regexp: '^MOUNTOPTIONS=.*'
--- a/roles/usb_lib/tasks/install.yml
+++ b/roles/usb_lib/tasks/install.yml
@@ -61,10 +61,18 @@
     - { src: 'iiab-usb_lib-show-all-off', dest: '/usr/bin/', mode: '0755' }
     - { src: 'iiab-clean-usb.sh', dest: '/usr/sbin/', mode: '0755' }
 
-- name: Enable exFAT and NTFS in /etc/usbmount/usbmount.conf
+- name: Add ' exfat fuseblk ntfs' to FILESYSTEMS var in /etc/usbmount/usbmount.conf
   lineinfile:
-    regexp: '^FILESYSTEMS.*'
+    regexp: '^FILESYSTEMS=.*'
     line: 'FILESYSTEMS="vfat ext2 ext3 ext4 hfsplus exfat fuseblk ntfs"'
+    path: /etc/usbmount/usbmount.conf
+
+# Setting 'umask=0000' for partic filesystems e.g. {fat, vfat, ntfs} appears to accomplish the exact same thing:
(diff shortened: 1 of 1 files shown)


jvonau:
Canoe, 1:09 PM Simply adding 'umask=0000' seems to be enough, when adding it to this line in /etc/usbmount/usbmount.conf and then rebooting... MOUNTOPTIONS="sync,noexec,nodev,noatime,nodiratime,umask=0000" 
As reconfirmed by Kolibri writing out any selected channels to: /media/usb0/KOLIBRI_DATA
Canoe, 1:20 PM Importing back into Kolibri (what I saved to a USB stick) worked for me...
Jerry, 1:27 PM should be in the FS_MOUNTOPTION for vfat only don't think using such a wide brush is a great idea when other filesystem types could be used also... would need to test ntfs and ext2/3/4 to ensure 0000 is correct/works for that filesystem whatever, rolling back security for one app is a bad idea when the now documented remount workaround is used.
Jerry 1:38 PM should of just kept my mouth shut, you don't want to punt this issue upstream to correct, lacking automount when used as a service or prompt for userid/password to remount the devices as needed. 
Canoe, 1:39 PM We should make sure NTFS and other filesystems work with Kolibri too.
Jerry 1:40 PM we don't have to do anything... the deployment needs to test not me.. 
Canoe, 2:00 PM mount supports 'umask=0000' for fat, vfat, ntfs and other filesystems so I don't see a huge danger in enabling that across the board for Kolibri folks that need it. But if we want to restrict that to some subset of filesystems like {fat, vfat, ntfs} that could be an option too.
Jerry, 2:02 PM Not across the board that is not a great idea.. see should of just keep my mouth shut, now you are off to put into effect this half backed idea for one badly coded app, those defaults are there for a reason... security. Canoe, 2:10 PM If something like the following is really better? (Certainly it's a lot longer, but if it works?) FS_MOUNTOPTIONS="-fstype=fat,umask=0000 -fstype=vfat,umask=0000 -fstype=ntfs,umask=0000"It's said that 'umask' settings do not affect ext4 & similar, so it seems there's likely no material difference in the end?
Canoe, 2:18 PM Outlines of a keep-it-simple PR...that I can certainly revise later today if there's a much better way: https://github.com/iiab/iiab/compare/master...holta:usbmount-with-kolibri
Jerry 2:20 PM should be fixed upstream... not here.. 
Canoe, 2:21 PM Kolibri is running under user 'kolibri' presumably...root@box:~# grep kolibri /etc/group 7:disk:x:6:kolibri 24:www-data:x:33:kolibri 65:kolibri:x:995:
Jerry 2:31 PM yes, but the import/export routines are based on running from a user's desktop, the desktop handles mounting the device so the desktop user can write to the device, so 'kolibri' running as a service doesn't really have the needed permissions to write to the device even when there is somebody logging into the desktop... That is why "/.kolibri" is hidden in a user's home directory,with service being started from there when this app was first developed. now it's time to take off the desktop diaper and fix it so kolibri works as a real service..
Canoe, 2:34 PM Ok. Feel free to spell out that recommendation here, as they may not understand its reasons/advantages: https://github.com/learningequality/kolibri/issues
Jerry, 2:36 PM I'm not the one suffering, I suggest the user file is issue, and link back to our issue.. I'll chime in when needed.. 
Jerry 2:44 PM I do not understand why all iiab users have to be exposed to this churn and possibly suffer some data loss from having the device being to set the way you propose, but your years of engineering experience tell you it's the right thing to do... Who am I to disagree with the self-appointed demi-god of IIAB.. best of luck in your future endevers, good luck 
Canoe, 2:49 PM Bryan's use of mount option 'users' is not necessary, as I've tested. But he's the 3rd person who asked how to make this work, so some kind of short-term solution is needed. If an upstream change is also appropriate in a future version of Kolibri, that makes perfect sense too.
Jerry. 2:51 PM then you file this issue upstream, this can be viewed as a security hole always using umask=0000 and kolibri returns the same error when there is no mounted usb device present, could prompt to mount the device, but would need a sudo userid/password combo to do the mounting.. same code path could prompt to remount asking for the userid/password to use 
Canoe, 2:56 PM Bryan just got home from Limpopo [South African province bordering Botswana, Zimbabwe and Mozambique]. He doesn't seem to have made progress on this issue, not having slept in 36h it appears.
Jerry, 2:56 PM it's an upstream issue the way I see it.. leave him be, bug upstream, that is where the shortsightedness is... "Kolibri needs access to the usb device but needs a user with permissions to do so, please enter userid/password when prompted" is what I would like to see... "
Canoe, 3:02 PM I have to run for the afternoon but we should write to https://github.com/learningequality/kolibri/issues asking about the security implications of Kolibri's USB-writing architecture. I certainly don't understand the desktop and associated implications, but others should be aware, and be made aware of other/better options.

PR #2715, opened 2021-03-21
Another way to set /etc/usbmount/usbmount.conf so that Kolibri exports work

holta:
Another alternative, more complex than PR #2714 but possibly a tiny bit safer.

Refs:

- #2713 "Let's provide a less confusing way for Kolibri communities to save [export Channels] to USB drives"
- http://FAQ.IIAB.IO > "4 Can teachers display their own content?"
- https://github.com/iiab/iiab/wiki/IIAB-Installation#external-usbdrive-content
- https://github.com/iiab/iiab/blob/master/roles/usb_lib/README.rst

### DIFF
--- a/roles/usb_lib/README.rst
+++ b/roles/usb_lib/README.rst
@@ -20,12 +20,11 @@ Automount is handled by usbmount, and scripts in this role look in the root of t
 
 USB drives must be formatted with one of the filesystems listed under "FILESYSTEMS=" at ``/etc/usbmount/usbmount.conf`` — these are specified on/around Line 67 of: `/opt/iiab/iiab/roles/usb_lib/tasks/install.yml <https://github.com/iiab/iiab/blob/master/roles/usb_lib/tasks/install.yml#L67>`_
 
+In March 2021, Kolibri exports were enabled, using ``umask=0000`` (also within /etc/usbmount/usbmount.conf) to override what was ``umask=0022`` by default: `PR #2715 <https://github.com/iiab/iiab/issues/2715>`_
+
 Official `usbmount <https://github.com/hfuchs/usbmount>`_ documentation, from 2010:
 
 * https://github.com/hfuchs/usbmount/blob/master/README
 * https://github.com/hfuchs/usbmount/blob/master/usbmount.conf
 
-In March 2021, Kolibri exports were enabled, using ``umask=0000`` (also within /etc/usbmount/usbmount.conf) to override what was ``umask=0022`` by default: `PR #2715 <https://github.com/iiab/iiab/issues/2715>`_
-
-Legacy: There is also a patch for problems with automount on Fedora 21+.
-Please Note that as of 4.1.8-200.fc22.x86_64 not all USB drives will mount, even with this patch.
+Legacy warning: There is also a patch for problems with automount on Fedora 21+.  Please note that as of 4.1.8-200.fc22.x86_64 not all USB drives will mount, even with this patch.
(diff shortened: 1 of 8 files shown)


holta:
FYI this PR was smoke-tested and functionallity tested on Raspberry Pi OS Lite on RPi 4.

holta:
New variable `usb_lib_umask0000_for_kolibri` is now introduced, per @tim-moody's recommendation during this morning's IIAB call (http://minutes.iiab.io).

With Kolibri usage growing broadly in 2021, this variable is intentionally set to `True` in all 3 local_vars.yml files {MIN-sized, MEDIUM-sized, BIG-sized} so that implementers and field operators can add Kolibri after the fact, and without infuriating confusion.

Conversely, the variable is intentionally set to `False` in [/opt/iiab/iiab/vars/local_vars_medical.yml](https://github.com/iiab/iiab/blob/master/vars/local_vars_medical.yml) for a small amount of additional security in implementations that intend never to install Kolibri (noting that USB drives are still mounted 'rw' in all situations, allowing root to read and write to the USB drive no matter how this flag is set!)

Finally as discussed on the call, Kolibri's playbook conditionally modifying usb_lib playbook's `/etc/usbmount/usbmount.conf` was very explicitly avoided, so as to avoid dependency/maintenance complexities in assorted situations such as usb_lib being installed/uninstalled/enabled/disabled before-or-after kolibri is installed/uninstalled/enabled/disabled.

Tested on RaspiOS Lite on RPi 4.

tim-moody:
I would have preferred a more generic name like usb_lib_permit_all

holta:
Further tested with `usb_lib_umask0000_for_kolibri: False` in /etc/iiab/local_vars.yml, by forcibly removing `usb_lib_installed: True` from `/etc/iiab/iiab_state.yml` and then running `cd /opt/iiab/iiab ; ./runrole usb_lib`

While I considered putting this logic into tasks/setup.yml or tasks/enable-or-disable.yml, I strongly prefer we not add complexity and over-engineer (after 7 days now, and in general) as IIAB communities have far higher priorities than these 10-characters in `/etc/usbmount/usbmount.conf` in the end.  Recap: as documented in http://FAQ.IIAB.IO > "4 Can teachers display their own content?" and at https://github.com/iiab/iiab/blob/master/roles/usb_lib/README.rst, those who prefer usbmount's default should set `usb_lib_umask0000_for_kolibri: False` prior to installing IIAB.

That being said, thank you to @fisherbryan for uncovering `umask=0000` core to this fix, **and for everyone who helped along the way (5 people) bringing this longstanding issue to a close finally after several years!**

jvonau:
Any reason why this new default is forced upon non-users of kolibri as only 'big' has the option enabled to install kolibri? https://github.com/iiab/iiab/search?q=kolibri_install
> I would have preferred a more generic name like usb_lib_permit_all

That would be preferable as there could be other situations that may require the same treatment

holta:
> Any reason why this new default is forced upon non-users of kolibri as only 'big' has the option enabled to install kolibri?

Summarized [above](https://github.com/iiab/iiab/pull/2715#issuecomment-807157872).

> there could be other situations that may require the same treatment

Let's change the variable name if that situation arises.

Other scenarios and contingencies could also arise over the years, for sure (in `/etc/usbmount/usbmount.conf` or in any number of very nearby/connected areas).

Generalizing too early is almost universally a software engineering mistake (not every time of course, but almost always!)  Things should be generalized when there's a demonstrable user/community/customer need.

jvonau:
> Further tested with `usb_lib_umask0000_for_kolibri: False` in /etc/iiab/local_vars.yml, by forcibly removing `usb_lib_installed: True` from `/etc/iiab/iiab_state.yml` and then running `cd /opt/iiab/iiab ; ./runrole usb_lib`
> 
> While I considered putting this logic into tasks/setup.yml or tasks/enable-or-disable.yml, I strongly prefer we not add complexity and over-engineer (after 7 days now, and in general) as IIAB communities have far higher priorities than these 10-characters in `/etc/usbmount/usbmount.conf` in the end. Recap: as documented in http://FAQ.IIAB.IO > "4 Can teachers display their own content?" and at https://github.com/iiab/iiab/blob/master/roles/usb_lib/README.rst, those who prefer usbmount's default should set `usb_lib_umask0000_for_kolibri: False` prior to installing IIAB.

Make sure that is loud and clear. The need to remove the *_installed state flag to have what amounts to editing a file action being preformed  points to having the editing in the incorrect location and there is a lack of recording of what is currently active. #2717 corrects these deficiencies.

PR #2716, opened 2021-03-21
selective remount

jvonau:
### Fixes bug:
#2713
Use KOLIBRI_DATA directory on the usb drive as a trigger to remount with umask 0000

### DIFF
--- a/roles/kolibri/tasks/main.yml
+++ b/roles/kolibri/tasks/main.yml
@@ -28,6 +28,17 @@
   include_tasks: install.yml
   when: kolibri_installed is undefined
 
+- name: Enable usbmount override for mounting
+  template:
+    src: 71-kolibri
+    dest: /etc/usbmount/usbmount.d/71-kolibri
+  when: kolibri_enabled
+
+- name: Disable usbmount override
+  file:
+    path: /etc/usbmount/usbmount.d/71-kolibri
+    state: absent
+  when: not kolibri_enabled
 
 - name: Enable & Start 'kolibri' systemd service, if kolibri_enabled
   systemd:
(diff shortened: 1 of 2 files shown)


holta:
This would appear not to allow Kolibri exports, until USB drives contain a KOLIBRI_DATA directory?  If so, leading to confusion for everyday Kolibri field operators when they insert a new USB stick/drive.

But this PR presumably works to import from multiple attached USB drives?

Is this tested on RaspiOS and/or Ubuntu on RPi?

Aside: we now have 3 PR's (#2714, #2715, #2716) towards solving this ongoing end-user confusion with Kolibri exports, that apparently affects imports too (#2713).

jvonau:
> This would appear not to allow Kolibri exports, until USB drives contain a KOLIBRI_DATA directory? If so, leading to confusion for everyday Kolibri field operators when they insert a new USB stick/drive.
>
Um, is creating the directory ahead of time that hard to do? Document fodder I would think, you are good at that. More confusing than using the Kolbri's built-in tools that don't work when installed as a service as per upstream's suggestions and help?
> But this PR presumably works to import from multiple attached USB drives?
> 
Does Kolibri allow what you suggest now? Should work
> Is this tested on RaspiOS and/or Ubuntu on RPi?
> 
Ubuntu Mate Desktop.
> Aside: we now have 3 PR's (#2714, #2715, #2716) towards solving this ongoing end-user confusion with Kolibri exports, that apparently affects imports too (#2713).

You would rather roll back security system wide for one misbehaving app, makes sense to me....

jvonau:
different approach taken.

PR #2719, opened 2021-04-01
Jupyter Notebooks

georgejhunt:
- Creates temporary user identities, but no unix permissions on the server that enable access outside a walled garden.
- Jupyter server runs as a service under systemd.
- Default is disabled in default_vars.yml
- Not yet tested on Rpi -- so not ready for merging into master
- See roles/jupyter/docs/README.md


tim-moody:
can we have the files in /library/jupyter/private instead of /var/lib/private?

georgejhunt:
Good idea

georgejhunt:
- I spent two weeks unsuccessfully, using print statements in site-packages/jupyterhub/(python programs> and pdb.set_trace() trying to understand all of the ways and places that permissions were being handled, and to get the private folder into /library. 
- I decided the benefit was not enough to justify more time, since running as root, and placing the files in /var/lib/private was already working.
- I verified that this PR runs on Rpi. But it's slow (takes about 15 seconds for the spawned single user python server to become available to the client). But then the application seems responsive, and adequate.
- I verified that although the systemd python-kernel-launcher runs as root, the spawned single-user-python kernel is running as a user that has privileges only in the "private" folder (and no access to anything in /home, and no entries in /etc/passwd or /etc/shadow).  That seems good enough to me.
- Regarding the use of /var/lib, python programs themselves are very compact. But the jupyter notebooks store the outputs from each cell. I was working on a TED-eD zim with 1800 videos. The notebook itself, with many output lists was  200KB. The videos themselves were on an external USB disk. I'd still like to move the notebooks to /library. But there are more moving parts than I have the patience for.

holta:
Ubuntu 21.04 prerelease VM test feedback:

```
TASK [network : Install clean copy of /usr/bin/iiab-gen-iptables from template] ***
fatal: [127.0.0.1]: FAILED! => {"changed": false, "msg": "AnsibleUndefinedVariable: 'jupyterhub_port' is undefined"}
```

@georgejhunt will work on it.

holta:
Possible Promotional Description:

> “High Schools may want to consider JupyterHub to integrate coding with dynamic interactive graphing — A New Way to Think About Programming allowing students to integrate science experiments results and program output within their notebook/document/blog.”

The `/etc/iiab/iiab.ini` description is currently:

> "Raspberry Pi Jupyter python programming environment"

PR #2667, opened 2020-12-27
use iptables legacy mode

jvonau:
### Fixes bug:
#2665

Issue #2665, opened 2020-12-26
iptables moved in ubuntu-20 to /usr/sbin/iptables


georgejhunt:
But iiab-gen-iptables  in roles/network/templates/gateway/iiab-gen-iptable says:
{% if is_debuntu %}
IPTABLES=/sbin/iptables
IPTABLES_DATA=/etc/iptables.up.rules
{% else %}
IPTABLES=/usr/sbin/iptables
IPTABLES_DATA=/etc/sysconfig/iptables
{% endif %}

This surfaces because gateway_enabled is no longer working on my development machine. Maybe just a symbolic link from /sbin/iptables to /usr/sbin/iptables (that's what I did to fix in the short term)

tim-moody:
on my x86 ub20.10 /sbin is linked to /usr/sbin and /sbin/iptables -S returns results

holta:
@georgejhunt should this be marked as a bug?

tim-moody:
I get the same result on my ub 20.04 vm as on my 20.10 nuc, i.e. that I don't see this problem.

holta:
VM 191-u2004-srv-BIG-1226 = 10.8.0.50 is now installing -- feel free to log in there if it helps?

holta:
> VM 191-u2004-srv-BIG-1226 = 10.8.0.50 is now installing -- feel free to log in there if it helps?

Just FYI on the above Ubuntu Server 20.04.1 (where BIG-sized IIAB has now finished installing) we have:

```
root@u20-srv:~# ls -l /sbin/iptables
lrwxrwxrwx 1 root root 26 Jul 31 12:29 /sbin/iptables -> /etc/alternatives/iptables
root@u20-srv:~# ls -l /usr/sbin/iptables
lrwxrwxrwx 1 root root 26 Jul 31 12:29 /usr/sbin/iptables -> /etc/alternatives/iptables
root@u20-srv:~# ls -l /bin/iptables
ls: cannot access '/bin/iptables': No such file or directory
root@u20-srv:~# ls -l /usr/bin/iptables
ls: cannot access '/usr/bin/iptables': No such file or directory

root@u20-srv:~# apt -a list iptables
Listing... Done
iptables/focal,now 1.8.4-3ubuntu2 amd64 [installed,automatic]

root@u20-srv:~# more /etc/issue.net
Ubuntu 20.04.1 LTS
root@u20-srv:~# uname -a
[...]

Issue #796, opened 2018-05-14
Rapid networking & other IIAB config changes that can't wait ~20min for Ansible


holta:
Am trying with @tim-moody to prioritize crucial features where implementers don't have time to wait ~20min for Ansible to finish.  Hopefully several of these improvements are possible for IIAB 6.6!?

Example things that (might!) be changeable in less than 1 minute, via Admin Console especially &mdash; or command-line interface if necessary &mdash; order VERY preliminary below!  Are these good ideas?  Are there other priorities needed from [local_vars.yml](http://wiki.laptop.org/go/IIAB/local_vars.yml) ?

0) Toggle student access to Internet (iiab_gateway_enabled: False/True)
1) Internal SSID Name Change (do implementers care about IIAB's internal Wi-Fi channel & password? Rarely...)
2) Connecting to / Disconnecting from upstream Wi-Fi with UX for SSID + password
3) VPN on/off + VPN handle
4) Fixed IP address to DHCP and back
5) Other network stuff...network sensing might be nec??
6) Turn services (server apps) on/off
7) iiab_home_url: /home or /wordpress or /mediawiki or /wiki (for DokuWiki)

holta:
Quick thoughts from @jvonau:

>1 & 3 easy. 4 & 5 & 6 network services (squid dhcpd/dnsmasq) are in iiab-network. 2 is easy on raspbian but the others not so easy. 6 could just be calls to systemctl to enable/start and disable/stop

m-anish:
5 would be good to have

katneer91:
Like

tim-moody:
I think that state and environment are two different things. I would like to see a state file (json/yml/ini) that holds name-value pairs for all the main features of iiab, but is not as chatty as iiab.ini, which was really intended for reporting. That file could be loaded as facts by ansible and also used (read/write) by non-ansible scripts/applications that change state. It would include networking (but not working variables), installed and (actually) enabled, and other things that define the current state of iiab at any point in time.

holta:
Important...but not achievable for [IIAB 6.6](https://github.com/iiab/iiab/wiki/IIAB-6.6-Release-Notes) so moved to [IIAB 6.7](https://github.com/iiab/iiab/milestone/4)

holta:
If any such redesign of /etc/iiab/iiab.env and related files moves forward, it should take into account that 3 different people got stuck as a result of iiab.env &mdash; i.e. not know that such a centrally-important file even exists:

[...]

PR #2727, opened 2021-04-12
Don't load legacy modules [iptables firewall] & New feature rapid changing of client internet access [iiab-internet-on, iiab-internet-off]

jvonau:
Eliminates this noted warning #2667 #2665
\# Warning: iptables-legacy tables present, use iptables-legacy to see them

### DIFF
--- a/roles/network/tasks/computed_network.yml
+++ b/roles/network/tasks/computed_network.yml
@@ -141,6 +141,13 @@
     line: 'IIAB_LAN_DEVICE={{ iiab_lan_iface }}'
     state: present
 
+- name: Record 'IIAB_GATEWAY_ENABLED={{ iiab_gateway_enabled }}' in {{ iiab_env_file }}
+  lineinfile:
+    path: "{{ iiab_env_file }}"
+    regexp: '^IIAB_GATEWAY_ENABLED=*'
+    line: 'IIAB_GATEWAY_ENABLED={{ iiab_gateway_enabled }}'
+    state: present
+
 - name: Add 'computed_network' variable values to {{ iiab_ini_file }}
   ini_file:
     dest: "{{ iiab_ini_file }}"
--- a/roles/network/tasks/enable_services.yml
+++ b/roles/network/tasks/enable_services.yml
@@ -205,6 +205,17 @@
     group: root
     mode: 0755
 
+- name: Install iiab-internet-on|off
+  template:
+    src: "{{ item.src }}"
+    dest: "{{ item.dest }}"
+    owner: root
+    group: root
+    mode: 0755
+  with_items:
+    - { src: 'gateway/iiab-internet-on', dest: '/usr/bin/iiab-internet-on' }
+    - { src: 'gateway/iiab-internet-off', dest: '/usr/bin/iiab-internet-off' }
+
 - name: Add 'squid' variable values to {{ iiab_ini_file }}
   ini_file:
     path: "{{ iiab_ini_file }}"
--- a/roles/network/tasks/restart.yml
+++ b/roles/network/tasks/restart.yml
@@ -53,11 +53,12 @@
     name: avahi-daemon
     state: restarted
 
-- name: Create gateway flag
-  shell: echo 1 > /etc/sysconfig/olpc-scripts/setup.d/installed/gateway
-  args:
-    creates: /etc/sysconfig/olpc-scripts/setup.d/installed/gateway
-  when: iiab_network_mode == "Gateway"
+- name: Record 'IIAB_GATEWAY_ENABLED={{ iiab_gateway_enabled }}' in {{ iiab_env_file }}
+  lineinfile:
+    path: "{{ iiab_env_file }}"
+    regexp: '^IIAB_GATEWAY_ENABLED=*'
+    line: 'IIAB_GATEWAY_ENABLED={{ iiab_gateway_enabled }}'
+    state: present
 
 #netplan de-configures pre-created bridged interfaces 
 #- name: Reload netplan when Wifi is not gateway on Ubuntu 18+
--- a/roles/network/tasks/restart.yml
+++ b/roles/network/tasks/restart.yml
@@ -53,13 +53,6 @@
     name: avahi-daemon
     state: restarted
 
-- name: Record 'IIAB_GATEWAY_ENABLED={{ iiab_gateway_enabled }}' in {{ iiab_env_file }}
-  lineinfile:
-    path: "{{ iiab_env_file }}"
-    regexp: '^IIAB_GATEWAY_ENABLED=*'
-    line: 'IIAB_GATEWAY_ENABLED={{ iiab_gateway_enabled }}'
-    state: present
-
 #netplan de-configures pre-created bridged interfaces 
 #- name: Reload netplan when Wifi is not gateway on Ubuntu 18+
 #  shell: netplan apply
--- a/roles/network/templates/gateway/iiab-gen-iptables
+++ b/roles/network/templates/gateway/iiab-gen-iptables
@@ -86,14 +86,6 @@ elif [ "$ports_externally_visible" -lt 0 ] || [ "$ports_externally_visible" -gt 
     exit 1
 fi
 
-#if [ "$wan" != "none" ] && [ "$network_mode" != "Appliance" ]; then
-# Load iptables kernel modules
-/sbin/modprobe ip_tables
-/sbin/modprobe iptable_filter
-/sbin/modprobe ip_conntrack
-/sbin/modprobe iptable_nat
-#fi
-
 # Delete all existing firewall rules
 $IPTABLES -F
 $IPTABLES -t nat -F
--- a/roles/network/templates/gateway/iiab-gen-iptables
+++ b/roles/network/templates/gateway/iiab-gen-iptables
@@ -37,6 +37,7 @@ IPTABLES_DATA=/etc/sysconfig/iptables
 source {{ iiab_env_file }}
 lan=$IIAB_LAN_DEVICE
 wan=$IIAB_WAN_DEVICE
+iiab_gateway_enabled=$IIAB_GATEWAY_ENABLED
 echo -e "\nLAN: $lan"
 echo -e "WAN: $wan\n"
 #network_mode=`grep iiab_network_mode_applied {{ iiab_ini_file }} | gawk '{print $3}'`
@@ -49,7 +50,6 @@ gw_block_https={{ gw_block_https }}
 sshd_port={{ sshd_port }}
 #gui_wan= [no longer needed]
 gui_port={{ gui_port }}
-iiab_gateway_enabled={{ iiab_gateway_enabled }}
 block_DNS={{ block_DNS }}
 
 azuracast_ports="{{ azuracast_port_range_prefix }}000:{{ azuracast_port_range_prefix }}100"
--- a/roles/network/templates/gateway/iiab-internet-off
+++ b/roles/network/templates/gateway/iiab-internet-off
@@ -1,0 +1,3 @@
+#!/bin/bash
+sed -i -e "s/^IIAB_GATEWAY_ENABLED.*/IIAB_GATEWAY_ENABLED=False/" {{ iiab_env_file }}
+/usr/bin/iiab-gen-iptables
--- a/roles/network/templates/gateway/iiab-internet-on
+++ b/roles/network/templates/gateway/iiab-internet-on
@@ -1,0 +1,3 @@
+#!/bin/bash
+sed -i -e "s/^IIAB_GATEWAY_ENABLED.*/IIAB_GATEWAY_ENABLED=True/" {{ iiab_env_file }}
+/usr/bin/iiab-gen-iptables


holta:
Should we test this on Ubuntu 21.04 Beta?  I personally don't know how to do that on RPi 4 (until the OS is released, in exactly 10 days).

Testing on Ubuntu 21.04 Beta on a VM I can do, if that helps here?

Any specific tests I should run...on RaspiOS on RPi 4...and/or any other OS's that are critically important to test here?

holta:
This PR looks great, thanks to @jvonau who knows a lot about firewalls:

@georgejhunt can you weigh in with your opinion and/or testing recommendations?

(If this is tested to be a clear improvement on ~2 key OS's / layouts, I'm happy to merge this prior to reconfirming on Ubuntu 21.04 on RPi 4 when that's released in 10 days?)

jvonau:
> Should we test this on Ubuntu 21.04 Beta? I personally don't know how to do that on RPi 4 (until the OS released, in exactly 10 days).
> 

I used `sudo do-release-upgrade -d` on the 20.10 sever release, haven't tried with mate desktop.

> Testing on Ubuntu 21.04 Beta on a VM I can do, if that helps here?
> 
> Any specific tests I should run...on RaspiOS on RPi 4...and/or any other OS's that are critically important to test here?

Not really, using modprobe is contained to the running boot session, reboot clears the loaded modules added during this session and loads the saved rules without issue. I just want to ensure nothing suggests that any old iptables modules are loaded, so the rules are saved in the correct format.

holta:
>> Should we test this on Ubuntu 21.04 Beta?

> `sudo do-release-upgrade -d` on the 20.10 [server] release

Excellent idea.

tim-moody:
is iiab.env expected to change independent of local vars? is it set somewhere outside of an install?

holta:
Just FYI I've created a link from https://github.com/iiab/iiab/wiki/IIAB-Networking#firewall-iptables to this PR as the associated discussion appears very promising for IIAB implementers, i.e. those who want to know more/finer details about their [iptables](https://en.wikipedia.org/wiki/Iptables) firewall in general.

jvonau:
> is iiab.env expected to change independent of local vars?

There is currently a disconnect between what is recorded in iiab.env from install time and what could be currently present in local_vars now. Toggle iiab_gateway_enabled and run .iiab-network, note iiab.env does not currently update, but the rest of the logic is fine. We currently rewrite iiab-gen-iptables with the value from {{ iiab_gateway_enabled }} hard-coded, just moving the source of the variable to reference to the same location as the network interfaces used.

>  is it set somewhere outside of an install?

The default value is False and may not be what is wanted in the field, could be toggled in the field to suit the needs of the deployment, with documented usage received via fieldback of that layout being deployed.

georgejhunt:
I think there are two issues on the table in this PR.

   - I ran ubuntu-20.04 in a vbox yesterday (21.04 failed to come up).
   20.04 does have symbolic links from /bin -> /usr/bin, and /sbin ->
   /usr/sbin. This link became missing in my upgrade from 18.04 to 20.04 on
   NUC hardware. I'm convinced now that the failure of masquerade that I
   observed was an artifact of the upgrade and not a flaw in our networking.
   - The second issue, which Jerry has helped to clarify, was the evolution
   of iptables. I'll be happy if the older platforms (thinking RaspiOS)
   function properly, when we no longer load the legacy modules. (I think
   Jerry has found that they are loaded as needed). And the risk of an
   udetected long term failure due to this PR is low, because we test almost
   every day on Rpi.

jvonau:
So this kind of morphed into a new feature, how can I quickly (dis)allow access to the internet, answer sudo iiab-internet-on|off

holta:
Good idea.  I greatly support these 2 new commands:

- iiab-internet-on
- iiab-internet-off

Personally I'd much prefer these commands operate on /etc/iiab/local_vars.yml directly, avoiding all ambiguities (https://en.wikipedia.org/wiki/Single_source_of_truth).

Allowing `./iiab-install --reinstall` and similar commands to run with zero confusion.

jvonau:
> Personally I'd much prefer these commands operate on /etc/iiab/local_vars.yml directly, avoiding all ambiguities (https://en.wikipedia.org/wiki/Single_source_of_truth).
>
I'm thinking local_vars more as the default authoritative state when ansible runs, with the commands being local overrides with no need to run ansible again, besides you set the stage with iiab-hotspot-on|off as the quick down and dirty end user method, just a cookie cutter of hotspot.
 
> Allowing `./iiab-install --reinstall` and similar commands to run with zero confusion.

That would set the state of the variable in iiab.env based on what is in local_vars, that was the **only** way to reset what appeared in iiab.env as the only place iiab_gateway_enabled was recorded was in 1-prep, that is a bug by itself.

/iiab-install --reinstall gains what? repeating stage 1-3? and I thought that is not supported so why does "--reinstall" even enter the picture in this disscussion?

tim-moody:
There should absolutely be a single source of truth. The problem is that the truth of the matter is misunderstood. All ansible variables are directives, a desired state. After an install the state may or may not be what was desired. Unfortunately a number of years ago these were conflated by using enabled instead of enable as an ansible var. The distinction between install and installed is correct and there is a state file to show where things stand. I think it should be used for all state. I don't view iiab.env as state but rather as a convenience for sourcing in scripts.

jvonau:
> There should absolutely be a single source of truth. The problem is that the truth of the matter is misunderstood. All ansible variables are directives, a desired state. After an install the state may or may not be what was desired. Unfortunately a number of years ago these were conflated by using enabled instead of enable as an ansible var. The distinction between install and installed is correct and there is a state file to show where things stand. I think it should be used for all state. I don't view iiab.env as state but rather as a convenience for sourcing in scripts.

Other than displaying the contents of iiab.ini I'm not sure how or where any of the entries in iiab.ini come into use in admin-console(sans for iiab_network_mode), all of the 'services enabled' are hardcoded in 20-configure.html treated as effective_vars derived from local|default vars. Would recording "enabled" in iiab-state,yml take the place of recording in iiab.ini or in addition to? Do we want iiab.ini to go away?

holta:
Certainly if any Ansible commands (`./iiab-network` and/or legacy commands like `./iiab-install --reinstall`) unintentionally cause confusion (as seems very likely, among many implementers) then `iiab-internet-on|off` commands should spit out a (reasonably loud) on-screen warning.  These commands should probably print out (on screen) something like...

```
WARNING: If you want to _permanently_ change your IIAB's default behavior
(i.e. to specify whether student client devices should have Internet or not,
in general!) then modify variable 'iiab_gateway_enabled' in
/etc/iiab/local_vars.yml — and finally run 'cd /opt/iiab/iiab ; ./iiab-network'
```

georgejhunt:
Computers are pretty good at changing variable names. Maybe we should
blanket-change enabled->enable in the whole repo. Then "enabled" could be
written to /etc/iiab/iiab_state.yml whenever appropriate.

iiab.ini would become a declaration of intention at the time of the last
run of the playbook (and also a record thereof).

holta:
Complete + Total Aside _(meditations definitely for another day, before this PR gets stretched any further!)_

I do wonder if the distinction between `_install` and `_enabled` variables has become pragmatically almost irrelevant more than 14 years after OLPC XS began.  Compared to when OLPC XS started coming together in January 2007 (and in ~2012 when the move to Ansible/XSCE began, almost a decade ago), IIAB implementers now live in a quite different world with (very) high-speed Internet, when they are provisioning services (typically an LMS-like service).  So while enabling/disabling services in the field was a nice theory back in the day, and very much well-intended, it did not pan out in practice.  Field usage does not really follow that adoption model, i.e. the notion that every service needs to have an `_install` and an `_enabled` variable meant very well, but in hindsight appears very much unneeded in almost every case.

In those rare post-deployment cases where a service's state really does need to be flipped on-and-off, familiar tools like `systemctl enable|disable <SVC-NAME>.service` seem to handle that state (setting) extremely well.  In most every case (though of course IIAB had to tune up some of those services' more stale/aging systemd unit files!)

No doubt we're stuck with 2 variables when 1 probably would have sufficed (so be it, normally one should design for the common case, ah well now we know!)  But we should also be honest that implementers universally today want the very latest version of any service they're installing, from the live Internet (for sure this sometimes happens a few weeks after their initial IIAB install, using `./runrole`, e.g. after they discover a new LMS/service they didn't quite realize they needed in earlier days/weeks).  Ideally this installation of a new service[*] might one day be driven by Admin Console &mdash; e.g. for those implementers not up for `./runrole <SERVICE/APP>` at the command-line.

[*] Typically a core LMS offering, very often one of the "Big Three" Kolibri/Moodle/Wordpress, thanks to those having very strong communities-of-practice/documentation/affordances/etc.

holta:
Above [operator warning](https://github.com/iiab/iiab/pull/2727#issuecomment-818225383) cleaned up, in case @jvonau wants to use that verbiage or similar.

holta:
Is the following error msg concerning, when testing this PR on RaspiOS Lite on RPi 4?

```
root@box:~# iptables -L -v -t nat4
iptables v1.8.2 (nf_tables): table 'nat4' does not exist
Perhaps iptables or your kernel needs to be upgraded.
```

A reboot does not help.  Anything else I should be trying instead?

jvonau:
> Is the following error msg concerning, when testing this PR on RaspiOS Lite on RPi 4?
> 
> ```
> root@box:~# iptables -L -v -t nat4
> iptables v1.8.2 (nf_tables): table 'nat4' does not exist
> Perhaps iptables or your kernel needs to be upgraded.
> ```
> 
> A reboot does not help. Anything else I should be trying instead?

try without the 4 on the end of nat

holta:
> try without the 4 on the end of nat

Exactly.

`iptables -L -v -t nat` works fine, and all looks good on RaspiOS Lite on RPi 4.

jvonau:
> > try without the 4 on the end of nat
> 
> Exactly.
> 
> `iptables -L -v -t nat` works fine, and all looks good on RaspiOS Lite on RPi 4.

Not sure why you thought nat4 was in use, get to know the loaded modules involved with `lsmod`

holta:
> Not sure why you thought nat4 was in use, get to know the loaded modules involved with `lsmod`

Good to know.

`-t nat4` was just a typo.  `-t nat` is correct.

jvonau:
> Complete + Total Aside _(meditations definitely for another day, before this PR gets stretched any further!)_
> 
> I do wonder if the distinction between `_install` and `_enabled` variables has become pragmatically almost irrelevant more than 14 years after OLPC XS began. Compared to when OLPC XS started coming together in January 2007 (and in ~2012 when the move to Ansible/XSCE began, almost a decade ago), IIAB implementers now live in a quite different world with (very) high-speed Internet, when they are provisioning services (typically an LMS-like service). So while enabling/disabling services in the field was a nice theory back in the day, and very much well-intended, it did not pan out in practice. Field usage does not really follow that adoption model, i.e. the notion that every service needs to have an `_install` and an `_enabled` variable meant very well, but in hindsight appears very much unneeded in almost every case.
>

In your narrow view, distinction between install & enable is one of requiring internet access to retrieve source files, and editing configuration files. Yes you took the idea way farther than the just for optional roles as I intended, then lopped validation on top of that so users don't shoot themselves in the foot editing, don't give them knobs to turn and injury can't result. The field usage is directly related to the instructions given, the best advice is usually discarded, the end user picks 'big' anyway and becomes unhappy with the result on weak hardware. Being able to toggle services on/off in the field is the best way tailer the machine to the environment, just like setting "high limits", especially when "big" was the starting point. 
 
> In those rare post-deployment cases where a service's state really does need to be flipped on-and-off, familiar tools like `systemctl enable|disable <SVC-NAME>.service` seem to handle that state (setting) extremely well. In most every case (though of course IIAB had to tune up some of those services' more stale/aging systemd unit files!)
> 

To some degree, depends if your ok with either 502(backend not working) 404(not found) from nginx when you try the now non-working url. lol It's a moot point, depends on the users skill level while un-ticking a box seem pretty easy. 

> No doubt we're stuck with 2 variables when 1 probably would have sufficed (so be it, normally one should design for the common case, ah well now we know!) 

And the design chose for *_installed lends itself very well to the single "enable" to do both. The issue is "enable True" implies "install True" but "enable False" does not automatically imply "install False" once the app/service is installed. Keep in mind this process is not a one shot wonder.

> But we should also be honest that implementers universally today want the very latest version of any service they're installing, from the live Internet (for sure this sometimes happens a few weeks after their initial IIAB install, using `./runrole`, e.g. after they discover a new LMS/service they didn't quite realize they needed in earlier days/weeks). Ideally this installation of a new service[*] might one day be driven by Admin Console — e.g. for those implementers not up for `./runrole <SERVICE/APP>` at the command-line.
> 

Just expose *_install in /opt/admin/console/htmlf/20-configure.html, presenting the list of roles based on *_install, having the checkbox pre-populated is pretty straight forward, I have that working. Hiding what is already checked so the user can't try the equivalent of "install False" on the already installed role thus only presenting a list of "new" apps/services is a work in progress but very do-able just not high on my list. Seems like "--reinstall" could be plumbed in also, that is just removing *_installed from iiab_state before calling "ICO"

Now only if some versioning was present, so the delta change could be applied automatically that was on the table but ignored, and why this PR even exists I'm not sure but this will be my last.

holta:
> Do we want [iiab.env similar to iiab.ini] to go away?

Just a note that command `source /etc/iiab/iiab.env` is used in several IIAB scripts and has led to serious bugs in the past, when some of those 11-to-15 variables collided with same-named variables in the script at hand.  If I recall correctly, this happened with the `OS` var.

root regularly running `source /etc/iiab/iiab.env` could lead to more serious security risks too.

Certainly `/etc/iiab/iiab.env` contains a lot of unnecessary duplication, similar to `/etc/iiab/iiab.ini` — but we're quite wedded to both files for the moment.  So we should likely live with these for now.

jvonau:
> > Do we want [iiab.env similar to iiab.ini] to go away?
> 
Are you for real? iiab.env is the bases for the staged install, where do you think STAGE= is stored? Suppose the value could be stored in iiab_state but I'm not going down that path.

> Just a note that command `source /etc/iiab/iiab.env` is used in several IIAB scripts and has led to serious bugs in the past, when some of those 11-to-15 variables collided with same-named variables in the script at hand. If I recall correctly, this happened with the `OS` var.

OS= is derived with OS={{ ansible_local.local_facts.os }}. What are you referring to?
 
> root regularly running `source /etc/iiab/iiab.env` could lead to more serious security risks too.
> 
Doing anything as root is dangerous, what is your point?

> Certainly `/etc/iiab/iiab.env` contains a lot of unnecessary duplication, similar to `/etc/iiab/iiab.ini` — but we're quite wedded to both files for the moment. So we should likely live with these for now.

As Tim pointed iiab.env is a "convenience for sourcing in scripts." and I pointed out is currently the holder of the stage information.

holta:
> > Just a note that command `source /etc/iiab/iiab.env` is used in several IIAB scripts and has led to serious bugs in the past, when some of those 11-to-15 variables collided with same-named variables in the script at hand. If I recall correctly, this happened with the `OS` var.
> 
> OS= is derived with OS={{ ansible_local.local_facts.os }}. What are you referring to?

This flared up in /opt/iiab/iiab/scripts/local_facts.fact if I'm remembering correctly.  Leading to erroneous output.

Anyway it's now fixed, no worries.

While iiab.env and iiab.ini each contain quite a bit of duplication, so be it.  Both are good enough for now (in my opinion).

Should they be revised or trimmed down at any time in future, we should just do our best to have a "Single Source Of Truth" (SSOT) wherever that's practical.

holta:
Public-facing documentation is now cleaner here:

https://github.com/iiab/iiab/wiki/IIAB-Networking#firewall-iptables

Recap after 4 months:

Thanks to everyone who contributed, and especially @jvonau who did the bulk of the actual work here, to figure this out and bring it together efficiently.

jvonau:
Four months? This feature request dates back to 2018 https://github.com/iiab/iiab/issues/796#issuecomment-389165271 with 0 not being implemented till now, guess my "0-7 above have been achieved." in 2020 was a bit premature, thought somebody with commit access would of taken the lead then, I was wrong.

PR #2728, opened 2021-04-13
PR #2727 w/ on-screen warnings for iiab-internet-on|off

holta:
WIP, refining PR #2727.

### DIFF
--- a/roles/network/tasks/computed_network.yml
+++ b/roles/network/tasks/computed_network.yml
@@ -141,6 +141,13 @@
     line: 'IIAB_LAN_DEVICE={{ iiab_lan_iface }}'
     state: present
 
+- name: Record 'IIAB_GATEWAY_ENABLED={{ iiab_gateway_enabled }}' in {{ iiab_env_file }}
+  lineinfile:
+    path: "{{ iiab_env_file }}"
+    regexp: '^IIAB_GATEWAY_ENABLED=*'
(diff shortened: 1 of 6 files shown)


holta:
Tested on RaspiOS Lite on RPi 4.

Thanks goes to @jvonau as this is just PR #2727 with a few cleanups.

holta:
FYI this was also tested on Ubuntu Server 20.10 on RPi 4.

holta:
> Thanks goes to @jvonau as this is just PR #2727 with a few cleanups.

Documentation is now cleaner here:

https://github.com/iiab/iiab/wiki/IIAB-Networking#firewall-iptables

jvonau:
The author of eeefe546e7f270c1afa74fd431e3513072a3b788 fails to follow form that was started from that start of this project from a time before the author became involved with ansible. Templates that have variables within them should have.j2 appended to the name to denote the use of variables, this automatically requires the use of full dest directives to be able to change the name, ie dropping the .j2 I chose the form I did because that was consistent with all the other roles use of template where the source file have .j2 appended to denote the use of variables but j2 needs to be removed when place on the target filesystem. It's actions like this that drive myself away. 

Should mention that the change will persist through a reboot until the opposite command is run as the rule set was changed and saved.

holta:
> Should mention that the change will persist through a reboot until the opposite command is run as the rule set was changed and saved.

Done.  That explanation is now added here:

https://github.com/iiab/iiab/wiki/IIAB-Networking#firewall-iptables

Issue #2721, opened 2021-04-06
ansible-core 2.11 RC1 & 2.11.0 (formerly known as ansible-base) have been tested successfully with IIAB on RaspiOS, Ubuntu 20.04.2 & 21.04


holta:
In short, `ansible-base` is being renamed to `ansible-core` so this quick change will shortly be made in these 3+ places:

- [scripts/ansible](https://github.com/iiab/iiab/blob/master/scripts/ansible)
- [.travis.yml](https://github.com/iiab/iiab/blob/master/.travis.yml)
- http://FAQ.IIAB.IO & related documentation

FYI ansible-core 2.11 spits out a colorful-but-harmless warning that Python < 3.8 will no longer be supported starting with ansible-core 2.12 (ETA very late 2021).  Good to know (e.g. with RaspiOS currently using Python 3.7.3) but this won't really affect IIAB as [all of our mainstream OS's](https://github.com/iiab/iiab/wiki/IIAB-Platforms#operating-systems) are either already on Python 3.8+ (Ubuntu 20+, Mint 20) or thankfully will be soon, well before then (RaspiOS 11, Debian 11).

Refs:

- #2659 (shows evolving version numbers of latest Ansible Collections required by IIAB)
- PR #2689 "Update & tighten up .travis.yml from Trusty (Ubuntu 14.04) + Python 2.7 to Focal (Ubuntu 20.04) + Python 3.8"
- PR #2690 "Install bionic-or-focal version of ansible-base, depending on OS"

holta:
Good news Ansible's growing up aiming to be more maintainable/sustainable — with this latest ansible-core 2.11 promising:

> "Role arguments can now be validated at runtime to ensure that roles are used correctly"

https://groups.google.com/g/ansible-devel/c/-hAc5RiGGiU

Ansible Documentation is still somewhat scattershot, but is emerging in places like:

- https://github.com/ansible/ansible/blob/stable-2.11/changelogs/CHANGELOG-v2.11.rst
- https://docs.ansible.com/ansible-core/devel/porting_guides/porting_guide_core_2.11.html somewhat related to...
  - https://docs.ansible.com/ansible/devel/porting_guides/porting_guide_4.html
  - https://toshio.fedorapeople.org/ansible/devel/porting_guides/porting_guide_base_2.11.html (obsolete with old name etc!)

holta:
ansible-core 2.11 RC2 is already out...which is mostly about fixing a showstopper bug with Python 3.5:

[...]

PR #2729, opened 2021-04-13
Correctly record Ansible versions like "[core 2.11.0rc2]" (e.g. when installing pre-releases using 'pip3 install ansible-core')

holta:
Tested on Ubuntu 20.04 with both of these Ansible versions:

- 2.10.8
- [core 2.11.0rc2]

More testing ongoing.

Ref: #2721


holta:
FYI the single whitespace on the right side...could be trimmed below if it really mattered...if we really cared:

```
root@box:~# scripts/local_facts.fact
...
"ansible_version"         : "[core 2.11.0rc2] ",
```

I don't think this trivial visual blemish matters in any way.

(In any case, this PR solves the core issue, making Ansible's version readable now!)

PR #2731, opened 2021-04-14
sysd-netd-debian.yml typo in comment?

holta:
@jvonau / others can confirm.

### DIFF
--- a/roles/network/tasks/sysd-netd-debian.yml
+++ b/roles/network/tasks/sysd-netd-debian.yml
@@ -62,7 +62,7 @@
     name: systemd-networkd
     state: restarted
 
-- name: Restart hostapd when WiFi is present but not when using WiFi as gateway$
+- name: Restart hostapd when WiFi is present but not when using WiFi as gateway
   systemd:
     name: hostapd
     state: restarted


jvonau:
Using ''blame" it's easy to deduce which commit changed the line, a95d4762874faac485f2c261eb347d86cf8f4ce5. The $ looks like a truncated line from a copy and paste situation where the line is longer than than what was shown in the terminal.

jvonau:
The full line was "Restart hostapd when WiFi is present but not when using WiFi as gateway with wifi_up_down False"

holta:
> The full line was "Restart hostapd when WiFi is present but not when using WiFi as gateway with wifi_up_down False"

Ok.
I'll assume this exact same language applies to (what's now) Line 71 of https://github.com/iiab/iiab/blob/master/roles/network/tasks/rpi_debian.yml#L71

jvonau:
> > The full line was "Restart hostapd when WiFi is present but not when using WiFi as gateway with wifi_up_down False"
> 
> Ok.
> I'll assume this exact same language applies to (what's now) Line 71 of https://github.com/iiab/iiab/blob/master/roles/network/tasks/rpi_debian.yml#L71

That is why I pointed out the commit to begin with and to suggest what the line was.

Issue #2739, opened 2021-04-18
Stop playing games with milestones


jvonau:
In September, 2020 #2531 - "Structural merges at the end of a release cycle are Not Ok", was pushed and I noticed that https://github.com/iiab/iiab/milestones now carries a new 'Expected Q2 2021' entry. What gives?
![0417](https://user-images.githubusercontent.com/5579710/115133387-6cf46480-9fcd-11eb-8453-f90745b9362f.png)

jvonau:
#2507

jvonau:
https://github.com/iiab/iiab/pull/1743#issuecomment-503822834

jvonau:
#2737

jvonau:
#1849

jvonau:
#1656

jvonau:
I'll close this when the linked issues are closed dating back to 2018

PR #2741, opened 2021-04-18
Make stage 6 less crowded - identify apps that have their own updater & Cleanup

jvonau:
First commit repurposes stage 5 as a note for roles that should **never** be reinstalled as each contain their own builtin updater.
Second commit removes mention of roles that are not present in /roles and are skipped in validation, this would cause an error if contained in local_vars. No functional change, grouping is for better info for the end-user and future developers, 2 roles move from stage 6 to stage 5. This would be on the lowest rung of release maintenance, with higher stages requiring upkeep of the current versions to install being a difference to note.


holta:
There are a lot of good ideas here.

Just to mention another idea I'm seeing a _lot_ of "During Covid".  This is the idea that an increasing number of NGO's want a "Major LMS" as part of IIAB — which almost always means one of the following:

- Kolibri (though smaller NGO's are still happy with KA Lite)
- Moodle
- WordPress

Whereas these 3 other LMS-like environments are certainly used occasionally, for somewhat similar purposes, but are just not in the same league:

- Calibre-Web
- Nextcloud
- Sugarizer

We need a better nickname than "Minor LMS's" but certainly the above taxonomy has proved extremely valuable in 2020 and 2021 and will likely become even more significant going forward.

It's not to say we absolutely need to speak the language of our customers e.g. sometimes internal infra will just not correspond to UX perceptions.  But it is to say NGO/community/adoption patterns are changing quite dramatically, much accelerated by things like Covid...

jvonau:
Think of this grouping as stuff in stage 6 and above requires more maintenance, while stage 5 is set the initial version used in the install and you are done.

jvonau:
I haven't looked at the local_vars_* files for adjustments yet.

jvonau:
ping.. 'question" tag was added, what is the question?

holta:
> ping.. 'question" tag was added, what is the question?

Design questions as to how to cluster stages 5 to 9 will be numerous this year.

On the one hand there's a strong argument that Stages 6-to-8 are very arbitrary.

On the other hand, there's a very strong argument that they should stay that way, until the dominant needs of implementers converge much more clearly.

In any case, this PR is absolutely a useful start to kicking off and eventually deciphering these complicated, rapidly evolving questions.

(Aside: several of these questions are 'artistic' but that does not mean they are culturally insignificant, e.g. bricklayers consider their work to be a sacred artform...be that as it may, we'll need to tease out diverse implementation trends over the coming decade...of course without succumbing to sectarian religiosities.)

jvonau:
> > ping.. 'question" tag was added, what is the question?
> 
> Design questions as to how to cluster stages 5 to 9 will be numerous this year.
> 
> On the one hand there's a strong argument that Stages 6-to-8 are very arbitrary.
> 
> On the other hand, there's a very strong argument that they should stay that way, until the dominant needs of implementers converge much more clearly.
>
With 7.2 due in 2021Q2 (moved from 2020Q3) you should make up your mind fast.

> In any case, this PR is absolutely a useful start to
> kicking off and eventually deciphering these complicated, rapidly evolving questions.
>
 
Where is this tracking issue for the design discussion? 
> (Aside: several of these questions are 'artistic' but that does not mean they are culturally insignificant, e.g. bricklayers consider their work to be a sacred artform...be that as it may, we'll need to tease out diverse implementation trends over the coming decade...of course without succumbing to sectarian religiosities.)

More dribble, smoke-screening, hand wavy BS... Stop trying to gaslight me, you will not be successful.

jvonau:
Now in 2021Q3 with no release in site

jvonau:
#2739 #2817

PR #2746, opened 2021-04-22
Update ubuntu-21.yml [for PostgreSQL 13]

jvonau:
shipping 13


jvonau:
It's actually a BUGFIX as moodle is rarely installed and not well tested.

holta:
In any case, the jump to PostgreSQL 13 was long-anticipated and documented in ubuntu-21.yml, so glad it's now official.

PR #2751, opened 2021-04-26
WIP: Lint & explain lesser-known packages in 2-common/tasks/packages.yml

holta:
Making IIAB more approachable to newcomers & outsiders who are less familiar.

PR #2693, opened 2021-02-25
Debian mate

jvonau:
### Fixes bug:
br0 handling in #2658 
### Description of changes proposed in this pull request:
enable systemd-networkd and install systemd-dispatcher for Debian 11 when NetworkManager is installed usually with a desktop to support br0, similar to Ubuntu desktops' layout that work fine just expanded coverage now.

PR #2752, opened 2021-04-26
replace 'brctl' with 'bridge'

jvonau:
### Fixes bug:
https://github.com/iiab/iiab/pull/2751/files#diff-0357db829e9ddded747840cd3375bfa85cded3a248d4aabda868012542f01d35R34


jvonau:
There is a packaging bug with iproute2
>ubuntu@box:\~$ which bridge
/usr/sbin/bridge
ubuntu@box:\~$ dpkg -S /usr/sbin/bridge 
dpkg-query: no path found matching pattern /usr/sbin/bridge
ubuntu@box:\~$ dpkg -S /sbin/bridge 
iproute2: /sbin/bridge

'bridge' is on the path so should be fine.

https://wiki.linuxfoundation.org/networking/net-tools
net-tools should go away also but there are legacy commands in https://github.com/iiab/iiab/blob/master/scripts/iiab-diagnostics#L181-L206 that some developers might miss.

holta:
Do we care that the new command is a bit different from the old command?  I don't know.  But a quick observation on Ubuntu 21.04 VM with IIAB:

```
root@box:~# brctl show br0
bridge br0 does not exist!
root@box:~# echo $?
1

root@box:~# bridge -d link
root@box:~# echo $?
0
```

Maybe slight changes (like the above) don't really matter...for https://github.com/iiab/iiab/pull/2752/files#diff-20015410c1c1ee6a49d6a514b3b4049493e516ebae40b0dea32dc9c411aa97feR68 ?

```
shell: bridge -d link | grep {{ iiab_wireless_lan_iface }} | wc -l
```

Aside: is the use of `brctl` on Line 15 of https://github.com/iiab/iiab/blob/master/roles/network/templates/network/br0.j2#L15 still ok, when apt package `bridge-utils` is no longer being installed?

```
    pre-up ip link set br0 down && brctl delbr br0
```

holta:
I guess the same question might apply to...Line 76 of iiab-network here: (https://github.com/iiab/iiab/blob/master/iiab-network#L76)

```
brctl show >> iiab-network.log
```

Should this become something like...?

```
bridge -d link >> iiab-network.log
```

jvonau:
sure whatever you are going to rewrite the PR anyway so just take this as a pointer, now ask your self if br0.j2 is used at all.

holta:
> if br0.j2 is used at all

Good point.

We should move https://github.com/iiab/iiab/blob/master/roles/network/templates/network/br0.j2 to `br0.j2.deprecated` as it's no longer in use, according to `grep -r br0.j2 /opt/iiab/iiab`

Ref: PR #2751

jvonau:
> Do we care that the new command is a bit different from the old command? I don't know. But a quick observation on Ubuntu 21.04 VM with IIAB:
> 
> ```
> root@box:~# brctl show br0
> bridge br0 does not exist!
> root@box:~# echo $?
> 1
> 
> root@box:~# bridge -d link
> root@box:~# echo $?
> 0

The shell code doesn't run when the correct environment is declared, note the conditional when a couple of lines down that requires br0 being present to prevent the noted "does not exist" and is filtered by the device name used with hostapd, returns 1 when present 0 otherwise. 
> 
> Maybe slight changes (like the above) don't really matter...for https://github.com/iiab/iiab/pull/2752/files#diff-20015410c1c1ee6a49d6a514b3b4049493e516ebae40b0dea32dc9c411aa97feR68 ?
> 
> ```
> shell: bridge -d link | grep {{ iiab_wireless_lan_iface }} | wc -l
>

The need for the conditional mentioned might go away as there is no real error if br0 is not present using `bridge`.

holta:
It's possible @deldesir might be interested in helping with this, if there aren't too many corner cases to verify (making sure he's testing in useful ways if so?)

holta:
@deldesir, if you want to test this, install IIAB as follows:

```
curl d.iiab.io/install.txt | sudo bash -s 2752
```

Preferably on a Raspberry Pi, i.e. a machine with at least 2 network interfaces (one of which is a working internal WiFi hotspot!)

jvonau:
The identification of the required network related packages could become clearer with the rename proposed in #2693? ie moved from packages.yml to network.yml.

jvonau:
testing for 2693 is in a "confirm not no harm" status...

holta:
@deldesir says he has a laptop where he may install Debian 11 RC1 (Desktop version of "Bullseye") this week, if that helps things like PR #2693 and/or #2752 here.

jvonau:
@deldesir same laptop as #2582 I presume? but don't use `curl d.iiab.io/install.txt | sudo bash -s 2752` there is a conflict that needs to be resolved.

holta:
> @deldesir same laptop as #2582 I presume?

I assume so, but hopefully @deldesir will respond here.

jvonau:
`for i in arp ifconfig ipmaddr iptunnel nameif mii-tool; do grep -r $i /opt/iiab/iiab/roles; done` pointed out https://github.com/iiab/iiab/blob/master/roles/captiveportal/templates/capture-wsgi.py.j2#L396 is using net-tools commands.

This PR has been updated, but discussion can continue about removing net-tools.

PR #2761, opened 2021-04-29
cups - drop apache

jvonau:
The port is open anyway, just lose the /cups redirect, not worth dragging apache in. Still
missing the nginx config to make apache work, nginx should really just use nginx directly.

### Fixes bug:
Incomplete nginx port
Just access via box:631vs box/cups, should /cups be wanted a nginx file could be crafted but to date there is no interest in doing so.

### DIFF
--- a/roles/cups/tasks/enable-or-disable.yml
+++ b/roles/cups/tasks/enable-or-disable.yml
@@ -1,11 +1,3 @@
-- name: Enable http://box/cups via Apache (MIGHT NOT WORK?)
-  command: a2ensite cups.conf
-  when: cups_enabled
-
-- name: Disable http://box/cups via Apache
-  command: a2dissite cups.conf
-  when: not cups_enabled
-
 - name: systemd daemon-reload
   systemd:
     daemon_reload: yes
--- a/roles/cups/tasks/install.yml
+++ b/roles/cups/tasks/install.yml
@@ -1,13 +1,3 @@
-- name: "Set 'apache_install: True' and 'apache_enabled: True'"
-  set_fact:
-    apache_install: True
-    apache_enabled: True
-
-- name: APACHE - run 'httpd' role
-  include_role:
-    name: httpd
-
-
 - name: Install 'cups' package
   package:
     name: cups
@@ -17,12 +7,6 @@
   template:
     src: cupsd.conf
     dest: /etc/cups/cupsd.conf
-
-- name: Install /etc/{{ apache_conf_dir }}/cups.conf from template
-  template:
-    src: cups.conf
-    dest: "/etc/{{ apache_conf_dir }}/"
-
 
 # RECORD CUPS AS INSTALLED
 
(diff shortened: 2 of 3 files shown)


holta:
@georgejhunt merging this seems very safe, but can you confirm?

Many thanks to @jvonau for pruning -- which is often the hardest work in the entire garden.

georgejhunt:
I think making http://box.lan/cups work should be part of the job.

holta:
> I think making http://box.lan/cups work should be part of the job.

Do you know how to do that?

Possibly with a synonym like http://box.lan/print ?

jvonau:
> I think making http://box.lan/cups work should be part of the job.

This is not a job, it's a contribution. Currently http://box/cups does not work as is, so there is no loss of current functionality, just removing the needless apache dependency from the equation.

georgejhunt:
Adam maybe this would help:
https://anthe.studio/blog/en/cups-nginx-reverse-proxy

jvonau:
I see no point in dragging in a "requirement" only to write/refine a nginx config file to remove the requirement. All this PR does open the door to install cups without apache, so a nginx config file could be worked on whenever in the future by some interested party. http://box:631 always is available for use in the meantime, seems like low hanging fruit but some new blood needs to take over.

holta:
> Adam maybe this would help: https://anthe.studio/blog/en/cups-nginx-reverse-proxy

@deldesir: I don't know if you care about NGINX?

But if so see the above, as well as status of IIAB's Apache -> NGINX transition begun in late 2019 &mdash; but still not quite entirely done:

- https://github.com/iiab/iiab/blob/master/roles/nginx/README.md
- #2762 "Notes: Stalled progress on nginx front"

holta:
On our IIAB weekly call (http://minutes.iiab.io) we observed that https://github.com/iiab-share/js-menu-files/blob/master/menu-defs/en-cups.json (and its logo https://github.com/iiab-share/js-menu-files/blob/master/images/CUPS.svg) are not auto-added to IIAB's main page http://box.lan (e.g. when a BIG-sized IIAB is installed).

@tim-moody says this appears to be a bug &mdash; I ran `/usr/bin/iiab-update-menus` to try to update `/library/www/html/home/menu.json` which should have worked (?) based on this list of things that should be auto-added:

https://github.com/iiab/iiab-admin-console/blob/master/roles/common/templates/adm_const.py.j2#L50-L67

Just FYI/FWIW these are the 20 that did appear in home/menu.json this morning:

```
  "menu_items_1": [
    "en-oob",
    "en-sample",
    "en-test_zim",
    "en-test_mod",
    "en-map_admin",
    "en-osm_viewer_v2",
    "en-usb",
    "en-elgg",
    "en-lokole",
    "en-mediawiki",
    "en-nodered",
    "en-nextcloud",
    "en-wordpress",
    "en-kalite",
    "en-kolibri",
    "en-moodle",
    "en-sugarizer",
    "en-internetarchive",
    "en-calibreweb",
    "en-credits"
  ]
```

Issue #2748, opened 2021-04-22
Testing of new 64-bit Raspberry Pi OS images from 2021-04-09


holta:
They are now downloadable here:

https://downloads.raspberrypi.org/raspios_lite_arm64/images/
https://downloads.raspberrypi.org/raspios_arm64/images/

Refs:

- #2422 "Test IIAB 7.2+ on 64-bit Raspbian (renamed to Raspberry Pi OS / RaspiOS) as this becomes more established? [UH-OH: WiFi firmware 7.45.18.0 not quite compatible; will a new kernel help?]"
- #2620 "Any guess how much faster (or slower!) 64-bit Raspberry Pi OS (RaspiOS) is?"
- #2744 "Testing Ubuntu 21.04 on (RPi+PC) x (Server+Desktop). Does WiFi firmware 7.45.18.0 work on RPi?"

holta:
FYI this builds off #2422 testing.

We can close that ticket as nec.

@deldesir you may not have time to work on this flood of so many eccentric OS's, but do consider those that might be strategic...

holta:
A BIG-sized IIAB install just completed on the newer 2021-04-09 **64-bit** RaspiOS Lite on RPi 4 at 10.8.0.38

FWIW I used the released version of ansible-core 2.11.0, installing IIAB as follows:

```
curl d.iiab.io/install.txt | sudo bash -s 2743
```

Ref: PR #2743

holta:
> A BIG-sized IIAB install just completed on the newer 2021-04-09 64-bit RaspiOS Lite on RPi 4 at 10.8.0.38

Preliminary But Very Promising News: last year's 10X-to-100X slowdown of pages loading over IIAB's WiFi hotspot...are not happening so far...when using high-capacity WiFi firmware 7.45.18.0 installed on the newer 2021-04-09's 64-bit RaspiOS **Lite** @ 10.8.0.38

Still To Be Tested: 2021-04-09's 64-bit RaspiOS **with Desktop** @ 10.8.0.50 (IIAB is currently being installed there for PR #2693).

BACKGROUND: 2020's buggy slowdown of WiFi is summarized under "Raspberry Pi OS (64-bit" at https://github.com/iiab/iiab/wiki/IIAB-Platforms and is pasted in here:

[...]

Issue #2658, opened 2020-12-06
IIAB 7.2 Testing with Debian 10.9+ [laptop, NetworkManager / networking questions]


shanti-bhardwa:

shanti-bhardwa:
testing Debian 10.7 with DELL lattitude E6230, i5 4 GB of RAM, 500GB Drive

Image: debian-live-10.7.0-amd64-mate.iso

The .iso image installs, but does not have wifi firmware

I had to install it separately as root with following commands.

echo “deb http://ft[.se.debian.org/debian buster non-free contrin” >> etcapt/sources.list

apt-get update
apt-get install firmware-iwlwifi

apt reboot

Then apt-get install curl

The run the the 1 line install 

IIAB medium installed without any issues

but cannot use the SSID as these two services do not run

networking-service & iiab-dnsmasq

main reason seem to be br0 not up

get Failed to start network interfaces at boot time

I recall we had this problem before when testing Ubuntu or may be I don’t have the correct WIFI drives

see http://spruge.us/FB1m4I?en

shanti-bhardwa:
sorry closed in error

holta:
> echo “deb http://ft[.se.debian.org/debian buster non-free contrin” >> etcapt/sources.list

echo “deb http://ftp.se.debian.org/debian buster non-free contrib” >> /etc/apt/sources.list

> see http://spruge.us/FB1m4I?en

http://sprunge.us/FB1m4I?en

jvonau:
This is sort of like Ubuntu, NetworkManager is used but I'm sure systemd-networkd is present as networkctl responds, lets just do the same as netplan.yml does and use systemd-networkd for br0. Give this branch a test spin:
`cd /opt/iiab/iiab/`
`sudo git checkout -b jv-test`
`sudo git pull https://github.com/jvonau/iiab debian_mate`
`sudo ./iiab-network`
`sudo reboot`

holta:
Thank you @shanti-bhardwa if you can help test @jvonau's recommendation just above.

shanti-bhardwa:
tested as above. iiab-dnsmaq service starts, br0 is up but networking.service does not start up, Status shows 

failed to bring up eth0

when in fact etho is UP and has an IP address assigned by my router.

root@box:~# systemctl status networking
● networking.service - Raise network interfaces
   Loaded: loaded (/lib/systemd/system/networking.service; enabled; vendor preset: enabled)
[...]

PR #2769, opened 2021-04-30
rename current_host_channel -> current_client_channel and record

jvonau:
This records the channel that the upstream SSID is using for access to the internet if connected via WiFi. Helpful to note when installing over wifi as 2.4 or 5 Ghz could be used but we only offer 2.4 for the hotspot. Tested on U21.04 with both wlan0 and ap0 active with no wired cable present.


jvonau:
Testable now.

jvonau:
Use case from https://github.com/iiab/iiab/issues/2658#issuecomment-741520350
> > They stop when I do iiab-hotspot-off and I can connect to upstream WIFI Router.

> With the machine in that state can you grab sudo /usr/sbin/iw dev and sudo systemctl status systemd-networkd posting the results in this issue please. Think your upstream wifi might be a 5GHz connection, the channel in use should tell the tale.

With dual mode on 2.4Ghz with the hostapd channel being adjusted to match what is in use this is a example output:
host_channel = 11
client_wifi_channel = 11

This would be an example of a 5Ghz upstream connection being detected and hostapd not being adjusted.
host_channel = 6
client_wifi_channel = 44

holta:
Tested on 183-rpi4-64bit-lite-BIG-2743 = 10.8.0.38 using:

```
cd /opt/iiab/iiab
git checkout -b jvonau-wifi_info master
git pull https://github.com/jvonau/iiab.git wifi_info
./iiab-network
```

Then I browsed around using http://box (clicking the links on the main page there) using an Android 11 WiFi client.

What additional testing is important, if any?

Ref: #2748

holta:
> Tested on 183-rpi4-64bit-lite-BIG-2743 = 10.8.0.38

[iiab-diagnostics](https://github.com/iiab/iiab/blob/master/scripts/iiab-diagnostics.README.md) output if it helps: http://sprunge.us/rWFSst?en

Issue #2766, opened 2021-04-30
NGINX advice on making http://box/cups and http://box/print work?


holta:
If a volunteer can be found to help here, @georgejhunt suggests this is a pretty easy job!  If one follows https://anthe.studio/blog/en/cups-nginx-reverse-proxy ?

Background:

- PR #2761 "cups - drop apache"
- https://github.com/iiab/iiab/tree/master/roles/cups
- https://github.com/iiab/iiab/tree/master/roles/nginx (README.md)

Ref:

- #2762 "Notes : Stalled progress on NGINX front"

deldesir:
Good. I'll do it.

deldesir:
See https://github.com/iiab/iiab/pull/2775.

holta:
> See #2775.

Let's finalize this one way or another.  Thanks @deldesir

holta:
Good Enough for now?

http://box/print and http://box/print/ now work in all cases, thanks to PR #2858 inspired by PR #2775.

(@deldesir thanks for pushing to make this to happen.  Please re-open this ticket — or any other — to push this or any similar further!)

Issue #2762, opened 2021-04-29
Notes: Stalled progress on nginx front


jvonau:
Just notes
>ubuntu@box:/opt/iiab/iiab$ grep "Set 'apache_install: True'" -r roles
roles/elgg/tasks/install.yml:- name: "Set 'apache_install: True' and 'apache_enabled: True'"
roles/moodle/tasks/install.yml:- name: "Set 'apache_install: True' and 'apache_enabled: True'"
roles/pbx/tasks/install.yml:- name: "Set 'apache_install: True' and 'apache_enabled: True'"
roles/lokole/tasks/install.yml:- name: "Set 'apache_install: True' and 'apache_enabled: True'"
roles/phpmyadmin/tasks/install.yml:- name: "Set 'apache_install: True' and 'apache_enabled: True'"
roles/nodered/tasks/install.yml:- name: "Set 'apache_install: True' and 'apache_enabled: True'"

Solutions in age of appearance

moodle - [moodle-nginx.conf.j2.native](https://github.com/iiab/iiab/blob/master/roles/moodle/templates/moodle-nginx.conf.j2.native) from day one of nginx to try and should work, nobody to champion the cause [upstream usage](https://github.com/moodlebox/moodlebox/blob/master/roles/webserver/templates/etc/nginx/sites-available/default.j2)
nodered - #2154 very basic layout self tested to splash, nobody to champion the cause 
lokole - #2507 has links, nobody to champion the cause
phpmyadmin - pending PR #2712

Pending moodle feedback, should point to a solution for these:
elgg - might be handled like nextcloud or moodle-nginx.conf.j2.native?
pbx - no nginx shim, might be handled like nextcloud or moodle-nginx.conf.j2.native?

Shame all install ok, apache could go way if the effort was spent.

holta:
https://github.com/iiab/iiab/blob/master/roles/nginx/README.md should link to this ticket.

jvonau:
http://learn.elgg.org/en/stable/intro/install/nginx.html

holta:
FYI the painstaking progress of removing Apache (legacy dependencies, etc) is also being documented here:

- [roles/3-base-server/tasks/main.yml#L11-L18](https://github.com/iiab/iiab/blob/master/roles/3-base-server/tasks/main.yml#L11-L18)

holta:
[...]

PR #2775, opened 2021-05-04
Reverse Proxy Cups with Nginx

deldesir:
### Fixes bug: 
CUPS is now accessible via http://box.lan/print/

Created nginx.yml task and corresponding templates

Ubuntu 20.10

@holta


holta:
Awesome.

@georgejhunt can you review this PR ?

Ref: #2766

tim-moody:
@deldesir thanks for your help on this. There's more where it came from.

deldesir:
> 
> 
> @deldesir thanks for your help on this. There's more where it came from.

Yes, you're right @tim-moody. The "additional" directives (the WebSocket part) from https://anthe.studio/blog/en/cups-nginx-reverse-proxy that I choosed to ignore would throw "Unable to encrypt connection: The TLS connection was non-properly terminated". I don't have any idea right now why it won't work with them. I also choose not to disable Cups authentication because I didn't notice any pages fail to load or load infinitely as stated in the aforementioned article. I'd be glad to learn from any improvements anyone could add to it.


Here is the output of /var/log/cups/error.log : https://paste.ubuntu.com/p/B7FSdfw3Nj/

tim-moody:
@deldesir IIAB doesn't use TLS (https) because self-signed certificates cause the browser to tell users not to use the site. I'm not sure if web sockets are the same, but web workers don't work unless there is TLS.

jvonau:
>E [04/May/2021:16:16:35 -0400] Hostname lookup for "{{ lan_ip }}" failed.
E [04/May/2021:16:16:35 -0400] Bad Listen address {{ lan_ip }}:631 at line 4.

{{ lan_ip }}:631 something is wrong with the cupsd.conf template that should be an ip address
>E [04/May/2021:16:16:35 -0400] Unable to open listen socket for address [v1.::1]:631 - Cannot assign requested address.

Looks to be a ipv6 notation, think that is because ipv6 is disabled globally.

holta:
> {{ lan_ip }}:631 something is wrong with the cupsd.conf template that should be an ip address

@deldesir can you look into this before this PR gets merged?

(Nobody is asking for perfection, but better to fix things now if they turn out to be fixable!)

deldesir:
> Looks to be a ipv6 notation, think that is because ipv6 is disabled globally.

Not an elegant solution, but I would suggest to comment out L4 in cupsd.conf for now since IIAB focus on IPv4.

holta:
> Not an elegant solution, but I would suggest to comment out L4 in cupsd.conf for now since IIAB focus on IPv4.

Does that also solve the following?

> E [04/May/2021:16:16:35 -0400] Hostname lookup for "{{ lan_ip }}" failed.
> E [04/May/2021:16:16:35 -0400] Bad Listen address {{ lan_ip }}:631 at line 4.

(Apologies I don't know where @jvonau's above 2 lines come from, but hopefully you or he can shed light to help others understand.)

holta:
@deldesir: what final testing is most needed before this PR gets merged?

(And/or summarize what testing you've already done, Thanks!)

deldesir:
> 
> Above two lines come from my own cups log. The timestamps from the "pastebin" look the same.
> I discover another thing. Changing from `Listen localhost:631;` to `Listen 127.0.0.1;` in cupsd.conf seems to supress error:
> E [04/May/2021:16:16:35 -0400] Unable to open listen socket for address [v1.::1]:631 - Cannot assign requested address.
> It shouldn't since I always knew IPV6 doesn't take precedence over IPV4 resoving from http://localhost.

This explains it (or, does it?):
https://unix.stackexchange.com/questions/530976/why-does-localhost-resolve-to-1-but-not-127-0-0-1

deldesir:
@jvonau and Do you think https://unit.nginx.org/ can be useful here? I am willing to invest some time in it if it can lead to some better handling of reverse proxying in general. I just want to be sure I am not doing anything that would go off the networking path settled for IIAB.

jvonau:
> @jvonau and Do you think https://unit.nginx.org/ can be useful here? I am willing to invest some time in it if it can lead to some better handling of reverse proxying in general. I just want to be sure I am not doing anything that would go off the networking path settled for IIAB.

https://nginx.org/en/docs/http/ngx_http_upstream_module.html
That might be worth while, you thinking of using something like?
upstream cups_backend
server unix:/var/run/cups/cups.sock

holta:
@deldesir let's not forget to update this NGINX status document:

https://github.com/iiab/iiab/blob/master/roles/nginx/README.md

(Either as part of this PR, or as follow-up soon after, either is fine!)

Ref: #2762

jvonau:
> > @jvonau and Do you think https://unit.nginx.org/ can be useful here? I am willing to invest some time in it if it can lead to some better handling of reverse proxying in general. I just want to be sure I am not doing anything that would go off the networking path settled for IIAB.
> 
> https://nginx.org/en/docs/http/ngx_http_upstream_module.html
> That might be worth while, you thinking of using something like?
> upstream cups_backend
> server unix:/var/run/cups/cups.sock

Kind of like the setup in #2776

holta:
@deldesir do you still recommend this PR?

(Removing Apache completely may be a long-term goal for 2021, but we also need quick/efficient decisions to move us forward here, Thanks!!)

holta:
We got burned by the trailing slash again:

- http://192.168.0.182/print/ works!
- http://192.168.0.182/print does not!

I'll now work on patching this up, as part of:

- PR #2858

Thanks @deldesir for your patience over 10 weeks.  Now the test plan is more clear.

Issue #2777, opened 2021-05-15
/var/mail/iiab-admin errors or warnings growing fast on Debian 10: (1) "Error while processing /etc/awstats/awstats.conf" (2) "Could not open includedir directory /etc/munin/conf.d: No such file or directory" (3) Monit has similar issues


holta:
@deldesir can you look into this hygiene issue in the coming week if you have time?  A copy of `/var/mail/iiab-admin` is here:

- http://sprunge.us/97EVYK?en

From your installation of PR #2776 onto 10.8.0.46, whose [iiab-diagnostics](https://github.com/iiab/iiab/blob/master/scripts/iiab-diagnostics.README.md) output is here:

- http://sprunge.us/ZCHBzB?en

(I don't know which OS's it affects, but maybe others can tell?)

Ref: #2767

holta:
Excerpt from http://sprunge.us/97EVYK?en, showing AWStats and Munin issues:

```
  55 From www-data@deb10-srv Fri May 14 22:00:01 2021
  56 Return-path: <www-data@deb10-srv>
  57 Envelope-to: root@deb10-srv
  58 Delivery-date: Fri, 14 May 2021 22:00:01 -0400
  59 Received: from www-data by box.lan with local (Exim 4.92)
  60 	(envelope-from <www-data@deb10-srv>)
  61 	id 1lhjaz-0000T2-Ky
  62 	for root@deb10-srv; Fri, 14 May 2021 22:00:01 -0400
  63 From: root@deb10-srv (Cron Daemon)
  64 To: root@deb10-srv
  65 Subject: Cron <www-data@box> [ -x /usr/share/awstats/tools/update.sh ] && /usr/share/awstats/tools/update.sh
  66 MIME-Version: 1.0
  67 Content-Type: text/plain; charset=UTF-8
  68 Content-Transfer-Encoding: 8bit
  69 X-Cron-Env: <MAILTO=root>
  70 X-Cron-Env: <SHELL=/bin/sh>
  71 X-Cron-Env: <HOME=/var/www>
  72 X-Cron-Env: <PATH=/usr/bin:/bin>
  73 X-Cron-Env: <LOGNAME=www-data>
  74 Message-Id: <E1lhjaz-0000T2-Ky@box.lan>
  75 Date: Fri, 14 May 2021 22:00:01 -0400
  76 
  77 Error while processing /etc/awstats/awstats.conf
[...]

PR #2781, opened 2021-05-16
Monit - do no harm

jvonau:
### Fixes bug:
#2777 
Drop use of old config file with invalid path
Ubuntu 20.10
/etc/monit/conf-available [lists](https://github.com/iiab/iiab/issues/2777#issuecomment-841849425) stock available options with /etc/monit/conf-enabled/ holding the enabled options.
No attempt is made to enable any options, that is future work.


holta:
This is really great news.

Similar to Munin's PR #2779.

Whereas modernizing/sanitizing AWStats[*] might be hard to pull off, given it currently requires (some parts of) Apache.

[*] As discussed on #2777.

PR #2787, opened 2021-05-18
add pid to unit file [KA Lite was failing to restart]

tim-moody:


holta:
This is a very important fix.

Thanks to @tim-moody who figured this out and tested it over recent hours, with @jvonau's systemd unit file expertise.

So `systemctl restart kalite-serve` should now once again work reliably.

CC: @deldesir so he can keep an eye this during general testing in coming weeks?

Issue #2797, opened 2021-05-22
cmdsrv fails to start [Raspberry Pi Zero W]


Mathazzar:
### Expected behavior and actual behavior:
cmdsrv isn't starting on its own, using the command "systemctl start iiab-cmdsrv.service" also fails to start the service. Looking at the logs generated by "journalctl -xe" indicates that the iiab_lib file is missing, which I have been unable to find anywhere on github, unlike the other files in the stacktrace, which are all python scripts.

[raspbianJournald.txt](https://github.com/iiab/iiab/files/6526931/raspbianJournald.txt)

### Steps to reproduce the problem:
I followed the instructions on https://github.com/iiab/iiab/wiki/IIAB-Installation#do-everything-from-scratch from a fresh install of Raspberry OS, and completed up to step 6 before recognizing the problem when attempting to log in as described on https://github.com/iiab/iiab/wiki/IIAB-Installation#configure-the-server.

### IIAB version, OS version, hardware details:
IIAB version: 7.2
OS version: Raspbian 10.9
hadrware: Raspberry Pi Zero W

### Please run the `iiab-diagnostics` command, to generate a URL summarizing your IIAB configuration, for volunteers seeking to help you: (Include that short URL here, Thanks!)
[raspbianDiag.txt](https://github.com/iiab/iiab/files/6526926/raspbianDiag.txt)

holta:
From above `iiab-diagnostics` output:

> TASK [www_base : Unarchive http://download.iiab.io/packages/php7.3-stem.x64.tar to / (x64)] ***
2021-05-21 23:00:12,245 p=1515 u=root n=ansible | skipping: [127.0.0.1]
2021-05-21 23:00:12,740 p=1515 u=root n=ansible | 
TASK [www_base : Symlink /etc/php/7.3/fpm/conf.d/20-stem.ini -> /etc/php/7.3/mods-available/stem.ini] ***
2021-05-21 23:00:19,653 p=1515 u=root n=ansible | fatal: [127.0.0.1]: FAILED! => {"changed": false, "msg": "src file does not exist, use \"force=yes\" if you really want to create the link: /etc/php/7.3/mods-available/stem.ini", "path": "/etc/php/7.3/fpm/conf.d/20-stem.ini", "src": "/etc/php/7.3/mods-available/stem.ini"}
2021-05-21 23:00:19,673 p=1515 u=root n=ansible | PLAY RECAP *********************************************************************
[...]

Issue #829, opened 2018-06-06
Is Spanish WikiHow's search fixable? (works on RACHEL site!)


holta:
Online Version: See the top-right of http://oer2go.org/mods/es-wikihow/content/Portada.html where it DOES work.

Offline Version: See the top-right of http:// iiab.me /modules/es-wikihow/content/Portada.html where it does NOT work.  Instead it shows: "Resultados de la búsqueda - wikiHow ES".  CLARIF: Autocomplete always works (in the search box, top-right of the uppermost page above) but the search itself does not.

Peru schools (and others) very much want this fixed, if that's possible!  @georgejhunt & @tim-moody do you have ideas?

holta:
Search Error is:

> how/content/search/search.php:72\nStack trace:\n#0 {main}\n  thrown in /library/www/html/modules/es-wikihow/content/search/search.php on line 72, referer: http://iiab.me/modules/es-wikihow/content/Categoria%3AArte-y-entretenimiento.html

> [Thu Jun 14 12:03:29.077326 2018] [php7:error] [pid 2639] [client 192.168.0.56:52294] PHP Fatal error:  Uncaught Error: Call to undefined function stem() in /library/www/html/modules/es-wikihow/content/search/search.php:72\nStack trace:\n#0 {main}\n  thrown in /library/www/html/modules/es-wikihow/content/search/search.php on line 72, referer: http://iiab.me/modules/es-wikihow/content/Categoria%3AVida-familiar.html

holta:
@georgejhunt & @tim-moody are investigating if locale or a PECL / PEAR issue might be the cause?  

Is a word stemming pkg like https://github.com/marc-mabe/php-stem missing?

holta:
@tim-moody observes that search works on PHP 5 sites like http://oer2go.org/mods/es-wikihow/content/Portada.html

Can we make this (php-stem etc, originally from Google) happen on PHP 7.x, if we can build it?  (There is discussion online that PHP 7 compatibility has been restored...so there is hope!)

Related:
https://github.com/hthetiot/php-stemmer
https://stackoverflow.com/questions/42764217/how-to-enable-stem-language-in-php

holta:
Ongoing/Related Links from @tim-moody and @georgejhunt:

https://bugs.php.net/patch-display.php?bug_id=71631&patch=update-for-php7&revision=latest
http://svn.php.net/viewvc/pecl/stem/
https://github.com/rachelproject/rachelplus/issues/51

[...]

PR #2799, opened 2021-05-23
php-stem - include all RaspOS variants

jvonau:
### Fixes bug:
#2797

### DIFF
--- a/roles/www_base/tasks/php-stem.yml
+++ b/roles/www_base/tasks/php-stem.yml
@@ -11,7 +11,7 @@
 - name: Set fact stem available php 7.3 - excludes i386
   set_fact:
     stem_available: True
-  when: not ansible_machine == "i386" and php_version == 7.3
+  when: (not ansible_machine == "i386" and php_version == 7.3) or is_raspbian_10
 
 - name: Set fact stem available php 7.4
   set_fact:
@@ -26,7 +26,7 @@
     group: root
     #mode: ????
     remote_src: yes
-  when: ansible_machine == "armv7l" and stem_available is defined
+  when: (ansible_machine == "armv7l" or is_raspbian) and stem_available is defined
 
 - name: Unarchive http://download.iiab.io/packages/php{{ php_version }}-stem.aarch64.tar to / (rpi)
   unarchive:


tim-moody:
the question is whether php7.3-stem.rpi.tar works on rpi W. if it does the approach should not be to exclude armv6 from the 'available' list but to install php7.3-stem.rpi.tar for armv6. As an aside I notice there is no version for arm for php7.4.

holta:
> further down the line nextcloud fails instead

FWIW Nextcloud says it requires 512 MB RAM (`nginx_high_php_limits: True` should really be set in [/etc/iiab/local_vars.yml](http://wiki.laptop.org/go/IIAB/FAQ#What_is_local_vars.yml_and_how_do_I_customize_it.3F) in advance of installing Nextcloud).

In short, MEDIUM-sized IIAB installs are not recommended on Raspberry Pi Zero W for several reasons, but this is one of them (512 MB happens to be the total hardware system memory!)

jvonau:
> editing php-stem.yml to either ignore armv6 or use the same source as armv7 appears to produce the same results. The check passes during installation of www_base,

That would be the part this pull request addresses

>and further down the line nextcloud fails instead; 
```
raspbian.raspberrypi.org/raspbian buster/main armhf libx11-xcb1 armhf 2:1.6.7-1+deb10u1", "  404  Not Found [IP: 93.93.128.193 80]"]}
```
You had a network glitch from your location. iiab is not fully installed, none of the options are installed.

> though cmdsrv does work now and the system can be managed through http://box/admin.

I'd slim down some of your install options or at least don't enable some of them. You can later turn those on and off at will post install with relative ease for the reasons stated above, scarce resources.

jvonau:
> the question is whether php7.3-stem.rpi.tar works on rpi W. if it does the approach should not be to exclude armv6 from the 'available' list but to install php7.3-stem.rpi.tar for armv6. As an aside I notice there is no version for arm for php7.4.

I have a vague recollection of the compilation was on ubuntu 32bit to get php7.3-stem.rpi, if the compilation was on RaspOS then the tag should be armv6 and 'should work' on armv7 or greater. Now given this feature fix is targeted at es-wikihow should this routine be part of mediawiki as this user is not even installing mediawiki?

tim-moody:
> I have a vague recollection of the compilation was on ubuntu 32bit to get php7.3-stem.rpi, if the compilation was on RaspOS then the tag should be armv6 and 'should work' on armv7 or greater. Now given this feature fix is targeted at es-wikihow should this routine be part of mediawiki as this user is not even installing mediawiki?

I think I labelled it rpi to indicate it would work for either arm, but it would have to be tried.

I also think there has been discussion about where to put it. But the name not withstanding es-wikihow is not a zim or mediawiki, but just a straight php module in /modules so served by nginx.

holta:
Discussed on today's call (http://minutes.iiab.io).  @jvonau will rework this PR to take that into account.

holta (commented):
This doesn't seen right, as php-stem is for RACHEL / OER2Go module http://box/modules/es-wikihow from http://oer2go.org/viewmod/es-wikihow

(It shouldn't have any connection to MediaWiki, that I know of.)

holta:
@tim-moody can you confirm this looks good?

Mathazzar:
nextcloud seems to consistently fail in the same way, regardless of how many modules are enabled or not, though no errors are stopping ./iiab-install early anymore, so I suspect it's probably the lack of physical resources inherent to the Zero W.

jvonau:
https://github.com/iiab/iiab/issues/829 mentions [this error](https://github.com/iiab/iiab/issues/829#issuecomment-397348765) at a point in time where there was movement from away from php5x and stem support being available. I have a feeling the entire website was rewritten since 2018. The d/l is 12GB, a bit much to look at one file, if es-wikihow deployed anywhere the contents of /library/www/html/modules/es-wikihow/content/search/search.php would be useful to know.

jvonau:
stem should just go away.

holta:
Not a postcard paradise for sure.

But we should try to see what's possible here, to mitigate ongoing install problems like #2797 and #2835.

Better yet if we can get search to "largely work" for WikiHow (given this OER2Go/RACHEL content module's massive popularity in high schools).

PR #2800, opened 2021-05-24
Lok super

jvonau:
### Fixes bug:
replaces #2790 #2792 

### Description of changes proposed in this pull request:
switch over to supervisor
### Smoke-tested on which OS or OS's:
ubuntu-21-04
@adamsclafani

PR #2802, opened 2021-05-25
Lok merge

jvonau:
### Fixes bug:
trimmed version of #2800


holta:
Thanks to live-testing by @nzola and @jvonau over the past 1.5 hours.

@jvonau will then rebase PR #2800 later today.

Aside: PR #2804 "Modernize iiab-support so that it only installs OpenVPN when truly nec"

PR #2743, opened 2021-04-19
Interim/Alternative to PR #2742 accelerates testing w/ "pip3 install ansible-core" (2.11+) in both places

holta:
Refs: PR #2689, #2721, PR #2742

PR #2803, opened 2021-05-25
ansible - install dependencies before pip

jvonau:
pip will install the same packages but in /local/lib if the dependency
is not already in the system path, fill the system path with all the
dependencies before doing the local pip install. Helps avoid
overlapping of apt vs pip managed files.


holta:
Refs: PR #2743, PR #2801

Issue #2820, opened 2021-06-19
Test Raspberry Pi OS 11 on its release. Does RPi WiFi firmware 7.45.18.0 [or its #2853 replacements!] work well?


holta:
Officially this is being released in the coming 10 days.  But it might possibly be delayed a bit in conjunction with the underlying OS (Debian 11, Bullseye) appearing a bit later than usual in its release cycle, right now tentatively scheduled for 2021-07-31 according to: https://release.debian.org

FYI Raspberry Pi OS is also known as RaspiOS.  And formerly known as Raspbian.

Refs:

- #823
- PR #2686
- #2749 "Testing IIAB on the new Debian 11 "Bullseye" Release Candidate(s) [and Haiti content discussion]"
- #2818 "Test IIAB on Ubuntu 21.10 pre-releases (RPi+PC) x (Server+Desktop). Does RPi WiFi firmware 7.45.18.0 work well?"
- PR #2819 "Raspberry Pi OS 11 is imminent, expected before Debian 11's own ETA "2021-07-31"

holta:
Related:

- https://github.com/moodlebox/moodlebox/issues/210
- #2853

martignoni:
FYI, Bullseye's release is [officialy scheduled for 2021-08-14](https://release.debian.org/).

holta:
Quick notes re: testing IIAB on pre-release RaspiOS Bullseye &mdash; @jvonau asked for the pastebins below:

- For now, `pip3 install ansible-core` fails with gcc compile problems.  Temporary hack / workaround is to first run `apt install ansible` (apparently to put cryptography stuff in place).
  - Top of errors: http://sprunge.us/X2I0vI?en
  - Bottom of errors: http://sprunge.us/8CYAYY?en

- [JupyterHub](https://github.com/iiab/iiab/tree/master/roles/jupyterhub) won't install, with similar looking pip3 / wheel compile problem: http://sprunge.us/jRGxrR?en

- [Calibre-Web](https://github.com/iiab/iiab/tree/master/roles/calibre-web) fails to install likewise due to some pip3 / wheel issue(s): http://sprunge.us/aZWlGk?en

Underlying issue appears to be:

- https://www.raspberrypi.org/forums/viewtopic.php?t=319114
- https://blog.piwheels.org/python-3-9-wheels-for-bullseye/

Related:

- #2978

jvonau:
[...]

PR #2819, opened 2021-06-19
Raspberry Pi OS 11 is imminent, expected before Debian 11's own ETA "2021-07-31"

holta:
Raspberry Pi OS 11 was announced as something that would be released in the coming 10 days.

Also known as RaspiOS 11, or Raspbian 11.

In any case, whether the Raspberry Pi Foundation meets their own deadline or not, let's be prepared for it.

Refs:

- #2749 "Testing IIAB on the new Debian 11 "Bullseye" Release Candidate(s) [and Haiti content discussion]"
- #2818 "Test IIAB on Ubuntu 21.10 pre-releases (RPi+PC) x (Server+Desktop). Does RPi WiFi firmware 7.45.18.0 work well?"

Issue #2818, opened 2021-06-18
Test IIAB on Ubuntu 21.10 (RPi+PC) x (Server+Desktop). Does RPi WiFi firmware 7.45.18.0 work well? [OS released 2021-10-14] [JupyterHub install pip fails intermittently?]


holta:
In the short-term [/opt/iiab/iiab/vars/ubuntu-21.yml (Line 26)](https://github.com/iiab/iiab/blob/master/vars/ubuntu-21.yml#L26) should be modified by hand to specify PHP 8.0 (and likely later Line 27, PostgreSQL 13 -> 14) as these versions are different from Ubuntu 21.04

Ubuntu 21.10 (Impish Indri) pre-release daily builds:

- https://cdimage.ubuntu.com/ubuntu-server/daily-live/current/ (Server)
- https://cdimage.ubuntu.com/daily-live/current/ (Desktop)

Refs: #2744, #2814

holta:
To get a better picture of PHP 8.0 released 7 months ago (#2814) a BIG-sized IIAB was installed on Ubuntu Server 20.10 pre-release daily build (VM) at 10.8.0.30 = u2110-srv-BIG-0625.  Using these 2 PHP 8.0 workarounds: (WIP)

1. Package `php8.0-xmlrpc` doesn't exist:

   > TASK [www_base : Install ~10 PHP packages (debuntu)] ***************************
   > fatal: [127.0.0.1]: FAILED! => {"changed": false, "msg": "No package matching 'php8.0-xmlrpc' is available"}

   If it turns out it's still needed by IIAB (?) fyi the "XML-RPC Extension Has Been Moved to PECL in PHP 8.0" :

   - https://lindevs.com/xml-rpc-extension-has-been-moved-to-pecl-in-php-8-0/
   - https://php.watch/versions/8.0/xmlrpc
   - https://pecl.php.net/package/xmlrpc

   For purposes of experimentation, I commented out the installation of `php8.0-xmlrpc` (on Line 18 of https://github.com/iiab/iiab/blob/master/roles/www_base/tasks/main.yml#L18) for now.

2. Package `php8.0-json` does not exist as it's now moved into PHP's core:

   > TASK [nextcloud : Install ffmpeg + libxml2 + 13 php packages] ******************
[...]

PR #2821, opened 2021-06-20
Specify is_debian & is_ubuntu for all supported OS's

holta:
There's no magical solution.  Legacy norms leave us with:

- Theoretically one variable should have sufficed (https://en.wikipedia.org/wiki/Single_source_of_truth) as `is_debian` is currently always the opposite of `is_ubuntu`
- Nevertheless both above variables are already in very common use, so it's better to give them accurate values across the board, avoiding a flood of long-winded and error-prone `is_<OS> is defined` tests (increasing software maintenance costs) all over the place.
- Currently `is_debuntu` is a useless variable, as it's always true.  So be it for now.
  - In future that might possibly change — whether other Linux distributions become supported or not!

Refs: #2749, #2818, PR #2819, #2820

### DIFF
--- a/vars/debian-10.yml
+++ b/vars/debian-10.yml
@@ -1,5 +1,6 @@
 is_debuntu: True
 is_debian: True
+is_ubuntu: False
 is_debian_10: True
 
 # 2019-01-31: These apply if-only-if named_install and/or dhcpd_install are True
--- a/vars/debian-11.yml
+++ b/vars/debian-11.yml
@@ -1,5 +1,6 @@
 is_debuntu: True
 is_debian: True
+is_ubuntu: False
 is_debian_11: True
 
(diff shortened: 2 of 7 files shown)


tim-moody:
The correct usage is only to put True values in the os files as default has all of the False. If is_debuntu is ALWAYS true then put it in defaults. It is not expected to have to revise an already working os file.

jvonau:
> The correct usage is only to put True values in the os files as default has all of the False. If is_debuntu is ALWAYS true then put it in defaults. It is not expected to have to revise an already working os file.

I agree, why was this pushed without discussion? Should one not file a PR, add it to the [agenda](https://docs.google.com/document/d/13akr0zXRbB5f0wFyACgQ_7QU6J99EaL9w1X7ZBZZReg/edit) and discuss its merits during the next meeting? Glad it was reverted, and why are @holta's PRs above review?

PR #2826, opened 2021-06-23
Update Kolibri supported languages from 16 to 25

holta:
2 languages have been dropped and 11 added.

Which brings the total number of supported languages up from 16 to 25.

In conjunction with https://github.com/iiab/iiab/blob/master/roles/kolibri/README.rst which has been improved in recent days.

Ref: learningequality/kolibri#8164


tim-moody:
hopefully no one is confused by support in the Kolibri app vs support in Kolibri rendered content

holta:
Kolibri has quite decent UI/UX-oriented documentation.

And for internationalization specifically:
https://kolibri-dev.readthedocs.io/en/develop/i18n.html

But in this case it was necessary to dig a bit deeper for:
https://github.com/learningequality/kolibri/blob/develop/kolibri/utils/i18n.py#L43-L70

Issue #2814, opened 2021-06-09
Testing IIAB with PHP 8.0, part of new OS's like Ubuntu 21.10 pre-releases


holta:
Ubuntu's PHP 8 progress report: (checklist)
https://wiki.ubuntu.com/ServerTeam/Transition/Php8.0

Ubuntu 21.10 (Impish Indri) daily pre-releases:
- https://cdimage.ubuntu.com/ubuntu-server/daily-live/current/ (Server)
- https://cdimage.ubuntu.com/daily-live/current/ (Desktop)

Longer term, Ubuntu 22.04 LTS is expected to ship with PHP 8.1.

Shorter term, PHP 8 can also be made to work on Debian 11 (Bullseye) whose release is expected very shortly in mid-2021.

Tangentially Related:

- #2749 Testing IIAB on the new Debian 11 "Bullseye" Release Candidate(s)
- #2799 php-stem so WikiHow search works
- #2808 Node.js 16.x

holta:
Just in case this context helps with the release of Nextcloud 22 in a few days, on "2021-07-06" :

"Nextcloud 21 and php8"
https://help.nextcloud.com/t/nextcloud-21-and-php8/104057

PS Nextcloud 21.0.3 was released earlier today.

holta:
Done!  Thanks to PRs like:

- #2830
- #2832 
- #2836 
- #2841

PR #2830, opened 2021-06-25
PHP 8.0 brings a couple changes—slightly affecting Moodle, Nextcloud etc: (1) php8.0-xmlrpc moved to PECL (2) php8.0-json is now baked into PHP itself

holta:
WIP for #2814 (PHP 8) and #2818 (Ubuntu 21.10).

This PR will be re-tested on Ubuntu Server 21.10 pre-release and other OS's.

### DIFF
--- a/roles/moodle/tasks/install.yml
+++ b/roles/moodle/tasks/install.yml
@@ -25,15 +25,15 @@
     name: postgresql
 
 
-- name: Install 8 php packages (debuntu)
+- name: Install 7 PHP packages (debuntu)
   package:
     name:
       - php{{ php_version }}-pgsql
       - php{{ php_version }}-curl
       - php{{ php_version }}-zip
-      - php{{ php_version }}-gd
+      #- php{{ php_version }}-gd         # 2021-06-25: Already installed by www_base/tasks/main.yml
       - php{{ php_version }}-mbstring    # 2020-06-15: Required by Moodle 3.9+
-      - php{{ php_version }}-cli         # 2020-06-15: In the past this included (above) mbstring? However this is not true on Ubuntu Server 20.04 LTS.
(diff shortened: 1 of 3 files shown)


holta:
Second smoketest also successful.

What functionality tests are most important to understand if the removal of `php{{ php_version }}-xmlrpc` matters?

holta:
Smoke-tested with a fresh install of IIAB (BIG-sized) onto 32-bit RaspiOS Lite on RPi 4.

holta:
@georgejhunt @tim-moody @jvonau 

> What functionality tests are most important to understand if the removal of php{{ php_version }}-xmlrpc matters?

So far I do not see any consequences from repeated install tests of this PR without php{{ php_version }}-xmlrpc.

Does anybody remember its purpose?

tim-moody:
The most recent agreement was that we would have php and various libraries as part of the web server role since they are of general interest. This was part of the general movement towards using ansible as a procedural language instead of a declarative language.

If you want to return to a declarative approach, this is just the tip of the iceberg.

Such changes to design strategy should be discussed as a group.

jvonau:
> @georgejhunt @tim-moody @jvonau
> 
> > What functionality tests are most important to understand if the removal of php{{ php_version }}-xmlrpc matters?
> 
> So far I do not see any consequences from repeated install tests of this PR without php{{ php_version }}-xmlrpc.
> 
> Does anybody remember its purpose?

Once upon a time an apt installation within a role was used to denote dependencies of an **optional** role. Now with both Nextcloud and Moodle roles falling into the optional category as both are **not** part of min-vars I see little value in moving the dependencies from being optional to being mandatory in www_base. The process is backwards www_base should be the file with entries being removed not the roles' dependencies, there is no harm in repeating dependencies amongst the optional roles. The analogy I'll use here is "buying snow tires for a car that is to be driven only in a warm climate is a waste of time, the snow tires will never will get used."

Issue #2831, opened 2021-06-27
Do Admin Console logins fail in general with the 'Debian Edu' distro? [Check passwords in /etc/shadow if LDAP absent?]


holta:
@shanti-bhardwa is unable to login to IIAB's Admin Console here, even after we verified that the username (iiab-admin) is part of group iiab-admin and the correct password is being typed in:

http://sprunge.us/ath2Ad?en (output of [iiab-diagnostics](https://github.com/iiab/iiab/blob/master/scripts/iiab-diagnostics.README.md))

Very strange, as Admin Console logins works on all similar Linux OS's to my knowledge:
- Hypothetically Admin Console may be using a very different technique to verify username/password ?
- Is this password-checking routine somehow broken on this pre-release build of Debian Edu 11 / Bullseye ?
  - Or within 'Debian Edu' in general?

A bit more about the 'Debian Edu' version of Debian 11 / Bullseye, that's now being actively promoted: https://blends.debian.org/edu/

e.g. see this download directory where it's one of only 2 choices:
http://cdimage.debian.org/cdimage/daily-builds/daily/arch-latest/amd64/iso-cd/

jvonau:
'Debian Edu' is a LTSP [Linux Terminal Server Project](https://wiki.debian.org/DebianEdu/BeforeGettingStarted) based solution using [LDAP](https://salsa.debian.org/debian-edu/debian-edu-install/-/blob/master/preseed-values/defaults.common) (not the local machine) authentication.

holta:
@jvonau has some promising suggestions to try to figure out why this is happening: (if @shanti-bhardwa's machine hopefully comes back online?)

- Seems like [this Debian Edu] machine was not fully setup before the iiab install was started [?]
  - Would have to join an LDAP server or run it's own
  - It's not looking at  /etc/passwd [and /etc/shadow] for the users
  - Should be transparent once LDAP is setup [File a bug upstream?]
- The pam stack could use pam_ldap, that should be transparent to the users of the stack
- If the machine comes back up check /etc/nsswitch.conf

holta:
@shanti-bhardwa do you have time to try installing Debian 10's 'Edu' version https://get.debian.org/cdimage/release/current/amd64/iso-cd/debian-edu-10.10.0-amd64-netinst.iso (406MB released 2021-06-19) to see if Admin Console logins also fail?

shanti-bhardwa:
yes /etc/nsswitch.conf exists and the contents are

```
[...]

PR #2834, opened 2021-06-29
Be a better Guest

jvonau:
### Fixes bug:
Ability to install on other closed-black-box-curl-like-installations that are floating around. 
Partial fix toward evaluation of #2831

New feature: ability to turn off automatic changes to hostname and networking. (be a good guest)
Further soft-code nginx variables to be able to run as a second instance of nginx on a foreign/unknown platform.
Exposes new variable nginx_dir for alternate location use. 
Supplies custom systemd file for use with nginx_dir is a non-stock location.
Allows further evaluation of the new platform with nginx on a different port   

lightly tested on Ubuntu 20.10 adding to local_vars
>is_guest: True
nginx_port: 2080
nginx_dir: /etc/iiab-nginx

### DIFF
--- a/iiab-stages.yml
+++ b/iiab-stages.yml
@@ -63,3 +63,4 @@
   - name: Network
     include_role:
       name: network
+    when: not is_guest
--- a/roles/0-init/tasks/main.yml
+++ b/roles/0-init/tasks/main.yml
@@ -108,7 +108,7 @@
 
 - name: Set hostname if FQDN_changed
   include_tasks: hostname.yml
-  when: FQDN_changed
+  when: FQDN_changed and not is_guest
 
 - name: Add 'runtime' variable values to {{ iiab_ini_file }}
   ini_file:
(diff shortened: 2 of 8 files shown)


jvonau:
- https://github.com/iiab/iiab/pull/2834/commits/dd5e9d03d4b369f7c7776173f80b1d8b28e1323d is quick hack. The install moves happily goes along restating the wrong service till the very end as to really have everything available app by app would require every roles' 'nginx' restart to be a soft-coded variable. Thought that could be a nice addition after this basic functionality is merged but otherwise is not required to explore an IIAB install on some other black box install. 

- https://github.com/iiab/iiab-admin-console/blob/master/roles/console/templates/admin-console-nginx.conf.j2 would need the 'include /etc/nginx/fastcgi_params' treatment. That is the only change for now, should somebody try this today I would suggest to hand restart iiab-nginx service, or reboot prior to trying **http://ip_addr:nginx_port/web_path/** otherwise exclude admin-console via the custom local_vars.yml that would be needed to trigger any of this workflow.

jvonau:
The second point no longer applies with the linking being done now.

Issue #2835, opened 2021-06-29
Quick installation script fails on [Raspberry] Pi 1 B+ (php-stem)


Hermanio:
### Expected result
Quick install script finishes the installation successfully on the Raspberry Pi 1.

### Actual result
installation fails when linking a config file related to php-stem:
```
TASK [www_base : Symlink /etc/php/7.3/fpm/conf.d/20-stem.ini -> /etc/php/7.3/mods-available/stem.ini] ****************************************************************************************************************************************
fatal: [127.0.0.1]: FAILED! => {"changed": false, "msg": "src file does not exist, use \"force=yes\" if you really want to create the link: /etc/php/7.3/mods-available/stem.ini", "path": "/etc/php/7.3/fpm/conf.d/20-stem.ini", "src": "/etc/php/7.3/mods-available/stem.ini"}
```
That file is indeed missing from the original path.

Going by past discussions that relate to this keyword, I understand that you intend to get rid of this soon, is that correct?

### Steps to reproduce the problem:

Run the installation script on the latest version of Raspberry Pi OS 32-bit: `curl d.iiab.io/install.txt | sudo bash`

### IIAB version, OS version, hardware details:

OS: latest Raspberry Pi OS 32-bit
Hardware: Raspberry Pi 1 B+ (single-core, 1GB RAM), 128GB Samsung microSD card.

IIAB diagnostics output here: http://sprunge.us/d52YZE

### Additional comments

Install script works fine on a ThinkPad X230 running Ubuntu 20.04.2 LTS desktop version so this seems to be Raspberry Pi specific.

Documentation hints at building IIAB on a more powerful Raspberry Pi model first (such as Raspberry Pi 4), and then moving the install to the slower Pi (such as the Zero W), is this recommendation made due to the slow performance or some fundamental incompatibility during the install process?

I'm just testing this out to see how well this project works so no need to treat this with any priority.

holta:
@Hermanio thanks for the detailed report — especially given the current dominance of Raspberry Pi 4 — largely due to its speed as you say.  Your report (and followup, if you have time in coming days) stands to give us a better understanding of older CPU's like the `armv6l` in your Raspberry Pi 1 (with 512 MB RAM according to Line 637 of http://sprunge.us/d52YZE?en).

[...]

PR #2837, opened 2021-06-30
Enable php-stem installation for armv6l machine type

Hermanio:
This allows the installation to continue on additional ARM devices, such as the original Raspberry Pi 1.

### Fixes bug:

https://github.com/iiab/iiab/issues/2835

Just an additional machine type added to the condition, which makes sure that this step is not accidentally skipped on Raspberry Pi 1 (and other machines with the same architecture).

Tested on the Raspberry Pi that is linked in the original issue. It got past stage 9 at this point, after which I ran into networking-related issues (probably due to the Pi not having native WiFi and me using an USB adapter), which are a separate topic to handle.

@holta

### DIFF
--- a/roles/www_base/tasks/php-stem.yml
+++ b/roles/www_base/tasks/php-stem.yml
@@ -26,7 +26,7 @@
     group: root
     #mode: ????
     remote_src: yes
-  when: ansible_machine == "armv7l" and stem_available is defined
+  when: (ansible_machine == "armv7l" or ansible_machine == "armv6l") and stem_available is defined
 
 - name: Unarchive http://download.iiab.io/packages/php{{ php_version }}-stem.aarch64.tar to / (rpi)
   unarchive:


holta:
@tim-moody @jvonau @georgejhunt can you review this PR please?

As compared to PR #2799 etc, for #2797 (another person who just got stuck here in essentially the same way).

Thanks!

georgejhunt:
What I understand from http://single-boards.com/armv6-vs-armv7/ suggests
that this PR should work. But has it been tested? That would be the gating
item for me.

holta:
> What I understand from http://single-boards.com/armv6-vs-armv7/ suggests
that this PR should work. But has it been tested?

I believe @Hermanio tested it but he can confirm.

Hermanio:
> I believe @Hermanio tested it but he can confirm.

I have the minimal install running, if you have a more specific test case I can run (using certain services) then I can enable those and give that a go as well.

holta:
@Hermanio do you have the ability to download the 12.6GB es-wikihow (e.g. using IIAB Admin Console) to verify that the search feature on page http://box/modules/es-wikihow works on armv6l ?

FYI if you don't use IIAB Admin Console, this can be downloaded at the command-line using the instructions at http://oer2go.org/viewmod/es-wikihow — specifically these 2 lines are all you need:

```
cd /library/www/html/modules
sudo rsync -Pavz rsync://dev.worldpossible.org/rachelmods/es-wikihow ./
```

jvonau:
Missing a kernel arch that could be ansible_machine types, see https://www.raspberrypi.org/documentation/configuration/boot_folder.md, kernel8 should banner as aarch64

Hermanio:
@holta The rsync finally finished and I could access wikihow over at `http://box.lan/modules/es-wikihow/`. Search seems to work and I could navigate the pages. 
Screenshot from 2021-07-01 09-44-04

> Missing a kernel arch that could be ansible_machine types, see https://www.raspberrypi.org/documentation/configuration/boot_folder.md, kernel8 should banner as aarch64

aarch64 should be handled by the separate entry in php-stem.yml, right?

holta:
@Hermanio: a debt of gratitude for pushing this forward, given the dominance of Spanish across dozens of countries!

@tim-moody: in a few "days" when Raspberry Pi OS 11 is released, how hard will it be to compile the needed http://d.iiab.io/packages/php7.4-stem.rpi.tar that appears currently missing?

Less Urgently: Should we generate PHP 8.0 php-stem package(s) for Ubuntu 21.10 etc?

holta:
FYI @tim-moody will try to post `php-stem` raw materials (and compilation instructions?) here:

https://github.com/iiab/php-stem

e.g. what's needed to build files like the following at http://d.iiab.io/packages/ :

| File | Date | Size |
| ------------- | ------------- | ------------- |
| php-stem.src.tar | 2018-06-14 15:43 | 3.4M |
| php-stem.x64.tar | 2018-06-14 16:45 | 640K |
| php7.2-stem.x64.tar | 2020-01-25 13:45 | 650K |
| php7.3-stem.aarch64.tar | 2020-03-02 11:16 | 770K |
| php7.3-stem.rpi.tar | 2020-01-25 13:44 | 580K |
| php7.3-stem.x64.tar | 2020-01-25 13:45 | 730K |
| php7.4-stem.aarch64.tar | 2020-04-25 15:42 | 770K |
| php7.4-stem.x64.tar | 2020-04-25 15:42 | 730K |

jvonau:
> > Missing a kernel arch that could be ansible_machine types, see https://www.raspberrypi.org/documentation/configuration/boot_folder.md, kernel8 should banner as aarch64
> 
> aarch64 should be handled by the separate entry in php-stem.yml, right?

Well for RaspiOS the common armv6l user space on the root partition is shared across the four kernels
- kernel.img armv6l
- kernel7.img armv7
- kernel7l.img armv7l
- kernel8.img aarch64

Think replacing 'ansible_machine' with just 'is_raspbian' is easier to cover all four, with Ubuntu 20+ moving to 7.4 or better that is the only supported combination anyway.

tim-moody:
>Think replacing 'ansible_machine' with just 'is_raspbian' is easier to cover all four, with Ubuntu 20+ moving to 7.4 or better that is the only supported combination anyway.

for some reason we thought we needed an aarch64, rather than using a 32 bit version, but we run 32 bit kiwix on all four.

I think we have been trying to fit the algorithm to the file names, but since we have complete control over the file names, let's just design the easiest algorithm that goes from one or more ansible variables and make the file names correspond. We can also use symlinks if it makes life easier.

I would also like to simplify the packaging and not create a zip. (see https://github.com/iiab/php-stem), just a static stem.ini, which could be in the role and a compiled stem.so that depends on arch and php version.

holta:
> * kernel.img armv6l
> * kernel7.img armv7
> * kernel7l.img armv7l
> * kernel8.img aarch64
> 
> Think replacing 'ansible_machine' with just 'is_raspbian' is easier to cover all four, with Ubuntu 20+ moving to 7.4 or better that is the only supported combination anyway.

To be safe, let's also beware of these 2 corner cases — and accommodate them if we can:

- `is_raspbian` (i.e. RaspiOS, a.k.a. Raspberry Pi OS) may be rare on x86_64 / amd64 (traditional PC's) but it does exist.  A few NGO's / implementers occasionally claim[*] they want this OS standardized across all their hardware (RPi & traditional PC's).
- `is_ubuntu` appears to be growing in popularity on RPi hardware.  Today we informally support that, and should continue to do so.  (Similarly the 32-bit kiwix-serve works on 64-bit Ubuntu on RPi 4, so that's what IIAB installs.)

[*] This is in a similar vein to [Ubermix](https://wiki.ubermix.org/index.php?title=Ubermix_Changelog).  So everyone involved in these organizations faces exactly the same GUI — to nurture a unified culture of collaboration — and reduce training costs.  Today there is less pressure for IIAB to support Ubermix (but there had been in the past, 3-4 years ago).  Likewise Raspbian on traditional PC's isn't exactly mainline.

However: if it's relatively easy to accommodate these sincere orgs, we should at least try.  And "gracefully fail" (with a very clear error message!) where we cannot.

tim-moody:
I did an experiment:

* mv stem.so to /usr/local/lib/php-stem/
* make /etc/php/7.3/mods-available/stem.ini point to it
* recreate symlink /etc/php/7.3/fpm/conf.d/20-stem.ini -> /etc/php/7.3/mods-available/stem.ini
* systemctl restart php7.3-fpm.service
* systemctl restart nginx
* search returns results

This has several benefits:

* stem.ini can be static across all environments
* the destination for stem.so can be static across all environments
* i.e. no need to figure out the magic date name for the path, which was the main reason for the zip file

tim-moody:
>However: if it's relatively easy to accommodate these sincere orgs

so can we avoid accommodating those that are not sincere or cease to be so?

holta:
My own opinion is that it's impossible to keep up with ever-changing distro wars.

So we should act on CPU arch where possible, so that we're not unnecessarily kneecapping an African deployment (etc) that's using an OS/distro we don't yet fulling understand.

tim-moody:
>So we should act on CPU arch where possible,

I agree. I think we can form the .so name as stem-{{ ansible_machine }}-{{ php_version }}.so (not sure if ansible_machine or ansible_architecture is better)

then use get_url to download it checking for failure. if failure skip the install of the stem.ini and symlink that enables it.

search will return results if there is a stem.so and not if there isn't, but the play will not break.

jvonau:
> > * kernel.img armv6l
> > * kernel7.img armv7
> > * kernel7l.img armv7l
> > * kernel8.img aarch64
> > 
> > Think replacing 'ansible_machine' with just 'is_raspbian' is easier to cover all four, with Ubuntu 20+ moving to 7.4 or better that is the only supported combination anyway.
> 
> To be safe, let's also beware of these 2 corner cases — and accommodate them if we can:
> 
>     * `is_raspbian` (i.e. RaspiOS, a.k.a. Raspberry Pi OS) may be rare on x86_64 / amd64 (traditional PC's) but it does exist.  A few NGO's / implementers occasionally claim[*] they want this OS standardized across all their hardware (RPi & traditional PC's).
>

The match for stem_available handles that nicely. 
 
>     * `is_ubuntu` appears to be growing in popularity on RPi hardware.  Today we informally support that, and should continue to do so.  (Similarly the 32-bit kiwix-serve works on 64-bit Ubuntu on RPi 4, so that's what IIAB installs.)
> 

Somebody needs to compile aarch32/armv7l for php 7.4 to support them and all php8.0
> 
> [*] This is in a similar vein to [Ubermix](https://wiki.ubermix.org/index.php?title=Ubermix_Changelog). So everyone involved in these organizations faces exactly the same GUI — to nurture a unified culture of collaboration — and reduce training costs. Today there is less pressure for IIAB to support Ubermix (but there had been in the past, 3-4 years ago). Likewise Raspbian on traditional PC's isn't exactly mainline.
>

Let me know if you get stuck with issue on Ubermix.

> I did an experiment:
> 
>     * mv stem.so to /usr/local/lib/php-stem/
> 
>     * make /usr/local/lib/php-stem/stem.ini point to it
> 
>     * recreate symlink /etc/php/7.3/fpm/conf.d/20-stem.ini -> /etc/php/7.3/mods-available/stem.ini
>

Don't think that was needed.

> This has several benefits:
> 
>     * stem.ini can be static across all environments
> 
>     * the destination for stem.so can be static across all environments
> 
>     * i.e. no need to figure out the magic date name for the path, which was the main reason for the zip file

The other part is the php version in '/etc/php/7.3/mods-available/stem.ini' for the installation path to use.

The .so file is already compiled and the structure that is needed to work is in the current zip file, you don't really gain anything but code churn for 7.3 or 7.4. This could be running change for 8.0 thou and could be seamless as the paths within stem.ini would just reference the new /usr/local/lib/php-stem/stem.so location. Would still need to know the path for the stem.ini file location. /etc/php/8.0/mods-available/

holta:
If testing is needed with PHP 8.0, feel free to log in to 10.8.0.62 (which is a very recent daily build of Ubuntu Server 21.10, whose final release will be 2021-10-14).

jvonau:
All I'm I'm saying is the path of least resistance **TODAY** to is to replace `ansible_machine == "armv7l" and stem_available is defined` with `is_raspbian and stem_available is defined` to cover all currently supported OSs that use php 7.3 with the only one being RaspiOS.

Hermanio:
> My own opinion is that it's impossible to keep up with ever-changing distro wars.

This is a bit off-topic, but has Docker ever been brought up as a way to better handle the deployment of all the different services? There are projects like [LinuxServer](https://www.linuxserver.io/) which have packaged some of the more popular services, and it's also possible to build your own with all the necessary customizations. 

You could target one OS (such as Debian) and not have to worry about various OS quirks that much. If it can run Docker, then it can run IIAB as well. 

There are some caveats, though:
* there is some performance overhead, which can in theory be a problem on some lower-end devices, such as Raspberry Pi. From my own testing it's not catastrophic, but on very restricted devices (Raspberry Pi 1, Orange Pi Zero with 512MB RAM) it's still noticeable.
* for some services you would have to build and publish your own images due to preexisting ones being out of date or not trustworthy
* there might be some networking-related changes that may or may not need to be done, as Docker does a lot of magic under the hood regarding that.

Feel free to move this out to a more suitable place if you feel like this can/should be discussed in depth.

jvonau:
> > My own opinion is that it's impossible to keep up with ever-changing distro wars.
> 
> This is a bit off-topic, but has Docker ever been brought up as a way to better handle the deployment of all the different services? There are projects like [LinuxServer](https://www.linuxserver.io/) which have packaged some of the more popular services, and it's also possible to build your own with all the necessary customizations.
> 
> You could target one OS (such as Debian) and not have to worry about various OS quirks that much. If it can run Docker, then it can run IIAB as well.
> 
> There are some caveats, though:
> 
>     * there is some performance overhead, which can in theory be a problem on some lower-end devices, such as Raspberry Pi. From my own testing it's not catastrophic, but on very restricted devices (Raspberry Pi 1, Orange Pi Zero with 512MB RAM) it's still noticeable.
> 
>     * for some services you would have to build and publish your own images due to preexisting ones being out of date or not trustworthy
> 
>     * there might be some networking-related changes that may or may not need to be done, as Docker does a lot of magic under the hood regarding that.
> 
> 
> Feel free to move this out to a more suitable place if you feel like this can/should be discussed in depth.

All the reasons I will not spend one second looking at using docker at anytime in the future.

holta:
@Hermanio if you can use Docker (or anything!) to make Internet-in-a-Box run on aging Nokia phones across the world's poorest countries (and the 1 Million schools that badly need this) you will receive a Nobel Peace Prize.

But even if you don't, I want to thank you for driving this challenging discussion forward.

Hermanio:
I see it mainly as a trade-off between a higher maintenance burden due to multiple OS targets and slightly higher resource requirements. At least on modern low-power machines (Raspberry Pi 4 and better), I would consider this option in the future if you feel like keeping up with OS and dependency changes takes up too much time in the project.

I wasn't aware that IIAB was running on aging Nokia phones, those are probably not a good candidate for this.

holta:
Certainly Docker is a very popular suggestion, and as such deserves a careful and deliberate answer:

https://github.com/iiab/iiab/discussions/2852

holta:
@Hermanio sorry this original fix PR of yours (a good one!) has unfortunately not been acted upon in 15 days now.  On the bright side:

@tim-moody did excellent work (also about 2 weeks ago) bringing together associated material at: https://github.com/iiab/php-stem

Likewise @jvonau laid out many of the tensions colliding here over the past 2 months!

In Conclusion: Hopefully all the above can spur on any outcome in one way or another, as suggested by @tim-moody if he has time as the original author here, to pull any/all these initiatives together, or offer suggestions?

In any case: we should try to document the family tree status quo in the coming week or so (in one way or another) to be just a bit more clear as we're now in the 9th month of PHP 8.0 era — with its crying baby brother/sister PHP 8.1 already due to be born — in 4 months from now (not that it should expect many Christmas presents, even if yes Dec 25th will be its 1-Month Birthday :-)

holta:
Good enough for now, as briefly discussed on this week's IIAB call: http://minutes.iiab.io

> @tim-moody: in a few "days" when Raspberry Pi OS 11 is released, how hard will it be to compile the needed http://d.iiab.io/packages/php7.4-stem.rpi.tar that appears currently missing?

Conclusion as discussed on call: we'll work on adding that in a week-or-two when Raspberry Pi OS 11 is very likely to be released.

Issue #1772, opened 2019-06-24
Can we make AzuraCast work on RPi?


holta:
@m-anish is there hope here?  So far `cd /opt/iiab/iiab; ./runrole azuracast` fails as follows on Raspbian Lite on RPi 3 at 10.8.0.62:

> TASK [azuracast : AzuraCast - Run the installer] *******************************
[...]

Issue #1766, opened 2019-06-23
AzuraCast fails to install on BIG-sized IIAB on "Debian 10" VM


holta:
@m-anish & @jvonau please take a quick look at the error below.

(VM/OS settings & configuration detailed within http://paste.debian.net/1088960/ thanks to/usr/bin/iiab-diagnostics.)

> TASK [azuracast : AzuraCast - Run the installer] *******************************
[...]

PR #2865, opened 2021-07-19
[NEEDSFIX] Azuracast fix

m-anish:
This should make the playbook run successfully on recent IIAB versions. 

Still we need to investigate why azuracast isn't accessible from anything but the host-server itself. Any help would be appreciated!

### DIFF
--- a/roles/azuracast/defaults/main.yml
+++ b/roles/azuracast/defaults/main.yml
@@ -16,7 +16,7 @@
 # All above are set in: github.com/iiab/iiab/blob/master/vars/default_vars.yml
 # If nec, change them by editing /etc/iiab/local_vars.yml prior to installing!
 
-docker_sh_url: https://raw.githubusercontent.com/AzuraCast/AzuraCast/master/docker.sh
-docker_compose_url: https://raw.githubusercontent.com/AzuraCast/AzuraCast/master/docker-compose.sample.yml 
+docker_sh_url: https://raw.githubusercontent.com/AzuraCast/AzuraCast/main/docker.sh
+docker_compose_url: https://raw.githubusercontent.com/AzuraCast/AzuraCast/main/docker-compose.sample.yml 
 docker_container_dir: /library/docker
 azuracast_host_dir: /opt/azuracast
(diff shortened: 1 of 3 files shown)


holta:
Thanks @m-anish !

Can you please do a quick revise to https://github.com/iiab/iiab/blob/master/roles/azuracast/README.rst to bring the following sentence up to date?

> As of 2019-08-04, this will only run on Ubuntu 18.04, and tentatively on Debian 10 "Buster" (#1766). Support for Raspberry Pi remains a goal for now — please if you can, consider helping us solve this critical challenge (#1772, AzuraCast/AzuraCast#332).

Refs:

- PR #1733
- #1772

jvonau:
>Still we need to investigate why azuracast isn't accessible from anything but the host-server itself. Any help would be appreciated!

I suspect some iptables rules that docker adds to the INPUT filter table might be missing, we flush the table and recreate the rules. 

mUqX6i?en is helpful, the /etc/systemd/network/IIAB-Slave-wlp0s20f3.network looks bogus, I'll look into that one but I don't understand 
```
1398 172.17.0.0/16 dev docker0 proto kernel scope link src 172.17.0.1 linkdown 
1400 172.19.0.0/16 dev br-ba308fa1b076 proto kernel scope link src 172.19.0.1 linkdown 
1401 172.20.0.0/16 dev br-2b05be063c3d proto kernel scope link src 172.20.0.1 
1402 172.21.0.0/16 dev br-90ba787062ed proto kernel scope link src 172.21.0.1 
1403 172.36.0.0/16 dev br-d2c42f7bfe61 proto kernel scope link src 172.36.0.1 
```
which dev-br* belongs to each of the networks defined in docker-compose.override.yml.j2 that defined the 'azure' network and docker-compose.sample that defines "frontend & backend'. 

https://docs.azuracast.com/en/getting-started/installation/ansible "Since AzuraCast installs its own radio tools, databases and web servers, you should always install AzuraCast on a "clean" server instance with no other web or radio software installed previously."

Given the above I would get AzuraCast running first before trying to adding anything else to capture what iptables rules are put into place before iiab gets a chance to muck with them https://github.com/iiab/iiab/pull/1753#issuecomment-504117039. Might entertain trying PR #2834 to suppress all network changes and move nginx to a different port.

jvonau:
or maybe https://github.com/AzuraCast/AzuraCast/issues/4155?

PR #2867, opened 2021-07-19
Clean kalite/install.yml, including softcode of venv

holta:


holta:
Tested on 32-bit RaspiOS Lite on RPi 4.

PR #2876, opened 2021-07-26
don't always run networking - now optional

jvonau:
### Fixes bug:
make networking opt-in

### DIFF
--- a/iiab-stages.yml
+++ b/iiab-stages.yml
@@ -60,6 +60,6 @@
       name: 9-local-addons
     when: ansible_local.local_facts.stage|int < 9
 
-  - name: Network
-    include_role:
-      name: network
+#  - name: Network
+#    include_role:
+#      name: network


holta:
Should [roles/network](https://github.com/iiab/iiab/blob/master/roles/network) declare its own dependencies, instead of [2-common/tasks/network.yml](https://github.com/iiab/iiab/blob/master/roles/2-common/tasks/network.yml) doing that?

Especially if some or all of those dependencies are no longer needed?

jvonau:
Should captiveportal and admin-console have their own setup routines for uwsgi instead of lumping uwsgi into nginx?

holta:
> Should captiveportal and admin-console have their own setup routines for uwsgi instead of lumping uwsgi into nginx?

That sounds very reasonable.  Noting that almost everybody installs roles/captiveportal and Admin Console at the moment.  So I guess uwsgi is a de facto requirement for the moment anyway.

jvonau:
The defaults are set to 'install' but the captiveportal role is completely optional as is admin-console but outside the scope of what this PR is trying to address, that is easier ability to install anywhere. Just a more automated version of the network workarounds mentioned in https://github.com/iiab/iiab/issues/2610 ie not running the network role at all, only on demand when asked.

OR 

You can curtail the soapbox broadcasting about iiab supporting every OS variant that somebody might install. Question becomes has this project moved from being primarily targeted at offline usage on the RPi hardware platform to becoming a portal/gateway to the lesser known locations of education content available via the curation of the various sources within admin-console's 'install content' section.

It's not like I haven't said in the past that networking could optional, and should be the last item configured to match the deployment's intended use case.

holta:
Excellent points.

Also Related: [roles/captiveportal](https://github.com/iiab/iiab/tree/master/roles/captiveportal) is installed almost everywhere, but good intentions only go so far &mdash; it's not really used much, due to usability problems on recent Androids etc &mdash; as such this remains an ever-serious and ongoing question that will need to be addressed.

holta:
> Should [roles/network](https://github.com/iiab/iiab/blob/master/roles/network) declare its own dependencies, instead of [2-common/tasks/network.yml](https://github.com/iiab/iiab/blob/master/roles/2-common/tasks/network.yml) doing that?

- iproute2
- iptables-persistent
- hostapd
- netmask

> Especially if some or all of those dependencies are no longer needed?

Oops I neglected to notice there are 5 additional networking utilities (below) installed across the board at https://github.com/iiab/iiab/blob/master/roles/2-common/tasks/packages.yml#L21-L45

- avahi-daemon
- inetutils-syslogd    # Error logging facility
- libnss-mdns    # Provides name resolution via mDNS (Multicast DNS) using Zeroconf/Bonjour e.g. Avahi
- wpasupplicant
- net-tools    # 2021-04-26: @jvonau suggests possibly deleting this...unless oldtimers really want these older commands in iiab-diagnostics output?

Should some-or-all of these 4 + 5 = 9 be moved to [roles/network](https://github.com/iiab/iiab/blob/master/roles/network) as we transition to non-lazy dependencies, towards all IIAB roles becoming a lot more understandable and maintainable going forward?

Conversely, should some of the above 9 remain globally installed?  (Perhaps starting with those that help [iiab-diagnostics](https://github.com/iiab/iiab/blob/master/scripts/iiab-diagnostics.README.md) be as understandable to all people across all possible failures, even early on during IIAB installs?)

jvonau:
> > Should [roles/network](https://github.com/iiab/iiab/blob/master/roles/network) declare its own dependencies, instead of [2-common/tasks/network.yml](https://github.com/iiab/iiab/blob/master/roles/2-common/tasks/network.yml) doing that?
> 
>     * iproute2
> 
>     * iptables-persistent
> 
>     * hostapd
> 
>     * netmask

These packages are used and managed by iiab-network exclusively.

> 
> > Especially if some or all of those dependencies are no longer needed?
> 
> Oops I neglected to notice there are 5 additional networking utilities (below) installed across the board at https://github.com/iiab/iiab/blob/master/roles/2-common/tasks/packages.yml#L21-L45
> 
>     * avahi-daemon
> 

Hold over from the XO days and used to advertise ssh/admin-console being available via avahi-daemon
>     * inetutils-syslogd    # Error logging facility
> 

Hold over from the XO days, journalctl has replaced this in newer distros
>     * libnss-mdns    # Provides name resolution via mDNS (Multicast DNS) using Zeroconf/Bonjour e.g. Avahi
> 

Client side library 
>     * wpasupplicant
> 

Client library for connections to a wifi AP
>     * net-tools    # 2021-04-26: @jvonau suggests possibly deleting this...unless oldtimers really want these older commands in iiab-diagnostics output?
> 
> 
None of these are used by iiab-network

> Should some-or-all of these 4 + 5 = 9 be moved to [roles/network](https://github.com/iiab/iiab/blob/master/roles/network) as we transition to non-lazy dependencies, towards all IIAB roles becoming a lot more understandable and maintainable going forward?
> 

When does admin-console join the clean out party?

> Conversely, should some of the above 9 remain globally installed? (Perhaps starting with those that help [iiab-diagnostics](https://github.com/iiab/iiab/blob/master/scripts/iiab-diagnostics.README.md) be as understandable to all people across all possible failures, even early on during IIAB installs?)

You forgot https://github.com/iiab/iiab/blob/master/roles/1-prep/tasks/raspberry_pi.yml and https://github.com/iiab/iiab/blob/master/roles/1-prep/tasks/main.yml

I'm inclined to do nothing, installs over the years have shown no ill effects from being installed even when not needed.

holta:
> You forgot https://github.com/iiab/iiab/blob/master/roles/1-prep/tasks/raspberry_pi.yml

Good point.  These 3 are also networking tools:

- iw    # configuring Linux wireless devices
- rfkill    # enabling and disabling wireless devices
- wireless-tools    # manipulating Linux Wireless Extensions

jvonau:
> > You forgot https://github.com/iiab/iiab/blob/master/roles/1-prep/tasks/raspberry_pi.yml
> 
> Good point. These 3 are also networking tools:
> 
>     * iw    # configuring Linux wireless devices
>

hard dependence for ap0 creation see clone-wifi.service.j2
 >     * rfkill    # enabling and disabling wireless devices
> 
>     * wireless-tools    # manipulating Linux Wireless Extensions

The 3 are provided by RaspiOS, Ubuntu|Debian on the other hand is hit and miss, desktops might have some/all 3 preinstalled while servers tend not to have these present at all but are needed to be installed if you want to take full advantage of wifi on Ubuntu and friends but is only enforced on RPi hardware where we know in advance of the likely hood of wifi being present. Now iproute2 should be preinstalled everywhere, iptables-persistent seems to be ok but netfilter is ever moving forward so keep an eye on it for now. Next, netmask is just a handy utility and hostapd has its service masked out of the box and only used when the network role detects the presence of wifi and an AP is desired.

holta:
1) The above clarifications on networking packages from @jvonau were massively helpful.  So I've added all those comments inline here:

   - [1-prep/tasks/raspberry_pi.yml](https://github.com/iiab/iiab/blob/master/roles/1-prep/tasks/raspberry_pi.yml#L34-L56)
   - [2-common/tasks/packages.yml](https://github.com/iiab/iiab/blob/master/roles/2-common/tasks/packages.yml)
   - [2-common/tasks/network.yml](https://github.com/iiab/iiab/blob/master/roles/2-common/tasks/network.yml#L1-L20)

2) At the moment [Line 6 of 1-prep/tasks/main.yml](https://github.com/iiab/iiab/blob/master/roles/1-prep/tasks/main.yml#L6#L6) says:

   > dnsmasq (install now, configure LATER in 'network', after Stage 9)

   Is this really true?

   And if so, does it matter that it will not happen when installing IIAB onto VM's (etc), where there is no WiFi device?

holta:
> 2. At the moment Line 6 of 1-prep/tasks/main.yml says:
>
> dnsmasq (install now, configure LATER in 'network', after Stage 9)

3. Isn't iptables within roles/network also important (firewall security) on IIAB systems that do not have WiFi?  For example:

   - network/templates/gateway/iiab-gen-iptables
   - network/tasks/enable_services.yml
   - network/tasks/restart.yml

jvonau:
> > 1. At the moment Line 6 of 1-prep/tasks/main.yml says:
> > 
> > dnsmasq (install now, configure LATER in 'network', after Stage 9)
> 
>     1. Isn't iptables within roles/network also important (firewall security) on IIAB systems that do not have WiFi?  For example:
>        
>        * network/templates/gateway/iiab-gen-iptables
>        * network/tasks/enable_services.yml
>        * network/tasks/restart.yml

I'm not going to try and second guess an enduser choice of placing his/her content on a public facing VPS, they should have the skills to secure the installation before ever considering curl iiab. Now iptables alone really gains you nothing other than a false sense of feeling secure, there are aspects of security that have received the ostrich treatment given the intended offline use of the server the attack vectors were deemed small. Public facing servers are a whole different ball game, you best show up with quality players or you'll find yourself outclassed by the bad actors out looking for some easy target. The moment the enduser takes the installer outside of its intended use of setting up a wifi hotspot to serve content the assumptions stop as to the intended use case, of course the enduser is free to run iiab-network at will post install should that functionalllty be wanted but I don't want to stand in the way of users trying out docker based solutions. The area of overlap between docker and iiab-gen-iptables would be the area of contention, I suggest saving a copy of the rules generated by docker for review before iiab-nework is run.

holta:
Related:

- #2921
- #2933

jvonau:
#2739

PR #2877, opened 2021-07-26
Update 4-server-options/tasks/main.yml explanations

holta:
Code maintenance for readability and possible future deprecating of things like {named / BIND, dhcpd, squid, DansGuardian}.


jvonau:
You can remove these BS lines also. https://github.com/iiab/iiab/blob/master/roles/4-server-options/tasks/main.yml#L3-L14

Issue #2873, opened 2021-07-21
JupyterHub + Node-RED + Moodle + Internet Archive install problems on RaspiOS-for-PC's Node.js 10.x -- lacking Node.js 16.x, which requires amd64 not i386 python3


holta:
Strangely I succeeded in installing most of BIG-sized IIAB onto RaspiOS-for-PC/VM (graphical desktop version, latest ISO is [2021-01-12](https://downloads.raspberrypi.org/rpd_x86/images/rpd_x86-2021-01-12/2021-01-11-raspios-buster-i386.iso)) _BUT_ Sugarizer was the only Node.js role that I cared to actually make work.  As explained below (just for the record!) as it's probably NOT worth solving these issues.

[iiab-diagnostics](https://github.com/iiab/iiab/blob/master/scripts/iiab-diagnostics.README.md) output: http://sprunge.us/xuBc6j?en

FWIW I did not try IIAB's 5th Node.js role (pbx) as it's not currently part of [BIG-sized IIAB](http://FAQ.IIAB.IO#What_services_.28IIAB_apps.29_are_suggested_during_installation.3F).

## Node.js 10.x PROBLEM NUMBER 1 OF 4: JupyterHub

```
TASK [jupyterhub : Use pip to install into a virtual environment: /opt/iiab/jupyterhub (~229 MB)] **********************************************************************
[...]

PR #2880, opened 2021-07-27
iiab-diagnostics: dpkg --print-architecture && dpkg --print-foreign-architectures

holta:
Great suggestion from @jvonau:

- https://github.com/iiab/iiab/issues/2873#issuecomment-887506529


holta:
Tested on Ubuntu Server 20.04

Issue #2875, opened 2021-07-21
Deprecating zzz_iiab.sh is a good idea (/usb/sbin $PATH hack) if scripts work well enough in /usr/bin + /usr/sbin Etc


holta:
In the interests of security (& making IIAB more readable, more trusted, and more understandable) as mentioned by @jvonau at: https://github.com/iiab/iiab/issues/2749#issuecomment-883870766

He mentions this excellent overall set of recommendations: https://www.geeksforgeeks.org/linux-file-hierarchy-structure/

Some scripts (or all scripts) might also be moved to `/usr/local/bin` and `/usr/local/sbin` as appropriate.

Reminder that zzz_iiab.sh is here:

- https://github.com/iiab/iiab/blob/master/roles/2-common/tasks/main.yml#L41-L44
- https://github.com/iiab/iiab/blob/master/roles/2-common/templates/zzz_iiab.sh

holta:
I think we should go ahead and try this, just as @jvonau has suggested.

In order to surface those commands that IIAB has placed in `/usr/sbin` and in hindsight might have found a better place:

```
root@box:/usr/sbin# ls iiab* piclone_cmd test-wifi

iiab                 iiab-clean-usb.sh     iiab-make-cp-servers.py  piclone_cmd
iiab-check-firmware  iiab-divert-to-nginx  iiab-rpi-max-rootfs.sh   test-wifi
```

```
root@box:/usr/bin# ls iiab*

iiab-cmdsrv-ctl          iiab-mongodb-repair-if-no-lock
iiab-diagnostics         iiab-refresh-wiki-docs
iiab-extend-sat.py       iiab-remote-off
iiab-gen-iptables        iiab-remote-on
iiab-get-kiwix-cat       iiab-summarize.sh
iiab-get-oer2go-cat      iiab-support
iiab-hotspot-off         iiab-support-off
iiab-hotspot-on          iiab-support-on
iiab-install-map-region  iiab-sync-menu-defs
iiab-install-oer2go-mod  iiab-update-map
iiab-internet-off        iiab-update-menus
iiab-internet-on         iiab-usb_lib-show-all-off
iiab-make-init.py        iiab-usb_lib-show-all-on
iiab-make-kiwix-lib      iiab-vpn-off
iiab-make-kiwix-lib.py   iiab-vpn-on
iiab-maps-finish.py

PR #2881, opened 2021-07-27
Experimentally do not add /usr/sbin to non-root users' $PATH, per distro usability/security norms [i.e. do not install /etc/profile.d/zzz_iiab.sh -- this affects Debian but does not affect RaspiOS & Ubuntu, where non-root users already have /usr/sbin in $PATH]

holta:
Towards advancing IIAB's usability/security footprint, per #2875.


holta:
@georgejhunt:

Which (if any) of these 3 [-> 6] commands really truly need to be run by non-root users?

- [[iiab-check-firmware](https://github.com/iiab/iiab/blob/master/roles/firmware/templates/iiab-check-firmware) installed by roles/firmware]
- [iiab-clean-usb.sh](https://github.com/iiab/iiab/blob/master/roles/usb_lib/templates/iiab-clean-usb.sh) installed by roles/usb-lib &mdash; as created by PR #2328 
- [iiab-divert-to-nginx](https://github.com/iiab/iiab/blob/master/roles/captiveportal/templates/iiab-divert-to-nginx) installed by roles/captiveportal
- [iiab-make-cp-servers.py](https://github.com/iiab/iiab/blob/master/roles/captiveportal/templates/iiab-make-cp-servers.py) installed by roles/captiveportal
- [[iiab-rpi-max-rootfs.sh](https://github.com/iiab/iiab/blob/master/roles/1-prep/templates/iiab-rpi-max-rootfs.sh) installed by roles/1-prep/tasks/raspberry_pi.yml]
- [[test-wifi -> iiab-test-wifi](https://github.com/iiab/iiab/blob/master/roles/network/templates/hostapd/test-wifi) installed by roles/network/tasks/hostapd.yml]

(I will move any [or all] of the 3 from `/usr/sbin` to `/usr/bin` as recommended, Thanks!)

holta:
This PR also moves:

- roles/firmware/templates/iiab-check-firmware from **/usr/sbin** to **/usr/bin** _(so non-root users can run it)_

And renames 2 commands + 3 systemd services:

- roles/firmware/templates/check-firmware.service to **iiab-check-firmware.service**
- roles/firmware/templates/fw_warn.sh to **iiab-firmware-warn.sh**
- roles/network/templates/hostapd/clone-wifi.service.j2 to **iiab-clone-wifi.service.j2**
- roles/network/templates/hostapd/test-wifi to **iiab-test-wifi.j2**
- roles/network/templates/hostapd/wifi-test.service.j2 to **iiab-wifi-test.service.j2**

holta:
Smoke-testing is ongoing on 183-rpi4-lite-MIN-2881 = 10.8.0.38

Thanks @georgejhunt for having looked into this, who agreed that `/usr/bin` is just as good as `/usr/sbin` when more access by non-root users is helpful.

For now I'm experimenting with moving [only 1 (that's iiab-check-firmware) of the above 6 commands](https://github.com/iiab/iiab/pull/2881#issuecomment-887696174) from `/usr/sbin` to `/usr/bin` &mdash; as it appears the other 5 are not generally needed by most all IIAB implementers/operators.

holta:
Tested on 32-bit RaspiOS Lite.

Testing is ongoing on a Debian 10 VM (10.8.0.62) where non-root users indeed do not have /usr/sbin in their $PATH by default.

jvonau:
> And renames 2 commands + 3 systemd services:
> 
>     * roles/firmware/templates/check-firmware.service to **iiab-check-firmware.service**
> 
> 
>     * roles/network/templates/hostapd/clone-wifi.service.j2 to **iiab-clone-wifi.service.j2**
> 
> 
>     * roles/network/templates/hostapd/wifi-test.service.j2 to **iiab-wifi-test.service.j2**

This just broke backwards compatibility... old installs that update will now have 2 .service files providing the same service....

roles/1-prep/tasks/raspberry_pi.yml
```diff
@@ -67,15 +67,15 @@
...
     src: "{{ item.src }}"
     dest: "{{ item.dest }}"
-    owner: root
-    group: root
+    #owner: root
+    #group: root
     mode: "{{ item.mode }}"
   with_items:
-    - { src: 'iiab-rpi-max-rootfs.sh', dest: '/usr/sbin/iiab-rpi-max-rootfs.sh', mode: '0755'}
+    - { src: 'iiab-rpi-max-rootfs.sh', dest: '/usr/sbin/', mode: '0755'}
```
jvonau, line 78: FWIW dest: using the full path is required when the template ends with .j2 (to denote softcoding) so it's generally preferred to keep using the full path anyway. Not sure why the full path was not good enough and now does not match the stanza directly below.

holta: When .j2 templating is not in use, there is the DRY principle (https://en.wikipedia.org/wiki/Don%27t_repeat_yourself) to consider, i.e. one can avoid specifying the exact same filename twice in a row

jvonau:
Please address upgrade breakage that is forthcoming given the stale advice of https://github.com/iiab/iiab/wiki/IIAB-Networking
```
If making changes to /etc/iiab/local_vars.yml, remember the general rule is to then run "cd /opt/iiab/iiab" 
followed by "./iiab-install --reinstall" (formerly "./runansible") 

```

roles/1-prep/tasks/raspberry_pi.yml
jvonau: At least be consistent apply that process to the line below also....

holta: Done:

- PR #2884

holta:
> Please address upgrade breakage

Windows and Android break very regularly as a result of upgrades, even despite multi-billion dollar budgets and extremely sophisticated automated testing regimes we don't yet have ;)

> that is forthcoming given the stale advice of https://github.com/iiab/iiab/wiki/IIAB-Networking

I just revised the advice at the bottom of https://github.com/iiab/iiab/wiki/IIAB-Networking

It needs more work for sure!  Hopefully everyone can weigh in and make this better.

Then once it's reasonably satisfactory, we can promote these suggestions to http://FAQ.IIAB.IO and to our [YouTube channel](https://www.youtube.com/channel/UC0cBGCxr_WPBPa3IqPVEe3g).

PR #2887, opened 2021-07-30
NGINX Security Risk: Remove /cgi-bin access to /usr/lib ?

holta:
Many thanks to @tim-moody for having identified this potentially serious hazard.

Needs testing to establish that most/all IIAB apps continue to work with this patch.


holta:
Briefly tested using `./runrole nginx`

So this seems like a low-risk fix.

More community testing after merging will be important, as always.

PR #2888, opened 2021-07-30
0-init/tasks/network.yml: Clarify var FQDN_changed falling into disuse?

holta:
Mostly documentation, in support of ongoing/future code maintenance choices.

Refs:

- PR #2876
- PR #2886

PR #2885, opened 2021-07-28
Move packages 'python3-*' & 'virtualenv' (for Python 2) from scripts/ansible to roles that need them, e.g. {mysql, postgresql, munin, kalite}

holta:
Per https://github.com/iiab/iiab/pull/2879#issuecomment-888327576

@jvonau might have further suggestions?

- PR #2527 
- PR #2655 
- PR #2659
- PR #2743

PR #2889, opened 2021-07-30
0-init/defaults/main.yml: reconcile & document many stale vars

holta:
Testing ongoing on multiple OS's.

Refs:

- PR #2885
- PR #2886 
- PR #2888

### DIFF
--- a/roles/0-init/defaults/main.yml
+++ b/roles/0-init/defaults/main.yml
@@ -5,29 +5,41 @@
 
 # These entries should never be changed in this file.
 # These are defaults for boolean routines.
+
+# 2021-07-30: This variable (first_run) is still used in 2 places:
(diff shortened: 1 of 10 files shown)


holta:
Tested together with PR #2885 on 3 OS's:

- 32-bit RaspiOS 10 on RPi 4
- Debian 10 non-desktop VM for nostalgia purposes (probably the last time, as Debian 11 is now where the action is, and Debian 12 Bookworm [Debian's testing branch, extremely popular on guru laptops] is about to go live in coming weeks...)
- Ubuntu Server 21.04 VM

That was Phase 1.  Further reconciliation of ambiguous variables in [0-init/defaults/main.yml](https://github.com/iiab/iiab/blob/master/roles/0-init/defaults/main.yml) and similar places, can happen in coming months.

Issue #1737, opened 2019-06-14
Who can help test IIAB 7.2 on Ubuntu 20.04+, Debian 10.10+, Linux Mint 20 & Raspberry Pi ? [Kolibri 502 Bad Gateway?]


holta:
Thanks to all who can help here, accelerating [IIAB 7.0](https://github.com/iiab/iiab/wiki/IIAB-7.0-Release-Notes)'s release in coming days/weeks!  Using http://d.iiab.io to install an IIAB 7.0 pre-release on the very latest build of Debian 10 Buster (generally several ISO's are published each day!) as can be found here:

https://cdimage.debian.org/cdimage/daily-builds/daily/arch-latest/amd64/iso-cd/debian-testing-amd64-netinst.iso

FYI "Debian 10's software package freeze takes place on Tuesday, June 25 at 13:00 UTC" and release date is expected to be Saturday July 6th.

WARNINGS/NOTES:

<strike>1) You might still need to set `VERSION_ID="10"` (or `VERSION_ID="10.0"`) in `/etc/os-release` as explained in #1387</strike>

2) If you're installing a BIG-sized IIAB (good idea during testing!) please remember to change your [/etc/iiab/local_vars.yml](http://wiki.laptop.org/go/IIAB/FAQ#What_is_local_vars.yml_and_how_do_I_customize_it.3F) Sugarizer variables to:

   > sugarizer_install: False
   > sugarizer_enabled: False

   As explained in #1437

<strike>3) Consider setting http://d2.iiab.io within Line 23 of your [/opt/iiab/iiab/vars/default_vars.yml](https://github.com/iiab/iiab/blob/master/vars/default_vars.yml#L23) before running `sudo iiab` to test IIAB's new download server, that should come online later in June.</strike>

4) Please download at least one Map Pack from http://box.lan/admin > Install Content > Get Map Region and let us know how that all works after you.  Then try out the resulting map on your IIAB home page (http://box) and also try out http://box/maps (should now work, as a result of PR #1724), Thanks!

Refs: #878, #1387, #1437, PR #1724

holta:
Please keep in touch with @floydianslips if you too can help here!

holta:
So far so good, http://d.iiab.io installed cleanly onto a 2019-06-15 pre-release snapshot of Debian 10 Buster, as documented at https://github.com/iiab/iiab/issues/1387#issuecomment-502345790

mitra42:
[...]

PR #2891, opened 2021-07-31
Desupport Debian 10 + lay groundwork for Debian 12 (Bookworm)

holta:
Arguably this should have been done 6 months ago, when the bulk of the Debian community moved from Debian 10 (Buster) to Debian 11 (Bullseye).

Not without reason nodesource.com began supporting Debian 11 last year in 2020.

In any case, IIAB should not and will not support nostalgia OS's from years ago.

CLARIFICATION 1: Those who insist on running nostalgia OS's can of course attempt that At Their Own Risk &mdash; by following the instructions under section 3. here: https://github.com/iiab/iiab/wiki/IIAB-Platforms#operating-systems

CLARIFICATION 2: IIAB will continue to support Raspberry Pi OS 10 for a few more weeks, until Raspberry Pi OS 11 is mainlined shortly at https://www.raspberrypi.org/software/operating-systems/ &mdash; _and confirmed to be good enough (:_

Ref:

- #1737
- #2658
- PR #2885

### DIFF
--- a/scripts/local_facts.fact
+++ b/scripts/local_facts.fact
@@ -52,6 +52,7 @@ OS_VER=$OS-$VERSION_ID
     #"fedora-22"    | \
     #"debian-8"     | \
     #"debian-9"     | \
+    #"debian-10"    | \
     #"ubuntu-16"    | \
     #"ubuntu-17"    | \
     #"ubuntu-18"    | \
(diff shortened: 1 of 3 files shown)


holta:
Recap: Debian 10 was desupported 2 weeks ago:

https://github.com/iiab/iiab/wiki/IIAB-Platforms/_compare/f5d91cb923fc9542cc02da67da747716653992b5...0f1d697af9251f6a6435f4344814205bc40af7f7

In Short: it's now time to start enforcing that, as few people read instructions during modern times.

PR #2895, opened 2021-08-02
Cleanup of roles/openvpn and roles/mysql

holta:
Making roles/openvpn's dependency on sudo more explicit, and other overdue cleanup in roles/mysql etc.

No functional changes.


holta:
Testing ongoing, on a Debian 11 Bullseye VM.

PR #2899, opened 2021-08-03
Experiment with FreePBX 16 Beta, as required by PHP 7.4

holta:
Refs:

- PR #1761
- PR #2489
- PR #2896
- PR #2897
  - https://www.freepbx.org/freepbx-16-beta-is-here/
  - https://wiki.freepbx.org/pages/viewpage.action?spaceKey=FOP&title=Installing+FreePBX+16+on+Debian+10.9
- PR #2898

PR #2900, opened 2021-08-03
Explain where/why sudo gets installed

holta:
No functional changes.


holta:
Tested on Debian 11, as part of PR #2899.

PR #2901, opened 2021-08-03
Lint pbx/tasks/asterisk.yml to a common baseline, as Apache extrication/isolation work continues

holta:
With @lemueldsouza and @m-anish to look this over if he has time:

Refs:

- PR #2850
- PR #2885 
- PR #2896 
- PR #2897
- PR #2898 
- PR #2899

PR #2896, opened 2021-08-02
pbx/defaults/main.yml: Asterisk 16.x -> 18.x

holta:
Verification is ongoing with @lemueldsouza for RPi etc, but long story short Asterisk 18.5.1 installed with FreePBX 15.0.17.37 on Ubuntu Server 20.04 with Node.js 16.6.0 seems to better support modern testing:

```
root@box:~# asterisk -r
Asterisk 18.5.1, Copyright (C) 1999 - 2021, Sangoma Technologies Corporation and others.
```

```
root@box:~# amportal a ma list | grep framework
| framework        | 15.0.17.37 | Enabled                           | GPLv2+  |
```

Asterisk 17.x is also possible, if anybody strongly prefers that for any reason in coming weeks ?

But for now, v18 seem would seem to be best, to harmonize IIAB community testing among all.

PR #2902, opened 2021-08-03
pbx/tasks/asterisk.yml: Experiment w/o aptitude

holta:
Thanks to @jvonau suggestion:

https://github.com/iiab/iiab/pull/2896#issuecomment-892061982

cc: @lemueldsouza @m-anish

- PR #2896
- PR #2899
- PR #2901

### DIFF
--- a/roles/pbx/tasks/asterisk.yml
+++ b/roles/pbx/tasks/asterisk.yml
@@ -38,10 +38,11 @@
     chdir: "{{ asterisk_src_dir }}"
     creates: addons/mp3/mpg123.h
 
-- name: Asterisk - Install aptitude (otherwise install_prereq fails?)
-  package:
-    name: aptitude
-    state: latest
+# 2021-08-03: Experiment w/o aptitude
+#- name: Asterisk - Install aptitude (otherwise install_prereq fails?)
+#  package:
+#    name: aptitude
+#    state: latest
 
 - name: Asterisk - Ensure all dependencies are resolved - CAN TAKE 5 MIN OR LONGER!
   shell: export DEBIAN_FRONTEND=noninteractive && ./contrib/scripts/install_prereq install


jvonau:
This does not really remove installing aptitude just moves the dependency back into `install_prereq`, you are missing the main part of the suggestion to use asterisk_dependencies.yml, after all that is what the name implies, in place of using install_prereq.

holta:
> This does not really remove installing aptitude just moves the dependency back into `install_prereq`, you are missing the main part of the suggestion to use asterisk_dependencies.yml, after all that is what the name implies, in place of using install_prereq.

Good point.

While also noting the people who maintain Asterisk's install_prereq seem to do a really good job.

So I strongly suspect we should rely on this well-maintained script of theirs going forward, instead of trying to reverse engineer and micromanage the dozens and dozens of requirements therein.

Yes they install the older `aptitude` rather than relying on the more modern apt ~= apt-get &mdash; but that's likely a small price to pay, to contain the larger Dependency Hell: https://github.com/asterisk/asterisk/blob/master/contrib/scripts/install_prereq#L262-L271

So it seems this PR #2902 might just be the right medicine after all ?  (A few more install tests should hopefully confirm!)

jvonau:
Now aptitude pulls in its own dependency chain altering what would be installed over a 'normal' base install I would suggest moving from 6-generic-apps where the new dependency might mask dependencies in other roles installing afterwards to be after calibre-web in 9-local-addons given this is the lone apache holdout.

holta:
> Now aptitude pulls in its own dependency chain altering what would be installed over a 'normal' base install I would suggest moving from 6-generic-apps where the new dependency might mask dependencies in other roles installing afterwards to be after calibre-web in 9-local-addons given this is the lone apache holdout.

Very good idea.  As Asterisk + FreePBX are humongous.

jvonau:
https://github.com/iiab/iiab/blob/master/roles/pbx/tasks/install.yml#L53-L56 should really be removed in light of the presence of build-essential within asterisk_dependencies.yml which would explain why 'make' is installed as a dependency of build-essential but otherwise build-essential would be installed via [install_prereq](https://github.com/iiab/iiab/pull/2896#issuecomment-892061982) anyway.

holta:
> https://github.com/iiab/iiab/blob/master/roles/pbx/tasks/install.yml#L53-L56 should really be removed

Good catch.  I'm now testing that (https://github.com/iiab/iiab/pull/2903/commits/e39aacbbab94bf2d7ffe89a4726ab66f269ff691) as part of PR #2903.

Of course Asterisk/FreePBX are gigantic so we can't exactly put them on a no-carb diet.

Still: this is absolutely tremendous progress &mdash; making [roles/pbx](https://github.com/iiab/iiab/tree/master/roles/pbx#pbx-readme) substantially more coherent + understandable + maintainable &mdash; so it'll actually be used!

jvonau:
still should be after calibre-web

holta:
> still should be after calibre-web

Really?  Ok, if that's truly best?

In the past our story was...that Calibre should be at the very end...due to testing branch wackiness.

But possibly nobody installs Calibre today.  In no small part b/c of its Eternal Dependency Hell (!)

jvonau:
if it the level of riskiness 
calibre-web
calibre might be better with newer distros?
pbx given the large number of devel packages pulled in by install_prereq

holta:
> if it the level of riskiness
> calibre-web
> calibre might be better with newer distros?
> pbx given the large number of devel packages pulled in by install_prereq

In the past I'd assumed that Calibre-Web should be installed after Calibre &mdash; so as to better take advantage of Calibre's own `/usr/bin/ebook-convert`

But possibly that's a non-issue?

e.g. if Calibre-Web very regularly monitors for such plugin-like programs, even if they appear some day later, after the fact.

jvonau:
https://github.com/iiab/iiab/blame/master/roles/calibre/tasks/install.yml should receive the axe apache treatment given just the port is now available.

> > if it the level of riskiness
> > calibre-web
> > calibre might be better with newer distros?
> > pbx given the large number of devel packages pulled in by install_prereq
> 
> In the past I'd assumed that Calibre-Web should be installed after Calibre — so as to better take advantage of Calibre's own `/usr/bin/ebook-convert`
> 
> But possibly that's a non-issue?
> 

Would that not depend on if calibre is selected to be installed?

> e.g. if Calibre-Web very regularly monitors for such plugin-like programs, even if they appear some day later, after the fact.

Would depend on if you require that functionality I would think, enduser's choice...

holta:
> Would that not depend on if calibre is selected to be installed?

Yes, I just want to make sure Calibre-Web finds `/usr/bin/ebook-convert` reliably, when the implementer has also installed Calibre.

But let's take the risk and see what happens: https://github.com/iiab/iiab/blob/master/roles/9-local-addons/tasks/main.yml#L22-L38 (that's PR #2903, now merged).

So [roles/calibre-web/README.rst](https://github.com/iiab/iiab/tree/master/roles/calibre-web#calibre-web-readme) can be cleaned up properly in short order.

jvonau:
Feels like a semi-firm dependency, could force calibre to install when calibre-web is selected to get the binary but not start the calibre service otherwise.

holta:
> get the binary but not start the calibre service otherwise

That could be very interesting.

Even if `/usr/bin/ebook-convert` might not be as powerful as https://pandoc.org, it's still an essential & powerful "legal scraping" tool in anybody's quiver.

_I suspect a lot of Calibre-Web users would be interested &mdash; and not just those using IIAB &mdash; if anybody knows how to make this happen !_

jvonau:
>apt-file search /usr/bin/ebook-convert
calibre: /usr/bin/ebook-convert           

>apt-file search /usr/bin/calibre-server
calibre: /usr/bin/calibre-server

apt install calibre is used in the calibre role but calibre-bin I think is the desktop widget having libqt5core5a as a dependency... Non desktops might benefit by omitting calibre-bin as calibre-server is provided by calibre and would drag in a desktop to meet the mentioned dependency of calibre-bin.

PR #2908, opened 2021-08-04
Does something like 'RestartSec=3' in /etc/systemd/system/freepbx.service help Asterisk start reliably? (WIP)

holta:
I don't know yet!

But @lemueldsouza and @jvonau may be able to guide us towards a reliable outcome?

Asterisk 18.x should certainly be stable after almost 10 months (is it possible FreePBX 16.x Beta is not yet stable in this very basic way?)

Refs:

- PR #2905 
- PR #2906

PR #2905, opened 2021-08-04
Update README.rst [roles/pbx for Asterisk & FreePBX]

lemueldsouza:
Draft changes, before we forget everything we've learned. It would be good to add pictures next. 
@holta please review and feel free to make changes as required

PR #2910, opened 2021-08-04
Revert 'RestartSec=3' in /etc/systemd/system/freepbx.service

holta:
It causes no harm, but far better to move forward towards a legit solution.

As discussed here:

- PR #2908

cc: @jvonau @lemueldsouza


holta:
Absolute worst case if we can't solve this in coming days we'll post a "Known Issue" in @lemueldsouza's pbx/README.rst here:

- PR #2905

PR #2916, opened 2021-08-06
Experiment: FreePBX on NGINX -- anybody can explore this, using flag 'pbx_try_nginx: True' [RENAMED TO... pbx_use_nginx: True] in /etc/iiab/local_vars.yml

holta:
This is @jvonau and @lemueldsouza's work, posted for others to try.  Building on:

- PR #2862 
- PR #2899
- PR #2911 
- PR #2913
- PR #2905
- #2914 
- PR #2915

Testing is ongoing.  This is a work-in-progress.

Issue #2914, opened 2021-08-05
FreePBX 15 + NGINX/IIAB quick recipe that worked for @lemueldsouza [let's gather Apache vs NGINX community experiences feedback]


holta:
Huge thanks to @lemueldsouza and @jvonau who've pushed so hard here &mdash; it seems like we're close:

1. Installed Asterisk 16.x and FreePBX 15.x and nodejs 12.x on Raspbian 10.
Followed steps from https://computingforgeeks.com/how-to-install-asterisk-16-with-freepbx-15-on-ubuntu-debian/
This worked fine, but was without IIAB, and nginx user was asterisk and not www-data.

2. Installed min version of IIAB on the same Raspberry Pi using [[IIAB's 1-line installer](http://download.iiab.io)] (I did not enable pbx as I had already installed it in step 1).

3. Added www-data to the asterisk group: `sudo usermod -a -G asterisk www-data`

4. Added another server entry/virtual host in /etc/nginx/server.conf to have freepbx listen on port 83:

   ```
   server {
       listen 83;
       listen [::]:83;

       root /var/www/html;
       index index.html index.htm index.nginx-debian.html index.php;
       server_name localhost;

       location / {
           try_files $uri $uri/ =404;
       }

       location ~ [^/]\.php(/|$) {
           fastcgi_split_path_info ^(.+?\.php)(/.*)$;

           if (!-f $document_root$fastcgi_script_name) {
               return 404;
           }

           fastcgi_pass unix:/run/php/php7.3-fpm.sock;
           fastcgi_index index.php;

           include fastcgi_params;
           fastcgi_param SCRIPT_FILENAME $document_root$fastcgi_script_name;
           fastcgi_param PATH_INFO $fastcgi_path_info;
           fastcgi_param PATH_TRANSLATED $document_root$fastcgi_path_info;
           fastcgi_param HTACCESS on;
           #fastcgi_param SERVER_NAME $host;
       }

       # disallows the things that the FreePBX .htaccess files disallow
       location ~ (/\.ht|/\.git|\.ini$|/libraries|/helpers|/i18n|/node|/views/.+php$) {
           deny all;
       }

       # from the api module .htaccess file
       rewrite ^/admin/api/([^/]*)/([^/]*)/?(.*)?$ /admin/api/api.php?module=$1&command=$2&route=$3 last;
   }
   ```

5. Changed back the entries in /etc/php/7.3/fpm/pool.d/www.conf to:

   ```
   user = www-data
[...]

PR #2850, opened 2021-07-07
WIP: Extricate IIAB from 2 web servers complexity [Apache -> NGINX]

holta:
Still Evolving.  To bring about a more sustainably focused software ecosystem using just 1 nervous system ([NGINX](https://github.com/iiab/iiab/tree/master/roles/nginx)).  Rather than 2 interneural/interwoven complexities (Apache + NGINX).  Deprecating takes time and a lot of patience even in the best of circumstances (imperial system, metric system, IPv6, etc).  Meantime:

- This is tested on 32-bit RaspiOS Lite (RPi 4).
- Testing is ongoing on 64-bit Ubuntu Desktop 20.04 (VM).

Eternal questions of course remain &mdash; as to when global learning communities will truly invest in the ongoing upkeep of lesser-used apps &mdash; e.g. Elgg and phpMyAdmin are the topics of the hour, whether or not they will be upgraded or gracefully put out to pasture.

But the very same questions of legit community impact apply far more broadly, than to just those 2 rather unmaintained apps/services.  Santa Claus does not provide glittering objects on demand.  And she too needs to eat.

Refs:

- #2762
- PR #2794
- PR #2832
- PR #2841 
- PR #2843 
- PR #2845 
- PR #2846 
- PR #2847 
- PR #2849

PR #2920, opened 2021-08-08
Install apache2 + its 'rewrite' mod in freepbx_dependencies.yml -- so roles/httpd can be phased out

holta:
Tested on Ubuntu Server 20.04 VM.  So that we're covered either way[*] &mdash; in preparation for:

- #2762
- PR #2850
- PR #2905

[*] Whether folks use `pbx_try_nginx: True` or `pbx_try_nginx: False` !  As outlined and implemented by:

- #2914
- PR #2916 
- PR #2917

### DIFF
--- a/roles/nginx/tasks/install.yml
+++ b/roles/nginx/tasks/install.yml
@@ -3,7 +3,7 @@
     name: "{{ apache_service }}"    # apache2 or httpd, per /opt/iiab/iiab/vars/<OS>.yml
     state: stopped
   ignore_errors: yes
-  # 'when: apache_installed is defined' insuff b/c mysql's php installs apache2
+  # 'when: apache_installed is defined' insuff b/c mysql's php (etc) can install apache2
 
 - name: 'Install 5 packages for NGINX: libnginx-mod-http-subs-filter, nginx-extras, php{{ php_version }}-fpm, uwsgi, uwsgi-plugin-python3'
   package:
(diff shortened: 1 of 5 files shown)


holta:
Just for the heck of it, fresh install testing of this PR #2920 is also underway on an RPi 4 running 64-bit Ubuntu Server 21.04, with FreePBX using Apache (default) instead of NGINX in this particular case.

(Just a reminder that FreePBX 16 cannot run on any OS with PHP < 7.4 -- which includes 32-bit RaspiOS.)

jvonau:
Go back to the complete user setup stanza exactly the way it was, the ansible user module is very finicky with the options selected, you have been warned. Make the user www-data addition to asterisk group conditional on pbx_try_nginx 

The installation of apache might fail as port 80 is already occupied by nginx when apt installs the package.

holta:
> ansible user module is very finicky with the options selected

No doubt.

Is there legit documentation that user 'asterisk' needs to be a system user?  (I don't see it, but if I'm looking in the wrong place LMK.)

Likewise, where do we have evidence that user 'asterisk' needs to be a member of groups 'audio' and 'dialout' ?  (There are bread crumbs from 2 decades ago, but so far I haven't seen anything that passes as authoritative.)

> Make the user www-data addition to asterisk group conditional on pbx_try_nginx

Good idea, and why not.  Done.

> The installation of apache might fail as port 80 is already occupied by nginx when apt installs the package.

Installing 'apache2' does not fail despite NGINX already using port 80, in 3 tests so far.

jvonau:
> > ansible user module is very finicky with the options selected
> 
> No doubt.
> 
> Is there legit documentation that user 'asterisk' needs to be a system user? (I don't see it, but if I'm looking in the wrong place LMK.)
> 
> Likewise, where do we have evidence that user 'asterisk' needs to be a member of groups 'audio' and 'dialout' ? (There are bread crumbs from 2 decades ago, but so far I haven't seen anything that passes as authoritative.)
> 

You are quite welcome to gain hands on experience with what works and what doesn't, why not just toss other's judgements out the window because you haven't discovered why the hard way yet. Go on continue, can't teach those who don't want to be taught. teacher's hat off, good luck.

holta:
There's circumstantial evidence that groups 'audio' and 'dialout' might still be relevant, still 2 decades later.  So there's an argument this might legitimately help a few implementers.

However adding user 'asterisk' as a system user appears to be a pure cargo cult.  But thankfully a relatively harmless cargo cult, likely having no ill effects.

jvonau:
How about stopping the malarkey about official howtos and keep what is working as is, if the setup is working with the apache webserver running as www-data keep the setup as is, we have already deviated from the official `sudo sed -i 's/^\(User\|Group\).*/\1 asterisk/' /etc/apache2/apache2.conf`and if it's working don't mess with it, we are one up on the others... Or this the apache install half-baked and we just are too naive to know it? Keep it objective, one variable change at a time.

PR #2926, opened 2021-08-09
[How should iiab-update-menus be integrated/automated into everyday workflows?] https://github.com/iiab/iiab/pull/2850#issuecomment-894821358

jvonau:
Not a bug but maybe a nice to have? https://github.com/iiab/iiab/pull/2850#issuecomment-894821358

### DIFF
--- a/roles/9-local-addons/tasks/main.yml
+++ b/roles/9-local-addons/tasks/main.yml
@@ -49,7 +49,7 @@
     path: /etc/iiab/install-flags/iiab-admin-console-complete
   register: iiab_admin_console_done
 
-# does this cobble user edits or removals?
+# does this clobber user edits or removals?
 - name: Run /usr/bin/iiab-update-menus when admin-console is installed and iiab-install was run
   command: /usr/bin/iiab-update-menus
   when: iiab_admin_console_done.stat.exists
--- a/roles/9-local-addons/tasks/main.yml
+++ b/roles/9-local-addons/tasks/main.yml
@@ -43,6 +43,17 @@
     state: present
   when: admin_console_install
 
+# could set the detection in local_facts.fact
+- name: Is iiab-admin-console installed by 'iiab'
+  stat:
+    path: /etc/iiab/install-flags/iiab-admin-console-complete
+  register: iiab_admin_console_done
+
+# does this cobble user edits or removals?
+- name: Run /usr/bin/iiab-update-menus when admin-console is installed and iiab-install was run
+  command: /usr/bin/iiab-update-menus
+  when: iiab_admin_console_done.stat.exists
+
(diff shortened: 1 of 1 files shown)


jvonau:
@tim-moody

holta:
Just another vote for making `/use/bin/iiab-update-menus` as ubiquitous as we can.

(Advertising & explaining it like crazy might be one way of doing that, certainly over the short term!)

Overnight ref:

- PR #2973

tim-moody:
running iiab-update-menus in iiab install does not make sense as menus are installed by admin console, so not necessarily present yet.

admin console will run iiab-update-menus when content changes and it can be run in invoke by clicking a button in the UI

jvonau:
> running iiab-update-menus in iiab install does not make sense as menus are installed by admin console, so not necessarily present yet.

Yup, fully agree, that is why there is a conditional test to see if admin-console is already installed.

> admin console will run iiab-update-menus when content changes and it can be run in invoke by clicking a button in the UI

Seems like new feature as there seems to be workflow that needed hand holding in the past, from the above (comment)

>Since renamed to 184-rpi4-u2104-srv-BIG-PBX-2850:
    cd /opt/iiab/iiab ; sudo ./iiab-configure is now running, as it upgrades itself from sub-MIN to BIG-sized.
    In about an hour, I'll run sudo /usr/bin/iiab-update-menus so that http://10.8.0.22 reflects this new reality.
 
If the need to 'click the button' goes away is that not one less thing for the user to do?

tim-moody:
>yup, fully agree, that is why there is a conditional test to see if admin-console is already installed.

sounds good. Be aware that the home menu is only updated if the flag to do so is set. it's a property of the menu that can be set through the gui.

>If the need to 'click the button' goes away is that not one less thing for the user to do?

if you take an action that would affect the home page, like add a zim or module, through the admin console it runs automatically. the button is only for the the cracks where it didn't get updated. maybe you copied a zim manually or installed a service with runrole.

>Seems like new feature as there seems to be workflow that needed hand holding in the past, from the above (comment)

I think there was a bug in update menus that was fixed some time ago.

tim-moody:
The role of update menus is that it should be used as an exception. if auto update was not checked and you decide to check it, then run update menus. otherwise there should be not need unless something fails.

Many people seem to get tripped up by attempting to update the home menu when the auto update flag is not checked. I think unchecked is the default on some presets and am thinking maybe checked should become the default.

On the OOB menu auto update flag is true.

holta:
> Many people seem to get tripped up by attempting to update the home menu when the auto update flag is not checked. I think unchecked is the default on some presets and am thinking maybe checked should become the default.

Sounds very reasonable!

holta:
FYI

- https://github.com/iiab/iiab-admin-console/pull/448

PR #2927, opened 2021-08-09
cleaner filesystem when apache is not installed

jvonau:
### Fixes bug:
don't pollute the filesystem when apache is not installed


jvonau:
The reason I left ports.conf in the nginx role is because I was going to do a conditional install of that file based upon the returned value from the systemd stop stanza, if 'changed' was returned (meaning it was installed and running) then install the ports file as it would be needed as an alternate to #2932.

Issue #2811, opened 2021-06-03
Test offline https/SSL/TLS NGINX with self-signed cert instead of LetsEncrypt.org ?


holta:
@tim-moody: @deldesir is interested if you or @jvonau make progress here figuring out what's possible?

For ZIM files from https://youzim.it and in general!

Ref: #2762

holta:
Four relevant tickets at https://github.com/kiwix/kiwix-hotspot :

- Added HTTPS support for all vhosts (PR 599, merged)
- Added HTTPS tutorial for SW ZIM (PR 603, merged)
  - Based on: Display tutorial to invite switching to HTTPs for SW based ZIM (546, closed)
- Hotspot does not read zimit-generated files (608)

tim-moody:
I tried the following:

```
openssl req -x509 -nodes -days 365 -newkey rsa:2048 -keyout /etc/ssl/private/nginx-selfsigned.key -out /etc/ssl/certs/nginx-selfsigned.crt

nano server.conf after listen 80;

listen 443 ssl;
      ssl_certificate /etc/ssl/certs/nginx-selfsigned.crt;
      ssl_certificate_key /etc/ssl/private/nginx-selfsigned.key;

```

The following on a medium vm were accessible over https:

•	Home
•	Kiwix
•	Module
•	Maps
•	Sugarizer
•	Nextcloud
•	WP
•	Calibre web

kalite was not accessible via the menu. It is not proxied and probably requires port 80. The problem is js-menu creates a link with the current protocol, so tried https when only http is available.

tim-moody:
I'm less sure about the kalite problem as it is not working with http on this vm, so there could be another issue.

jvonau:
kalite runs on port 8008 and never requires port 80, I'll go with the curl script was not used, now this machine has an incomplete kalite install https://github.com/iiab/iiab/issues/2863#issuecomment-887426453 due to the kalite role|iiab split of commands needed.

tim-moody:
>I'll go with the curl script was not used, now this machine has an incomplete kalite install 

probably right

PR #2928, opened 2021-08-09
kalite content script - How should kalite be integrated/automated into everyday workflows?

jvonau:
### Fixes bug:
https://github.com/iiab/iiab/issues/2863#issuecomment-887426453 #2811

### DIFF
--- a/roles/9-local-addons/tasks/main.yml
+++ b/roles/9-local-addons/tasks/main.yml
@@ -43,6 +43,15 @@
     state: present
   when: admin_console_install
 
+# Taken directly from 'iiab' has builtin checking via /etc/iiab/install-flags/ so safe to rerun. Now
+# 'iiab' could just reference this new file like scripts/ansible or the file can live in iiab-factory
+# use stage 9 so the stanzas from 'iiab' are used on the first pass, installs that add the kalite later
+# will get picked up during iiab-configure or ICO in admin-console but not preset's runroles but could
+# call this script to get the job done.
+- name: Install kalite support if iiab-install was completed but kalite was not installed on the first pass during 'iiab'
+  command scripts/kalite_support.sh
+  when: iiab_stage == 9 and kalite_installed is defined
+
 - name: Recording STAGE 9 HAS COMPLETED ====================
   lineinfile:
     path: "{{ iiab_env_file }}"
--- a/scripts/kalite_support.sh
+++ b/scripts/kalite_support.sh
@@ -1,0 +1,41 @@
+#!/bin/bash
+
+# K. KA Lite prep
+if [ -d /library/ka-lite ]; then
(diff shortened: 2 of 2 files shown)


jvonau:
@tim-moody

jvonau:
Taken directly from 'iiab' has builtin checking via /etc/iiab/install-flags/ so safe to rerun. Now 'iiab' could just reference this new file like how scripts/ansible currently is used. Used stage 9 as a conditional so the stanzas from 'iiab' are used on the first pass where the stage would be less than 9 when the install starts or continues. Installs that add the kalite role later will get picked up during iiab-configure or ICO in admin-console automatically but not preset's runroles nor runrole but one could run this script to get the job done.

tim-moody:
presets can include kalite and will install/enable it if not already done. 
will also install the appropriate language pack(s): EN and other.
can include list of nodes, such as khan/math and will install that content in required language

this is based on an underlying cmdsrv verb that has not been exposed other than in presets, but could be invoked with iiab-cmsrv-ctl

jvonau:
Um, think the point is being missed or ignored, for the most part the exact same code is carried in a couple of places, first in the [iiab](https://github.com/iiab/iiab-factory/blob/master/iiab#L455-L487) curl script, second since May [admin-console](https://github.com/iiab/iiab-admin-console/blob/master/roles/cmdsrv/files/scripts/kalite_install_lang.sh). Though it would be nice to have a common place that both routines could reference, extend the routine to be generally available for cmdline use and make the inner-workings clearer with a better readme for the kalite role.

tim-moody:
>Um, think the point is being missed or ignored

I don't usually look for a point in a question, so quite probably I missed the point.

I mistakenly thought the adm cons functionality was more recent than May. and more recent that this posting. 

>Installs that add the kalite role later will get picked up during iiab-configure or ICO in admin-console automatically but not preset's runroles nor runrole but one could run this script to get the job done.

was this the point and is it that this fixes a defect in presets?

jvonau:
> > Installs that add the kalite role later will get picked up during iiab-configure or ICO in admin-console automatically but not preset's runroles nor runrole but one could run this script to get the job done.
> 
> was this the point and is it that this fixes a defect in presets?

Depends on the if the preset was suppose to install the en.zip version also when it wrote kalite_install: True and should Spanish be be running on a different port?

tim-moody:
>should Spanish be be running on a different port?

I understand this to be a manual configuration in the case where both EN and ES kalite are desired, mostly a demo machine. I didn't think it could be done via ansible. Perhaps there are those who use a different port for ES even when not also installing EN.

If the preset includes kalite, EN language pack is always installed and a second language pack is optionally installed. (Non-EN language packs require EN as well.)

PR #2929, opened 2021-08-09
Cleaner filesystem when Apache is not installed, using roles/pbx/tasks/apache.yml + local_vars_unittest.yml

holta:
Building upon @jvonau's:

- PR #2927

Testing is underway.

Ref:

- PR #2850


holta:
Bare bones `/opt/iiab/iiab/vars/local_vars_unittest.yml` has been added to greatly speed up unit testing:

```
root@deb11-srv:/opt/iiab/iiab/vars# diff local_vars_min.yml local_vars_unittest.yml
33,34c33,34
< admin_console_install: True
< admin_console_enabled: True
---
> admin_console_install: False
> admin_console_enabled: False
59c59
< host_ssid: Internet in a Box
---
> host_ssid: unittest
105c105
< bluetooth_install: True
---
> bluetooth_install: False
118c118
< openvpn_enabled: False
---
> openvpn_enabled: True
120c120
< openvpn_handle: MIN-sized - Put Your Name Here
---
> openvpn_handle: unittest
232,233c232,233
< kalite_install: True
< kalite_enabled: True
---
> kalite_install: False
> kalite_enabled: False
240,241c240,241
< kiwix_install: True
< kiwix_enabled: True
---
> kiwix_install: False
> kiwix_enabled: False
250,251c250,251
< osm_vector_maps_install: True
< osm_vector_maps_enabled: True
---
> osm_vector_maps_install: False
> osm_vector_maps_enabled: False
285,286c285,286
< awstats_install: True
< awstats_enabled: True
---
> awstats_install: False
> awstats_enabled: False
309c309
< captiveportal_install: True
---
> captiveportal_install: False

**Comment by holta** [comment]
e.g. the currently running unit test (10.8.0.34) is elegantly summarized using:

```
root@deb11-srv:/opt/iiab/iiab# diff vars/local_vars_unittest.yml /etc/iiab/local_vars.yml
120c120
< openvpn_handle: unittest
---
> openvpn_handle: 194-deb11-srv-2929
352,353c352,353
< pbx_install: False
< pbx_enabled: False
---
> pbx_install: True
> pbx_enabled: True

holta:
> Bare bones `/opt/iiab/iiab/vars/local_vars_unittest.yml` has been added to greatly speed up unit testing

Companion PR also merged — to support more rapid/efficient unit testing of IIAB Apps: (roles, services)

- PR https://github.com/iiab/iiab-factory/pull/176

jvonau:
Could rename *unittest.yml to be *base.yml, the name is better, could be used for|with images|presets support, not just runrole testing.

holta:
> Could rename *unittest.yml to be *base.yml, the name is better, could be used for|with images|presets support, not just runrole testing.

Or `local_vars_core.yml` similar to Ansible which rejected to word 'base', in favor of the word 'core'.

Nothing to do with free-basing.  But yes definitely something to think over.

jvonau:
Think 'base' or 'core' is more accurate, 'unittest' looks like a single use case.

jvonau:
Again there is no PR/design discussion workflow being carried out, not even a tag for others to review the changes. I can see why apache.yml was added but fail to understand why freepbx_dependencies.yml. just had to go away in favor making freepbx.yml longer. FWIW, every install section could have a dependencies file, that would make commits more meaningful when looking for real changes with all the 'touchups' being bantered about.

holta:
> fail to understand why freepbx_dependencies.yml. just had to go away

In principle separate dependencies.yml files might (sometimes!) be a great idea, but when these only contain 1 command, they add complexity rather than clarity.

Certainly it's true that there remain other implied/implicit/unacknowledged spaghetti dependencies.  Even things like NGINX, and many others.

For sure, surfacing (reconciling, auditing, testing, documenting) all these unacknowledged and forgotten interdependencies can **_greatly_** help code maintainability.

But the flip side is that any good idea ([SoC](https://en.wikipedia.org/wiki/Separation_of_concerns) modularizing) can be taken too far &mdash; e.g. a blizzard of additional 1-command files (e.g. packages.yml or dependencies.yml, etc) that can easily subtract from rather than adding clarity.

jvonau:
You should run for office, you speak from both sides of your mouth to fit the current situation.
>a blizzard of additional 1-command files

You think it's better to just run a random new script to get the job done without fully understanding what that process is going to do? I can't speak for the original author but the pre-installing of packages that would be installed during the upstream installer script _**is**_ helpful to know that massive changes are going to be preformed and what the dependencies are upfront rather hiding them within some other process, makes chasing down things like why 'make' and 'aptitude'(another package manager) are now installed a bunch easier. I think the original intent was to install the same package set as what `install_prereq` installs but without including aptitude and dropping the call to install_prereq at some point.

holta:
> I think the original intent was to install the same package set as what `install_prereq` installs but without including aptitude and dropping the call to install_prereq at some point.

Ok.  I didn't realize that.  No question that [aptitude from the 1990s](https://en.wikipedia.org/wiki/Aptitude_(software)) is showing its age.

Meaningful code maintenance strategies are definitely the challenge here and elsewhere.

Personally I'm in favor of the upstream crew (Asterisk in that case) taking responsibility of dependency maintenance here ([install_prereq](https://github.com/asterisk/asterisk/blob/master/contrib/scripts/install_prereq) lines [21-35](https://github.com/asterisk/asterisk/blob/master/contrib/scripts/install_prereq#L21-L35) and [262-271](https://github.com/asterisk/asterisk/blob/master/contrib/scripts/install_prereq#L262-L271)).

PR #2932, opened 2021-08-11
Apache service [let's try to disable apache2 systemd service across the board, when pbx doesn't need it, even on desktop OS's like Debian 11!]

jvonau:
### Fixes bug:
fbf7141 is to address https://github.com/iiab/iiab/pull/2850#issuecomment-882093007
Disable apache for the next boot if installed.

### DIFF
--- a/roles/nginx/tasks/install.yml
+++ b/roles/nginx/tasks/install.yml
@@ -2,6 +2,7 @@
   systemd:
     name: "{{ apache_service }}"    # apache2 or httpd, per /opt/iiab/iiab/vars/<OS>.yml
     state: stopped
+    enabled: no
   ignore_errors: yes
   # 'when: apache_installed is defined' insuff b/c mysql's php (etc) can install apache2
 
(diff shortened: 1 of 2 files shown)


holta:
This is somewhat aggressive e.g. if desktop/graphical versions of Debian 11 had Apache enabled for a reason.

But now's the right time to see if Apache can be turned off, across the board during IIAB installs, without too much blowback.

Then if desktop/graphical OS folks complain too much that they needed Apache enabled for one reason or another, we can always reverse this later in coming months, if absolutely necessary.

- PR #2929

jvonau:
or supply the ports file to run apache on an alternate port or entertain the idea of running nginx on another port #2933

jvonau:
> This is somewhat aggressive e.g. if desktop/graphical versions of Debian 11 had Apache enabled for a reason.
>

As it was without this PR, nginx and apache would of been in a race for control of port 80 upon boot, only one of them can win that race. The user wanted 'iiab', the user gets the intended layout of 'iiab'.

holta:
> As it was without this PR, nginx and apache would of been in a race for control of port 80 upon boot, only one of them can win that race. The user wanted 'iiab', the user gets the intended layout of 'iiab'.

Good point.  We can slog thru those boot-time race conditions / timing issues if really nec.

But best to avoid them entirely if possible.

Indeed our multi-month efforts to disentangle years of legacy code no longer in use had more than one purpose &mdash; as outlined here:

- PR #2850

PR #2921, opened 2021-08-08
half of #2834 with a different name

jvonau:
### Fixes bug:
#2904 via local_vars.yml
### Description of changes proposed in this pull request:
new on demand variable
### Smoke-tested on which OS or OS's:
NA
Just because it meets half of the mentioned objective, suppress changing the hostname on demand

PR #2933, opened 2021-08-11
replacement for #2834

jvonau:
#2834


holta:
Related:

- #2921

jvonau:
Also related: #2876

jvonau:
#1743

jvonau:
Why is completing **_unfinished_** soft-coding that by itself would not alter any workflow considered to be for 8.0?

Issue #1802, opened 2019-07-01
MySQL (MariaDB 10.1.40) fails on every boot, Ubuntu Server 18.04 VM, fresh IIAB BIG-sized install [so http://box/mediawiki "Cannot access the database" etc]


holta:
This is a fresh install of BIG-sized IIAB 7.0 pre-release (http://download.iiab.io) on Ubuntu Server 18.04 (VM) that was rebooted per the install instructions.

Even 10 min after booting, http://10.8.0.30/mediawiki/ and http://192.168.0.x/mediawiki/ keep showing:
```
Sorry! This site is experiencing technical difficulties.
Try waiting a few minutes and reloading.

(Cannot access the database)
```

@tim-moody do you have suggestions?

holta:
@tim-moody can you tell what might be causing these underlying MariaDB failures on 10.8.0.30, by logging in?  (Ubuntu Server 18.04 VM, with BIG-sized IIAB 7.0 pre-release freshly installed)

```
root@box:~# systemctl status mysql
● mariadb.service - MariaDB 10.1.40 database server
   Loaded: loaded (/lib/systemd/system/mariadb.service; enabled; vendor preset: enabled)
   Active: failed (Result: timeout) since Sun 2019-06-30 23:22:49 EDT; 2min 39s ago
     Docs: man:mysqld(8)
           https://mariadb.com/kb/en/library/systemd/
  Process: 639 ExecStart=/usr/sbin/mysqld $MYSQLD_OPTS $_WSREP_NEW_CLUSTER $_WSREP_START_POSITION (code=exited, status=0/SUCCESS)
  Process: 450 ExecStartPre=/bin/sh -c [ ! -e /usr/bin/galera_recovery ] && VAR= ||   VAR=`/usr/bin/galera_recovery`; [ $? -eq 0 ]   && systemctl set-environment _WSREP
  Process: 447 ExecStartPre=/bin/sh -c systemctl unset-environment _WSREP_START_POSITION (code=exited, status=0/SUCCESS)
  Process: 430 ExecStartPre=/usr/bin/install -m 755 -o mysql -g root -d /var/run/mysqld (code=exited, status=0/SUCCESS)
 Main PID: 639 (code=exited, status=0/SUCCESS)
   Status: "MariaDB server is down"

Jun 30 23:19:27 box.lan systemd[1]: Starting MariaDB 10.1.40 database server...
[...]

PR #2936, opened 2021-08-11
The OSs have finally caught up - the default timeout value is 900 seconds now

jvonau:


holta:
Related:

- #1802

holta:
Let's risk it and evaluate closely all IIAB install tests over the coming days.

holta:
Just FYI this PR #2936 was further clarified with a paragraph on Lines 14-19 here: https://github.com/iiab/iiab/blob/d2064f0ef07089439bd88562a9d03b5360279d0a/roles/mysql/tasks/install.yml#L14-L19

Issue #2924, opened 2021-08-08
Review phpMyAdmin security + practicality


holta:
Who can help review https://github.com/iiab/iiab/tree/master/roles/phpmyadmin so that it serves a clear + pragmatic public purpose?

Thanks!

(Worst case if nobody is willing, this role should be deprecated in coming weeks.)

Refs:

- #2122
- PR #2712
- PR #2923
- http://wiki.laptop.org/go/IIAB/Security

cc: @georgejhunt, @tim-moody, @jvonau, @deltacloud9

holta:
Does IIAB really need this?

_"phpMyAdmin is a free software tool written in PHP, intended to handle the administration of MySQL over the Web.  phpMyAdmin supports a wide range of operations on MySQL and MariaDB.  Frequently used operations (managing databases, tables, columns, relations, indexes, users, permissions, etc) can be performed via the user interface, while you still have the ability to directly execute any SQL statement."_

Please weigh in.

I've heard ambiguous sentiments from @georgejhunt and @jvonau.

Either decision is fine, but we need a decision.

georgejhunt:
I have found it useful. But security concerns make it hazardous. Let's drop
it.

jvonau:
What does 2948 have to do with this subject? Keep the cross linking directly relevant please.

holta:
Deprecating stale software appears to be the very hardest[*] work of all.

In that sense, these are intimately related, from a software lifecycle perspective:

- #2948

[*] And possibly the most important.  As well as the most controversial.

jvonau:
#2122 #2712 My personal choice is to never install the role anyway, made a note about maybe using the socket for the connection in the PR. If nobody want to keep an eye on the role the lets drop it and move on.

holta:
As a start, I've removed phpMyAdmin from https://github.com/iiab/iiab/wiki/IIAB-7.2-Release-Notes

holta:
Villagers in remote northern India apparently need this.

Please see the ongoing discussion here:

- #2712
- #2812

Issue #2525, opened 2020-09-15
Validate vars errors are non viewable in Admin Console -> Server Utilities -> Display Job Status


jvonau:
### Steps to reproduce the problem:
Use invalid combinations in local_vars.yml, when the error hits ref: #2502
The error message goes flying by, have to scroll back to view, and is not clearly visible at the end of the logfile. 
I do all of my installs over ssh, so no issue for me, but scrolling back at a keyboard with a stage2 install while "iiab" was typed after a bad edit to local_vars.yml can be problematic.

jvonau:
Here is one example using admin-console, note one cannot scroll the feedback. 
![squid_fail](https://user-images.githubusercontent.com/5579710/114728836-135a1480-9d05-11eb-9ade-68d0e6a925ed.png) induced this by selecting (Squid Cache) in
![squid_select](https://user-images.githubusercontent.com/5579710/114729641-c9256300-9d05-11eb-8b21-205e3516825d.png)
Being able to "enable" something that is not "installed" is a bug. Should I file an issue for this if so where here or iiab-admin?

tim-moody:
>Being able to "enable" something that is not "installed" is a bug. Should I file an issue for this if so where here or iiab-admin?

was install True? I admit to not checking the installed state, but if install is false a service should not appear, but not sure about some of these old ones.

was the inconsistency edited manually or through Admin Console?

tim-moody:
>Here is one example using admin-console, note one cannot scroll the feedback.

are you saying the scroll bar under results does not work?

jvonau:
> > Here is one example using admin-console, note one cannot scroll the feedback.
> 
> are you saying the scroll bar under results does not work?

Gives very little past history, same for completed jobs

tim-moody:
I'll have to check. I vaguely remember truncating it to improve performance, but I thought that was only before the job terminates.

tim-moody:
Really needs reworking to make it expandable on demand rather than pulling a huge history every time.

jvonau:
> > Being able to "enable" something that is not "installed" is a bug. Should I file an issue for this if so where here or iiab-admin?
> 
> was install True? I admit to not checking the installed state, but if install is false a service should not appear, but not sure about some of these old ones.
>
Install was False
 
> was the inconsistency edited manually or through Admin Console?

Editing strictly by the admin-console only.

I have a mockup to expose the *_install variables for this pane
[...]

Issue #1879, opened 2019-07-16
ISSUE WHILE SAVING PERMITTED URLS [Admin Console, Squid]


vinaybissa:
### Expected behavior and actual behavior.
Expected Behavior could be while saving the permitted urls it should whitelist those urls, but it return with error that SET -WHLIST generate a command error
### Steps to reproduce the problem.
After logging in to admin console and setting the permitted url it does not allow to save those url it returns with a error alert. (I have already set the IIAB GATEWAY True in Local Vars File)
### Specifications like the version of the IIAB, operating system version, or hardware details.
IIAB 7.0, Ubuntu 18.04 LTS, INTEL NUC

holta:
Thanks @vinaybissa for the report.

Can you provide the exact clicks/actions you took within IIAB's Admin Console (http://box.lan/admin) so others can reproduce this?

vinaybissa:
Step 1 - I opened the admin console with credentials (http://box.lan/admin)
Step 2 - I Clicked on Configure from Menus
Step 3 - From left navigation pane, i clicked on Edit Permitted Urls
Step 4 - In the section where i can enter the permitted urls, i entered one url (eg. http://www.domian.com)
Step 5- Then i Clicked on Update Permitted URls list

After this step i got error in alert box like this (SET-WHLIST FAILED and reported Unexpected error in Command SET-WHLIST)

m-anish:
Is squid/dansguardian installed?

I haven't been down that road in a while, so it may well be broken. 

Can you paste the output of:
```

/etc/iiab/iiab.env
/etc/iiab/iiab.ini
/etc/iiab/local_vars.yml
```

m-anish:
tagging @jvonau @tim-moody

m-anish:
FWIW, it looks like it may easily be possible with dnsmasq (if we're not already using it)
https://superuser.com/questions/625885/dns-block-all-websites-except-for-whitelisted-sites

vinaybissa:
Hi Anish - Seems SQUID/ DANSGUARDIAN Both are installed, PFA for your reference.

Regards
Vinay
________________________________

vinaybissa:
############ OUTPUT OF IIAB.ENV###########################
# This is a configuration file for IIAB
# It can be sourced in a shell script or read into an application
IIAB_BASE_PATH=/opt/iiab
IIAB_DIR=/opt/iiab/iiab
IIAB_RELEASE=7.0
IIAB_REVISION=0
[...]

Issue #1367, opened 2019-01-09
Set squid_install: & squid_enabled: False in BIG-sized local_vars.yml ?


holta:
Squid is the source of so many ongoing/costly problems (e.g. https://github.com/iiab/iiab/issues/1323#issuecomment-452574543 is just the latest) that I hope we can please discuss this during Thursday's http://minutes.iiab.io community/team call (and on ticket here!)

PR #2948, opened 2021-08-14
Deprecate DansGuardian & WonderShaper + overhaul Squid -- in future let's consider fortifying, removing or replacing Squid (some prefer Varnish as a proxy/cache)

holta:
These have not been used in about half a decade.

At a certain point carrying around unmaintained legacy code becomes dangerous.

It's time for some legit pruning.

Testing is ongoing to verify.

Related:

- #6
- #1323
- #1367
- #1879
- PR #1893
- #2525
- #2924


holta:
Background: IIAB has made headway removing these 23 roles...

...but needs to go further removing legacy bitrot &mdash; that in the end gets in everybody's way &mdash; and sometimes can be downright dangerous.

SEE ALSO: https://github.com/holta/iiab/blob/deprecate-wondershaper-etc/tests/test.yml#L84-L95

jvonau:
Might be premature, #1879 is hardly a half a decade and can't say unmaintained, rather under used in field or it works ok with no feedback needed.

jvonau:
Given I'm the maintainer of the network code please remove the debug stanzas, it's not my style to leave that laying around in the code.

holta:
Tested with a BIG-sized IIAB installed onto a 32-bit Raspberry Pi OS on RPi 4 (10.8.0.42) with these 4 variables in [/etc/iiab/local_vars.yml](http://faq.iiab.io/#What_is_local_vars.yml_and_how_do_I_customize_it.3F) changed from False to True:

```
squid_install: True
squid_enabled: True
gw_squid_whitelist: True    # Only HTTP sites can be blocked, not HTTPS sites
gw_block_https: True
```

jvonau:
Thought one has to understand the underling software stack to write documentation. #1323 #1367

jvonau:
In [#1779](https://github.com/iiab/iiab/pull/1779/files#diff-0601aa25f583c27785a361849e35797da9633fc2903f1bd6f86766cd2450a272L144) a regression was introduced in that HTTPCACHE_ON is not written to /etc/iiab/iiab.env causing a variable to be absent that [iiab-gen-iptables](https://github.com/iiab/iiab/blob/master/roles/network/templates/gateway/iiab-gen-iptables#L206) expects to always be defined. Hasn't caused any issues seeing that it's the last command, -e isn't set.

jvonau:
> Tested with a BIG-sized IIAB installed onto a 32-bit Raspberry Pi OS on RPi 4 (10.8.0.42) with these 4 variables in [/etc/iiab/local_vars.yml](http://faq.iiab.io/#What_is_local_vars.yml_and_how_do_I_customize_it.3F) changed from False to True:
> 
> ```
> squid_install: True
> squid_enabled: True
> gw_squid_whitelist: True    # Only HTTP sites can be blocked, not HTTPS sites
> gw_block_https: True
> ```

Block 'gw_block_https: True' would only be effective when `iiab_gateway_enabled: True` is set but would otherwise allow all other connections to the internet on other ports.

jvonau:
Think the [user creation](https://github.com/iiab/iiab/pull/2948/files#diff-1479e35d0a0d888bc928d5072b7e85334efaeb1f636b809197541e1d8339e4f4R19-R22) can go away, should be handled by the apt package when installed. Has roots to similar code #6 that was used in the now dead apache role from years ago.

holta:
> [iiab-gen-iptables](https://github.com/iiab/iiab/blob/master/roles/network/templates/gateway/iiab-gen-iptables#L206) expects [$HTTPCACHE_ON] to always be defined.

Expects but doesn't require!

If the undefined var had not been surrounded by quotes, that line would indeed have failed:

```bash
pi@box:~ $ if [ $SANTACLAUS == "True" ]; then echo CELEBRATE; fi
-bash: [: ==: unary operator expected
pi@box:~ $ [ $SANTACLAUS == "True" ]
-bash: [: ==: unary operator expected
pi@box:~ $ [ $SANTACLAUS = "True" ]
-bash: [: =: unary operator expected
```

With quotes or double brackets however, it never fails:

```bash
pi@box:~ $ [ "$SANTACLAUS" = "True" ]
pi@box:~ $ [[ $SANTACLAUS = "True" ]]
pi@box:~ $ [ "$SANTACLAUS" == "True" ]
pi@box:~ $ [[ $SANTACLAUS == "True" ]]
pi@box:~ $
```

So it's easy...to deceive:

```bash
pi@box:~ $ if [ "$SANTACLAUS" != "False" ]; then echo CELEBRATE; fi
CELEBRATE
```

Likewise undefined variables, when tested against each other, can *(very confusingly appear to)* contain the same value:

```bash
pi@box:~ $ if [ "$CHRISTMAS" = "$RAMADAN" ]; then echo CELEBRATE; fi
CELEBRATE
```

holta:
> Think the [user creation](https://github.com/iiab/iiab/pull/2948/files#diff-1479e35d0a0d888bc928d5072b7e85334efaeb1f636b809197541e1d8339e4f4R19-R22) can go away, should be handled by the apt package when installed. Has roots to similar code #6 that was used in the now dead apache role from years ago.

That's what I thought.

Turns out we were both wrong.

I added brief explanations, to make maintenance more clear, should we decide to preserve Squid: https://github.com/iiab/iiab/pull/2948/commits/e2bf2578e2f1f47f2832fefc024a5a361d51bd5f

jvonau:
```
sudo iiab-gen-iptables
snip
+ '[' False == True ']'
+ /sbin/iptables -A FORWARD -i br0 -o eth0 -j ACCEPT
+ /sbin/iptables -A FORWARD -i eth0 -o br0 -j DROP
+ echo 1
+ '[' 3 -lt 5 ']'
+ /sbin/iptables -A INPUT -i eth0 -j DROP
+ '[' False == True ']'
+ '[' '' == True ']'
+ netfilter-persistent save
run-parts: executing /usr/share/netfilter-persistent/plugins.d/15-ip4tables save
run-parts: executing /usr/share/netfilter-persistent/plugins.d/25-ip6tables save
+ exit 0
```
The behavior is correct but the point of having the variable written in the first place was to show the relationship between iiab-gen-iptables, /etc/iiab/iiab.env, iiab-hotspot-on|off iiab-iiab-internet-on|off and the network role. Now with squid being optional absent in iiab.env would mean the squid part of the network role was never selected, ok I'll go with that. 

Guess I should explain what block_DNS does, it uses the same intercept like rule as squid uses but for 53 in place of 80, this should intercept from a client on the LAN any dns request out to the internet lets say the client is using 8.8.8.8 as nameserver in the clients config to our own nameserver now dnsmasq. Back in the xsce days one of the deployments setup a nice forwarding dns layout where the lookups could go to specific servers for spam/blocklist/porn scraping without having to set that up locally but would be easily defeated with the client not using our dhcp's server supplied setting to the LAN clients for nameserver to use.

jvonau:
> Turns out we were both wrong.
> 
> I added brief explanations, to make maintenance more clear, should we decide to preserve Squid: [e2bf257](https://github.com/iiab/iiab/commit/e2bf2578e2f1f47f2832fefc024a5a361d51bd5f)

In a 'stock out of the box sample of config files' where does 'cache_dir' point to? Just thinking why are we defining this in /library, that is a holdover from the original schoolserver where at that point in time squid was known to eat all the available freespace on whatever partition the cache was hosted on so it was moved to /library with nothing much more residing on /library to keep the system from locking up when without the move / would of otherwise been filled up.

holta:
This PR was tested in 2 scenarios to salvage[*] Squid:

- Squid 4.6 on 32-bit Raspberry Pi OS Lite
- Squid 4.13 on 64-bit Ubuntu Server 21.04

(Both setups / OS's on were tested on Raspberry Pi 4 hardware.)

[*] Squid is now salvaged, with a very basic security audit being a start (thanks @jvonau for probing) as part of this PR's quite substantial overhaul of `/etc/squid/squid.conf` etc &mdash; which was necessary for several reasons:

1. Squid had not been working for years, with the old `/etc/squid/squid-iiab.conf` completely ignored by Squid, which was ignoring both of IIAB's attempts to point it towards this alternate .conf file.
2. To make matters worse, the contents of squid-iiab.conf had become far from workable, including many obsolete and deprecated settings from the OLPC era (beginning of 2007) that make no sense a decade and a half later.

Hence this PR quite radically overhauls IIAB's Squid cache and basic allowlist / whitelist functionality &mdash; so as to allow communities to at least evaluate the big questions of the day &mdash; _what is most desired going forward, during this 3rd decade of XS / XSCE / IIAB?_

holta:
> In a 'stock out of the box sample of config files' where does 'cache_dir' point to?

Officially, the default is to use memory not disk:

_"Default Value:  No disk cache. Store cache ojects only in memory."_
( According to http://www.squid-cache.org/Doc/config/cache_dir/ )

This is software originally released more than a quarter century ago.

They still seems to emphasize/recommend the owner/operator craft their own disk layout / account / etc...

> Just thinking why are we defining this in /library, that is a holdover from the original schoolserver where at that point in time squid was known to eat all the available freespace on whatever partition the cache was hosted on so it was moved to /library with nothing much more residing on /library to keep the system from locking up when / filled up.

I agree: `/library/cache` might not be the best place for this.

jvonau:
>http_port 3129
http_port 3128 intercept

I take it 3128 is intended for the transparent proxy with 3129 being a standard proxy?
What is implied by "Squid auto-creation of cache_dir (or the old way, 'squid -z') both fail:"?
cache_effective_user seems to imply that should maybe be set to 'squid' in the squid.conf?

holta:
> > http_port 3129
> > http_port 3128 intercept
> 
> I take it 3128 is intended for the transparent proxy with 3129 being a standard proxy?

Apparently `transparent` began to be deprecated in 2010.  Using `intercept` is similar, but requires both `http_port` lines.

My assumption is that port 3128 is for the LAN side, and port 3129 for the WAN/Internet side, but I might not have that quite right.

In any case, I've recorded both to https://github.com/iiab/iiab/wiki/IIAB-Networking#list-of-ports--services

> What is implied by "Squid auto-creation of cache_dir (or the old way, 'squid -z') both fail:"?
> cache_effective_user seems to imply that should maybe be set to 'squid' in the squid.conf?

When you say 'squid', do you mean the user `proxy_user: proxy` ?

As mentioned in [network/tasks/squid.yml#L25-L26](https://github.com/holta/iiab/blob/deprecate-wondershaper-etc/roles/network/tasks/squid.yml#L25-L26), neither directive (cache_effective_user, cache_effective_group) is recommended.

(I'd prefer that we stick with OS maintainers' choices.  Unless strong evidence emerges that these truly should and need to be overridden.)

holta:
> This PR was tested in 2 scenarios to salvage[*] Squid:
> 
> * Squid 4.6 on 32-bit Raspberry Pi OS Lite
> * Squid 4.13 on 64-bit Ubuntu Server 21.04

Squid 4.13 just happens to be significant, as Debian Bullseye and Raspberry Pi OS 11 will both probably be running that version for the next 2 years:

https://packages.debian.org/bullseye/squid

jvonau:
> > > http_port 3129
> > > http_port 3128 intercept
> > 
> > 
> > I take it 3128 is intended for the transparent proxy with 3129 being a standard proxy?
> 
> Apparently `transparent` began to be deprecated in 2010. Using `intercept` is similar, but requires both `http_port` lines.
> 
> My assumption is that port 3128 is for the LAN side, and port 3129 for the WAN/Internet side, but I might not have that quite right.
> 
> In any case, I've recorded both to https://github.com/iiab/iiab/wiki/IIAB-Networking#list-of-ports--services
> 
> > What is implied by "Squid auto-creation of cache_dir (or the old way, 'squid -z') both fail:"?
> > cache_effective_user seems to imply that should maybe be set to 'squid' in the squid.conf?
> 
> When you say 'squid', do you mean the user `proxy_user: proxy` ?
> 
> As mentioned in [network/tasks/squid.yml#L25-L26](https://github.com/holta/iiab/blob/deprecate-wondershaper-etc/roles/network/tasks/squid.yml#L25-L26), neither directive (cache_effective_user, cache_effective_group) is recommended.
> 
> (I'd prefer that we stick with OS maintainers' choices. Unless strong evidence emerges that these truly should and need to be overridden.)

I'm of the 'if you broke it you bought it' frame of mind, given it's not on the default install path do as you wish.

jvonau:
As an improvement ports_externally_visible=3 could live in iiab.env, that would allow external adjustment without the need to run the complete network role to alter the variable in iiab-gen-iptables, just edit iiab.env and iiab-internet-on|off.

jvonau:
> > Think the [user creation](https://github.com/iiab/iiab/pull/2948/files#diff-1479e35d0a0d888bc928d5072b7e85334efaeb1f636b809197541e1d8339e4f4R19-R22) can go away, should be handled by the apt package when installed. Has roots to similar code #6 that was used in the now dead apache role from years ago.
> 
> That's what I thought.
```
ubuntu@box:/opt/iiab/iiab$ grep proxy /etc/passwd
proxy:x:13:13:proxy:/bin:/bin/false
```
Tells me this is a very old and longstanding name uid gid [assignment from 2011](https://community.spiceworks.com/topic/127693-default-linux-accounts), I'll bet that the account is present in a clean out of the box install(money is on this one) or if not then apt created the user on demand when squid was asked to be installed.
```
ubuntu@box:/opt/iiab/iiab$ cat /usr/lib/systemd/system/squid.service
## Copyright (C) 1996-2020 The Squid Software Foundation and contributors
##
## Squid software is distributed under GPLv2+ license and includes
## contributions from numerous individuals and organizations.
## Please see the COPYING and CONTRIBUTORS files for details.
##

[Unit]
Description=Squid Web Proxy Server
Documentation=man:squid(8)
After=network.target network-online.target nss-lookup.target

[Service]
Type=notify
PIDFile=/run/squid.pid
Group=proxy
RuntimeDirectory=squid
RuntimeDirectoryMode=0775
ExecStartPre=/usr/sbin/squid --foreground -z
ExecStart=/usr/sbin/squid --foreground -sYC
ExecReload=/bin/kill -HUP $MAINPID
KillMode=mixed
NotifyAccess=all

[Install]
WantedBy=multi-user.target
```
This was problematic as who is running the show as there also is present /etc/init.d/squid as part of the stock apt install also but is now ignored you are saying but that can vary depending on how systemd is configured to handle /etc/init.d which can vary between distros. Once upon a time the systemd unit file just called /etc/init.d/squid and iiab-conf was respected but that can go by the wayside, Group=proxy tells me that proxy group name is in use. I'll bet there is not a 'changed' response on that stanza to create the user, could you pastebin iiab-install.log please.

jvonau:
> > This PR was tested in 2 scenarios to salvage[*] Squid:

Well if the iiab-conf was not being read then the stock distribution configuration would of been used, when that changed I'm not sure cause it's a seldom used option but you noted that the disk caching was not enabled out of the box the service should of started right up but without the custom config. What kind of disk limits on the size of the cache was set?

jvonau:
How is the new squid.conf being used? I don't see it being copied into place so there would be no effective change in behavior. Are you implying that disk caching is not yet working?

holta:
```
 proxy:x:13:13:proxy:/bin:/bin/false
```

Good point.  IIAB's [squid.yml](https://github.com/iiab/iiab/blob/master/roles/network/tasks/squid.yml#L17-L21) isn't really creating pre-installed user `proxy` &mdash; but rather just changing the rightmost field of /etc/passwd (user's shell path) from Debian/Ubuntu norm `/usr/sbin/nologin` to UNIX/Linux norm `/bin/false`

Last but not least, this stanza helps enforce a clear + common + visible Squid mandate across OS's.  No harm is done.  Stating these assumptions upfront and on-top is really important for software maintenance.  So I'll mention this useful observation inline.

As a bonus, it's great that norms across Linux distros are (a bit) more stable and consistent as compared to 10-20 years ago.

holta:
Just FYI Debian 11's `squid.service` systemd unit file is 100% identical to the one from Ubuntu, that @jvonau pasted into this ticket ~5 hours ago.

And with Raspberry Pi OS 10 / Buster going away in not so many weeks or days, its own `squid.service` unit file doesn't much matter, but here it is while it lasts: _(this deviates from Ubuntu/Debian above, but still works!)_

```
root@box:~# more /usr/lib/systemd/system/squid.service
## Copyright (C) 1996-2019 The Squid Software Foundation and contributors
##
## Squid software is distributed under GPLv2+ license and includes
## contributions from numerous individuals and organizations.
## Please see the COPYING and CONTRIBUTORS files for details.
##

[Unit]
Description=Squid Web Proxy Server
Documentation=man:squid(8)
After=network.target network-online.target nss-lookup.target

[Service]
Type=forking
PIDFile=/var/run/squid.pid
ExecStartPre=/usr/sbin/squid --foreground -z
ExecStart=/usr/sbin/squid -sYC
ExecReload=/bin/kill -HUP $MAINPID
KillMode=mixed

[Install]
WantedBy=multi-user.target
root@box:~#
```

holta:
> /etc/init.d/squid and iiab-conf

Also being ignored across all IIAB distros was... `/etc/sysconfig/squid` that [squid.yml](https://github.com/iiab/iiab/blob/master/roles/network/tasks/squid.yml#L61-L62) had been installing from https://github.com/iiab/iiab/blob/master/roles/network/templates/squid/squid.sysconfig.unused

My guess is that all this init.d stuff (and /etc/sysconfig which may be a Red Hat legacy?) was ignored starting 2019-07-06 when Debian 10 Buster was released, and then even more so after 2020-04-23 when Ubuntu 20.04 was released.

> I'll bet there is not a 'changed' response on that stanza to create the user, could you pastebin iiab-install.log please.

Recap: as outlined above, Ansible changed user proxy's shell path (gratuitously, no doubt) but this explains why /opt/iiab/iiab/iiab-install.log shows `changed` instead of `skipping` there.

holta:
> service should of started right up but without the custom config.

Sure it started pointlessly, doing nothing useful with the default (nearly empty, security-oriented) `/etc/squid/squid.conf`

> What kind of disk limits on the size of the cache was set?

There's no real answer to this question, as the default is for cache_dir to use memory not disk, as mentioned above: https://github.com/iiab/iiab/pull/2948#issuecomment-899801155

holta:
> How is the new squid.conf being used? I don't see it being copied into place so there would be no effective change in behavior.

`/etc/squid/squid.conf` is installed here:

https://github.com/iiab/iiab/blob/a8114e65dd2f7d81902f083cb07b7b27b20fab4b/roles/network/tasks/enable_services.yml#L150-L157

holta:
> Are you implying that disk caching is not yet working?

No.  Disk caching is restored ~2 years later, as a result of IIAB's new `/etc/squid/squid.conf` installed from this template:
https://github.com/iiab/iiab/blob/e2bf2578e2f1f47f2832fefc024a5a361d51bd5f/roles/network/templates/squid/squid.conf.j2#L10-L30

(Whether people really need this approach is another question, that may be much more important...)

jvonau:
> (Whether people really need this approach is another question, that may be much more important...)

OMG and LOL, lot of irons in the fire, forgot where I put that stanza for the squid.conf file, was a volatile file that had two setting to handle dansguardian. Given the location where the conf file is put into place it would be *very easy* to make the disk cache optional.

jvonau:
Still don't see the need to alter the distro provided default shell, most of the listings in /etc/passwd are using nologin why deviate from the norms as provided by the OS? 
>(I'd prefer that we stick with OS maintainers' choices. Unless strong evidence emerges that these truly should and need to be overridden.)

PR #2906, opened 2021-08-04
Spring Cleaning: pbx, nodejs, openvpn, local_vars, default_vars.yml

holta:
Tightens up pbx/tasks/asterisk.yml and pbx/tasks/freepbx.yml etc.

Thanks to @jvonau and @lemueldsouza suggestions.  cc: @m-anish

In conjunction with ongoing tuneups to our explanation of **all 3 php.ini files** here:
https://github.com/iiab/iiab/blob/master/roles/www_options/tasks/main.yml#L53-L133

Tested on Ubuntu Server 20.04 VM.

Refs:

- PR #2903
- PR #2905

PR #2644, opened 2020-11-23
Clean/Shorten openvpn/templates/announcer.j2

holta:
To improve maintainability in preparation for [OpenVPN 2.5.0](https://openvpn.net/community-downloads-2/) ("major release with many new features", "Connections setup is now much faster") etc.

Refs: #318, #1124, #2361, PR #2565, PR #2643

Issue #2361, opened 2020-04-25
install wipes my openvpn_handle


tim-moody:
during install my openvpn_handle was renamed to openvpn_handle.6756.2020-04-25@16\:58\:40~ (by backup I assume) and the original was wiped (no value in local vars). 

the template directive should be force: no

holta:
FWIW Ansible variable openvpn_handle has always been the "master location" to store this value.

It sounds like you're proposing to change the master location to be `/etc/iiab/openvpn_handle` instead (if it exists!)

FYI things like /usr/bin/iiab-support (sourced from https://github.com/iiab/iiab/blob/master/iiab-support) would have to be re-architected if indeed this design change moves forward.

tim-moody:
I guess I never hit this before, because I add a handle to images after they are built, but replacing a non-empty openvpn_handle file with an empty one doesn't make much sense to me.

jvonau:
> during install my openvpn_handle was renamed to openvpn_handle.6756.2020-04-25@16:58:40~ (by backup I assume) and the original was wiped (no value in local vars).
> 
> the template directive should be force: no

Any reason to force a freshly created image back though 1-prep? Or to use runrole --reinstall openvpn? install.yml is where the template is copied into place. You didn't by chance remove or alter /etc/iiab/iiab_state.yml? Provisioning a new image should not really need to reinstall anything, fiddle a few files/services maybe but no real need to start at stage 1. Can you go into more detail about your provisioning routine, might be able to understand better what steps you are taking.

tim-moody:
I have started creating pre-install images and probably put a handle in the prep code. You could also need to do a software upgrade. What would be the reason to overwrite a user's file with an empty file under any circumstances? What would break if the one line change I proposed were applied?

holta:
If as part of the IIAB 7.2 release cycle we rethink and redesign how-and-where OpenVPN's handle should be stored (thruout the complete life cycle of diverse IIAB machines' installation->implementation->ops->renaming/moving to another school, etc) we should keep in mind not only the future[*] (but also the past/evolving mechanics) of this de facto IIAB naming scheme.  Here is one example:

```
# openvpn_handle variable can no longer be left completely undefined [as] of August 2018 (EMPTY STRING "" IS TOLERATED, in which case OpenVPN server should use /etc/iiab/uuid in lieu of the handle)
```
From: https://github.com/iiab/iiab/blob/master/roles/openvpn/tasks/main.yml#L45

[...]

PR #2957, opened 2021-08-19
Major cleanup of roles/openvpn, openvpn_handle SSOT

holta:
Testing is ongoing to verify.  Building on:

- #318
- #1124
- #2361
- #2564
- PR #2644 
- PR #2804 
- PR #2895 
- PR #2906


holta:
Tested on Debian VM's especially.

holta:
While there is no rush (OpenVPN is not intended to be promoted) eventually these Admin Console lines will also need to change, as the file `/etc/iiab/openvpn_handle` will no longer be supported going forward:

- https://github.com/iiab/iiab-admin-console/blob/fdb904e244950a84c24904705c9d86d635a8f424/roles/cmdsrv/files/iiab-cmdsrv3.py#L1209
- https://github.com/iiab/iiab-admin-console/blob/fdb904e244950a84c24904705c9d86d635a8f424/roles/cmdsrv/files/iiab-cmdsrv3.py#L1400-L1407

Some adjustments to https://github.com/iiab/iiab-admin-console/blob/master/roles/console/files/js/admin_console.js might also be appropriate in due course.

Whie personally I'd rather not have OpenVPN promoted within Admin Console, that is just my opinion.

So if this functionality is truly important to others, the `openvpn_handle` variable value could be extracted from /etc/iiab/local_var.yml, similar to the way [/usr/bin/iiab-support](https://github.com/holta/iiab/blob/openvpn-cleaner/roles/openvpn/templates/iiab-support) will now do that going forward.

holta:
This PR was also tested with a fresh install onto Linux Mint 20.2

Issue #2949, opened 2021-08-14
Upgrade IIAB's password-tester to support 'yescrypt' not just SHA-512


holta:
As a result of Debian 11 Bullseye[1] released an hour ago, yescrypt[2][3] is now the default:

1. https://www.debian.org/News/2021/20210814
2. https://www.debian.org/releases/bullseye/amd64/release-notes/ch-information.en.html#pam-default-password
3. https://manpages.debian.org/testing/libcrypt-dev/crypt.5.en.html#yescrypt

However invoking `yescrypt` from Python is not so easy.  Converting [our code](https://github.com/iiab/iiab-factory/blob/master/iiab#L76-L86) may take some time:

```bash
    [ $(python3 -c "import crypt; print(crypt.crypt('$2', '\$$meth\$$salt'))") == "\$$meth\$$salt\$$hash" ]
```

Why?  Maintainers of Python's crypt.crypt do not seem eager to add additional methods as of 2021-06-05: https://bugs.python.org/issue44309

The claim is that `libcrypt` and [libxcrypt](https://github.com/besser82/libxcrypt) are unreliable &mdash; _and not only that..._

> "We still plan to deprecate and remove the crypt module because it's not reliable"

Related:

- iiab/iiab-factory#182

jvonau:
You could drop pre-creation of the iiab-admin user in the 'iiab' script and just let the ansible [role](https://github.com/iiab/iiab/tree/master/roles/iiab-admin) handle the situation for you.

holta:
> You could drop pre-creation of the iiab-admin user in the 'iiab' script and just let the ansible [role](https://github.com/iiab/iiab/tree/master/roles/iiab-admin) handle the situation for you.

That approach is non-optimal &mdash; as the implementer's attention is very short-lived &mdash; before they walk away from the keyboard (for dinner, etc, as IIAB installs).

So best to get them to deal with published passwords _immediately_ during those very first few seconds.

jvonau:
Or not install iiab-admin by default is another option but that just moves the issue to within iiab-support to deal with but that is a better option to install when asked rather than the false sense that the user is forced to change the password but can be bypassed by hitting enter twice. Move that action to where we have the user's complete attention, they want help and are now asking for that help making the out of the box install more secure with one less account on the machine.

jvonau:
[...]

PR #2970, opened 2021-08-29
Check passwds hashed w/ 'yescrypt' etc too, in /etc/shadow -- /etc/profile.d/sshpwd-profile-iiab.sh and /etc/xdg/lxsession/LXDE-pi/sshpwd-lxde-iiab.sh

holta:
In support of:

- #2949

Building on:

- PR iiab/iiab-factory#183
- PR iiab/iiab-factory#184


holta:
Tested on Debian 11 Bullseye.

PR #2971, opened 2021-08-29
Fix internet_available

jvonau:
### Fixes bug:
https://github.com/iiab/iiab/pull/2959#issue-716065778
### Description of changes proposed in this pull request:
Finishes the iiab-state/installed logic where internet access is assumed to be present for the initial install.

PR #2959, opened 2021-08-19
No net exit

jvonau:
### Fixes bug:
No bug just better operation when internet is not present, prevents role operation rather than having the play fail at some later point requiring internet. Well this might prevent other bugs from showing up as some role's install.yml are peppered with "internet_available" which should be assumed to be available during the install and would be skipped. That style is a holdover from the pre iiab_state.yml days and should most likely be removed.

>grep internet_available roles/*/tasks/install.yml
roles/azuracast/tasks/install.yml:  when: internet_available
roles/azuracast/tasks/install.yml:  when: internet_available
roles/calibre-web/tasks/install.yml:  when: internet_available
roles/calibre-web/tasks/install.yml:  when: internet_available
roles/gitea/tasks/install.yml:  when: internet_available
roles/gitea/tasks/install.yml:  when: internet_available
roles/internetarchive/tasks/install.yml:  when: internet_available
roles/jupyterhub/tasks/install.yml:  when: internet_available
roles/kalite/tasks/install.yml:  when: internet_available
roles/kalite/tasks/install.yml:  when: internet_available
roles/kiwix/tasks/install.yml:  when: internet_available
roles/kolibri/tasks/install.yml:  when: internet_available
roles/lokole/tasks/install.yml:    - internet_available
roles/lokole/tasks/install.yml:    - internet_available
roles/lokole/tasks/install.yml:    - internet_available
roles/mediawiki/tasks/install.yml:  when: internet_available
roles/nextcloud/tasks/install.yml:  when: internet_available
roles/nodejs/tasks/install.yml:#  when: internet_available and is_debuntu
roles/nodejs/tasks/install.yml:  #when: internet_available    # 2021-08-04: Better to fail & notify implementer!
roles/nodejs/tasks/install.yml:  #when: internet_available and (is_debian_8 or is_debian_9 or is_ubuntu_16 or is_ubuntu_17)
roles/nodejs/tasks/install.yml:  #when: internet_available    # 2021-08-04: Better to fail & notify implementer!
roles/nodejs/tasks/install.yml:  #when: internet_available and (is_debian_8 or is_debian_9 or is_ubuntu_16 or is_ubuntu_17)
roles/nodejs/tasks/install.yml:#   when: internet_available and not (is_debian_8 or is_debian_9 or is_ubuntu_16 or is_ubuntu_17)
roles/nodered/tasks/install.yml:  when: nodered_install and internet_available
roles/nodered/tasks/install.yml:  when: nodered_install and internet_available and is_raspbian
[...]

PR #2972, opened 2021-08-29
Remove 'when: internet_available' from 25 files [preserving it in network/tasks/netplan.yml + www_options/tasks/main.yml for now]

holta:
Building on @jvonau's:

- PR #2971

This is years overdue, given that `when: internet_available` was never used as intended, as explained at #2959.

Testing is ongoing to verify.

### DIFF
--- a/roles/1-prep/tasks/hardware.yml
+++ b/roles/1-prep/tasks/hardware.yml
@@ -14,4 +14,4 @@
     url: "{{ iiab_download_url }}/iwlwifi-8000C-13.ucode"
     dest: /lib/firmware
     timeout: "{{ download_timeout }}"
-  when: internet_available and usb_NUC6.stdout|int > 0
+  when: usb_NUC6.stdout|int > 0
--- a/roles/azuracast/tasks/install.yml
+++ b/roles/azuracast/tasks/install.yml
@@ -25,7 +25,6 @@
     dest: "{{ azuracast_host_dir }}/"
     timeout: "{{ download_timeout }}"
     mode: 0755
-  when: internet_available
 
 - name: AzuraCast - Download AzuraCast's docker-compose.yml sample from GitHub to {{ azuracast_host_dir }}
   get_url: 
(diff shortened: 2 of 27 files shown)


jvonau:
Too far, network and www_options lack install.yml those roles still need internet_available to be present..

jvonau:
whatever your ball now you run with with it..

holta:
> Too far, network and www_options lack install.yml those roles still need internet_available to be present..

That makes sense for [www_options](https://github.com/iiab/iiab/tree/master/roles/www_options), Ok:

https://github.com/iiab/iiab/blob/1f94166a232e3d6bc0b496b010dbe27e28bb77bd/roles/www_options/tasks/main.yml#L159-L161

Note that [Groovy](https://releases.ubuntu.com/20.10/) was EOL'd on [2021-07-22](https://wiki.ubuntu.com/Releases#End_of_Life), so this code needs to be updated regardless:

https://github.com/iiab/iiab/blob/a581d6ebf4dec6b41a3b61fa578507c9ff33db8b/roles/network/tasks/netplan.yml#L62-L66

Should it be changed to something like `is_ubuntu` ?  Or are post-Groovy particulars like `hirsute` and `impish` still important?

Ref:

- #2585

jvonau:
> Note that Groovy was EOL'd on 2021-07-22, so this code needs to be updated regardless:
> 
> https://github.com/iiab/iiab/blob/a581d6ebf4dec6b41a3b61fa578507c9ff33db8b/roles/network/tasks/netplan.yml#L62-L66
> 
> Should it be changed to something like `is_ubuntu` ? Or are post-Groovy particulars like `hirsute` and `impish` still important?

The fix is 'Groovy' specific, 21.04 and later should have the fix baked into a newer apt package installed by default.

jvonau:
back out the two mentioned roles, (re network - the match covers 20.10 as nothing stops the user from starting at that point..) ok  done I see, thanks

holta:
Testing is ongoing with a BIG-sized IIAB install onto a Debian 11 (server) VM.

Issue #2919, opened 2021-08-07
JupyterHub "administrator's page" not accessible


holta:
Hopefully this is easy to fix, and @georgejhunt can at least lead us in the right direction?

Quick Summary: Teachers (i.e. Admin users) cannot currently access the very helpful "administrator's page" discussed at [JupyterHub FAQ >> "How do I manage users?"](https://jupyterhub.readthedocs.io/en/stable/getting-started/institutional-faq.html#how-do-i-manage-users) and [roles/jupyterhub/templates/jupyterhub_config.py#L1049-L1054 >> "Admin users have extra privileges"](https://github.com/iiab/iiab/blob/d0e8e048347bf46c02a2cdb0da9c5cd0c489fe40/roles/jupyterhub/templates/jupyterhub_config.py#L1049-L1054).

Refs:

- PR #2892 
- #2918

holta:
[roles/jupyterhub/README.md](https://github.com/iiab/iiab/blob/master/roles/jupyterhub/README.md) typo corrected: https://github.com/iiab/iiab/commit/b036c1c7b6b5632ce1acbaad6f40bae766a70d6a

Issue #2918, opened 2021-08-07
JupyterHub password-changing page shows "500 : Internal Server Error"


holta:
FYI this is true whether or not the user is a JupyterHub administrator[*].

Example &mdash; even after logging in, http://box/jupyterhub/hub/auth/change-password shows this in browser:

```
500 : Internal Server Error
```

As I've briefly mentioned here: https://github.com/iiab/iiab/tree/master/roles/jupyterhub#known-issues
I don't know the best way forward, but hopefully @georgejhunt has a moment to look into this over the coming week?

[*] As specified within `/opt/iiab/jupyterhub/etc/jupyterhub/jupyterhub_config.py` ([Line 67](https://github.com/iiab/iiab/blob/master/roles/jupyterhub/templates/jupyterhub_config.py#L67) and [Line 1060](https://github.com/iiab/iiab/blob/master/roles/jupyterhub/templates/jupyterhub_config.py#L1060)).

RELATED: Password [dbm file](https://github.com/jupyterhub/firstuseauthenticator#firstuseauthenticatordbm_path) `/passwords.dbm.db` should be moved from the top of the filesystem to a better place — e.g. to `/opt/iiab/jupyterhub/etc/` ?  PR https://github.com/iiab/iiab/pull/2892#issuecomment-890579789

- PR #2892
- #2919
- https://github.com/jupyterhub/firstuseauthenticator/issues/36

holta:
[roles/jupyterhub/README.md](https://github.com/iiab/iiab/blob/master/roles/jupyterhub/README.md) typo corrected: https://github.com/iiab/iiab/commit/b036c1c7b6b5632ce1acbaad6f40bae766a70d6a

PR #2974, opened 2021-08-31
Admin panel and change password [JupyterHub]

georgejhunt:
Fixes bug: #2919 and #2918  

Tested on clean lite RaspiOS 5/7/21 with min IIAB on Rpi4 hardware


holta:
Wow.  This is tremendous news.

See also @georgejhunt's upstream PR's:

- https://github.com/jupyterhub/firstuseauthenticator/pull/37
- https://github.com/jupyterhub/firstuseauthenticator/pull/38

roles/jupyterhub/templates/patch_await.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash -x
+# add await to asyncio change password function
+
+cat /opt/iiab/jupyterhub/lib/python3.7/site-packages/firstuseauthenticator/firstuseauthenticator.py |grep 'await self.render'
```
holta: @georgejhunt doesn't `python3.7` need to be softcoded using {{ python_ver }} so this works on different versions of different OS's?

Ideally with `patch_await.sh.j2` if so?

holta: If this is true (I don't know, I'm asking?) then Line 7 would need it too of course.

georgejhunt:
This has been tested.  Please explain exactly what you think should be
changed.
The "cat . . ." line you quoted is only to test whether the patch has been
applied, in order to only apply it once

holta:
@georgejhunt I'm asking if this path is correct, across all the different OS's that IIAB supports:

```
/opt/iiab/jupyterhub/lib/python3.7/site-packages/firstuseauthenticator/firstuseauthenticator.py
```

It doesn't look correct.  But I'm just trying to understand if `python3.7` needs to be softcoded.  So that this works on different OS's, each of which brings a different version of Python.

PR #2898, opened 2021-08-02
Update roles/pbx/README.rst, announcing Asterisk 16.x -> 18.x but also warning about PHP 7.4 etc! [Announce transition to FreePBX 16 Beta is now underway, moving away from PHP 7.3]

holta:
@lemueldsouza can you help refine this 1st pass?

(What else should be added/removed/clarified !)

Refs:

- #1467
  - #1493
- PR #1761
  - https://github.com/FreePBX/framework/pull/71#issuecomment-494390926
- PR #2896 
- #2897
  - PR #2899

Issue #2864, opened 2021-07-15
php-stem for PHP 7.4 (e.g. Raspberry Pi OS 11, Debian 11, Ubuntu 21.04) and PHP 8.0 (e.g. Ubuntu 21.10) -> PHP 8.1 (e.g. Ubuntu 22.04) -> PHP 8.2 Dec 2022 [for wikiHow in Spanish]


holta:
As discussed on this week's http://minutes.iiab.io call-

We'll work on adding these soon (e.g. when Raspberry Pi OS 11 is released in not so many days) which will need this or similar:

- http://d.iiab.io/packages/php7.4-stem.rpi.tar

We should probably also compile php-stem files for PHP 8.0 (released November 2020) as IIAB has been well-tested on Ubuntu 21.10 daily builds over the past month:

- #2814
- #2818

(Creating these should be easier than in the past thanks to https://github.com/iiab/php-stem that @tim-moody set up recently!)

Recap: http://oer2go.org/viewmod/es-wikihow needs this for full-text searching.

Refs:

- #2797
- PR #2799
- #2835
- PR #2837
- https://github.com/iiab/php-stem/issues/2

jvonau:
How about using the pinned issue for FAQ type subjects in place of frivolous issues that are targeted at a single fix of a 5 year old php website scraping. Use this real estate wisely.

holta:
@deldesir we should help @tim-moody who is interested in solving this more methodically in coming days, for PHP 7.4 and 8.0 initially, and then for PHP 8.1 sometime later, after that is released [2021-11-25](https://stitcher.io/blog/new-in-php-81).

For PHP 8.0, Ubuntu 21.10 (released 5 days ago) should be sufficient.

For PHP 7.4, we can use Debian 11 "Bullseye" on a VM.  But let's also upgrade Raspberry Pi OS from PHP 7.3 to 7.4 &mdash; by manually changing *both* these Raspberry Pi OS files from `buster` to `bullseye`:

- `/etc/apt/sources.list`
- `/etc/apt/sources.list.d/raspi.list`

And then upgrading its ~500 OS packages by running something like:

```
apt update
apt -y dist-upgrade
reboot
apt -y autoremove
```

This is to simulate and test the ~6 new Raspberry Pi OS 11 "Bullseye" installers arriving very shortly in coming weeks &mdash; to replace the 6 older Raspberry Pi OS 10 "Buster" installers below: 

[...]

PR #2984, opened 2021-09-13
Support PHP 8.0 for Ubuntu 21.10 ETA 2021-10-14, in preparation for "PHP 8.1" on Ubuntu 22.04 ETA 2022-04-21

holta:
With apologies to those having residual affections for Ubuntu 21.04, which will no longer be supported.

The old cliche remains as true as ever: "If we try to please everyone all the time, we will in the end accomplish nothing."

Refs:

- #2814
- #2818 
- #2864
- PR #2898 
- PR #2899
- #2914

### DIFF
--- a/vars/ubuntu-21.yml
+++ b/vars/ubuntu-21.yml
@@ -23,7 +23,7 @@ mysql_service: mariadb
 apache_log: /var/log/apache2/access.log
 sshd_package: openssh-server
 sshd_service: ssh
-php_version: 7.4    # 2021-04-22: Will Ubuntu 21.10 require 8.0?
-postgresql_version: 13    # 2021-04-22: Will Ubuntu 21.10 require 14?
+php_version: 8.0    # IIAB will no longer support non-LTS Ubuntu 21.04, as Ubuntu 21.10 (ETA 2021-10-14) is more important, in preparation for Ubuntu 22.04 (ETA 2022-04-21).
+postgresql_version: 13
 systemd_location: /lib/systemd/system
 python_ver: 3.9


holta:
FYI this PR is expected to be merged in coming weeks, likely _well before_ Ubuntu 21.10's 2021-10-14 release day.

Again apologies to those having undying affections for Ubuntu 21.04 &mdash; who will need to modify [Line 26 of their /opt/iiab/iiab/vars/ubuntu-21.yml](https://github.com/holta/iiab/blob/hello-21.10-goodbye-21.04/vars/ubuntu-21.yml#L26) (from 8.0 to 7.4) as will be documented within: https://github.com/iiab/iiab/wiki/IIAB-Platforms

holta:
ADVANCE WARNING that IIAB will be desupporting Ubuntu 21.04 very shortly, in coming days, when this PR is merged.

Thanks everyone for their understanding, as IIAB has a very limited number of developers, and [Ubuntu 21.10 Beta](https://releases.ubuntu.com/21.10/) is fundamentally now more important to our community.

jvonau:
Why drop a currently supported OS for the next shiny thing to come along that has the great unknown attached to it? Seems shortsighted to me. Dropping 21.04 is a mistake, not to mention in the current scheme 20.10 would be allowed but now is an unmaintained OS. All of the recent validation of applications has been against 21.04, with 20.04 being not considered current enough for the RPi hardware platform. A simple conditional can detect then assign the correct value needed there is no need to drop support or hand edit.
```
{% if ansible_distribution_release == "hirsute" %}
php_version: 7.4
{% else %}
php_version: 8.0
{% endif %} 
```

holta:
Anybody raising a storm about Ubuntu 21.04 is creating a tempest in a teapot.

As is extensively explained at https://github.com/iiab/iiab/wiki/IIAB-Platforms, it's a walk in the park to install IIAB on 21.04 for those obsessed with holding onto old times, rather than improving the quality of IIAB for 2022.

Let us all save our alligator tears for something useful.

jvonau:
> Let us all save our alligator tears for something useful.

I will not be acting upon any issues or 'calls for testing' going forward, now the cause is down another developer due to the continuing abuse of people's opinions.

holta:
The following is a better approach, to reduce the "twice-per-year" confusion and bikeshedding &mdash; that are in fact both inherent to Ubuntu's release cycle:

- PR #2988

Issue #2993, opened 2021-10-01
Language Code missing in lang_codes.json can cause Admin Console init to fail


holta:
How can we help @rickgurmet work through this failure to access Admin Console?

Output of `sudo iiab-diagnostics` is here: http://sprunge.us/rI2zR9?en

![image](https://user-images.githubusercontent.com/2458907/135568052-449e5c05-0843-4c33-8ecc-b8eb4d84bda1.png)

tim-moody:
reboot and try again. if still fails to initialize:

1. ssh into server as root
2. systemctl stop iiab-cmdsrv
3. cd /opt/admin/cmdsrv
4. ./iiab-cmdsrv3.py
5. try again and note any errors

rickgurmet:
I did all the steps you mention but after step 4, there is a line that says "in job_minder_thread" and nothing happens after that, just stays there. 
Here is the sprunge.us URL, http://sprunge.us/rgkho1

tim-moody:
and you reran admin console?

tim-moody:
"in job_minder_thread" is waiting for a command

tim-moody:
This was in js console

jquery.min.js:2 jQuery.Deferred exception: Cannot read properties of undefined (reading 'locname') TypeError: Cannot read properties of undefined (reading 'locname')
    at String.<anonymous> (http://10.8.0.38/admin/js/zim_functions.js:139:39)
    at Function.each (http://10.8.0.38/common/js/jquery.min.js:2:3003)
    at procZimGroups (http://10.8.0.38/admin/js/zim_functions.js:135:5)
    at procZimCatalog (http://10.8.0.38/admin/js/zim_functions.js:409:3)
    at Array.initDone (http://10.8.0.38/admin/js/admin_console.js:2307:5)
    at e (http://10.8.0.38/common/js/jquery.min.js:2:30038)
    at t (http://10.8.0.38/common/js/jquery.min.js:2:30340) undefined

lbj is missing from lang_codes.json

holta:
Should we close this now, or keep it as a placeholder for other languages that also need fixing?

holta:
Here are 36 other missing languages that @tim-moody mentioned:

(that might also need to be addressed in future)

```
Language code lbj not in langCodes.
admin_console.js:2156 Language code ban not in langCodes.
admin_console.js:2156 Language code lfn not in langCodes.
admin_console.js:2156 Language code din not in langCodes.
[...]

PR #2994, opened 2021-10-01
add lbj lang code [for Ladakhi language, in India]

tim-moody:


holta:
https://en.wikipedia.org/wiki/Ladakhi_language mentions both...

> Language codes
> ISO 639-3
> 
> Either:
> lbj – Ladakhi
> zau – Zangskari

...that roles/www_base populates into:

```
/library/www/html/common/assets/lang_codes.json
```

Ref:

- #2993

PR #2998, opened 2021-10-18
Update provision.yml [minetest/tasks/provision.yml: change include: to include_tasks: for ansible-core 2.16]

deldesir:


holta:
@deldesir this looks great.

Is it tested?  On which OS/HW if so?

holta:
> Smoke-tested on: Ubuntu 21.10

Got it!

PR #3010, opened 2021-10-29
Refine & Clarify remote.it (remote support) building on PR #3009

holta:
This is mostly documentation + clarifications + small fixes after testing @jvonau's PR #3009 on Ubuntu Server 20.04 LTS VM.

Refs refs:

- PR #3003
- PR #3005
- #3006
- PR #3007
- PR #3008
- PR #3009
- PR #3026

PR #3009, opened 2021-10-28
Remoteit2

jvonau:
### Fixes bug:
#3008

PR #3007, opened 2021-10-28
use gitea as a template for remoteit

jvonau:
### Fixes bug:
#3006
### Description of changes proposed in this pull request:
map archs to file names in defaults/main.yml like gitea does
### Smoke-tested on which OS or OS's:
not yet.

Issue #3006, opened 2021-10-28
Enhance [and test!] roles/remoteit so it works on RaspiOS Bullseye, Ubuntu, Mint & Debian too [user testing & documentation for remote.it remote support]


holta:
If you can help, please see:

- https://github.com/iiab/iiab/pull/3005#issuecomment-953502502

holta:
@jvonau has thankfully pointed out that this new [roles/remoteit](https://github.com/iiab/iiab/tree/master/roles/remoteit) also fails on the very-soon-to-be-released Raspberry Pi OS 11 Bullseye:

- https://github.com/iiab/iiab/pull/3005#issuecomment-953505445

In Short: the situation really needs to be cleaned up and verified across [all 4 mainline OS's](https://github.com/iiab/iiab/wiki/IIAB-Platforms) and/or the most common CPU architectures. And hopefully very soon in coming days.

holta:
This #3006 might already be solved a few hours later:

Please test on different OS's & architectures, those who can!

Huge thanks to @jvonau for this PR which is now merged:

- PR #3007

holta:
1) I've changed the title (subject line) of this ticket to make sure our focus going forward is practical and real-world usability testing — across different OS's — thanks to these 3 remote.it PR's now merged:

   - #3003
   - #3005
   - #3007

2) Thanks in advance to everyone who can quickly review our brand new README.md here!

   - https://github.com/iiab/iiab/tree/master/roles/remoteit

cc: @deldesir

jvonau:
Not sure stage1 is the right location for this new role as unlike openvpn there is user interaction required to allow the service to function.

holta:
Absolutely it's a complicated question for sure.

Both OpenVPN and remote.it (each installed but disabled by default) can be real lifesavers when an install goes wrong, permitting efficient remote coaching / intervention as needed.

For these kinds of reasons, my preference at the moment is to try to help people to "call 911" from the get-go — so they can quickly enable either remote support option as necessary — e.g. with a phone call to a Linux expert who quickly guides them through enabling/activating either one.

PR #3011, opened 2021-10-29
roles/remoteit/tasks/install.yml: Set 'force: yes' for both downloads

holta:
So that the following might in future help with very basic upgrading of remote.it to a new/higher version number:

```
cd /opt/iiab/iiab
sudo ./runrole --reinstall remoteit
```

Building on:

- PR #3003
- PR #3005 
- #3006
- PR #3007 
- PR #3009
- PR #3010

### DIFF
--- a/roles/remoteit/tasks/install.yml
+++ b/roles/remoteit/tasks/install.yml
@@ -12,6 +12,7 @@
   get_url:
     url: "{{ remoteit_device_url }}"
     dest: "{{ downloads_dir }}/"
+    force: yes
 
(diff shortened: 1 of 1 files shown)


jvonau:
um.. for the binary using 'latest' ok, but the apt package is versioned that will need manual upkeep anyway so there is no naming overlap when the version is bumped. Wonder if there is a 'latest' path to the newest apt package?

holta:
> Wonder if there is a 'latest' path to the newest apt package?

Excellent Question...

PR #3012, opened 2021-10-29
ansible_is_chroot

jvonau:
### Fixes bug:
Long standing issue building images within a chroot environment. 
```
ansible -m setup localhost --connection=localhost | grep chroot
[WARNING]: No inventory was parsed, only implicit localhost is available
        "ansible_is_chroot": false,
```
ansible now provides a builtin variable for such a situation, use it in place of having to maintain custom branches for a workaround.
within a chroot and needs normal install testing but will be fine.


holta:
> ### Smoke-tested on which OS or OS's:
> within a chroot and needs normal install testing but will be fine.

Ok.  Let's keep an eye on it to be sure.

PR #3004, opened 2021-10-27
Wsl [Windows Subsystem for Linux]

georgejhunt:
Smoke tested on a windows 10 insider version that was auto updated to windows 11.
Requires manual installation of Ubuntu from Microsoft cloud.
Requires manual cloning of IIAB, and command line execution of script/wsl2 (See script/wsl2/README.md).

PR #3014, opened 2021-10-30
Is guest

jvonau:
Replacement for part of #2834, looks like #3004 wants the same functionality, pick a name for the variable and go with that. Ether way I think hostname.yml should be suppressed.


jvonau:
place keeper, revisit once #3173 and #3015 land.

jvonau:
#3298

jvonau:
network_install landed

PR #2990, opened 2021-09-30
Pmychanges [phpMyAdmin]

georgejhunt:
Still need to use a socket to connect to MySQL And create a low capability Admin/changeme user. But there's a timing issue in using iiab_feedback table.  So manbe I need to create a useless table which would let phpmyadm  log on out  of the box.

PR #2712, opened 2021-03-17
phpmyadmin was never migrated to nginx -- missing shim config file

georgejhunt:
The current master has access to mysql "root" user (with all privileges) wide open. But phpmyadmin fails to open a mysql server with a blank password. (The mysql client does provide root privilege without password) -- so it is possible for our end users to remedy the situation.

I do question whether we want to leave mysql root user wide open in the long run since mysql is now a standard service. 

I searched our docs, FAQ, default_vars.yml, and mysql role, without finding any reference to mysql security.

I was able to log into phymyadmin by executing the following string in mysql client:
```
 grant all on * to ghunt identified by "<password>"
```
and then I could start phpmyadin on a clean install of current raspi-os lite.

PR #3016, opened 2021-10-30
 found in #3004? [roles/phpmyadmin/templates/config.inc.php]

jvonau:
@georgejhunt @holta this code fragment found in #3004 should of been part of #2990?


holta:
Quickly mentioned during this morning's IIAB call (http://minutes.iiab.org).  @georgejhunt recommends this PR be tested across all ~4 OS's before moving forward.

He also mentions that this last line doesn't seem to have any effect either way:

```
$cfg['Servers'][$i]['AllowNoPassword'] = '';
```

Ref:

- #2712

PR #3013, opened 2021-10-30
Disable the use of pip cache system wide

jvonau:
Disable the use of pip cache system wide, leaves no files hanging around in /root/.pip when disabled.

PR #3021, opened 2021-11-06
scripts/ansible: pip -> 'pip3 config set global.cache-dir false' (PR #3013)

holta:
Fixes this small bug (is this fix [good enough](https://github.com/iiab/iiab/pull/3013#issuecomment-962509987) ?) that crept in a week ago:

- https://github.com/iiab/iiab/pull/3013#issuecomment-962508868

Which didn't affect all OS's, but did however affect the dominant one (Raspberry Pi OS).


jvonau:
Right RaspiOS still ships python2 out of the box while everybody else just has python3 installed.

Issue #3027, opened 2021-11-11
#3023 breaks existing installs


tim-moody:
Anyone with an install on raspian 10 who does a pull will now get ERROR! vars file vars/{{ ansible_local.local_facts.os_ver }}.yml was not found when doing anything that involves ansible install.

holta:
Debian Buster was desupported 4 months ago.

Very Intentionally So: after Debian Bullseye was very carefully tested, and to keep our community from balkanizing.

Raspberry Pi OS Buster was desupported 3 days ago.

Very Intentionally So: after Raspberry Pi OS Bullseye was very carefully tested, and to keep our community from balkanizing.

IIAB supports 9 OS versions within https://github.com/iiab/iiab/blob/master/scripts/local_facts.fact#L72-L85 which is already too many.

Please focus on the positive **(people who've spent hundreds of hours in recent months making sure IIAB runs well on Raspberry Pi OS Bullseye).**

Finally any individual that absolutely must attempt an older, legacy OS is of course free to do so with a 1-line change to the file mentioned above, as is mentioned repeatedly in: https://github.com/iiab/iiab/wiki/IIAB-Platforms

holta:
If the "OS Not Supported" message from Line 83 of https://github.com/iiab/iiab/blob/master/scripts/local_facts.fact#L83 is not clear, I will work on making that more prominent, more visible and more explanatory.

tim-moody:
>Debian Buster was desupported 4 months ago.
Very Intentionally So: after Debian Bullseye was very carefully tested, and to keep our community from balkanizing.
Raspberry Pi OS Buster was desupported 3 days ago.

>IIAB supports 9 OS versions

so what is the recommendation for an existing install, throw it away and start over?

at the very least, be careful never to do a pull

You're only looking at this from the standpoint of doing an install, not an existing server.

>IIAB supports 9 OS versions

Ubuntu from 2020, but not raspios that was only superseded 3 days ago. Even ubuntu-2104 is older than May raspios. And linux mint from 2020.

holta:
http://FAQ.IIAB.IO has for many years made clear that IIAB software upgrades are entirely "At Your Own Risk" (e.g. `git pull` is definitely not supported as a mainline practice or ongoing recommendation).

[...]

PR #3029, opened 2021-11-11
IIAB's limited community/testing resources need to be focused on the present not the past

holta:
Refs:

- PR #3023
- #3027
- PR #3028

### DIFF
--- a/scripts/local_facts.fact
+++ b/scripts/local_facts.fact
@@ -78,7 +78,6 @@ case $OS_VER in
     "ubuntu-2204"  | \
     "linuxmint-20" | \
     "linuxmint-21" | \
-    "raspbian-10"  | \
     "raspbian-11")
         ;;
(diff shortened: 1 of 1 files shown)


tim-moody (commented):
why no concern about balkanizing with ubuntu 2004, 2104, linux mint 20?

holta:
> why no concern about balkanizing with ubuntu 2004, 2104, linux mint 20?

It is a very real concern.

No question.

Especially among people who give hundreds of hours of their lives, contributing intensive testing (quality engineering) for others:

- Ubuntu 20.04 LTS we're forced to support until 22.04 LTS is released around 2022-04-21.
- Ubuntu 21.04 should be desupported soon (Ubuntu 21.10's PHP 8.0 has risks, but is proving to work well enough).
- Linux Mint 20 is almost identical to 20.04 LTS, so the testing coverage workload here has so far thankfully been minimal — in service to a growing number of schools and NGO's who insist they need this distro above all others.

PR #3033, opened 2021-11-12
scripts/ansible: pip3 config --global set global.no-cache-dir false

holta:
@jvonau can you review/confirm?

(I assume this bug fix is sufficient, as outlined on...)

- https://github.com/iiab/iiab/pull/3013#issuecomment-967162092


holta:
This will be merged as it was (quickly) tested on:

- Ubuntu Server 20.04 LTS (VM)
- Raspberry Pi OS Lite "Bullseye" (RPi 4)
- Raspberry Pi OS with desktop "Bullseye" (RPi 4)

Thanks @jvonau if you can confirm this is working as it should on RPi 400 and/or similar.

PR #3031, opened 2021-11-12
Update iiab-install & scripts/local_facts.fact to explain more clearly when OS / kernel / Ansible are unsupported

holta:
Refs:

- PR #3023 
- https://github.com/iiab/iiab/issues/3027#issuecomment-966601309
- PR #3028
- PR #3029
- PR #3030

PR #3034, opened 2021-11-12
scripts/local_facts.fact: 15 days transitional support for LTS OS's (e.g. both raspbian-10 & raspbian-11 in this case)

holta:
Raspberry Pi OS 10 Buster is not really [LTS](https://en.wikipedia.org/wiki/Long-term_support) (but that's ok, close enough, the Raspberry Pi Foundation is looking into that possibility especially for industrial/IoT sales in the future) will be de-supported in `/opt/iiab/iiab/scripts/local_facts.fact` on 2021-11-23.

After this 15-day transition to [Raspberry Pi OS 11 Bullseye](https://www.raspberrypi.com/news/raspberry-pi-os-debian-bullseye/) (released on 2021-11-08) is complete.

Thank you to @icebox827 for his spirit of compromise.  _Trying to forge meaningful balance here &mdash; out of very different hard-working and sincere people, each bringing together overlapping but very different experiences and perspectives from each of their lives and viewpoints._

Refs:

- PR #3023 
- #3027 
- PR #3028
- PR #3029 
- PR #3031

### DIFF
--- a/scripts/local_facts.fact
+++ b/scripts/local_facts.fact
@@ -64,7 +64,6 @@ OS_VER="$OS-$VERSION_ID"
     #"centos-7"     | \
     #"raspbian-8"   | \
     #"raspbian-9"   | \
-    #"raspbian-10"  | \
 
 # 2021-09-27: With Debian 12 (Bookworm) pre-releases, please manually add
 # this line to its /etc/os-release before installing IIAB: VERSION_ID="12"
@@ -78,6 +77,7 @@ case $OS_VER in
     "ubuntu-2204"  | \
     "linuxmint-20" | \
     "linuxmint-21" | \
+    "raspbian-10"  | \
     "raspbian-11")
         ;;
     *) echo -e "\n\e[41;1mOS '$OS_VER' IS NOT SUPPORTED. Please read:\e[0m\n\n\e[1mhttps://github.com/iiab/iiab/wiki/IIAB-Platforms\e[0m\n" ; exit 1    # Used by /opt/iiab/iiab/iiab-install


tim-moody:
So the policy is that support for an os is removed (even for existing installs) 15 days after its successor is released?

Then please remove support for mint 20 and ubuntu 2004, 2010, and 2104 as all have had successors for more than 15 days. That will greatly reduce the burden of supporting many os versions.

Also please add a warning that IIAB does not guarantee support for your install for more than 15 days.

holta:
> please remove support for mint 20 and ubuntu 2004, 2010, and 2104 as all have had successors for more than 15 days.

- Linux Mint 20 does not have a successor that is released and widely available (mint-21 if it exists at all, would be in very early testing as of today).  Mint 20 is still in the midst of its currently active LTS-like support cycle as of right now (November 2021).  And thankfully has been (generally) a lesser workload to test than other distros!

- Ubuntu 20.04 LTS does not have an LTS successor.  [As mentioned yesterday](https://github.com/iiab/iiab/pull/3029#issuecomment-966684053) this will be de-supported soon after Ubuntu 22.04 LTS is released ETA 2022-04-21.

- IIAB Support for Ubuntu 20.10 was removed approx 6 months ago, and it was end-of-life'd by Ubuntu on 2021-07-22.

- I would agree that IIAB should not support Ubuntu 21.04 (a non-LTS OS) much longer, as it's now almost 1 month after Ubuntu 21.10 (also a non-LTS OS) was released on 2021-10-14.  However there was concern that PHP 7.4 -> 8.0 was such a radical shift over the past few weeks, that comparison testing of both (might) be useful and/or necessary?  In any case, 21.04 will be end-of-life'd by Ubuntu in ~2 months from now in January 2022.  @jvonau might still be testing IIAB on Ubuntu 21.04 but he can clarify if there is any remaining value there.  I suspect not, but I might be wrong, and he can clarify.

PR #3037, opened 2021-11-20
Rename local_vars_min.yml to local_vars_small.yml, local_vars_big.yml to local_vars_large.yml (ETC)

holta:
_It will take Many Days to fix documentation in countless different places,_ to correspond to the {S, M, L} lingo (just like T-Shirt sizes) that essentially everyone prefers.

But these 2 PR's are the 1st Step.

At Long Last &mdash; see also this Companion PR:

- https://github.com/iiab/iiab-factory/pull/193

### DIFF
--- a/roles/pbx/README.adoc
+++ b/roles/pbx/README.adoc
@@ -303,9 +303,9 @@ Node.js applications like Asterisk/FreePBX, Node-RED and Sugarizer won't work on
 ////
 == Raspberry Pi Known Issues
 
-As of 2019-02-14, "systemctl restart freepbx" failed more than 50% of the time when run on a http://wiki.laptop.org/go/IIAB/FAQ#What_services_.28IIAB_apps.29_are_suggested_during_installation.3F[BIG-sized] install of IIAB 6.7 on RPi 3 or RPi 3 B+.
+As of 2019-02-14, "systemctl restart freepbx" failed more than 50% of the time when run on a http://wiki.laptop.org/go/IIAB/FAQ#What_services_.28IIAB_apps.29_are_suggested_during_installation.3F[LARGE-sized] install of IIAB 6.7 on RPi 3 or RPi 3 B+.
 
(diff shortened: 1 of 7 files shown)


tim-moody:
how about renaming unittest to local_vars_zero. should it have apache_allow_sudo: True? And I don't think vpn should be enabled.

holta:
Good questions.

So a clarification is definitely in order:

local_vars_unittest.yml has been an absolute lifesaver for unit-testing (of individual Ansible roles especially) over the past 2-3 months &mdash; it has allowed for a tremendously accelerated workflow used by many testing folk old and new &mdash; who need VPN access right away so that collaboration is nearly completely hassle-free.  (Allowing IIAB testing feedback to become increasingly far more efficient-and-meaningful than it was the past, a big step forward compared to even 3 months ago.)  This push to make the testing community growth as completely obstacle-free and hassle-free as possible is indeed critical towards allowing IIAB to scale.  In that light, as much as possible anyway, the particular filename should really should reflect what it's being used for most (which is in fact unit testing, in this particular case).  Whereas another stock local_vars.yml file might be advisable, for those who want something more completely bare bones.

holta:
> should it have apache_allow_sudo: True?

I'm assuming this default value remains fine across the board at this time, and is also useful to allow everyone to try & test the http://box/common/services/power_off.php "OFF BUTTON" hidden near the bottom of many/most IIAB main pages (http://box.lan).

But possibly this variable should be renamed to `nginx_allow_sudo` in future?

tim-moody:
>Whereas another stock local_vars.yml file might be advisable, for those who want something more completely bare bones.

I'll add one

jvonau:
How about a heads-up that the primary local_vars.yml files were to be renamed, kind of messes up automated workflows where `cp vars/local_vars_min.yml /etc/iiab/local_vars.yml` was used. Easy to change on my end but a hey this is changing would be nice in the future.

Issue #3039, opened 2021-11-21
kiwix-serve doesn't start on Ubuntu 20.04 ARM64 (UTM [& Multipass & Parallels VMs] on OSX 12.0.1) [on M1 Mac]


avni:
### Expected behavior 
- Have Kiwix-serve start successfully after installation via sudo iiab

### System Information

- **Host OS:** OSX Monterey v12.0.1 M1 processor
- **VM:** UTM VM: https://getutm.app
- **uname -a**
`Linux box 5.4.0-90-generic #101-Ubuntu SMP Fri Oct 15 20:02:26 UTC 2021 aarch64 aarch64 aarch64 GNU/Linux`
- **dpkg --print-architecture** 
`arm64`
- **ansible_machine**
`aarch64`
- **Kiwix version:** Kiwix was downloaded from: http://download.iiab.io/packages/kiwix-tools_linux-armhf-3.1.2-5.tar.gz

### Actual Behavior: When running, sudo iiab, I get the following error related to kiwix

- **iiab-diagnostics output:** http://sprunge.us/GykyNH?en
- **systemctl status kiwix-serve.service** 
```
root@box:/opt/iiab/iiab# systemctl status kiwix-serve.service 
kiwix-serve.service - Provides the Kiwix ZIM File Server
     Loaded: loaded (/etc/systemd/system/kiwix-serve.service; enabled; vendor preset: enabled)
     Active: failed (Result: exit-code) since Sun 2021-11-21 12:57:26 EST; 25min ago
    Process: 49546 ExecStart=/opt/iiab/kiwix/bin/kiwix-serve --daemon --port 3000 --nolibrarybutton --library /library/zims/library.xml --urlRootLocation=/kiwix/ --threads 4 (cod>

Nov 21 12:57:26 box systemd[1]: Starting Provides the Kiwix ZIM File Server...
Nov 21 12:57:26 box systemd[49546]: kiwix-serve.service: Failed to execute command: Exec format error
Nov 21 12:57:26 box systemd[49546]: kiwix-serve.service: Failed at step EXEC spawning /opt/iiab/kiwix/bin/kiwix-serve: Exec format error
Nov 21 12:57:26 box systemd[1]: kiwix-serve.service: Control process exited, code=exited, status=203/EXEC
Nov 21 12:57:26 box systemd[1]: kiwix-serve.service: Failed with result 'exit-code'.
[...]

PR #3050, opened 2021-11-26
Fix Ansible rpi_model != "none" boolean test for Raspberry Pi hardware [other systems like Mac Parallels VMs also use /proc/device-tree/model !]

holta:
Example of the bug in action:

Parallels (VM software on Mac) sets `/proc/device-tree/model` to "Parallels ARM Virtual Machine" (in VM's, e.g. an Ubuntu VM).

IIAB's current logic (prior to this PR) then erroneously interprets that to be Raspberry Pi hardware &mdash; as a result of: https://github.com/iiab/iiab/blob/032436421b9ecc24e701f89e6e41920d2f321f3d/scripts/local_facts.fact#L97-L98

Ref:

- #3039

cc: @avni, @jvonau

### DIFF
--- a/scripts/local_facts.fact
+++ b/scripts/local_facts.fact
@@ -94,7 +94,7 @@ tmp=$(git rev-parse --verify HEAD) &&
 #tmp=$(cat /proc/device-tree/mfg-data/MN) &&
 #    XO_MODEL=$tmp
 
-tmp=$(cat /proc/device-tree/model) &&
+tmp=$(grep -ai raspberry /proc/device-tree/model) &&
     RPI_MODEL=$tmp
 
 tmp=$(ansible --version) &&


holta:
As a bonus, this PR cleans up the CRLF (Windows newline) at the end of every Raspberry Pi's binary file `/proc/device-tree/model` &mdash; i.e. saving the value more intelligently into Ansible var `rpi_model`

avni:
Thank you, this allows the install to move forward after pulling in this change. There seems to, however, be lingering logic which outputs the same variable twice and has 2 different values on this VM:

http://sprunge.us/jdHlFC?en

```
 101 rpi_model = Parallels ARM Virtual Machine
 115 rpi_model = none

```

holta:
> Thank you, this allows the install to move forward after pulling in this change. There seems to, however, be lingering logic which outputs the same variable twice and has 2 different values on this VM:
> 
> http://sprunge.us/jdHlFC?en
> 
> ```
>  101 rpi_model = Parallels ARM Virtual Machine
>  115 rpi_model = none
> ```

FWIW this code below will only run once...the 1st time 'sudo iiab' is run:

https://github.com/iiab/iiab/blob/master/roles/0-init/tasks/create_iiab_ini.yml#L40-L41

So indeed `/etc/iiab/iiab.ini` is ugly (when a variable appears to have 2 different values only 14 lines apart!) but arguably...this is not a serious bug (!)

jvonau:
So what is the bug other than an ill named variable? Using arm_model might be more accurate in the description of the information returned. The routine went to returning useful information and recording in iiab.ini to hiding behind 'none'... Ask yourself what actually acts upon rpi_model being set? https://github.com/iiab/iiab/search?q=rpi_model&type= if you what to base conditionals on OS use is_raspbian in place of rpi_model != "none" otherwise it it's a test of arm vs non-arm when testing against 'none'. Once upon a time XOs were the only arm hardware in use then along comes the Rpi family, just an evolution forward in the adaption of arm hardware.

jvonau:
> Thank you, this allows the install to move forward after pulling in this change. There seems to, however, be lingering logic which outputs the same variable twice and has 2 different values on this VM:
> 
> http://sprunge.us/jdHlFC?en
> 
> ```
>  101 rpi_model = Parallels ARM Virtual Machine
>  115 rpi_model = none
> ```

Different git hashes means different code was used between first run and the last run of ansible.

holta:
> So what is the bug other than an ill named variable? Using arm_model might be more accurate in the description of the information returned.

Correction/Clarification: Parallels is just one of 3 major VM systems on macOS, and the only 1 of the 3 that uses `/proc/device-tree/model` within Ubuntu VMs &mdash; where it stores the string `Parallels ARM Virtual Machine`
Whereas the other 2 VM systems (UTM, Multipass) do not use the file `/proc/device-tree/model` at all, within their Ubuntu VMs.

Further complicating the matter, there's no way of knowing when/where/if another non-ARM system also uses `/proc/device-tree/model` in its own way (to record hardware or similar info) also causing IIAB installs to fail.

Conclusion: IIAB install was failing as a result of erroneous conditional tests acting on `rpi_model != "none"` (as if this was supposed to test YES/NO for RPi hardware, but in fact was generating False Positives).  Example: prior to this PR, IIAB was failing to install on Ubuntu VM's that use the Parallels VM system.  The bug is now fixed, so those kinds of IIAB installs can now proceed.

If `/proc/device-tree/model` should also be recorded in iiab-diagnostics (or anywhere else) that's indeed definitely an interesting idea.  But a separate idea.

Recap: The point of this PR was to fix the boolean test to verify YES/NO whether there's Raspberry Pi hardware.  (This was buggy prior to this PR &mdash; preventing install of IIAB on those non-RPi systems that also happen to use `/proc/device-tree/model`)

holta:
FWIW `/sys/firmware/devicetree/base/model` is already recorded in https://github.com/iiab/iiab/blob/master/scripts/iiab-diagnostics#L122 which in the case of Raspberry Pi hardware, appears to be a file with the identical contents:

```
root@box:/opt/iiab/iiab# cat /proc/device-tree/model
Raspberry Pi 4 Model B Rev 1.4root@box:/opt/iiab/iiab#

root@box:/opt/iiab/iiab# cat /sys/firmware/devicetree/base/model
Raspberry Pi 4 Model B Rev 1.4root@box:/opt/iiab/iiab#

root@box:/opt/iiab/iiab# ls -l /proc/device-tree/model
-r--r--r-- 1 root root 31 Nov 25 09:13 /proc/device-tree/model

root@box:/opt/iiab/iiab# ls -l /sys/firmware/devicetree/base/model
-r--r--r-- 1 root root 31 Nov 25 09:13 /sys/firmware/devicetree/base/model
```

@avni can you check within a Parallels VM, to see if both above files are also identical?

i.e. do both above files contain the string "Parallels ARM Virtual Machine" and nothing else?

jvonau:
> > So what is the bug other than an ill named variable? Using arm_model might be more accurate in the description of the information returned.
> 
> Correction/Clarification: Parallels is just one of 3 major VM systems on macOS, and the only 1 of the 3 that uses `/proc/device-tree/model` within Ubuntu VMs — where it stores the string `Parallels ARM Virtual Machine` Whereas the other 2 VM systems (UTM, Multipass) do not use the file `/proc/device-tree/model` at all, within their Ubuntu VMs.
> 
> Further complicating the matter, there's no way of knowing when/where/if another non-ARM system also uses `/proc/device-tree/model` in its own way (to record hardware of similar info) also causing IIAB installs to fail.
> 

Where is this other failures you speak of? The bluetooth role is before kiwix and didn't stop there from what is reported. Did it stop in 1-prep?
> Conclusion: IIAB install was failing as a result of erroneous conditional tests acting on `rpi_model != "none"` (as if this was supposed to test YES/NO for RPi hardware, but in fact was generating False Positives). Example: prior to this PR, IIAB was failing to install on Ubuntu VM's that use the Parallels VM system. The bug is now fixed, so those kinds of IIAB installs can now proceed.

From what I see 3039 has not improved leaving mongodb and other roles that would not of been reached yet.
> 
> If `/proc/device-tree/model` should also be recorded in iiab-diagnostics (or anywhere else) that's indeed definitely an interesting idea. But a separate idea.
> 

If it is in iiab.ini it was already being recorded by iiab-diagnostics, why suppress that information? I would prefer to record the actual value in iiab.ini as it is useful information to know and have the conditionals move to ansible, testing for 'raspberry' there with something like [when: "variable1.find('v=' ~ value) == -1"](https://stackoverflow.com/questions/36496911/run-an-ansible-task-only-when-the-variable-contains-a-specific-string/36497471.) To clear up any use-case confusion rename the variable to arm_model at least in the user facing iiab.ini's recording.

> Recap: The point of this PR was to fix the boolean test to verify YES/NO whether there's Raspberry Pi hardware. (This was buggy prior to this PR — preventing install of IIAB on those non-RPi systems that also happen to use `/proc/device-tree/model`)

It is not a boolean value, what returned is a string `the != none` acts on the string as a boolean, needs to be more selective now.

avni:
> @avni can you check within a Parallels VM, to see if both above files are also identical?

confirmed they are identical. This is a Parallels VM with Ubuntu 20.04 ARM64

```
avni@box:~$ sudo cat /proc/device-tree/model
Parallels ARM Virtual Machine

avni@box:~$ sudo cat /sys/firmware/devicetree/base/model
Parallels ARM Virtual Machine
```

avni:
@jvonau 
>> The bluetooth role is before kiwix and didn't stop there from what is reported. 

Prior to the patch that @holta  applied with #3050, the install on Parallels did fail at the bluetooth role. I was able to proceed after applying this patch. The install then failed as the installs on UTM and multipass had at the kiwix role. (Note that installs on UTM and multipass did _not_ fail at the bluetooth role interstingly enough!)

I hope this clarifies! Happy to jump on a call if needed. Thanks Jerry!

jvonau:
Using `bluetooth_install: False` would of worked around that issue equally as well in the mean time, perhaps the default True for this for this role should be revisited and not just be papered over with a big hammer?

jvonau:
> I hope this clarifies! Happy to jump on a call if needed. Thanks Jerry!

@avni Not tonight but hit me up tomorrow to pencil in a time.

holta:
> arm_model at least in the user facing iiab.ini's recording.

`/proc/device-tree/model` equivalent (`/sys/firmware/devicetree/base/model`) is already recorded as part of [iiab-diagnostics](https://github.com/iiab/iiab/blob/master/scripts/iiab-diagnostics.README.md).

But it could also be recorded to `/etc/iiab/iiab.ini` if that's desired for additional diagnostic redundancy?

But it should not be called `arm_model` as this file's contents could convey a new Chinese CPU — or any other kind of new VM system unrelated to ARM.

Possibly `cpu_or_vm_model` would be a more accurate variable name?

holta:
Changing the core `rpi_model` variable name would create a large amount of unnecessary complexity.

Fine-tuning diagnostics is a good idea and we will do that, wherever it helps.

But the larger point is that `rpi_model != "none"` was a buggy test for Raspberry Pi OS hardware used across 2 repos (iiab/iiab and iiab/iiab-admin-console).

This bug is now fixed, across both repos.

Issue #3057, opened 2021-12-02
Internal WiFi [IIAB hotspot] not working on a Old PC (Panasonic Toughbook CF-52)


icebox827:
# Internal WiFi Hotspot issue

Unable to use iaab internal WiFi hotspot on a Old PC (Panasonic CF-52) running Ubuntu desktop 21.10

## Diagnostic links

- http://sprunge.us/9zTw1Z?en

holta:
I see on Line 467 of http://sprunge.us/9zTw1Z?en that you also tried `wifi_up_down: false`

holta:
> http://sprunge.us/9zTw1Z?en

@jvonau might possibly know whether this laptop has any hope (or not) to run **AP Mode** ?

Some say Mint has somewhat better driver/firmware support than Ubuntu on older laptops.  But I don't know if that's proven true.

Certainly the Toughbook is unlikely to be able to run [AP+STA](https://github.com/iiab/iiab/issues/2853#issuecomment-934372625) if indeed its internal WiFi cannot be put into AP Mode...?

jvonau:
No duel mode available when the upstream wifi is using 5GHz as shown below, while the hotspot is always 2.4GHz on channels 1-12 only.
```
 395 host_channel = 6
 396 client_wifi_channel = 149
1365 		channel 149 (5745 MHz), width: 40 MHz, center1: 5755 MHz
```

holta:
> No [dual] mode available when the upstream wifi is using 5GHz as shown below, while the hotspot is always 2.4GHz on channels 1-12 only.
> 
> ```
>  395 host_channel = 6
>  396 client_wifi_channel = 149
> 1365 		channel 149 (5745 MHz), width: 40 MHz, center1: 5755 MHz
> ```

Thanks @jvonau for spotting that:

@icebox827 can you clarify &mdash; did you run `sudo ./iiab-network` (in /opt/iiab/iiab) and then reboot &mdash; after you set `wifi_up_down: false` in /etc/iiab/local_vars.yml ?

(I'd assume that would be your best shot at achieving a 2.4GHz internal hotspot?  If there's any remaining chance in the universe...that this Toughbook CF-52 from almost a decade ago...might actually be capable of AP Mode ;-)

jvonau:
> I see on Line 467 of http://sprunge.us/9zTw1Z?en that you also tried `wifi_up_down: false`

[...]

PR #3070, opened 2021-12-17
Detect, record if wifi can be AP and suppress installing supporting files

jvonau:
### Fixes bug:
hostapd restart error noted in https://github.com/iiab/iiab/issues/3057#issue-1069824427

Untested

### DIFF
--- a/roles/network/tasks/detected_network.yml
+++ b/roles/network/tasks/detected_network.yml
@@ -101,6 +101,16 @@
   set_fact:
     num_wifi_interfaces: "{{ count_wifi_interfaces.stdout|int }}"
 
+- name: Check for Access Point capablility with 'iw list'
+  command: iw list | grep AP | wc -l
+  register: look_for_ap
+  when: discovered_wireless_iface != "none"
+
+- name: Setting can_be_ap
+  set_fact:
+    can_be_ap: True
+  when: look_for_ap|int != "0"
+
 - name: Detect wifi gateway active
   shell: ip r | grep default | grep {{ discovered_wireless_iface }} | wc -l
   register: wifi_gateway_found
@@ -260,6 +270,8 @@
       value: "{{ iiab_lan_iface }}"
     - option: iiab_wan_iface
       value: "{{ iiab_wan_iface }}"
+    - option: can_be_ap
+      value: "{{ can_be_ap }}"
 
 # well if there ever was a point to tell the user things are FUBAR this is it.
 # limit 2 network adapters wifi wired
--- a/roles/network/tasks/detected_network.yml
+++ b/roles/network/tasks/detected_network.yml
@@ -102,14 +102,14 @@
     num_wifi_interfaces: "{{ count_wifi_interfaces.stdout|int }}"
 
 - name: Check for Access Point capablility with 'iw list'
-  command: iw list | grep AP | wc -l
+  command: iw list | grep -v AP: | grep AP | wc -l
   register: look_for_ap
   when: discovered_wireless_iface != "none"
 
 - name: Setting can_be_ap
   set_fact:
     can_be_ap: True
-  when: look_for_ap|int != "0"
+  when: look_for_ap.stdout|int != 0
 
 - name: Detect wifi gateway active
   shell: ip r | grep default | grep {{ discovered_wireless_iface }} | wc -l
--- a/roles/network/defaults/main.yml
+++ b/roles/network/defaults/main.yml
@@ -59,6 +59,7 @@ virtual_network_devices: "-e ap0 -e lo -e br0 -e tun -e br- -e docker -e bridge0
 # Set defaults for discovery process as strings
 wifi1: "not found-1"
 wifi2: "not found-2"
+can_be_ap: False
 exclude_devices: none
 device_gw: none
 prior_gw_device: unset
--- a/roles/network/tasks/hostapd.yml
+++ b/roles/network/tasks/hostapd.yml
@@ -4,10 +4,10 @@
     enabled: no
     masked: no
 
-- name: Disable hostapd when not using ap0 and wifi gateway present, or no WiFi hardware present
+- name: Disable hostapd when not using ap0 and wifi gateway present, or no WiFi hardware present or support not detected
   set_fact:
     hostapd_enabled: False
-  when: (not wifi_up_down and discovered_wireless_iface == iiab_wan_iface) or discovered_wireless_iface == "none"
+  when: (not wifi_up_down and discovered_wireless_iface == iiab_wan_iface) or discovered_wireless_iface == "none" or not can_be_ap
 
 - name: Detect current Wifi channel
   shell: iw {{ discovered_wireless_iface }} info | grep channel | cut -d' ' -f2
@@ -29,7 +29,7 @@
   with_items:
     - { src: 'hostapd/hostapd.conf.j2', dest: '/etc/hostapd/hostapd.conf' }
     - { src: 'hostapd/hostapd.conf.j2', dest: '/etc/hostapd/hostapd.conf.iiab' }
-  when: discovered_wireless_iface != "none"
+  when: can_be_ap
 
 - name: Generate new random mac address for ap0
   shell: tr -dc A-F0-9 < /dev/urandom | head -c 10 | sed -r 's/(..)/\1:/g;s/:$//;s/^/02:/'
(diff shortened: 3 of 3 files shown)


holta:
@jvonau are the 1+6 mentions of "AP" in bold below extraneous &mdash; not offering any hope at all to enable AP mode? (as an example, apparently from a Kingdel PC)

Paste below from file `/etc/NetworkManager/system-connections/Wi-Fi connection 1.nmconnection` :

> [wifi]
band=bg
channel=9
mac-address=00:1E:65:41:55:61
mac-address-blacklist=
mode=**ap**
ssid=SmartBox

Output below from command `/usr/sbin/iw list` :

> Supported interface modes:
          * IBSS
          * managed
          * monitor
 Band 1:

...

> Supported commands:
          * new_interface
          * set_interface
          * new_key
          * **start_ap**
          * new_station

...

> Supported TX frame types:
          * IBSS: 0x00 0x10 0x20 0x30 0x40 0x50 0x60 0x70 0x80 0x90 0xa0 0xb0 0xc0 0xd0 0xe0 0xf0
          * managed: 0x00 0x10 0x20 0x30 0x40 0x50 0x60 0x70 0x80 0x90 0xa0 0xb0 0xc0 0xd0 0xe0 0xf0
          * **AP:** 0x00 0x10 0x20 0x30 0x40 0x50 0x60 0x70 0x80 0x90 0xa0 0xb0 0xc0 0xd0 0xe0 0xf0
          * **AP/VLAN:** 0x00 0x10 0x20 0x30 0x40 0x50 0x60 0x70 0x80 0x90 0xa0 0xb0 0xc0 0xd0 0xe0 0xf0

...

> Supported RX frame types:
          * IBSS: 0x40 0xb0 0xc0 0xd0
          * managed: 0x40 0xd0
          * **AP:** 0x00 0x20 0x40 0xa0 0xb0 0xc0 0xd0
          * **AP/VLAN:** 0x00 0x20 0x40 0xa0 0xb0 0xc0 0xd0

...

> Driver supports full state transitions for **AP/GO** clients

(Above excerpts are from iiab-diagnostics http://sprunge.us/NMOYJl?en)

jvonau:
Now what has been presented is a user configured WiFi network, given I shown you how to use NetworkManager's connection sharing on an XO-4 what in 2014-2015, running iiab-network would cause a clash over control of the wifi device causing the crash noted at `Restart hostapd when WiFi is present` in the pastebin. Easiest course of action would be to set "hostapd_enable False" in local_vars.yml as a start and look at #3014 for those experts that know how to configure a network properly and don't need hand holding.

jvonau:
```
1677 	software interface modes (can always be added):
1678 		 * monitor
1679 	interface combinations are not supported
```
https://github.com/iiab/iiab/issues/3057#issuecomment-987081624 won't support ap0, I'd would try with wifi_up_down: False.
`sudo iiab-hotspot-off`
edit local_vars 
`cd /opt/iiab/iiab`
`sudo ./iiab-network`
`sudo iib-hotspot-on`
Or 
Just use the GUI to manage the network but the trouble is admin-console will run the network role anyway, is_guest would need to be added to iiab-from-console.yml also.

```
1098 -IIAB--------------------------------------------------------------------------
1099 -rw------- 1 root root 343 Dec 11 10:39 /etc/NetworkManager/system-connections/Wi-Fi connection 1.nmconnection
1100                         ...ITS LAST 100 LINES FOLLOW...
1101 
1102 [connection]
1103 id=Wi-Fi connection 1
1104 uuid=aee7d61d-bfbf-4e4a-840d-5daab2601e6d
1105 type=wifi
1106 interface-name=wlp3s0
1107 permissions=
1108 
1109 [wifi]
1110 band=bg
1111 channel=9
1112 mac-address=00:1E:65:41:55:61
1113 mac-address-blacklist=
1114 mode=ap
1115 ssid=SmartBox
1116 
1117 [ipv4]
1118 dns-search=
1119 method=shared
1120 
1121 [ipv6]
1122 addr-gen-mode=stable-privacy
1123 dns-search=
1124 ip6-privacy=0
1125 method=shared
```
Did this connection actually become active and working? That would be useful information to know.

jvonau:
> @jvonau are the 1+6 mentions of "AP" in bold below extraneous — not offering any hope at all to enable AP mode? (as an example, apparently from a Kingdel PC)

I would think none.
> > Driver supports full state transitions for **AP/GO** clients
> 
> (Above excerpts are from iiab-diagnostics http://sprunge.us/NMOYJl?en)

 see https://wiki.gentoo.org/wiki/Hostapd & https://gist.github.com/Semant1ka/ee087c2bd1fbf6b0287c3307b8d4f291

roles/network/tasks/detected_network.yml
```diff
@@ -101,6 +101,16 @@
   set_fact:
     num_wifi_interfaces: "{{ count_wifi_interfaces.stdout|int }}"
 
+- name: Check for Access Point capablility with 'iw list'
+  command: iw list | grep -v AP: | grep AP | wc -l
```
holta, line 105: Sometimes `AP` appears in comments like the following &mdash; e.g. on a Raspberry Pi 4 Model B Rev 1.1 in this case:

```
root@box:~# iw list | grep -vn AP: | grep AP
21:              * AP
164:             * #{ managed } <= 1, #{ AP } <= 1, #{ P2P-client } <= 1, #{ P2P-device } <= 1,
188:            * [ 4WAY_HANDSHAKE_AP_PSK ]: AP mode PSK offload support
```

Should we... perhaps instead... grep for any string(s) `* AP` that appear at the end of any lines?

(If the existence of at least one of these... is indeed the decider/clincher?)

```suggestion
  command: iw list | grep '* AP$' | wc -l
```

jvonau: Lines 164 188 will not exist without line 21 and the test below is against an integer, what return value is provided without wc -l?

holta: > Lines 164 188 will not exist without line 21

For sure, I agree.

I was just suggesting a failsafe &mdash; e.g. in case the string 'AP' mysteriously appears in comments arising from diverse/wonky WiFi cards in future?

Towards possibly making things even more resilient?

> what return value is provided without wc -l?

I restored the `wc -l` into the suggestion above a few minutes ago.

bash can do the counting, if counting's required, that works too.

Whereas another option is to avoid bash and have Ansible do the counting &mdash; e.g. using code like `REGISTER_VAR.stdout_lines | length > 0` here:

https://github.com/iiab/iiab/blob/db1531eef57091083dd40f9612c309b7af64019e/roles/1-prep/tasks/hardware.yml#L23-L33

holta:
Every year, lots of people make clear they want/need this!
(Of course with a clean UX + clean documentation, that might come later if the stars align...)

What test cases / HW / OS should be prioritized to smoke-test this PR?

holta:
Can this PR be considered prior to...?

- PR #3173
- PR #3179 related to #3097

jvonau:
>Untested

missing quotes corrected in [3216](https://github.com/iiab/iiab/pull/3216/commits/799e7140c71a2fed27575d939ff48957a9b8cd79)
didn't like command corrected in [3216](https://github.com/iiab/iiab/pull/3216/commits/f55f4f3f63ffb48a21a596378bd3b078b3c1678f)

jvonau:
> Can this PR be considered prior to...?
> 
>     * PR [Wifi net2 #3173](https://github.com/iiab/iiab/pull/3173)
> 
>     * PR [set 'host_country_code' to value found in wpa_supplicant.conf [on Raspberry Pi OS only, deferring Ubuntu/Debian upstream issues] #3179](https://github.com/iiab/iiab/pull/3179) related to [Should host_country_code in local_vars.yml update wpa_supplicant.conf not just hostapd.conf ? #3097](https://github.com/iiab/iiab/issues/3097)

Shouldn't the PR number tell you the age?

jvonau:
Found this while running ./iiab-network
>TASK [firmware : Symlink /lib/firmware/brcm/brcmfmac43455-sdio.bin.iiab -> brcmfmac43455-sdio.bin_2021-11-30_minimal (as rpi3bplus_rpi4_wifi_firmware is "19")] ***
[WARNING]: Cannot set fs attributes on a non-existent symlink target. follow should be set to False to avoid this.
ok: [127.0.0.1]

holta:
> brcmfmac43455-sdio.bin_2021-11-30_minimal

Can you tell if there might have been some kind of a network glitch during [firmware/tasks/download.yml](https://github.com/iiab/iiab/blob/master/roles/firmware/tasks/download.yml#L12-L24) &mdash; when it was trying to download http://d.iiab.io/packages/brcmfmac43455-sdio.bin_2021-11-30_minimal to `/lib/firmware/brcm/` ?

(Did several of those 7 files possibly fail to download, if you can tell what happened?)

Issue #3072, opened 2021-12-18
IIAB Kiwix proxy fails to proxy certain articles?


holta:
Apparently there's a bug in our 11-line Kiwix proxy...

https://github.com/iiab/iiab/blob/master/roles/kiwix/templates/kiwix-nginx.conf.j2

...that prevents very real articles like ".apple" from being displayed.  As you can see this leads to *403 Forbidden*:

http://iiab.me/kiwix/wikipedia_en_all_maxi_2021-03/A/.apple

Whereas it works without our NGINX Kiwix proxy:

http://10.8.0.14:3000/kiwix/wikipedia_en_all_maxi_2021-03/A/.apple

As mentioned here:

- kiwix/kiwix-tools#513

jvonau:
http://iiab.me/kiwix/search?content=wikipedia_en_all_maxi_2021-03&pattern=apple

holta:
> http://iiab.me/kiwix/search?content=wikipedia_en_all_maxi_2021-03&pattern=apple

The above is a "Full Text Search" for "apple" with 97,264 search results.

Whereas the easier way to experience this bug is using a "Title Search" for "apple" (a search dropdown should appear almost instantly, before you hit Enter...showing the 10 choices outlined at https://github.com/kiwix/kiwix-tools/issues/513#issue-1083485408).

jvonau:
http://iiab.me/kiwix/search?content=wikipedia_en_all_maxi_2021-03&pattern=TLD Ok I see it is the leading . in the final part of the url. The nginx kwix template passes everything though to the port, the filtering might be in one of the other templates in use.

jvonau:
Seems more of a security feature of nginx, https://www.reddit.com/r/nginx/comments/i6et27/displaying_hidden_files/ A normal doc_root could have sensitive data within the .file or .directory, that is the point of hiding them there to begin with.

jvonau:
```
jvonau@box:/media/usb1/git/iiab$ ls -lh
total 3.2M
-rwxr-xr-x  1 root root  444 Oct 20 17:25 ansible.cfg
-rwxr-xr-x  1 root root   23 Oct 20 17:25 ansible_hosts
-rwxr-xr-x  1 root root  574 Oct 20 17:25 collections.yml
-rwxr-xr-x  1 root root  139 Oct 20 17:25 CONTRIBUTING.md
-rwxr-xr-x  1 root root 1.9K Oct 20 17:25 iiab-configure
[...]

PR #3073, opened 2021-12-18
Test adding allow all

jvonau:
for the hidden file issue
### Fixes bug:
#3072 
systemctl restart nginx is clean


holta:
Thanks.

Let's merge this experimentally and I'll test it right away.

holta:
Unfortunately applying this PR (`cd /opt/iiab/iiab` then `sudo ./runrole --reinstall kiwix`) did not improve the situation.

( http://10.8.0.14/kiwix/wikipedia_en_all_maxi_2021-03/A/.apple still shows **403 Forbidden** )

jvonau:
Now I wonder why this was not hand tested before merging, and if this is not the fix why is it not reverted yet?

holta:
This PR was tested immediately after it was merged.

It did not help, as explained above.

So it was reverted by:

- PR #3074

jvonau:
Still does not explain why a single one line change was not hand tested on the machine with the dataset installed before merging.

PR #3082, opened 2021-12-26
www_base/tasks/php-stem.yml needs phpX.Y-sqlite3 (es-wikihow search was failing when osm-vector-maps was not installed)

holta:
Searching [es-wikihow](https://rachel.worldpossible.org/viewmod/es-wikihow) was not working when [osm-vector-maps](https://github.com/iiab/iiab/tree/master/roles/osm-vector-maps) was not installed.

This PR fixes the missing dependency bug.

Background reports:

- https://github.com/iiab/php-stem/pull/3#issuecomment-1001070050
- https://github.com/iiab/iiab/pull/3080#issuecomment-1001083415

Tangentially related:

- PR #3081

Issue #3077, opened 2021-12-21
(A) "size" erroneous in map-catalog.json (B) NEW OSM dataset from Oct/Nov 2020 (C) Prep for future datasets incl Satellite Photos


holta:
@georgejhunt

1) Should we fix `"size": 16841445376` that appears 3 times in http://download.iiab.io/content/OSM/vector-tiles/map-catalog.json which is apparently drawn from https://github.com/iiab/maps/blob/master/public/2021-01-27/map-catalog.json ?

   - [Line 21](https://github.com/iiab/maps/blob/master/public/2021-01-27/map-catalog.json#L21) under osm-planet_z0-z10_2019.mbtiles
   - [Line 47](https://github.com/iiab/maps/blob/master/public/2021-01-27/map-catalog.json#L47) under satellite_z0-z9_v3.mbtiles
   - [Line 75](https://github.com/iiab/maps/blob/master/public/2021-01-27/map-catalog.json#L75) under osm_africa_z11-z14_2019.mbtiles

2) Is [2020-10-06/modify_catalog.py](https://github.com/georgejhunt/maps/blob/addbase/public/2020-10-06/modify_catalog.py) the most recent and best way of doing this?

   Do you have any recommendations here, in general?

3) In contrast, 16841445376 appears only once (on [Line 124](https://github.com/iiab/iiab-admin-console/blob/master/roles/common/files/map/adm-map-catalog.json#L124), under osm_africa_z11-z14_2019.mbtiles) in `/library/www/html/common/assets/adm-map-catalog.json` ~= https://github.com/iiab/iiab-admin-console/blob/master/roles/common/files/map/adm-map-catalog.json &mdash; apparently because it's auto-created by something like https://github.com/iiab/iiab-admin-console/blob/master/roles/cmdsrv/files/scripts/osm-vect_v2_finish_install.py ?

4) NEW OpenStreetMap datasets from Oct/Nov 2020 are now posted to https://archive.org/details/osm-vector-mbtiles that we should take a look at in coming months (when we can!) in service to low-income communities worldwide and their students who will greatly benefit...

   We should lay out a rough plan for 2022 so these communities get what they want (is Sentinel satellite data...essentially auto-updated at the moment?!)

Thanks!

Historical footnote:

- PR #1612

Tangentially related:

- #3064
- PR #3083

tim-moody:
there are two catalogs because the osm module and adm cons do the math differently, and the difference with respect to size reflect the different math.

[...]

PR #3083, opened 2021-12-26
osm_vector_maps was repeatedly failing to install, so use 'timeout: "{{ download_timeout }}"' with get_url (in 10 roles!) [& change IIAB's 'download_timeout: 200' to 100 sec]

holta:
IIAB's OpenStreetMap installer ([osm-vector-maps/tasks/install.yml](https://github.com/iiab/iiab/blob/master/roles/osm-vector-maps/tasks/install.yml)) is very fragile when GitHub and/or your ISP (Internet) is overwhelmed.

As I discovered repeatedly yesterday, the problem is especially acute when trying to install osm-vector-maps as a result of this role downloading so many different files.

This PR fixes the problem, by avoiding Ansible's default timeout for `get_url:` of 10 seconds, and using 100 seconds instead.  

This PR also establishes `'timeout: "{{ download_timeout }}"'` essentially across the board, for almost all of repo iiab/iiab's ~50 uses of get_url, as this was missing from almost 10 roles.

The exception (where we do not want to use `timeout: "{{ download_timeout }}"`) is: https://github.com/iiab/iiab/blob/2634fa207bece346d6f40e07ced388966293f783/roles/0-init/tasks/network.yml#L22-L31

Note that the Internet is comparatively more stable in major cities around the world in 2021, as compared to 5 years ago, so this PR also lowers `download_timeout: 200` to `download_timeout: 100` (seconds), to try to avoid pointlessly frustrating delays.

**In future years we (might!) possibly want to further lower this IIAB setting (in default_vars.yml etc) from 100 seconds, down to 60 seconds or so?**

Tangentially related:

- #3077
- PR #3082

### DIFF
--- a/roles/azuracast/tasks/install.yml
+++ b/roles/azuracast/tasks/install.yml
@@ -23,15 +23,15 @@
   get_url:
     url: "{{ docker_sh_url }}"
     dest: "{{ azuracast_host_dir }}/"
-    timeout: "{{ download_timeout }}"
     mode: 0755
+    timeout: "{{ download_timeout }}"
 
 - name: AzuraCast - Download AzuraCast's docker-compose.yml sample from GitHub to {{ azuracast_host_dir }}
   get_url: 
     url: "{{ docker_compose_url }}"
     dest: "{{ azuracast_host_dir }}/docker-compose.yml"
-    timeout: "{{ download_timeout }}"
     mode: 0755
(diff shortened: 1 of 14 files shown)


holta:
This PR was tested over the past 2 hours with a LARGE-sized install on RasPiOS 11 on RPi 4.

All looks great.

This should help make IIAB installs substantially more reliable.

PR #3104, opened 2022-01-14
"Power Off" button leads to big classroom problems, even when /library/www/html/home/menu.json contains: "allow_poweroff": false;

holta:
This is an ongoing problem, year after year.

Apparently this longstanding issue results from partially loaded or cached versions (of an installed IIAB's home page) which reveal to students how they can effectively sabotage an entire classroom's use of IIAB — using the "Power Off" button shown below:

IMG-20220114-WA0000~2

So this PR resolves default_vars.yml and local_vars files to do what most all schools want — put an end to these disruptive problems.

At the same time this precautionary default can easily be overridden — for family-sized IIAB installs if necessary, among a very tightknit group of people where routine sabotage is less of a risk, by setting `apache_allow_sudo: True` — or using local_vars_medical.yml to do exactly that, e.g. for a small clinic wanting the "Power Off" button available to all, to save energy when IIAB is not in use.

### DIFF
--- a/vars/default_vars.yml
+++ b/vars/default_vars.yml
@@ -324,7 +324,7 @@ nginx_high_php_limits: False
 # https://github.com/iiab/iiab/blob/master/roles/www_options/tasks/main.yml#L53-L133
 # ...ARE SUITABLE FOR YOUR HARDWARE, as saved in: /etc/php/<VERSION>/*/php.ini
 
-# Make this True to enable http://box/common/services/power_off.php button:
+# Make this True to enable http://box/js-menu/menu-files/services/power_off.php
 apache_allow_sudo: False
 
 # Toggle iiab-refresh-wiki-docs scraping for offline docs (http://box/info)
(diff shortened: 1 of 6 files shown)


holta:
Context of how var `apache_allow_sudo` works, to enable-or-disable this general "Power Off" capability:

https://github.com/iiab/iiab/blob/master/roles/www_options/tasks/main.yml#L145-L156

(Not the best-named variable for sure, but it works!)

holta:
Discussed on IIAB's weekly community call (http://minutes.iiab.io).

@tim-moody will also create a companion PR to greatly reduce the number of situations where the "Power Off" button is being accidentally displayed:

- PR iiab/iiab-admin-console#458

i.e. it should almost never appear accidentally going forward, except in some rare situations e.g. when the browser isn't hard-refreshed / hard-reloaded when holding down the [SHIFT KEY] while clicking the [RELOAD BUTTON] in your web browser.

Issue #2552, opened 2020-09-29
Usb booted Rpi hardware can display the contents of "/"


jvonau:
### Expected behavior and actual behavior:
Boot Rpi hardware capable of usb boot from an usb drive and not be able to see contents of '/' at http://box/usb

from: #2497 
>981 =IIAB==========================================================================
 982 COMMAND: /bin/df -h    # Disk usage
 983 
 984 Filesystem      Size  Used Avail Use% Mounted on
 985 /dev/root       917G   11G  860G   2% /
 991 /dev/sda1       253M   54M  199M  22% /boot

> 995 =IIAB==========================================================================
 996 COMMAND: /bin/lsblk    # Partition mount points
 997 
 998 NAME   MAJ:MIN RM   SIZE RO TYPE MOUNTPOINT
 999 sda      8:0    0 931.5G  0 disk 
1000 ├─sda1   8:1    0   256M  0 part /media/usb0
1001 └─sda2   8:2    0 931.3G  0 part /media/usb1

### Steps to reproduce the problem:
see above and note the boot device is mounted at the usb share point

### IIAB version, OS version, hardware details:
master as of 2497

Suggestions:
1. toggle local_vars iiab_usb_lib_show_all: to False, and document the effects of True
or
2. add code to override the selection of True when usb-boot is detected, and note why the override is present

holta:
@pumpkinpi85 have you too observed this?

CONTEXT: http://FAQ.IIAB.IO > "4 Can teachers display their own content?"

While USB boot of IIAB machines is not common, this is definitely undesirable behavior in those cases!

holta:
For now I've added this as an [IIAB 7.2 Known Issue](https://github.com/iiab/iiab/wiki/IIAB-7.2-Release-Notes#known-issues) : 

- If you're booting from a USB device (e.g. an external disk or flash drive) please set `iiab_usb_lib_show_all: to False` in [/etc/iiab/local_vars.yml](http://wiki.laptop.org/go/IIAB/FAQ#What_is_local_vars.yml_and_how_do_I_customize_it.3F) so that your filesystem is not viewable at http://box/usb  <sub><sub>[#2552](https://github.com/iiab/iiab/issues/2552)</sub></sub>

I've also posted an ALL CAPS warning here:

- http://FAQ.IIAB.IO > "45 Can I install IIAB onto a USB drive?"

[...]

PR #3121, opened 2022-02-17
Use 'raspi-config --expand-rootfs' in 1-prep/templates/iiab-rpi-max-rootfs.sh (for external USB disks, that also need their filesystem expanded)

holta:
Background: `/etc/systemd/system/iiab-rpi-root-resize.service` invokes `/usr/sbin/iiab-rpi-max-rootfs.sh` on boot, which tries to expand IIAB's root filesystem if the file `/.resize-roofs` exsits.

The shortcoming of `/usr/sbin/iiab-rpi-max-rootfs.sh` to date, is that until now it assumed... 

```
# Assumes sd card style partition name like <device>p<partition number>
# Only works on F22 + where resizepart command exists
```

So with this PR, external IIAB boot disks/devices are now possible, at least in the case where Raspberry Pi OS is being used (or any OS that makes available the command `/usb/bin/raspi-config --expand-rootfs` !)

Further improvement to make this work on other OS's would be great in future &mdash; e.g. on Ubuntu, Mint and Debian?

In any case, this PR was tested and confirmed to work with 32-bit Raspberry Pi OS Lite on RPi 4, using an external USB disk.

### DIFF
--- a/roles/1-prep/templates/iiab-rpi-max-rootfs.sh
+++ b/roles/1-prep/templates/iiab-rpi-max-rootfs.sh
@@ -1,20 +1,32 @@
 #!/bin/bash -x
-# Resize rootfs and its partition on the rpi SD card to maximum size
-# To be used by systemd service on boot
-# Only resizes if /.resize-rootfs exists
-# Assumes root is last partition
-# Only works on F22 + where resizepart command exists
-# Assumes sd card style partition name like <device>p<partition number>
 
-if [  -f /.resize-rootfs ];then
-  echo "$0: maximizing rootfs partion"
-  # Calculate root partition
-  root_part=`lsblk -aP -o NAME,MOUNTPOINT|grep  'MOUNTPOINT="/"' |awk -F\" '{ print $2 }'`
-  root_dev=${root_part:0:-2}
-  root_part_no=${root_part: (-1)}
+# Resize rootfs and its partition on the rpi SD card (or external USB
+# disk if possible, e.g. with Raspberry Pi OS) to maximum size.
 
-  # Resize partition
-  growpart /dev/$root_dev $root_part_no
-  resize2fs /dev/$root_part
-  rm /.resize-rootfs
+# To be used by /etc/systemd/system/iiab-rpi-root-resize.service on boot.
+# Only resizes if /.resize-rootfs exists.
+# Assumes root is last partition.
+
+if [ -f /.resize-rootfs ]; then
+    echo "$0: maximizing rootfs partion"
+
+    if [ -x /usr/bin/raspi-config ]; then
+        # 2022-02-17: Works in many more situations, e.g. with USB disks (not
+        # just microSD cards).  IF ONLY THIS ALSO WORKED ON Ubuntu/Mint/etc !
+        raspi-config --expand-rootfs
+    else
+        # Assumes sd card style partition name like <device>p<partition number>
+        # Only works on F22 + where resizepart command exists
+
+        # Calculate root partition
+        root_part=`lsblk -aP -o NAME,MOUNTPOINT | grep  'MOUNTPOINT="/"' | awk -F\" '{ print $2 }'`
+        root_dev=${root_part:0:-2}
+        root_part_no=${root_part: (-1)}
+
+        # Resize partition
+        growpart /dev/$root_dev $root_part_no
+        resize2fs /dev/$root_part
+    fi
+
+    rm /.resize-rootfs
 fi


tim-moody:
How was this tested? In the past the existing code solved the problem that raspi-config would not expand a rootfs compacted with min-sd. The comments that led to the proposed change also appear to be out of date. I don't see how this command assumes partition names like `<device>p<partition number>`

`lsblk -aP -o NAME,MOUNTPOINT | grep  'MOUNTPOINT="/"' | awk -F\" '{ print $2 }'` just returns the partition mounted on /

holta:
> How was this tested?

Using a Raspberry Pi 4 with IIAB on an external USB boot disk (that had been minimized using using min-sd, and so contained `/.resize-rootfs` to request expanding of the filesystem on next boot).

> `lsblk -aP -o NAME,MOUNTPOINT | grep 'MOUNTPOINT="/"' | awk -F\" '{ print $2 }'` just returns the partition mounted on /

True.

That being insufficient when IIAB is on an external boot device, which typically auto-mounts to `/media/usb0` and `/media/usb1` (or similar!)

Thankfully the new command (`raspi-config --expand-rootfs`) takes cares of everything very simply, on Raspberry Pi OS anyway.

tim-moody:
Well I have to say it surprises me that the rootfs could be mounted on /media/usbx. Does webroot become /media/usbx/library/www/html?

There are, of course, two tests. Does the new feature work in the new environment and does it work in the old one?

holta:
> Well I have to say it surprises me that the rootfs could be mounted on /media/usbx. Does webroot become /media/usbx/library/www/html?

Yes.

(But normally there should be no need to use the above long-winded path, as the canonical `/library/www/html` also works.)

Typical RPi-IIAB-on-microSD output:

```
root@box:~# lsblk -aP -o NAME,MOUNTPOINT | grep /
NAME="mmcblk0p1" MOUNTPOINT="/boot"
NAME="mmcblk0p2" MOUNTPOINT="/"

root@box:~# lsblk -aP -o NAME,MOUNTPOINT | grep 'MOUNTPOINT="/"' | awk -F\" '{ print $2 }'
mmcblk0p2
```

Typical RPi-IIAB-on-external-USB-disk output:

```
root@box:~# lsblk -aP -o NAME,MOUNTPOINT | grep /
NAME="sda2" MOUNTPOINT="/media/usb1"
NAME="sda1" MOUNTPOINT="/media/usb0"

root@box:~# lsblk -aP -o NAME,MOUNTPOINT | grep 'MOUNTPOINT="/"' | awk -F\" '{ print $2 }'
[ PRODUCES NO OUTPUT ]
```

Here's some of the USB external disk mounting happening behind the scenes FWIW:

```
root@box:~# mount | grep sda
/dev/sda2 on / type ext4 (rw,noatime)
/dev/sda1 on /boot type vfat (rw,relatime,fmask=0022,dmask=0022,codepage=437,iocharset=ascii,shortname=mixed,errors=remount-ro)
/dev/sda1 on /media/usb0 type vfat (rw,nodev,noexec,noatime,nodiratime,fmask=0022,dmask=0022,codepage=437,iocharset=ascii,shortname=mixed,errors=remount-ro)
/dev/sda2 on /media/usb1 type ext4 (rw,nodev,noexec,noatime,nodiratime)
```

I've put a short summary in the code (PR #3122, just merged) in case further improvements are possible in future.

The issue of http://box/usb displaying the contents of the entire Linux filesystem is related: (arguably this is a feature rather than a bug, for certain kids trying to learn Linux if they get a vote anyway, but of course not all teachers agree ;)

- https://github.com/iiab/iiab/issues/2552

> There are, of course, two tests. Does the new feature work in the new environment and does it work in the old one?

Yes, I did spot-testing using a microSD too.  But further testing never hurts (raspi-config is heavily relied upon, but its code changes monthly if not weekly, and of course it does nothing to help Ubuntu/Mint/Debian etc!)

jvonau:
just use root_part=`findmnt -n / | awk '{ print $2 }' | awk -F\/ '{ print $3 }'` in place of https://github.com/iiab/iiab/blob/master/roles/1-prep/templates/iiab-rpi-max-rootfs.sh#L27

holta:
> just use root_part=`findmnt -n / | awk '{ print $2 }' | awk -F\/ '{ print $3 }'` in place of https://github.com/iiab/iiab/blob/master/roles/1-prep/templates/iiab-rpi-max-rootfs.sh#L27

Extremely constructive suggestion, thanks @jvonau.

The other challenge was [identifying the device name on Line 28](https://github.com/iiab/iiab/blob/master/roles/1-prep/templates/iiab-rpi-max-rootfs.sh#L28):

- Sometimes 2 characters need to be removed (e.g. `mmcblk0p2` to `mmcblk0`)
- Whereas other times just 1 character needs to be removed (e.g. `sda2` to `sda`)

The solution (that I borrowed from raspi-config) was:

```
root@box:~# lsblk -no pkname /dev/mmcblk0p2
mmcblk0
```

In the end I decided to borrow several other improvements from raspi-config's `do_expand_rootfs()` as well, to modernize with a few of their safety checks as well:

- PR #3137

Ref:

- #3136

jvonau:
Note the double mount of /dev/sda2, I'll bet if you unmount /media/sda2 (use `sudo systemctl stop usbmount*`) and retest the failing "lsblk [ PRODUCES NO OUTPUT ]" will now show the correct results.

holta:
> Note the double mount of /dev/sda2, I'll bet if you unmount /media/sda2 (use `sudo systemctl stop usbmount*`) and retest the failing "lsblk [ PRODUCES NO OUTPUT ]" will now show the correct results.

Yes, I assume that's true, per https://github.com/iiab/iiab/pull/3121#issuecomment-1045036899

(In any case, the solution borrowed from raspi-config at PR #3136 bypasses all such issues thankfully!)

Issue #3136, opened 2022-03-14
IIAB boot devices should auto-expand for all OS's (e.g. Ubuntu, Mint, Debian) not just RasPiOS


holta:
Currently this works for microSD cards, but not for external USB boot devices which are growing more common for reasons of cost and reliability!

Modifying IIAB's current code here should be relatively straightforward, if someone can volunteer their time to make it happen:

- #3121
- #3122

holta:
Done:

- PR #3137 

(And evolving based on...)

- PR #3138
- PR #3139

PR #3137, opened 2022-03-15
iiab-expand-rootfs for more HW/OS's w/ safety checks, per raspi-config's do_expand_rootfs()

holta:
Building off @jvonau's excellent suggestion at:

- https://github.com/iiab/iiab/pull/3121#issuecomment-1067579437

Several safety checks are also added, e.g. multi-digit partition numbers, and verification that the rootfs partition is indeed the last partition, etc.  In the end, most improvements were borrowed (almost verbatim) from `raspi-config`'s well-tested `do_expand_rootds()` here: https://github.com/RPi-Distro/raspi-config/blob/master/raspi-config

In summary, this modernizes:

- 1-prep/templates/iiab-expand-rootfs originally iiab-rpi-max-rootfs.sh
- 1-prep/templates/iiab-expand-rootfs.service originally iiab-rpi-root-resize.service

Allowing for use on other OS's (e.g. Ubuntu/Mint/Debian) with other HW (e.g. external boot disks, not just Raspberry Pi microSD cards).

Requirements are re-organized a bit, given this is no longer just for Raspberry Pi, and docs are clarified e.g. to explain flag `/.expand-rootfs` which matches the new name, in keeping with Raspberry Pi norms.

Related:

- PR #3122
- #3136


roles/1-prep/templates/iiab-expand-rootfs
```diff
@@ -0,0 +1,57 @@
...
+        ROOT_PART="$(findmnt / -o SOURCE -n)"    # e.g. /dev/sda2 or /dev/mmcblk0p2
+        ROOT_DEV="/dev/$(lsblk -no pkname "$ROOT_PART")"    # e.g. /dev/sda or /dev/mmcblk0
+
+        ROOT_PART_NUM="$(echo "$ROOT_PART" | grep -o "[[:digit:]]*$")"    # e.g. 2
+        # SLOW (~10 seconds) but it works!
+        LAST_PART_NUM=$(parted "$ROOT_DEV" -ms unit s p | tail -n 1 | cut -f 1 -d:)
+
+        if [ $ROOT_PART_NUM -ne $LAST_PART_NUM ]; then
+            echo "ERROR: $ROOT_PART partition ($ROOT_PART_NUM) is not the last partition ($LAST_PART_NUM). Don't know how to expand."
+            return 0
```
jvonau: would exit 1 be better here?

holta: Thanks.

I borrowed `return 0` from https://github.com/RPi-Distro/raspi-config/blob/684fbe183a53f60cccee21c341d5754357cf1288/raspi-config#L170 but `exit 1` is a useful fix and improvement I'll add now.

roles/1-prep/templates/iiab-expand-rootfs
```diff
@@ -0,0 +1,57 @@
...
+        # https://github.com/RPi-Distro/raspi-config/blob/master/raspi-config
+        raspi-config --expand-rootfs
+    else
+        # 2022-03-15: Borrows from above raspi-config URL's do_expand_rootfs()
+        ROOT_PART="$(findmnt / -o SOURCE -n)"    # e.g. /dev/sda2 or /dev/mmcblk0p2
+        ROOT_DEV="/dev/$(lsblk -no pkname "$ROOT_PART")"    # e.g. /dev/sda or /dev/mmcblk0
+
+        ROOT_PART_NUM="$(echo "$ROOT_PART" | grep -o "[[:digit:]]*$")"    # e.g. 2
+        # SLOW (~10 seconds) but it works!
+        LAST_PART_NUM=$(parted "$ROOT_DEV" -ms unit s p | tail -n 1 | cut -f 1 -d:)
```
jvonau, line 25: should really add 'parted' as a dependency, same as cloud-guest-utils

roles/1-prep/tasks/hardware.yml
```diff
@@ -1,17 +1,38 @@
+- name: Install package 'cloud-guest-utils' so growpart is available for expanding rootfs partition during boot
+  package:
+    name: cloud-guest-utils    # 2022-03-15: For RasPiOS especially.  Ubuntu has still pre-installed, for use with cloud-init.
+    state: present
+
```
jvonau, line 7: You are using raspi-config on RaspOS why install universally unless the use of raspi-config is going to go away?

roles/1-prep/templates/iiab-expand-rootfs
holta: Done: 85eb059

roles/1-prep/templates/iiab-expand-rootfs
```diff
@@ -34,6 +31,8 @@ if [ -f /.expand-rootfs ] || [ -f /.resize-rootfs ]; then
         growpart $ROOT_DEV $ROOT_PART_NUM
         resize2fs $ROOT_PART
 
+        # 2022-03-15: Legacy code below for microSD cards but NOT USB boot disks
+
```
jvonau, line 38: The use of /dev/sda is not just a 'usb' boot disk but any boot device that uses sdX naming, this should work nicely with 'normal' spinning disks also. (think older sata drives)

roles/1-prep/tasks/hardware.yml
holta: Good point.  Installing `cloud-guest-utils` is indeed currently overkill on RasPiOS.

`/usr/bin/growpart` is only 22 kB however, and the total package size is also tiny:

```
root@box:~# apt show cloud-guest-utils | grep Size

WARNING: apt does not have a stable CLI interface. Use with caution in scripts.

Installed-Size: 62.5 kB
Download-Size: 23.7 kB
```
```
root@box:~# dpkg-query -L cloud-guest-utils
/.
/usr
/usr/bin
/usr/bin/ec2metadata
/usr/bin/growpart
/usr/bin/vcs-run
/usr/share
/usr/share/doc
/usr/share/doc/cloud-guest-utils
/usr/share/doc/cloud-guest-utils/NEWS.Debian.gz
/usr/share/doc/cloud-guest-utils/changelog.Debian.gz
/usr/share/doc/cloud-guest-utils/changelog.gz
/usr/share/doc/cloud-guest-utils/copyright
/usr/share/man
/usr/share/man/man1
/usr/share/man/man1/growpart.1.gz
```

Currently I'm inclined to continue installing it for all OS's, in case such infra utilities later help others.

But if others feel otherwise (e.g. if growpart proves to be unmaintained and/or causing problems?) then we could later try to remove it completely?

roles/1-prep/templates/iiab-expand-rootfs
```diff
@@ -0,0 +1,59 @@
...
+# Expand rootfs partition to its maximum size, if /.expand-rootfs exists.
+# Used by /etc/systemd/system/iiab-expand-rootfs.service on IIAB boot.
+
+# Should work with all Linux OS's boot disks -- regardless whether Raspberry Pi
+# microSD cards, external USB drives, internal spinning disks or SSD's, etc.
+
+# Verifies that rootfs is the last partition.
+
+if [ -f /.expand-rootfs ] || [ -f /.resize-rootfs ]; then
```
jvonau, line 11: Not sure why there would be a need to introduce another flag to act upon, all the tools in iiab-factory know nothing about the new 'expand-rootfs' flag.

holta:
Tested with Ubuntu Server 21.10 on Raspberry Pi 4.

roles/1-prep/templates/iiab-expand-rootfs
holta: I've implemented consistent `/.expand-rootfs` naming across the board in both repos, per raspi-config norms:

- PR iiab/iiab-factory#208

(Legacy flag name /.resize-rootfs will still work during the interim.)

PR #3138, opened 2022-03-15
generalize wifi dependencies based on hardware present

jvonau:
Untested, jiggled things around, detection is a rehash of what is in network, will need a rebase once #3137 is pushed. I'll test then..


jvonau:
Can this be tested, I will be unavailable for the next while. More so on a VM without wifi where the exclusion was based on hardware (rpi_model != "none") previously and on RPi hardware running ubuntu-server where the packages are noted to be missing.

holta:
Please see this follow-on PR, that builds upon this PR #3138:

- PR #3139

PR #3139, opened 2022-03-15
1-prep/tasks/hardware.yml: Cleaner modularization & assoc fixups

holta:
Building on @jvonau's:

- PR #3138

and

- PR #3137

Proposes that var `has_wireless` be renamed to `internal_wifi`.

Testing continues.


holta:
Tested on machines with and without internal WiFi hardware.

@jvonau feel free to suggest improvements after this is merged, if any such are necessary.

jvonau:
Other than 'internal' not being accurate as the code will pickup external devices such as usb wifi devices, hence the reason for using the more general 'has_wifi_device', has_wireless and wifi.yml being used in the original commit. Whatever, just another example of not really understanding the underling code with a hint of 'I must be the latest commit' tossed in for good measure. Keep it up you will run out of fingers to stick into the leaking dike. Great way to push away developers that have the skills that are lacking from your portfolio. Smells a lot like sugarlabs around here from where I'm sitting.

PR #3145, opened 2022-03-24
Wifi net

jvonau:
Reorder/grouping of network components with the future eye of being made optional and/or only required with wifi hardware.
The fly in the ointment would be admin-console relying on network entries in iiab.ini being present.


holta:
I've added the 'Question' tag so others can hopefully weigh in here?

(Presumably commands like the following...will not change?)

- [./iiab-network](https://github.com/iiab/iiab/blob/master/iiab-network)
- [iiab-hotspot-off](https://github.com/iiab/iiab/blob/master/roles/network/templates/network/iiab-hotspot-off)
- [iiab-hotspot-on](https://github.com/iiab/iiab/blob/master/roles/network/templates/network/iiab-hotspot-on)
- [iiab-internet-off](https://github.com/iiab/iiab/blob/master/roles/network/templates/gateway/iiab-internet-off)
- [iiab-internet-on](https://github.com/iiab/iiab/blob/master/roles/network/templates/gateway/iiab-internet-on)
- ...

holta:
Only tangentially related, but fyi Linux too is ever-evolving here:

"Linux 5.18 Brings Many Networking Changes, New WiFi / Wired Hardware Support"
https://www.phoronix.com/scan.php?page=news_item&px=Linux-5.18-Networking

holta:
Two questions to help understand this PR:

1) Currently [./iiab-configure](https://github.com/iiab/iiab/blob/master/iiab-configure) and [./iiab-network](https://github.com/iiab/iiab/blob/master/iiab-network) are separate and entirely distinct[*].

   This PR would require changing everyone's mental model here (and exhautively changing all documentation over coming months, until people understand their assumptions are no longer valid!)  Be that as it may: what's the primary/overall benefit of this proposed new approach, of encapsulating roles/network fully within roles/9-local-addons ?

   [*] I've told a lot of people in recent years, directly and also via documentation, about these 2 commands being completely separate.  (To date, `./iiab-configure` runs [Stage 0, followed by Stages 4-9](https://github.com/iiab/iiab/blob/master/iiab-from-cmdline.yml), WITHOUT running the ['network' role](https://github.com/iiab/iiab/tree/master/roles/network#network-readme).)  So if this is indeed changing, let's just offer a clean explanation as to why, so we don't leave people confused.

2) If `when: has_wifi_device is defined` is implemented in future (as a condition for running [roles/network](https://github.com/iiab/iiab/tree/master/roles/network)) this would mean that iptables firewall package(s) would not be installed or set up on machines lacking {internal WiFi, USB WiFi adapter, or similar}.

   Does this make sense (i.e. to dispense with IIAB's firewall altogether in such situations, i.e. is this safe) and if so what are the reasons?

   (A USB WiFi adapter might possibly be plugged into the IIAB at any time in future, so this lack of firewall appears concerning, when packages like `iptables-persistent` cannot easily be added in the field.  Specifically: one would normally run `./iiab-network` after inserting the USB WiFi adapter, even if in a remote/offline location where apt packages are no longer installable.)

jvonau:
> Two questions to help understand this PR:
> 
>     1. Currently [./iiab-configure](https://github.com/iiab/iiab/blob/master/iiab-configure) and [./iiab-network](https://github.com/iiab/iiab/blob/master/iiab-network) and separate and entirely distinct[*].

That behavior is unchanged, /iiab-configure does **not** use iiab-stages.yml, and if the install does not detect wifi hardware the role is skipped, could just make the call to install the prerequisites for network and add an override for those who don't want/need the full blown network dominance that IIAB forces upon every user.

>        This PR would require changing everyone's mental model here (and exhautively changing all documentation over coming months, until people understand their assumptions are no longer valid!)  Be that as it may: what's the primary/overall benefit of this proposed new approach, of encapsulating roles/network fully within roles/9-local-addons ?

Think you are blowing smoke up everybody's behind, running around like chicken little... The world is not changing, just getting more selective on where we need to meddle with the underling OS.
  
>        [*] I've told a lot of people in recent years, directly and also via documentation, about these 2 commands being completely separate.  (To date, `./iiab-configure` runs [Stage 0, followed by Stages 4-9](https://github.com/iiab/iiab/blob/master/iiab-from-cmdline.yml), WITHOUT running the ['network' role](https://github.com/iiab/iiab/tree/master/roles/network#network-readme).)  So if this is indeed changing, let's just offer a clean explanation as to why, so we don't leave people confused.

Think a simple conditional on < stage 9 should suffice, that would not change iiab-configure at all.

> 
>     2. If `when: has_wifi_device is defined` is implemented in future (as a condition for running [roles/network](https://github.com/iiab/iiab/tree/master/roles/network)) this would mean that iptables firewall package(s) would not be installed or set up on machines lacking {internal WiFi, USB WiFi adapter, or similar}.

Think I want to add network_install to local_vars.yml to make networking fully optional under the end user's control, see little use in installing wifi related components on a machine running as a VM, will never have wifi added into the mix, or is not part of the end user's deployment due to using external access points.    
>        Does this make sense (i.e. to dispense with IIAB's firewall altogether in such situations, i.e. is this safe) and if so what are the reasons?

I'd hardly call iiab-gen-iptables a firewall, the only protected interface is the one detected during installation if the name of the device that has the default gateway changed you are on the network butt naked, that is why connections across openvpn allows access by default.

>        (A USB WiFi adapter might possibly be plugged into the IIAB at any time in future, so this lack of firewall appears concerning, when packages like `iptables-persistent` cannot easily be added in the field.  Specifically: one would normally run `./iiab-network` after inserting the USB WiFi adapter, even if in a remote/offline location where apt packages are no longer installable.)

I'd ensure the hardware was part of the initial install or ensure internet is available when re-configuring the network with what could become the "new" default gateway, most wifi devices require at the very least linux-firmware or some other firmware file/package to be downloaded but we do **not** install any such packages.

jvonau:
> > Two questions to help understand this PR:
> > ```
> > 1. Currently [./iiab-configure](https://github.com/iiab/iiab/blob/master/iiab-configure) and [./iiab-network](https://github.com/iiab/iiab/blob/master/iiab-network) and separate and entirely distinct[*].

> >    [*] I've told a lot of people in recent years, directly and also via documentation, about these 2 commands being completely separate.  (To date, `./iiab-configure` runs [Stage 0, followed by Stages 4-9](https://github.com/iiab/iiab/blob/master/iiab-from-cmdline.yml), WITHOUT running the ['network' role](https://github.com/iiab/iiab/tree/master/roles/network#network-readme).)  So if this is indeed changing, let's just offer a clean explanation as to why, so we don't leave people confused.
> > ```
> 
> Think a simple conditional on < stage 9 should suffice, that would not change iiab-configure at all.

Think we both missed the fact that the only place 'has_wifi_device' can be defined is within 1-prep, that automatically excludes iiab-from-cmdline.yml and iiab-from-console.yml from ever matching the 'when: has_wifi_device is defined' in 9-local-addons as both run as noted, without 1-prep. The stages < 9 part is a bit overkill but clear on the intention, run once only during the installation.

> 
> > ```
> > 2. If `when: has_wifi_device is defined` is implemented in future (as a condition for running [roles/network](https://github.com/iiab/iiab/tree/master/roles/network)) this would mean that iptables firewall package(s) would not be installed or set up on machines lacking {internal WiFi, USB WiFi adapter, or similar}.
> > ```
> 
> Think I want to add network_install to local_vars.yml to make networking fully optional under the end user's control, see little use in installing wifi related components on a machine running as a VM, will never have wifi added into the mix, or is not part of the end user's deployment due to using external access points.
> 

I've made this opt-out via network_install in default_vars.yml, so this would require manual user intervention to activate.

> > ```
> >    Does this make sense (i.e. to dispense with IIAB's firewall altogether in such situations, i.e. is this safe) and if so what are the reasons?
> > ```
> 
> I'd hardly call iiab-gen-iptables a firewall, the only protected interface is the one detected during installation if the name of the device that has the default gateway changed you are on the network butt naked, that is why connections across openvpn allows access by default.
> 
> > ```
> >    (A USB WiFi adapter might possibly be plugged into the IIAB at any time in future, so this lack of firewall appears concerning, when packages like `iptables-persistent` cannot easily be added in the field.  Specifically: one would normally run `./iiab-network` after inserting the USB WiFi adapter, even if in a remote/offline location where apt packages are no longer installable.)
> > ```
> 
> I'd ensure the hardware was part of the initial install or ensure internet is available when re-configuring the network with what could become the "new" default gateway, most wifi devices require at the very least linux-firmware or some other firmware file/package to be downloaded but we do **not** install any such packages.

By default the same package set is installed, the only thing that is changing is if the hotspot is automatically configured/created based on wifi hardware being present. Single interface installs would not receive any network mods, none are required to serve out content but the builtin firewall can be setup just by running iiab-network at any time.

holta:
> missed the fact that the only place 'has_wifi_device' can be defined is within 1-prep

Question: Wouldn't it make a lot more sense to move the `has_wifi_device` logic, from 1-prep to roles/network ?

- Correct logic would then result, when a USB WiFi adapter is later added (or removed) e.g. after IIAB software install or in the field.
- All (or almost all!) network logic would then be modularized / encapsulated into roles/network, as it probably should be.
- If other roles or even iiab-diagnostics (etc) really need to preview things like `has_wifi_device` in advance, they can invoke this logic (e.g. something like roles/network/tasks/check4wifidevices.yml ?) 
  - Similar to how other network tasks have sometimes been run early on during IIAB installations.

holta:
> a VM, will never have wifi added into the mix

Just a thought that there (might!) be some occasional exceptions, e.g. someone who specially configures an IIAB VM on Windows or any other OS, to take advantage of the chassis's WiFi or USB WiFi adapter.

tim-moody:
I have never gotten builtin wifi to work on a vm, though I haven't tried in a few years. But a router attached over ethernet works.

Issue #2584, opened 2020-10-21
Sierra Leone: "can no longer log into KA Lite and Admin Console" [RaspiOS Lite on RPi 4 disk was full: can a warning be made more visible?]


holta:
I'm trying to understand the situation that (might) have been caused by the implementer/operator installing additional packages after everything was working?

Output of [sudo iiab-diagnostics](https://github.com/iiab/iiab/blob/master/scripts/iiab-diagnostics.README.md) is here: http://sprunge.us/HdPJ80?en

jvonau:
>1014 /dev/root       217G  216G     0 100% /

That needs attention  Related #2157 #2205

holta:
>I was able to delete some KA-Lite videos to free up some space and that solved the problem . 
I plan to deploy this installation in a school in Sierra Leone.

Where should/could IIAB make a "DISK FULL" warning much more visible/colorful at http://box and/or on the Admin Console page (http://box.lan/admin) _prior to login_ ?

i.e. to give operators/educators a head's up when this relatively common situation occurs...

jvonau:
If the disk is that full you will never be able to login, or as reported access the webserver, that is why there was the 5% fudge factor for root normally present to help to prevent that kind of issue, but was tossed to the wayside in favor of 1% later raised to 2% on RPi hardware.

PR #3148, opened 2022-03-25
Remove 500M restriction on file uploads. Use 10G instead. [affects NGINX, PHP, Calibre-Web, Moodle, Nextcloud, WordPress, PBX, ETC]

holta:
Building on:

- PR #3096 
- PR #3142


holta:
This PR is especially for "LMS's" like WordPress, Nextcloud, Moodle, as large files (videos, educational documentaries, etc) become increasingly common.

But also potentially for Kolibri, PBX and others that might be affected by NGINX?

While Calibre-Web might be internally restricted to 200M for now?  As mentioned here:
https://github.com/janeczku/calibre-web/wiki/FAQ#what-is-the-maximum-file-size-which-can-be-uploaded-through-calibre-web

holta:
CLARIFICATION: One should **ignore** the claim at https://www.php.net/manual/en/ini.core.php#ini.post-max-size that "Generally speaking, memory_limit [512M when nginx_high_php_limits: True] should be larger than post_max_size [10000M when nginx_high_php_limits: True]"

REASON: Several stackoverflow.com articles **confirm** that uploaded files are cached in /tmp not memory (during upload).

jvonau:
Think using 10G is a bit much, good way to run the machine out of storage space when the filesystem is near maximum capacity causing a DOS situation as noted in #2584

jvonau:
Might be better to expose 'client_max_body_size' directly as a end user editable variable via local_vars.yml?

holta:
@EMG70 @deldesir it's time to move forward with this PR or similar.  As too many grassroots/indigenous communities are now getting blocked from uploading their very own multi-gigabyte phone videos to Calibre-Web, Nextcloud, Moodle, WordPress, etc.

Of course there will be glitches en route, e.g. when local educators misunderstand that they really do need to compress their own phone videos prior to sharing them.

But let's not make their lives impossible.  So let's fine tune as we go forward — figuring it out en route to help as many educators as we can — and adjusting as needed to help them across the thousands of communities who want this:

1) Local educators having authentic ownership refining their community's curricular content (or their own family's personal videos) is the goal — it's high time that we stop blocking local voices / local contributors by default.
2) IIAB's main page will eventually in fact need a colorful "Disk Nearly Full" warning right on top (something similar to RACHEL's perhaps?) making "Vital Signs" visible to everyone without a password — e.g. things like {Temperature, WiFi status, Disk Almost Full} would be really great: #3309

Related:

- iiab/calibre-web#61

deldesir (approved):
Looks good to me.

holta:
> Might be better to expose 'client_max_body_size' directly as a end user editable variable via local_vars.yml?

@jvonau's above suggestion is a really good one as we later fine-tune.

But first let's set the default to something far higher (personally I advocate for 10000M to see how this works out in coming months, in keeping with modern phone videos and educational documentaries that schools/educators need in all countries, ETC!)

deldesir (approved):

holta:
Thanks @deldesir:

Let's go with this for starters, gauge impact, and adjust en route where that proves necessary.  As @EMG70 & others do further testing.  FWIW this has been repeatedly tested on Raspberry Pi OS and Ubuntu 24.04 pre-releases over the past 2 years — now's the time to start welcoming more communities to battle-test this with their own grassroots/indigenous use cases — helping us/all to evolve!

tim-moody:
while 500M is a little low, I find 10G pretty big. I would have gone for 1 or 2 G.

PR #3151, opened 2022-03-30
9-local-addons: Do not install php-pear

holta:
See accompanying:

- PR iiab/iiab-admin-console#473

### DIFF
--- a/roles/9-local-addons/tasks/main.yml
+++ b/roles/9-local-addons/tasks/main.yml
@@ -37,12 +37,6 @@
     name: pbx
   when: pbx_install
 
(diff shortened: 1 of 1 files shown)


jvonau:
Are going to clean up/note/edit/remove the now unused admin_console_install from the local_vars files?

holta:
> Are going to clean up/note/edit/remove the now unused admin_console_install from the local_vars files?

Variables `admin_console_install` and `admin_console_enable` have other uses, so those files should be ok for now I think (I took a quick look anyway!)

Just FWIW the 2 vars above were not implemented for apt package `php-pear` &mdash; but it's true they were very useful in this situation &mdash; thanks to your (@jvonau) excellent suggestion here about a year ago.

PR #3152, opened 2022-03-30
awstats/tasks/install.yml: 'command' safer than 'shell'

holta:
Code cleanup; no functional changes.


holta:
Tested on 64-bit RasPiOS Lite on RPi 4.

PR #3156, opened 2022-04-01
Always install the latest remote.it Device Package (instead of hard-coding a new version# every 1-2 months!)

holta:
This PR also:

- Fixes a longstanding bug with the CPU arch specified incorrectly for remote.it's CLI package &mdash; whose filename in fact requires `x86_64` instead of `amd64`.
- Better documents what's happening in {remoteit/defaults/main.yml, remoteit/tasks/install.yml}

Tested on 64-bit RasPiOS Lite on RPi 4.  Testing is ongoing on an Ubuntu Desktop 22.04 VM.

Building on:

- #3006
- PR #3116
- PR #3128
- PR #3154

### DIFF
--- a/roles/remoteit/defaults/main.yml
+++ b/roles/remoteit/defaults/main.yml
@@ -6,15 +6,14 @@
 # All above are set in: github.com/iiab/iiab/blob/master/vars/default_vars.yml
 # If nec, change them by editing /etc/iiab/local_vars.yml prior to installing!
 
-# This should (generally) be the latest from https://remote.it/download/ -- as
-# of Feb 2022 there are 4 such "Device Package" .deb files that may be useful:
-# - Debian Linux (ARM64)
-# - Debian Linux (x86_64)
-# - Pi OS (ARM)
-# - Pi OS (ARM64)
-remoteit_version: 4.14.1
 
-# See https://docs.remote.it/device-package/installation to refine URL below:
+# 2022-03-31: https://remote.it/download/ offers 4 relevant "Device Packages"
+# 1) Raspberry Pi (ARM)    = armhf.rpi
+# 2) Raspberry Pi (ARM64)  = arm64.rpi
+# 3) Debian Linux (ARM64)  = arm64
+# 4) Debian Linux (x86_64) = amd64
+
+# See https://docs.remote.it/software/device-package/installation to refine URL below:
 device_suffixes:
   armv6: armhf.rpi
   armv6l: armhf.rpi
@@ -24,11 +23,22 @@ device_suffixes:
   aarch64: arm64
   x86_64: amd64
 remoteit_device_suffix: "{{ device_suffixes[ansible_architecture] | default('unknown') }}"
-remoteit_deb: remoteit-{{ remoteit_version }}.{{ remoteit_device_suffix }}.deb
-remoteit_device_url: https://downloads.remote.it/remoteit/v{{ remoteit_version }}/{{ remoteit_deb }}
-# Example...         https://downloads.remote.it/remoteit/v4.13.5/remoteit-4.13.5.armhf.rpi.deb
+remoteit_device_url: https://downloads.remote.it/remoteit/latest/remoteit.{{ remoteit_device_suffix }}.deb
 
-# See https://docs.remote.it/cli/overview to refine URL below:
+# 2022-03-31: Use "latest" above, instead of ever-changing version below
+# remoteit_version: 4.14.1
+# remoteit_deb: remoteit-{{ remoteit_version }}.{{ remoteit_device_suffix }}.deb
+# remoteit_device_url: https://downloads.remote.it/remoteit/v{{ remoteit_version }}/{{ remoteit_deb }}
+# # Example...         https://downloads.remote.it/remoteit/v4.14.1/remoteit-4.14.1.armhf.rpi.deb
+
+
+# 2022-03-31: https://remote.it/download/ offers 4 relevant "CLI" installs:
+# 1) Debian Linux (ARM v6) OR Raspberry Pi (ARM)   = armv6
+# 2) Debian Linux (ARM v7)                         = armv7
+# 3) Debian Linux (ARM64)  OR Raspberry Pi (ARM64) = arm64
+# 4) Debian Linux (x86_64)                         = x86_64
+
+# See https://docs.remote.it/software/cli/overview to refine URL below:
 cli_suffixes:
   armv6: armv6
   armv6l: armv6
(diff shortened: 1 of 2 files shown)


holta:
> Testing is ongoing on an Ubuntu Desktop 22.04 VM.

Functionality test confirmed.

In short, this PR also fixes amd64 -> x86_64 in the filename to be downloaded, so a non-bogus CLI (https://downloads.remote.it/cli/latest/remoteit_linux_x86_64 is the correct URL) is installed at `/usr/bin/remoteit`

jvonau:
A 'reinstall' will be needed to activate should anybody be tardy in registering the machine and is only good for a week. Given 'The claim code must be used within 24 hours' is there any point to always installing the apt package? Think the remoteit installation needs the same treatment as openvpn's iiab-support, be an optional on-demand operation. 

Side question, when did iiab-support move to be a template under the openvpn role? Seems a little less than thought all the way through... The intention was to be able to install openvpn without openvpn already being [installed](https://github.com/iiab/iiab/blob/master/roles/openvpn/templates/iiab-support#L114-L122), when did the script move out of /opt/iiab/iiab/ where it was executable as a standalone script to only being in /usr/bin/ when the role is installed?

holta:
> A 'reinstall' will be needed to activate should anybody be tardy in registering the machine and is only good for a week.

Do you mean 24h?

(Or is there also some kind of week-long deadline, that I'm not aware of it?)

> Given 'The claim code must be used within 24 hours' is there any point to always installing the apt package? Think the remoteit installation needs the same treatment as openvpn's iiab-support, be an optional on-demand operation.

It's an idea.  I prefer that remote.it be installed by default so that newcomers can explore running `/usr/bin/remoteit help` immediately right out-of-the-box &mdash; and are very actively encouraged (in all sorts of ways) to start taking responsibility for their own remote support needs.

(FWIW remote.it CLI offers a few more features than their Device Package, and can help some folks build an understanding of their options.)

> Side question, when did iiab-support move to be a template under the openvpn role? Seems a little less than thought all the way through... The intention was to be able to install openvpn without openvpn already being [installed](https://github.com/iiab/iiab/blob/master/roles/openvpn/templates/iiab-support#L114-L122), when did the script move out of /opt/iiab/iiab/ where it was executable as a standalone script to only being in /usr/bin/?

No solution will please everybody.

Sometimes there's a need to de-clutter and simplify.

So I spent a week working on this in August.  The new approach also works, but yes (as you say, and intentionally so) in a slightly different way:

- PR #2957

jvonau:
On a side note I sure hope remoteit was not installed on any of the testing images released. I do not see any cleanup code in iiab-factory to deal with remoteit...

holta:
> On a side note I sure hope remoteit was not installed on any of the testing images released. I do not see any cleanup code in iiab-factory to deal with remoteit...

It's a good question.

Possibly the claim code in `/etc/remoteit/config.json` should be auto-deleted, or some such similar approach for IIAB images that are published or circulated.

jvonau:
I'm not comfortable with playing fast and loose with what really needs a paid account to work over the long term. Given this project can be deemed a 'reseller' by redistributing the remoiteit code, think that is a bit of a grey area at best. I prefer to stay clear of any forced use, and yes I turn the role off for that reason.

Pristine image creation should really omit the role given a 'reinstall' is most likely needed to create the install's unique identifier when the image is written to the SDcard and would insure no leftovers are present to clean up. `iiab-install-remoteit` could just be a wrapper to 'runrole remoteit' kind of like how iiab-support is setup.

holta:
> needs a paid account to work over the long term

This is not true, among personal and non-commercial users of remote.it

We're going to have to agree to disagree here.

holta:
A big thank you to @jvonau who surfaced the core security/privacy issue with `/etc/remoteit/config.json` &mdash; which definitely does not belonging in published IIAB images &mdash; and is now addressed by:

- PR iiab/iiab-factory#209

jvonau:
> > needs a paid account to work over the long term
> 
> This is not true, among personal and non-commercial users of remote.it
> 
> We're going to have to agree to disagree here.

Yes we are, for other's benefit have a read of https://remote.it/legal/fair-use-policy/

holta:
FWIW paid and unpaid plans can both be shut down if usage is excessive &mdash; excerpt from https://remote.it/legal/fair-use-policy/ dated November 7, 2020:

> You are unlikely to be affected by this Fair Use Policy if:
>
> - Your Free Subscription Plan does not exceed 10 Devices and/or 250 daily Connections and there are no connected Devices for Business use.
> - Your Paid Subscription Plan has the correct number of Devices or seats purchased in relation to the number of actual Devices in your account that are for Business use. On average, that means about 500 Devices per seat.
> - Your Paid Subscription Plan does not have excessive daily connections to your Devices.

jvonau:
> A Paid Subscription Plan is the standard licensing option for organizations and Business entities.  A Paid Subscription Plan is purchased by the organization or Business and can be used by a single person referred to as users, within the Business. Users in the Business must have their own unique account on the Service.

>This includes but is not limited to LLC’s, Inc.’s, partnerships, DBA’s, schools, non-profits, clubs, groups, teams, religious organizations, and other organized groups or legal entities.

>A Free Subscription Plan is for private individuals who are granted a license when using a Free Plan, and is solely for their own use. A Free Subscription Plan may not be purchased, refunded or in any way financed by a Business. A Free Subscription Plan may only be used by a Business if the number of Devices are less than 10 and the Business is evaluating remote.it prior to Commercial use. A Business or organization may use a Free Subscription Plan to test, evaluate and for Proof Of Concepts that lead to a commercial deployment.

jvonau:
https://remote.it/legal/legal-definitions-and-terms/ dated October 8, 2019
>“Personal” means an individual person that is for non-commercial, and is not a corporation, business, school, team, company, LLC,  partnership, government, organization or a nonprofit.

>“Business” means commercial use, such as by corporations, LLC’s, organizations, schools, teams, partnerships, charities, non-profits, governments, religious establishments or any other entity that is not Personal.

>“Commercial” means you are earning revenue while using remote.it directly or indirectly, either for profit or not for profit.

As I mentioned above, I'm not OK with playing fast and loose with upstream's policies, this role should be opt-in and not expose all potential uninformed users of IIAB to any potential legal blowback, just don't force the installation by default. The instructions already note how to enable the role and install after the main install is complete, why have the phone-home running for everybody, some that may not even know remoteit's purpose and requires further configuration to be useful.

holta:
I humbly disagree.

This legalese is certainly entertaining, but at the same time it's a completely different ballgame from Haitian families trying to help their very own village at a distance using IIAB, and 100% legal.

If remote.it was so idiotic as to sue people like this (they are not!) this would go nowhere fast.

Issue #3159, opened 2022-04-02
New resizing logic requires an extra reboot


tim-moody:
write image from latest master prepared with min and cp sd to sd card and boot up the image

in the old version card resizing would begin and could be seen in progress in ssh

now reboot after first boot is required to start resizing.

holta:
Apparently `raspi-config --expand-rootfs` would seem to now require an additional reboot.

[Line 17 of 1-prep/templates/iiab-expand-rootfs](https://github.com/iiab/iiab/blob/212d1fa765887ba3830099dd6bb8f7ffc289f5b1/roles/1-prep/templates/iiab-expand-rootfs#L17) tries to delegate the details to RasPiOS so they can manage all the evolving details year-by-year.

But if a reboot is too annoying, we have an alternative: [Lines 20-34](https://github.com/iiab/iiab/blob/212d1fa765887ba3830099dd6bb8f7ffc289f5b1/roles/1-prep/templates/iiab-expand-rootfs#L20-L34) suddenly works far better (as a result of PR #3137 2.5 weeks ago) on most Debuntu-style OS's.  So 2 months later we can consider reverting to this older approach (now refined to deal with external boot disks, etc...)

Refs:

- PR #3121
- #3136
- PR #3137 
- #3147

holta:
Resolved by:

- PR #3160

jvonau:
Guess someone missed the part of raspi-config that uses an older fdisk editing [method](https://github.com/RPi-Distro/raspi-config/blob/master/raspi-config#L160-L196) where the partition needs to be unmounted while editing when you liberated some of the underling functions when working up 2131.  A reboot is not that bad and is what RaspiOS does when booting a 'fresh' image out of the box, just a change in behavior when using the builtin tools. Using growpart is way more flexible, and is the standard on Ubuntu based arm64 images. Now the question becomes should growpart be installed on all installations to take advantage of the revised root device detection that should work on all /dev/sd* root filesystems? That would be a bonus if cloning an intel based machine could be achieved with the tools within iiab-factory.

holta:
> should growpart be installed on all installations

Yes.  I agree.  Clarified by 4e2e4c8.

(With respect to `raspi-config --expand-rootfs` whose open source code has allowed all this progress over the past 2 months!)

Issue #3147, opened 2022-03-24
small imager image does not enlarge file system into expanded partition [unable to reproduce]


georgejhunt:
Experienced on https://d.iiab.io/8.0/iiab-8.0beta1-220203-SMALL-raspios-11-lite-g6598f3e1.img.zip

fixed by logging on as root and doing "resize2fs /dev/mmcblk0p2"

holta:
@georgejhunt I did a complete test and was unable to reproduce the problem.

For me, the partition /dev/mmcblk0p2 expands to its fullsize about 30 seconds after I logged in (which was about 60 seconds after boot).

Question: Roughly how long did you wait after booting?

As an aside, the [code that expands IIAB's rootfs](https://github.com/iiab/iiab/blob/master/roles/1-prep/templates/iiab-expand-rootfs) has since been completely replaced:

- PR #3121 (Feb 17/18)
- #3136 (Mch 14)
- PR #3137 (Mch 15)

georgejhunt:
I will repeat and wait for 5 min.  I  did not realize that I needed to wait. And there was what I interpreted as a stall during the initial boot.

georgejhunt:
I failed to replicate when I waited.  It was resized in 2 or 3 minutes. But was really sluggish during the resize process.

PR #3160, opened 2022-04-02
1-prep/templates/iiab-expand-rootfs: Avoid reboot

holta:
This will be more work to maintain year-by-year regrettably, but at least it resolves this annoyance:

- #3159 

Building on:

- PR #3121
- #3136
- PR #3137 
- PR #3147


holta:
@georgejhunt apologies the reboot is necessary for now, to avoid what appears to be a race condition with fsck:

- #3325

So this PR #3160 has been reverted below &mdash; allowing for reliable operation on Raspberry Pi OS:

- PR #3336

PR #3161, opened 2022-04-03
New command /usr/bin/iiab-remoteit to quickly obtain a new Claim Code for https://remote.it (allows remote connections to manage an IIAB)

holta:
Thanks to @jvonau's ideas at #3158 and building on:

- #3006
- PR #3154
- PR #3156
- PR #3157

This PR also overhauls and dramatically simplifies the installation of remote.it's Device Package, using their official 1-line installer ("curl -L https://downloads.remote.it/remoteit/install_agent.sh") as recommended by their 2022-01-24 document ["Streamlined installation for Linux and Raspberry Pi platforms"](https://support.remote.it/hc/en-us/articles/4422773654669-Streamlined-installation-for-Linux-and-Raspberry-Pi-platforms).

Note that `apt purge remoteit*` is regrettably necessary at this time [prior to running remote.it's `install_agent.sh`], <strike>in order to get a new Claim Code in many corner cases, and</strike> for several reasons.  One of those reasons is that the Device Package .deb files (same thing you get at https://remote.it/download/) are oddly each built in slightly different ways.  Without going into all the ugly details, the situation and its resolution are summarized here:

https://github.com/holta/iiab/blob/5854ab7c619b3c01446951f925c5f27eb37ed077/roles/remoteit/templates/iiab-remoteit#L32-L34

FYI this PR intentionally avoids Ansible, so that the `/usr/bin/iiab-remoteit` command runs as quickly and resiliently as possible, quite similar to `/usr/bin/iiab-support` with OpenVPN.

This PR is tested on the latest pre-release of Ubuntu Desktop 22.04, with more testing ongoing on other OS's too, including 32-bit and 64-bit RasPiOS on Raspberry Pi 4.

### DIFF
--- a/roles/remoteit/templates/iiab-remoteit
+++ b/roles/remoteit/templates/iiab-remoteit
@@ -1,0 +1,51 @@
+#!/bin/bash -e
+
+# Run 'sudo iiab-remoteit' to (re)install & enable remote.it -- GENERAL TIPS:
+# http://FAQ.IIAB.IO -> "How can I remotely manage my Internet-in-a-Box?"
+
+# /usr/bin/remoteit CLI is already be installed by:
+# https://github.com/iiab/iiab/blob/master/roles/remoteit/tasks/install.yml
+
+echo -e "\nhttps://remote.it can help you remotely manage this IIAB:"
+echo -e "https://github.com/iiab/iiab/blob/master/roles/remoteit/README.md\n"
+
+echo -en "\e[1mInstall remote.it Device Package after purging all prior versions? [Y/n]\e[0m "
+read ans < /dev/tty    # Strips outer whitespace, whether we like it or not!
+echo
+[ "$ans" = "n" ] || [ "$ans" = "N" ] && exit 1
+
+if grep -q '^remoteit_install:' /etc/iiab/local_vars.yml; then
+    sed -i "s/^remoteit_install:.*/remoteit_install: True/" /etc/iiab/local_vars.yml
+else
+    echo "remoteit_install: True" >> /etc/iiab/local_vars.yml
+fi
+
+if grep -q '^remoteit_enabled:' /etc/iiab/local_vars.yml; then
+    sed -i "s/^remoteit_enabled:.*/remoteit_enabled: True/" /etc/iiab/local_vars.yml
+else
+    echo "remoteit_enabled: True" >> /etc/iiab/local_vars.yml
+fi
+
+# 2022-04-02: Full Path Avoids problematic /usr/local/bin/apt on Linux Mint
+/usr/bin/apt -y purge remoteit*
+
+# Why the brutal purge?  Even 'apt -y reinstall remoteit.*.deb' is much stronger
+# than 'install -y' in install_agent.sh below, but still insufficient.  Maybe in
+# future years their /usr/bin/remoteit CLI might seed a new claim code when nec?
+
+# apt install & enable "latest" remote.it Device Package for your CPU/OS
+curl -L https://downloads.remote.it/remoteit/install_agent.sh | sh
+
+if grep -q '^remoteit_installed:' /etc/iiab/iiab_state.yml; then
+    sed -i "s/^remoteit_installed:.*/remoteit_installed: True/" /etc/iiab/iiab_state.yml
+else
+    echo "remoteit_installed: True" >> /etc/iiab/iiab_state.yml
+fi
+
+echo -e "\e[44;1mNEXT STEPS...\e[0m\n"
+
+echo -e "\e[1m1) Install the remote.it Desktop Application on your own laptop/computer:"
+echo -e "   https://remote.it/download/\n"
+
+echo -e "2) Use the above 8-character claim code within 24h as shown here:"
+echo -e "   https://docs.remote.it/software/device-package/installation#3.-claim-and-register-the-device\e[0m\n"
--- a/roles/remoteit/templates/iiab-remoteit
+++ b/roles/remoteit/templates/iiab-remoteit
@@ -36,19 +36,19 @@ else
     # '|| true' overrides 'bash -e' so script continues if config.json missing
     mv /etc/remoteit/config.json /etc/remoteit/config.json.$(date +%F_%T_%Z) || true
 
-    echo -e "In just a few seconds, all 3 services should be enabled + started.\n"
+    echo -e "In just a few seconds, all 3 services should be enabled/started.\n"
 
     systemctl restart connectd    # Claim Code logic + kickstarts 2 svcs below
-    systemctl enable connectd     # 3 enable lines, like enable-or-disable.yml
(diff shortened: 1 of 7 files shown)


roles/remoteit/templates/iiab-remoteit
```diff
@@ -0,0 +1,51 @@
...
+# Why the brutal purge?  Even 'apt -y reinstall remoteit.*.deb' is much stronger
+# than 'install -y' in install_agent.sh below, but still insufficient.  Maybe in
+# future years their /usr/bin/remoteit CLI might seed a new claim code when nec?
+
+# apt install & enable "latest" remote.it Device Package for your CPU/OS
+curl -L https://downloads.remote.it/remoteit/install_agent.sh | sh
+
+if grep -q '^remoteit_installed:' /etc/iiab/iiab_state.yml; then
+    sed -i "s/^remoteit_installed:.*/remoteit_installed: True/" /etc/iiab/iiab_state.yml
+else
```
jvonau, line 57: _installed: True are the only entries found in iiab_state, `if !` and just the echo if missing seems more appropriate, does away with the unneeded sed alteration shorting the code

holta: Makes sense.

(Unless `/etc/iiab/iiab_state.yml` hypothetically changes in future, to store other kinds of state...)

jvonau: Other new states sure I can see that, I doubt you are refactoring every role to drop is defined for any reason other than code churn.

jvonau:
The forced reinstall is a pretty big hammer and might be able to go away. I've discovered that removing /etc/remoteit/config.json and then restarting connectd results in a new config.json with a different 'id' & 'claim' being generated, looks like 'name' is generated from the hostname of the machine.

>FYI this PR intentionally avoids Ansible, so that the /usr/bin/iiab-remoteit command runs as quickly and resiliently as possible, quite similar to /usr/bin/iiab-support with OpenVPN.

Just to clarify, the current logic in iiab-support that skips using ansible relies on ansible being run at least once to install the role otherwise ansible will be used used to install openvpn to place all the template files in to their correct locations.

holta:
Aside; It appears remote.it is hardly bug-free on Raspberry Pi.

If so this is unfortunate and a bit surprising, as the company has a very long history working with Raspberry Pis.

Still, let's try to get more feedback from more implementers to see what many different people's experiences are in coming weeks and months.

jvonau:
On ubuntu-22.04 running `curl -L https://downloads.remote.it/remoteit/install_agent.sh | sudo sh` by hand results in the restart services dialog box coming to the forefront and not being able to tab over to 'ok', can't see how ansible will be able to proceed given this stall   
Screenshot from 2022-04-03 09-30-40

holta:
> On ubuntu-22.04 running `curl -L https://downloads.remote.it/remoteit/install_agent.sh | sudo sh` by hand results in the restart services dialog box coming to the forefront and not being able to tab over to 'ok', can't see how ansible will be able to proceed

It's a good question as to why this in fact works without stalling from scripts and from Ansible.

I'd love to know the answer &mdash; as this happens in other situations too &mdash; specifically: such scripts do consistently work even on graphical desktop OS's, e.g. when invoked from a script or from Ansible, instead of from a graphical command-line.

(Possibly this has something to do with the particular environment variables or similar, within a graphical [I meant interactive] terminal, [versus non-interactive scripts] ?)

holta:
> I've discovered that removing /etc/remoteit/config.json and then restarting connectd results in a new config.json with a different 'id' & 'claim' being generated, looks like 'name' is generated from the hostname of the machine.

This is very promising.  I've seen 1 instance where this did not work, but it would seem to work a vast majority of the time.

jvonau:
> > On ubuntu-22.04 running `curl -L https://downloads.remote.it/remoteit/install_agent.sh | sudo sh` by hand results in the restart services dialog box coming to the forefront and not being able to tab over to 'ok', can't see how ansible will be able to proceed
> 
> It's a good question as to why this in fact works without stalling from scripts and from Ansible.
> 
> I'd love to know the answer — as this happens in other situations too — specifically: such scripts do consistently work even on graphical desktop OS's, e.g. when invoked from a script or from Ansible, instead of from a graphical command-line.

This was from a ssh session with no desktop installed.

> 
> (Possibly this has something to do with the particular environment variables or similar, within a graphical terminal ?)

no graphics just a cmdline over ssh

holta:
> no graphics just a cmdline over ssh

Indeed.  I should not have said a literal graphical desktop (in the style of X Windows, etc).  Certainly it works from scripts/Ansible.

But I don't know the exact dynamics.  I (suspect?) it has something to do with shell environment variables — that somehow establish whether it's an interactive session or not.

Likewise: The `needrestart` command (that recent Ubuntu-like OS's use when you apt update) shows the same behavior — it prompts you colorfully if you're running from an interactive terminal / shell — but somehow is wise enough _not to get stuck_ when the same command runs from a script.

jvonau:
> > no graphics just a cmdline over ssh
> 
> Indeed. I should not have said a literal graphical desktop (in the style of X Windows, etc). Certainly it works from scripts/Ansible.
> 
> But I don't know the exact dynamics. I (suspect?) it has something to do with shell environment variables — that somehow establish whether it's an interactive session or not.
> 
> Likewise: The `needrestart` command (that recent Ubuntu-like OS's use when you apt update) shows the same behavior — it prompts you colorfully if you're running from an interactive terminal / shell — but somehow is wise enough _not to get stuck_ when the same command runs from a script.

You still have to tab over to OK and then enter to proceed when using /usr/bin/iiab-remoteit, same as doing a `wget of install_agent.sh` then `sudo bash install_agent.sh`(yea I could of chmod it) only using curl directly doesn't allow one to tab over to OK. Must have an interactive tty flag somewhere that ansible and/or needrestart makes use of.

jvonau:
https://github.com/iiab/iiab/pull/3011#issuecomment-954441316 additional inspiration for this PR. 
> > I've discovered that removing /etc/remoteit/config.json and then restarting connectd results in a new config.json with a different 'id' & 'claim' being generated, looks like 'name' is generated from the hostname of the machine.
> 
> This is very promising. I've seen 1 instance where this did not work, but it would seem to work a vast majority of the time.

Think the only time this won't work if there is a /etc/remoteit/registration file present.

holta:
Heavily tested on Ubuntu 22.04 pre-release (VM).

Time to push this and seek out broader community feedback as more people bang on this and use it in many different ways.

Many thanks to @jvonau whose insight into deleting `/etc/remoteit/config.json` and then restarting systemd service `connectd` was instrumental to the new command `/usr/bin/iiab-remoteit` quickly obtaining a new claim code.

(Hopefully this undocumented feature[*] will remain working as part of remote.it Device Packages for years to come, if we're lucky!)

[*] Hinted at within https://support.remote.it/hc/en-us/articles/360061228252-Oops-I-cloned-an-SD-card-

PR #3169, opened 2022-04-06
remoteit - visual feedback and recording claim code in iiab.ini & New feature auto registeration for early connections

jvonau:
Record initial claim code, updates to blank when registered if ansible is run again.. Becomes viewable via iiab.ini thus available within iiab-diagnostics.

### DIFF
--- a/roles/remoteit/defaults/main.yml
+++ b/roles/remoteit/defaults/main.yml
@@ -49,3 +49,4 @@ cli_suffixes:
   x86_64: x86_64
 remoteit_cli_suffix: "{{ cli_suffixes[ansible_architecture] | default('unknown') }}"
 remoteit_cli_url: https://downloads.remote.it/cli/latest/remoteit_linux_{{ remoteit_cli_suffix }}
+remoteit_REGISTRATION_CODE: none #eg "592AA9BB-68C8-520A-AACA-6E27654C3DF6" generated in the desktop or web portal
--- a/roles/remoteit/tasks/enable-or-disable.yml
+++ b/roles/remoteit/tasks/enable-or-disable.yml
@@ -1,3 +1,7 @@
+- name: Using pre-created REGISTRATION_CODE from desktop or web portal
+  shell: echo {{ remoteit_REGISTRATION_CODE }} > /etc/remoteit/registration
+  when: remoteit_REGISTRATION_CODE is not none
+
 - name: Enable & Restart remote.it "parent" service connectd, which exits after spawning 2 "child" services/daemons below
   systemd:
     name: connectd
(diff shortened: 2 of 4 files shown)


holta:
It feels too risky to put claim codes in iiab-diagnostics, as these are generally publicly published (i.e. to pastebins and GitHub tickets).

holta:
Just FYI there are quite a few situations where `/etc/remoteit/config.json` does not contain a claim code.

(e.g. if `/usr/bin/remoteit` was run in almost any way, which is rather mysterious and unpleasant, but confirmed many times.)

jvonau:
If it is not apparent the claim code is removed once registered, as an alternate to the apt claim code approach /usr/bin/remoteit is designed to do what the desktop can do once your account is setup, login, register a device (without using a claim code), define a local service, start connections to remote services but I can see slip ups using https://docs.remote.it/developer-tools/cli-usage. If the user is not going to use the claim code within the allotted 24 hours why is the claim code being generated in advance of the user requiring/requesting it?

holta:
> If the user is not going to use the claim code within the allotted 24 hours why is the claim code being generated

Many will use the claim code within 24h of installing IIAB, and their time is valuable too (no implementer wants their time wasted).

And those who suddenly need a claim code after 24h also greatly benefit, as they can run `sudo iiab-remoteit` to generate a new claim code literally within mere seconds, instead of waiting around for software to ploddingly install.

jvonau:
> > If the user is not going to use the claim code within the allotted 24 hours why is the claim code being generated
> 
> Many will use the claim code within 24h of installing IIAB, and their time is valuable too (no implementer wants their time wasted).
>

This is a question of the needs of the many vis the needs of the few. How many? What stats do have to back up that claim? How many of the users know there is an application that phones home with the claim code to use at the webportal just by installing it? You force the users to know in advance on how to turn this install off making your beta testing opt-out. 

> And those who suddenly need a claim code after 24h also greatly benefit, as they can run `sudo iiab-remoteit` to generate a new claim code literally within mere seconds, instead of waiting around for software to ploddingly install.

Get off your high horse if I didn't poke you would still be doing a blanket apt purge.

jvonau:
[296e8c2](https://github.com/iiab/iiab/pull/3169/commits/296e8c20326e16708cddea4359822ec2529c7b44) introduces remoteit_REGISTRATION_CODE: when populated with a valid key, with 'remoteit_enabled: True', in local_vars.yml will auto register and allow ssh connections right after the play ends. This is kind of like how openvpn is available during the install, just extending that feature to remoteit.

jvonau:
I created the key within the desktop +Add device -> Linux & Raspberry Pi -> click on the box to 'copy' -> paste into a text editor -> save file. The key is R3_REGISTRATION_CODE value without the " "
Screenshot from 2022-04-06 09-45-04

Issue #2542, opened 2020-09-22
Design flaw in mysql database password changing


jvonau:
Required reading:
https://docs.ansible.com/ansible/latest/modules/mysql_db_module.html
https://docs.ansible.com/ansible/latest/modules/mysql_user_module.html

The setup.yml in a role is where the passwords should be changed but setup.yml are now being called from within install.yml but were not prior to: 
1. f5180d99a2ebefff64e30e3976cbe360fcc0f372#diff-a45913a4ffde64b946e0db1c03591a25 for nextcloud Jan 29, 2020
2. 027832889be31438306dbf6353f85a31e4699a33#diff-1460b7c549aa9ce1a2a72a38aed3cd65 for elgg  Feb 4. 2020 
3. mediawiki lacks setup.yml 
4. wordpress has setup.yml in its main.yml 

I believe these efforts above were pushed at that time to purposely suppress or attempt to sabotage my #1937 pi-gen branch work opened on Aug 25, 2019 then continuity updated with #2254 -> #2381 -> #2503 to stay current with master.

There is no easy way to update the passwords post-install without "reinstalling" the role while wordpress still has setup.yml in main.yml. Now #2503 attempts to bring the same standardization of operations that the *_installed effort has obtaining by:
1. - Using setup.yml in the same way across the 4 popular apps above.
2. - Ensure mysql is running and records the current running state of mysql.
3. - Introduce *_provisioned for use with the setup.yml routines, like install.yml currently has.
4. - Introduce the optional ability to suppress the running of setup.yml.

Now the stage is set to be able to more easily change the mysql passwords in the future, and allows any image builder an easy way to accommodate user changing passwords in local_vars on prefabricated images build with 'provision_active True' 
at a later point in time.

holta:
If there is a serious bug with our treatment of MySQL / MariaDB passwords, there aren't many but there still are a few days left before IIAB 7.2 is released, if folks can confirm that some kind of emergency fix is needed.

Ref: PR #2488

jvonau:
As a matter of fact as of today it is not possible to change the default mysql passwords without editing vars/default_vars.yml, but a crafty implementer could populate local_vars before the initial install. Post install, changing in the field prior to deployment, is not possible.

jvonau:
http://wiki.laptop.org/go/IIAB/FAQ#What_are_the_default_passwords.3F
[...]

PR #3172, opened 2022-04-06
Allow remote.it license key to be pasted into local_vars.yml & document both registration options

holta:
Building on @jvonau's PR #3169, using Ansible rather than bash approaches where possible.

FWIW I'm uncomfortable with the claim code being disseminating as part of `/etc/iiab/iiab.ini` or similar files, so that part is commented out.

Earlier refs:

- #3006
- PR #3156
- PR #3161
- PR #3167

### DIFF
--- a/roles/remoteit/README.md
+++ b/roles/remoteit/README.md
@@ -8,24 +8,29 @@ For other approaches, please see http://FAQ.IIAB.IO -> "How can I remotely manag
 
 ## Getting Started
 
-### Create a remote.it account + install its desktop application
+### Create a remote.it account + consider its desktop application
 
 1. Browse to [https://remote.it](https://remote.it) (Web Portal) and sign up for an account.
 
-2. Download and install the remote.it [desktop application](https://remote.it/download/) (e.g. for Windows, macOS or Linux) on your own laptop/computer.  Their https://remote.it Web Portal and [mobile apps](https://docs.remote.it/introduction/get-started/readme#installation-packages) are also sometimes sufficient, but less functional.
+2. Consider downloading and installing the remote.it [desktop application](https://remote.it/download/) (e.g. for Windows, macOS or Linux) on your own laptop/computer.  Their https://remote.it Web Portal and [mobile apps](https://docs.remote.it/introduction/get-started/readme#installation-packages) are also sometimes sufficient, but less functional.
 
    COMPARISON: "The Desktop and [CLI](https://docs.remote.it/software/cli) can [each] support both peer to peer connections and proxy connections [whereas] the Web Portal and API can only support proxy connections" according to https://docs.remote.it/software/device-package/usage
 
 <!-- ### Install remote.it onto an IIAB + register it + authorize services/ports -->
 ### OPTION #1: Generate a remote.it claim code for your IIAB + register it + authorize services/ports
+<!-- ### OPTION #1: Run `sudo iiab-remoteit` to enable remote.it -->
 
-Prerequisite: Find any IIAB with `remoteit_installed: True` in `/etc/iiab/iiab_state.yml` (this is the default!)  It means that the remote.it [Device Package](https://docs.remote.it/software/device-package) is installed on your IIAB, most important.  It also means the _strictly optional_ `/usr/bin/remoteit` [command-line interface (CLI)](https://docs.remote.it/software/cli) is installed, offering [a few more features](https://support.remote.it/hc/en-us/articles/4412786750861-Install-the-remoteit-agent-on-your-device) than the Device Package.
+Prerequisite: Find any IIAB with `remoteit_installed: True` in `/etc/iiab/iiab_state.yml` (this is the default!)  This means that the remote.it [Device Package](https://docs.remote.it/software/device-package) is installed on your IIAB, most important.  It also means the _strictly optional_ `/usr/bin/remoteit` [command-line interface (CLI)](https://docs.remote.it/software/cli) is installed, offering [a few more features](https://support.remote.it/hc/en-us/articles/4412786750861-Install-the-remoteit-agent-on-your-device) than the Device Package.
 
 1. Run `sudo iiab-remoteit` to enable remote.it on your IIAB.
 
-   Hit `[Enter]` (repeatedly if necessary, to accept all defaults) if you want to quickly generate a new claim code for your IIAB.
+   Hit `[Enter]` <!-- (repeatedly if necessary, to accept all defaults) --> if this is a fresh install, to quickly generate a claim code for your IIAB.
 
    (The claim code is put in `/etc/remoteit/config.json` and must be used [within 24 hours](https://docs.remote.it/device-package/installation#2.-update-your-package-manager-and-install).)
+
+   NOTE: If a remote.it license key is found in `/etc/iiab/local_vars.yml` or `/etc/remoteit/registration`, that will be tried first (prior to generating a claim code).
+
+   *If the license key works, you will not get a claim code, as the IIAB device auto-registers to your remote.it account (skip to Step 3. below).*
 
 <!--
 1. Connect your IIAB device to the Internet.
@@ -103,14 +108,16 @@ Prerequisite: Find any IIAB with `remoteit_installed: True` in `/etc/iiab/iiab_s
 
 4. If your IIAB software is already installed, make sure your IIAB is online.
 
-   Now register your IIAB to your remote.it account (while enabling remote.it on your IIAB device) by running:
+   Now register the IIAB device to your remote.it account, by running `sudo iiab-remoteit` (this also enables remote.it on your IIAB).
+
+   Or, if you prefer the legacy approach, run:
 
    ```
    cd /opt/iiab/iiab
    sudo ./runrole remoteit
    ```
 
-   If you also want to reinstall the (latest) remote.it Device Package on your IIAB, you can instead run:
+   The legacy approach can also reinstall the (latest) remote.it Device Package on your IIAB, if you instead run:
 
    ```
    cd /opt/iiab/iiab
--- a/roles/remoteit/README.md
+++ b/roles/remoteit/README.md
@@ -101,11 +101,20 @@ Prerequisite: Find any IIAB with `remoteit_installed: True` in `/etc/iiab/iiab_s
 
    When that's complete, skip to Step 5.
 
-4. If your IIAB software is already installed, make sure your IIAB is online, and then run:
+4. If your IIAB software is already installed, make sure your IIAB is online.
+
+   Now register your IIAB to your remote.it account (while enabling remote.it on your IIAB device) by running:
 
    ```
    cd /opt/iiab/iiab
    sudo ./runrole remoteit
+   ```
+
+   If you also want to reinstall the (latest) remote.it Device Package on your IIAB, you can instead run:
+
+   ```
+   cd /opt/iiab/iiab
+   sudo ./runrole --reinstall remoteit
    ```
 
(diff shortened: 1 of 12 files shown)


holta:
Tested with the latest pre-release of Ubuntu 22.04

jvonau:
The presents of a claim code is still useful information, just hide the value behind 'wc' and record 1 or 0 for present or absent. 
https://docs.remote.it/software/device-package/installation#3.-claim-and-register-the-device\ I intentionally use the same wording as what is used in the desktop for the variable, to help people visually make the connection that it's the correct value to use. Think there is a difference in behavior between which key you use, the user's account key or the user generated key like I used. I haven't tried the user key method, the second method configures ssh access without the need to authorize, it's implied because the key is already associated with a user account.

jvonau:
I populated defaults with the default value for the key as none, with the hopes that this method becomes the standard and advertised in local_vars. Now with the use of 'is defined' suggests the 'not invented here' attitude is alive and well and will not be promoted.

>Tested with the latest pre-release of Ubuntu 22.04

Can I be so bold to ask how this was tested? Which key did you use? What else did you have to do to make ssh active?

jvonau:
https://docs.remote.it/oem-and-bulk-provisioning/registration-into-a-users-account Think we should always be setting the registration file to blank, that suppress the creation of the config.json file with a claim code. Now iiab-remoteit might be able to just remove the registration file after checking if it is blank 'in place of'/'in addition to' (haven't checked) removing the config.json to generate the new claim code on demand.

jvonau:
Still need to note that enabled needs to be True with the license key.

holta:
> Still need to note that enabled needs to be True with the license key.

The need for `remoteit_enabled: True` is documented near the top here:

https://github.com/holta/iiab/tree/remoteit_license_key/roles/remoteit#option-2-paste-your-remoteit-license-key-to-iiab--enable-it--authorize-servicesports

holta:
This was also lightly tested with a LARGE-sized fresh install on Ubuntu Server 20.04

(In addition to unit-testing on Ubuntu Desktop 22.04)

jvonau:
> Passwords and license keys in /etc/iiab/local_vars.yml are not a healthy
\# precedent :/  (Going forward let's try to keep credentials in their own apps)

```
grep pass vars/default_vars.yml 
iiab_admin_published_pwd: g0adm1n    # Default password.  For pwd warnings too.
# Obtain a password hash - NEW MORE SECURE WAY:
# Obtain a password hash - OLD WAY:
hostapd_secure: False    # 2021-03-02 WiFi EAPOL fails if hotspot passwords,
hostapd_password: changeme    # espec if WiFi firmware patched below?  #2696
# Set True if client machines should have "passthrough" access to WAN/Internet:
# elgg_mysql_password: $6$iiab51$jeTwnATcbaa92xo0QBTgjLBU.5aVDDrbKeNyyC99R/TAWz6pvfzj.L7lfnOVVjD78nxqT.gkNn6XZmuRV0W3o1
elgg_mysql_password: elgg4kids
transmission_password: changeme
# xovis_db_password: admin
```
```
grep pass roles/*/defaults/main.yml
roles/calibre-web/defaults/main.yml:# calibreweb_admin_password: changeme
roles/iiab-admin/defaults/main.yml:# iiab_admin_published_pwd: g0adm1n   # Default password. For pwd warnings too.
roles/kalite/defaults/main.yml:kalite_admin_password: changeme
roles/kolibri/defaults/main.yml:# provisions Kolibri with facility name, admin acnt / password, preset type,
roles/kolibri/defaults/main.yml:kolibri_admin_password: changeme
roles/lokole/defaults/main.yml:lokole_admin_password: changeme
roles/mediawiki/defaults/main.yml:mediawiki_db_user_password: changeme2020
roles/mediawiki/defaults/main.yml:mediawiki_admin_user_password: changeme2020
roles/mediawiki/defaults/main.yml:# 2020-01-17: MediaWiki 1.34.0 NO LONGER ACCEPTS 'changeme' as a password.
roles/mediawiki/defaults/main.yml:# 2019-09-30: http://box/mediawiki asked for a stronger password on login, per:
roles/moodle/defaults/main.yml:moodle_db_pass: changeme
roles/mosquitto/defaults/main.yml:mosquitto_password: changeme
roles/munin/defaults/main.yml:munin_password: changeme
roles/network/defaults/main.yml:# hostapd_password: changeme
roles/network/defaults/main.yml:py_captive_portal_password: changeme
roles/nextcloud/defaults/main.yml:nextcloud_dbpassword: nextcloudmysql
roles/nextcloud/defaults/main.yml:nextcloud_admin_password: 'changeme'
roles/nodered/defaults/main.yml:nodered_password: changeme    # REMOVE THIS PASSWORD REMINDER LINE AS NECESSARY
roles/nodered/defaults/main.yml:nodered_password_hash: $2b$08$oxgvoU9et3deSbXY8UNVTOWHSTQAyEASIal86RHVMqYQJhpPMNz7q
roles/nodered/defaults/main.yml:# the new password.  Paste the resulting hash as above, but into your own:
roles/nodered/defaults/main.yml:# username and password hash are stored in:
roles/pbx/defaults/main.yml:asterisk_db_password: asterisk
roles/samba/defaults/main.yml:smbpassword : $6$51iiab$VXzRJK88k9k8SXd5sjs37bEbVQ9x4ob1ng7A5PSGyoVXTKQrhu.89BRuXZAgn8a2DPqZkKDcCpqCZVOs.cieT/ 
roles/transmission/defaults/main.yml:# transmission_password: changeme
roles/wordpress/defaults/main.yml:wp_db_user_password: changeme
```
Using local_vars.yml is the normal way of overriding default passwords present within roles. What has been missing over the years was a way to remove the credentials when provided #2542. Now you have a cookie cutter to backfit the other roles.

Now that you are done stop using my personal machine reg_key for testing, the owner of the key will see the machine come online but can't connect to it.
Screenshot from 2022-04-07 23-53-39

holta:
The main risk today is leaking of credentials for online/cloud services (e.g. remote.it, VPN-like, etc) that can have many unintended ramifications, and so can become especially fraught in quasi-public places like [/etc/iiab/local_vars.yml](https://wiki.iiab.io/go/FAQ#What_is_local_vars.yml_and_how_do_I_customize_it?) and `/etc/iiab/iiab.ini` that are often shared and sometimes publicly posted using [iiab-diagnostics](https://github.com/iiab/iiab/blob/master/scripts/iiab-diagnostics.README.md).

Auto-redaction mechanisms can help for sure, but these are inherently brittle as the years go by.

As such drawing a line in the sand (even if that line certainly shifts year-by-year, as new threats necessarily emerge) likely makes sense going forward.  With a preference towards stepping back, and prominently documenting the industry-wide approach for each App, e.g. endorsing the most minimal/mainline 2-line social practice that applies to everyone:

```
sudo mkdir /etc/remoteit
sudo nano /etc/remoteit/registration    # Put your remote.it license key here & save!
```

Thereby keeping the onus (security costs, and associated UX/documentation/design costs) fully with the implementer whenever possible &mdash; as the folks who inherently need to take the dominant responsibility, to understand and learn about their own security, to protect their community.

For sure I'm acquiescing in this case, as indeed this is a very worthwhile experiment, as we and others examine what's possibly coming and what's possibly best over future years.  Because as you say (@jvonau) there is definitely plenty to learn here &mdash; to refine the "trust hand-off" to offline communities.

As more and more "offline" apps gain associated "online" functionality year-by-year (e.g. Kolibri Studio pub/sub exchange of channels/classes/student analytics, and Kolibri sharing with peer machines being just one example among many) which will inherently change security assumptions + paradigms along the way.  No question.

So I'd favor erring on the side of caution, i.e. try to completely avoid intricate security complexities in the first place (where possible!) e.g. the bureaucracy of redacting from subtle/evolving security bits from `/etc/iiab/local_vars.yml` &mdash; and similar security engineering nuance &mdash; which risks burdening everyone with ugly/costly/regular and unnecessary security audits down the road (a huge bureaucratic tax on everybody, that can be completely avoided in many/most cases hopefully!)

In any case: to see what makes most sense in the years to come, we will all need to evolve, necessarily.  "Fully offline" credentials similar to [Admin/changeme](https://wiki.iiab.io/go/FAQ#What_are_the_default_passwords?) aren't much of any risk generally, certainly not in 2022.  But that might possibly change in future, as more and more IIAB's inevitably start to appear online.  While I'm personally Not in favor of `/etc/iiab/local_vars.yml` being used for (most) highly sensitive passwords, license keys, and claim codes etc (especially those that are commonly used online not just offline).  But certainly: yes this is an ongoing series of human design questions &mdash; with a very strong Precautionary Principle to protect the more vulnerable en route.

There's no magical answer here for sure!  As offline operators in low-income communities inherently + urgently need to build (believable, and increasing) trust that they are being fully protected by the "digital missionaries" who drop off digital learning boxes in their village.  All together monitoring and learning about (intended + unintended) consequences as this happens organically year-by-year....

As is fundamentally a Human question more than a Technological question...
(Associated explanatory videos will certainly be needed in due course, in many languages, whenever people's lives/futures are at stake.  Oftentimes ["Silicon Valley" has set a very low bar in protecting the global public](https://washingtonpost.com/world/2022/03/12/russia-putin-google-apple-navalny/).  Certainly we can do better!)

holta:
Coincidentally the new Raspberry Pi OS from ["April 4th"](https://www.raspberrypi.com/software/operating-systems/) re-architects username/password security substantially:

- '“pi” no more: Raspberry Pi OS ditches longtime user account for security reasons'

  https://arstechnica.com/gadgets/2022/04/raspberry-pi-os-axes-longstanding-default-user-account-in-the-name-of-security/

- '[The "pi" user account] could potentially make a brute-force attack slightly easier, and in response to this, some countries are now introducing legislation to forbid any Internet-connected device from having default login credentials'

  https://www.raspberrypi.com/news/raspberry-pi-bullseye-update-april-2022/

jvonau:
The last gotcha the user can hit is with runrole --reinstall, using [purge](https://github.com/iiab/iiab/blob/master/roles/remoteit/tasks/install.yml#L22-L27) will wipe out any prior registration history as /etc/remoteit will be removed. Without the forced removal the routine just would try to update whatever version is present to the next revision if needed otherwise the user's data should be left as is.

holta:
Indeed.  Definitely `apt purge` is harsh.

Hopefully in future years `apt purge` will no longer be necessary, after remote.it hopefully cleans up their `apt install` .deb packages.

As outlined in these 2 places:

https://github.com/iiab/iiab/blob/1a9025050085f394cce9d8f1caaaaf523da52a33/roles/remoteit/templates/iiab-remoteit#L61-L63

https://github.com/iiab/iiab/blob/1a9025050085f394cce9d8f1caaaaf523da52a33/roles/remoteit/tasks/install.yml#L18-L21

ASIDE: While `apt purge` remains in use, a side benefit is that it promises a completely clean slate ([not that IIAB supports automated uninstalling of apps!](https://github.com/iiab/iiab/blob/1a9025050085f394cce9d8f1caaaaf523da52a33/vars/default_vars.yml#L7-L8))  But occasionally manually wiping is desired &mdash; so I've also documented that here:

https://github.com/iiab/iiab/blob/master/roles/remoteit/README.md#how-to-i-disable-remoteit-on-my-iiab

jvonau:
The use of apt purge during what should be a clean install starting point is really more than what is required to perform what would be just an upgrade of a registered system as originally provided for. Are you suggesting that local_vars be repopulated with the key prior to running runrole --reinstall?, seems less user friendly if you do. Point is any --reinstall should not blow away any user information period, since you want to go the local hands path that power should be left in the hands of the end user at the command line ie run apt purge by hand. The trouble is runrole for some reason always suggest using --reinstall if you don't state it, that is a bit confusing for what should be the normal action, just don't want to set the user up for a fail if --reinstall is mistakenly used.

PR #3174, opened 2022-04-08
Risk order [let's consider stage "5-never-reinstall" instead of 5-xo-services]

jvonau:
Updated version of #2741


holta:
I find this suggestion very promising.

Let's give it some thought and discuss what fine-tuning might make it even better?

e.g. during an upcoming Thursday community call (http://minutes.iiab.io) and/or right here.

holta:
Personally I'm not onboard with the idea that almost every new role/app (which is inherently riskier, as a result of its newness) should be put on the bottom of [9-local-addons/tasks/main.yml](https://github.com/iiab/iiab/blob/master/roles/9-local-addons/tasks/main.yml)

Among other problems, this leads to roles constantly moving around (e.g. where they are positioned within Stage 5, 6, 7, 8 and 9) which would be insufferable (and lead to endlessly pointless re-ordering debates & my-app-is-better-than-yours 100% useless dramas).

My preference is that the specific risks/benefits (of any app/role) be articulated publicly in English, as to the specific consequences (as best we can determine!) of running each role.

jvonau:
Stage 5 is pretty self-explanatory, as is the cleanup of dead roles. The reset is more a grading of degrees of risk to the underling OS, difficulties in configuration of the app for the end-user, and potential network breakage by selecting the role is what I'm using for the criteria for what is in stage 9.

holta:
Thanks for spelling out these 3 criteria:

> grading of degrees of risk to the [underlying] OS

> difficulties in configuration of the app for the end-user

> potential network breakage

There might possibly be others?
And the above 3 might be re-ordered possibly?
In any case, spelling these out is a huge help.

(Anything to avoid a perpetual game of Musical Chairs, highly opinionated and overdramatized...)

Towards working through clear criteria (that everyone may not agree with, but at least everyone can see their reasoning & rational clearly published) &mdash; i.e. evidence that a very thoughtful analysis took place &mdash; with explanation as to how those guidelines will at least help keep things from "perpetually" shifting around every few months, or whatever is "too frequently."

jvonau:
Think a better grouping in place of the arbitrary 6-generic-apps, 7-edu-apps, 8-mgmt-tools might be by the use of supporting technologies, by that I mean the use of php, nodejs, docker.

Issue #3097, opened 2022-01-03
Should host_country_code in local_vars.yml update wpa_supplicant.conf not just hostapd.conf ?


holta:
Should https://github.com/iiab/iiab/blob/master/roles/network/tasks/rpi_debian.yml#L23-L33 change the country code in /etc/wpa_supplicant/wpa_supplicant.conf when someone proactively changes `host_country_code` in local_vars.yml and then runs ./iiab-network ?

(I just noticed that https://github.com/iiab/iiab/blob/master/roles/network/templates/hostapd/hostapd.conf.j2#L18 does indeed change it in /etc/hostapd/hostapd.conf — in these same circumstances — whereas wpa_supplicant.conf does not get changed.)

holta:
FYI I just spoke with @jvonau who indicates he doesn't care, either way.

holta:
FYI whatever is decided, this should be aligned with http://FAQ.IIAB.IO -> "23 Any other networking tips?"

Which currently says: (as the #1 recommendation on top)

> Please change "host_country_code: US" to the correct country code, in `/etc/iiab/local_vars.yml`, prior to installing IIAB. If IIAB is already installed, also run `cd /opt/iiab/iiab` then `sudo ./iiab-network` and then reboot.

jvonau:
"Use the same country code as you used with raspi-config"

holta:
For now I've added a context note to:

[https://wiki.iiab.io/go/FAQ#Any_other_networking_tips?](https://wiki.iiab.io/go/FAQ#Any_other_networking_tips?)

jvonau:
> "Use the same country code as you used with raspi-config"

We can automate some of the detection for those who properly provision their Rpis by using raspi-config before running the curl script, otherwise fall back to the provided default.

holta:
A different approach is implemented by:

- PR #3179

In short:

https://github.com/iiab/iiab/blob/743d8f260157178faa588870efda4f215a1ad153/vars/default_vars.yml#L109-L120

PR #3179, opened 2022-04-14
set 'host_country_code' to value found in wpa_supplicant.conf [on Raspberry Pi OS only, deferring Ubuntu/Debian upstream issues]

jvonau:
host_country_code becomes a fallback value if undetected

#3097


holta:
Does any other IIAB country code logic also need a sanity check &mdash; and/or some kind of user confirmation output (ideally?!)

jvonau:
Not really, the wpa_supplicant.conf becomes authoritative, so there is no mismatch between the two daemons running/talking to the wifi-chipset. The user sets the country code great we run with that or fallback, either way the value **must** be the same for both daemons.

holta:
Discussed on our call (http://minutes.iiab.io) &mdash; this PR looks great, needs testing, and should be merged soon if possible!

jvonau:
grep handling might need a bit of work, the value needed is to the right of the =

holta:
> grep handling might need a bit of work, the value needed is to the right of the =

https://github.com/iiab/iiab-factory/blob/8626f10160b6588d7ee7acb0adda4fde9cf41ebf/box/rpi/min-sd#L88-L94 is a bit related?

```bash
# sanitize wpa_supplicant.conf but preserve country code
COUNTRY_CODE_LINE=`cat /mnt/sdcard/etc/wpa_supplicant/wpa_supplicant.conf | grep country=`
cat <<EOF > /mnt/sdcard/etc/wpa_supplicant/wpa_supplicant.conf
ctrl_interface=DIR=/var/run/wpa_supplicant GROUP=netdev
update_config=1
EOF
echo $COUNTRY_CODE_LINE >> /mnt/sdcard/etc/wpa_supplicant/wpa_supplicant.conf
```

A side question is what to do when a different country code appears in `/etc/wpa_supplicant/wpa_supplicant.conf` (for example...)

```
country=HT
```

...clashing with the country code that appears in `/etc/hostapd/hostapd.conf` (for example...)

```
country_code=CU
```

tim-moody:
how does this impact building a server in one country and deploying in another.

jvonau:
How did min-sd get dragged in to this discussion? The narrow discussion is if the user has provided the working country to unlock the wifi by whatever means (raspi-config, rpi-imager, files in /boot) on RaspiOS that should be taken as the seed value for the templates and not just rely on the user to edit local_vars correctly. Other distros should still edit local_vars.

holta:
What high-level goals make most sense?

Coincidentally [min-sd's existing logic](https://github.com/iiab/iiab-factory/blob/8626f10160b6588d7ee7acb0adda4fde9cf41ebf/box/rpi/min-sd#L88-L94) and [Tim's question](https://github.com/iiab/iiab/pull/3179#issuecomment-1100141528) are both very helpful in surfacing workflow assumptions here &mdash; which common scenarios truly most need something like this?

If in fact propagation of country code makes sense on RasPiOS (e.g. from wpa_supplicant.conf, or from anywhere else!) shouldn't this equally apply on other OS's also &mdash; with Ubuntu and Mint now becoming more popular?

A very small point, but whatever is decided it seems likely we'll end up needing to at least check for 2 (or 3) different country codes possibly appearing in 2 (or 3) different places like...

- `country=XX` in `/etc/wpa_supplicant/wpa_supplicant.conf`
- `country_code=YY` in `/etc/hostapd/hostapd.conf`
- `host_country_code: ZZ` in `/etc/iiab/local_vars.yml`

jvonau:
Whenever Ubuntu gets their act together we can extent the auto detection. https://bugs.launchpad.net/netplan/+bug/1951586

jvonau:
> A side question is what to do when a different country code appears in `/etc/wpa_supplicant/wpa_supplicant.conf` (for example...)
> 
> ```
> country=HT
> ```
> 
> ...clashing with the country code that appears in `/etc/hostapd/hostapd.conf` (for example...)
> 
> ```
> country_code=CU
> ```

Fix the user's provisioning routine, the only way the above currently occurs is with some user interaction and the lack of editing local_vars.yml to the correct value to match the user action then applying with the role. Just want to promote the existing value if found in wpa_supplicant.conf as the authoritative value as that is a user set variable, relying less on what is in local_vars as not to set the user up for a fail the same way as 'channel' can be ignored when needed. You can have the info in the docs all you want but if the keyboard needs to be touched chances are the lazy/uninformed won't make the edit, then complain it not working correctly. Taking the edit local_vars step out of the picture seems like a win to me.

holta:
> Taking the edit local_vars step out of the picture seems like a win to me.

Killing off IIAB's `host_country_code` variable (or hiding it from all new implementers, which is almost the same thing) would be a drastic change from our past recommendations for sure?

Ok, definitely this needs some thinking (hopefully several can help!) whatever's decided here in the end.

Either way, and I might be 100% wrong here: but I was assuming that onboarding Ubuntu/Mint folk to IIAB (i.e. graphical desktop OS's especially) among those who've already connected their laptops to their home's WiFi (such that wpa_supplicant.conf or such is already populated with country code?) &mdash; is likely to be an extremely common newbie scenario.
And if so effectively equivalent to raspi-config setting the WiFi country code?
If there's a way to easily act on that in a somewhat uniform way?
(Or does every desktop manager store country code in its own eccentric way, instead of abiding by wpa_supplicant.conf rather universally, making this all unrealistic?)

jvonau:
> > Taking the edit local_vars step out of the picture seems like a win to me.
> 
> Killing off IIAB's `host_country_code` variable (or hiding it from all new implementers, which is almost the same thing) would be a drastic change from our past recommendations for sure?
>

Um, I do not see it like that, host_country_code is a fallback value as it is now, just the fallback might be incorrect once the user environment is localized and local_vars becomes out of sync. Running the role with the user env and local_vars out of sync will cause the mis-match as the fallback is applied shooting the user in the foot. That is why there is the warning to ensure the value is correct in the local_vars in the docs but we all know that step is usually omitted leading to further confusion.
 
> Ok, definitely this needs some thinking (hopefully several can help!) whatever's decided here in the end.
> 
> Either way, and I might be 100% wrong here: but I was assuming that onboarding Ubuntu/Mint folk to IIAB (i.e. graphical desktop OS's especially) among those who've already connected their laptops to their home's WiFi (such that wpa_supplicant.conf or such is already populated with country code?) 

Dig into NetworkManager's innards and it becomes clear that the wpa_supplicant.conf is generated on the fly and not saved.
> — is likely to be an extremely common newbie scenario. And if so effectively equivalent to raspi-config setting the WiFi country code? 

No, follow up in some of the related links in the upstream bugtracker

>If there's a way to easily act on that in a somewhat uniform way? (Or does every desktop manager store country code in its own eccentric way, instead of abiding by wpa_supplicant.conf rather universally, making this all unrealistic?)

Unrealistic, NetworkManager does not store the value, see the double connect comment while netplan upstream is struggling on how to handle this long standing situation as 22.04 nears the release date.

holta:
Thanks @jvonau for the explanations above.

holta:
> how does this impact building a server in one country and deploying in another.

I'm realizing @tim-moody's question is crucial if not the crux of the matter:

As more implementers/deployers/communities get in the habit of changing their WiFi country code in various ways...

1) Running `/usr/bin/raspi-config`
2) Manually populating country code into `/etc/wpa_supplicant/wpa_supplicant.conf`
3) "Raspberry" pull-down menu (top-left) > Preferences > "Raspberry Pi Configuration" leads to...
image
4) WiFi/Networking graphical widget (top-right of screen, left-click)
image

_...the more they'll (later on) go back to the same place(s) where they began (like salmon, we all return to our spawning grounds ;) trying to change the country code once again (in later weeks/months)._  e.g. Typically in support of any kind of linguistic/cultural/ethnic/diaspora spread across many countries.

How should we explain this...

- What workflow/clickflow habits + recommendations should we instill?
- IIAB's are almost inevitably moving across borders, and offline devices will not be magically auto-configured on arrival (and few people care about WiFi country code...but still many implementers do want to respect local laws...if it's made very easy for them to do so!)
- Not easy given the final destination is _Not_ always known in advance.  Especially among refugees / asylum seekers continuously being pushed from one country to another (all too common on all continents, Antarctic penguins being pushed out too now...)

_Of course we Cannot help everybody!_  e.g. folks who can't resist and so right-click in the top-right of their "Raspberry Pi OS with desktop" screens &mdash; which is their problem if they decide to mess up critical "Wireless & Wired Network" bits below &mdash; e.g. if temptation sets in too often, they'd better get to know `dd`-like backups often!
image

jvonau:
Fiddle with the network setting you bought the layout as is. Want to return the 'stock' layout run the network role, editing local_vars to match the country first. Running the network role at least once post install ensures distributed images receive a new unique bssid/mac address for the wifi's ap0 interface.

holta:
> Fiddle with the network setting you bought the layout as is. Want to return the 'stock' layout run the network role, editing local_vars to match the country first. Running the network role at least once post install ensures distributed images receive a new unique bssid/mac address for the wifi's ap0 interface.

Thanks!

I posted summary instructions to that effect, at the top of:

https://github.com/iiab/iiab/wiki/IIAB-Networking#common-customizations

jvonau:
> Want to return the 'stock' layout run the network role, editing local_vars to match the country first. 

Just trying to use KISS. This PR is geared to detect what the user has already configured using whatever means from above then respecting that configuration in place of just always using the fallback value which requires the user editing in advance to keep the value in sync. I toyed with the idea of recording the 'detected country' as 'host_country_code' in iiab_state,yml to denote there is a local override to the supplied value in local_vars but the code path is pretty clear that the source of the variable can be within wpa_supplicant.conf or the fallback value for the variable is used.

holta:
So everyone can help smoke-test this PR, are there any OS's or test cases that should be prioritized?

jvonau:
supersedes PR 3226, testing on 181-raspios64-lite-UNITTEST-country, needs appliance like vm testing to ensure iiab-hotspot-on contains only "AP support was not detected please see"

roles/network/templates/hostapd/iiab-hotspot-on
```diff
@@ -1,4 +1,7 @@
 #!/bin/bash
+{% if not can_be_ap %}
+AP support was not detected please see https://github.com/iiab/iiab/pull/3179
```
holta:
```suggestion
echo -e "\nUH-OH: Your Wi-Fi firmware doesn't support AP mode, according to 'iw list'\n"
echo -e "If you add Wi-Fi hardware, run 'cd /opt/iiab/iiab' then 'sudo ./iiab-network'\n"
echo -e "For details, see: https://github.com/iiab/iiab/pull/3179\n"
exit 1
```

Ref:

- PR #3222

holta, line 3:
```suggestion
echo -e "\nUH-OH: Your Wi-Fi firmware doesn't support AP mode, according to 'iw list'\n"
echo -e "If you add Wi-Fi hardware, run 'cd /opt/iiab/iiab' then 'sudo ./iiab-network'\n"
echo -e "For details, see: https://github.com/iiab/iiab/pull/3179\n"
exit 1
```

jvonau:
iiab-network tested on a 22.04 multipass VM - appliance summary of important info;
```
Please REBOOT to fully verify your network -- graphical desktops MUST reboot!
ubuntu@box:/opt/iiab/iiab$ cat /usr/bin/iiab-hotspot-on
#!/bin/bash
echo -e "\nUH-OH: Your Wi-Fi firmware doesn't support AP mode, according to 'iw list'\n"
echo -e "If you add Wi-Fi hardware, run 'cd /opt/iiab/iiab' then 'sudo ./iiab-network'\n"
echo -e "For details, see: https://github.com/iiab/iiab/pull/3179\n"
exit 1
ubuntu@box:/opt/iiab/iiab$ sudo /usr/bin/iiab-hotspot-on

UH-OH: Your Wi-Fi firmware doesn't support AP mode, according to 'iw list'

If you add Wi-Fi hardware, run 'cd /opt/iiab/iiab' then 'sudo ./iiab-network'

For details, see: https://github.com/iiab/iiab/pull/3179

```
- template is correct

holta:
Looks great: is any other testing required?

Or is this ready to merge?

jvonau:
Famous last words, 'should be good to go'

holta:
Quick Recap of this PR...for those who haven't been following details:

https://github.com/iiab/iiab/blob/743d8f260157178faa588870efda4f215a1ad153/vars/default_vars.yml#L109-L120

jvonau:
Movement on the Ubuntu front https://github.com/canonical/netplan/pull/281

PR #3184, opened 2022-04-19
Moodle 4.0 is now official

holta:
Moodle 4.0 is now official: https://moodle.com/news/moodle-4-is-here/

This PR could use broad community testing on different OS's (after merging too) while we spot-check Moodle 4.0's official requirements[*] here:

- https://docs.moodle.org/dev/Moodle_4.0_release_notes#Server_requirements
- #3182

[*] Whether or not it yet works with Ubuntu 22.04's PHP 8.1!

Building on:

- PR #2841
- PR #2869

PR #3185, opened 2022-04-19
workaround php8.1 [e.g. for Moodle 4.0 on Ubuntu 22.04]

jvonau:
https://moodle.org/mod/forum/discuss.php?d=429511

### Fixes bug:
Not sure if this is a fix or not #3184
Allows install to continue to allow exploring what else might be broken with the shift to php8.1
multipass-2204vm


jvonau:
It's an experimental hack..

PR #3165, opened 2022-04-05
Branch Maps7.3 -- includes 2020 OSM data and 2020 satellite images

georgejhunt:
Tested on Ubuntu vbox, and rpi hardware, San_Jose, Spanish_speakers, Central_America, and tested on my repo. Needs to be coordinated with branch maps7.3 into the iiab/maps rep. This is a test version and cannot be pulled into master without modification to roles/osm-vector-maps/defaults/main.yml.

Code for generation is a jupyter notebook at https://github.com/georgejhunt/maps/blob/maps7.3/osm-source/jupyter-code/maps.ipynb.

Notes: For 2020 (this version) are at  https://github.com/georgejhunt/maps/blob/maps7.3/osm-source/jupyter-code/README.md

PR #3192, opened 2022-04-26
host maps on github

jvonau:
Related https://github.com/iiab/iiab-admin-console/pull/483#issuecomment-1108922895 "I propose that we put the map-catalog.json on githubusercontent" 
Needs cities1000.sqlite and map-catalog.json added to [maps](https://github.com/iiab/maps) and regions.json corrected as per https://github.com/jvonau/maps/pull/1/files
Would need the last commit rolled back before committing to master, it's provided for testing convenience of the new github location scheme until the above changes are implemented in maps.
Tested on 22.04 VM

### DIFF
--- a/roles/osm-vector-maps/defaults/main.yml
+++ b/roles/osm-vector-maps/defaults/main.yml
@@ -15,6 +15,7 @@ maps_branch: 'master'    # Quotes not required
 
 # soft code sources
 archive_org_url: https://archive.org/download
+iiab_map_url: http://download.iiab.io/content/OSM/vector-tiles/maplist/hidden
 #map_catalog_url: http://download.iiab.io/content/OSM/vector-tiles
 map_catalog_url: http://timmoody.com/iiab-files/maps
 satellite_version: satellite_z0-z9_v3.mbtiles    # 2021-12-20: Var unused, but hard-coded in 11 places within https://github.com/iiab/iiab-admin-console -- #3077 discusses map-catalog.json & adm-map-catalog.json
--- a/roles/osm-vector-maps/defaults/main.yml
+++ b/roles/osm-vector-maps/defaults/main.yml
@@ -8,10 +8,10 @@
 # If nec, change them by editing /etc/iiab/local_vars.yml prior to installing!
 
 # The following soft coded variables allow testing, before pulling PR's into master
-# osm_repo_url: https://raw.githubusercontent.com/iiab/maps
-osm_repo_url: https://raw.githubusercontent.com/georgejhunt/maps
-#maps_branch: 'master'    # Quotes not required
-maps_branch: 'maps7.3'
+osm_repo_url: https://raw.githubusercontent.com/iiab/maps
+maps_branch: 'master'    # Quotes not required
+#osm_repo_url: https://raw.githubusercontent.com/georgejhunt/maps
+#maps_branch: 'maps7.3'
 
 # soft code sources
 archive_org_url: https://archive.org/download
--- a/roles/osm-vector-maps/defaults/main.yml
+++ b/roles/osm-vector-maps/defaults/main.yml
@@ -15,7 +15,7 @@ maps_branch: 'master'    # Quotes not required
 
 # soft code sources
 archive_org_url: https://archive.org/download
-iiab_map_url: http://download.iiab.io/content/OSM/vector-tiles/maplist/hidden
+#iiab_map_url: http://download.iiab.io/content/OSM/vector-tiles/maplist/hidden
 #map_catalog_url: http://download.iiab.io/content/OSM/vector-tiles
 map_catalog_url: http://timmoody.com/iiab-files/maps
 satellite_version: satellite_z0-z9_v3.mbtiles    # 2021-12-20: Var unused, but hard-coded in 11 places within https://github.com/iiab/iiab-admin-console -- #3077 discusses map-catalog.json & adm-map-catalog.json
--- a/roles/osm-vector-maps/tasks/install.yml
+++ b/roles/osm-vector-maps/tasks/install.yml
@@ -10,6 +10,16 @@
     - "{{ vector_map_path }}/viewer/tiles"
     - "{{ vector_map_path }}/installer"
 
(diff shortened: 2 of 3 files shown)


jvonau:
https://github.com/iiab/maps/pull/54

roles/osm-vector-maps/defaults/main.yml
```diff
@@ -8,13 +8,17 @@
...
-# osm_repo_url: https://raw.githubusercontent.com/iiab/maps
-osm_repo_url: https://raw.githubusercontent.com/georgejhunt/maps
+#osm_repo_url: https://raw.githubusercontent.com/iiab/maps
 #maps_branch: 'master'    # Quotes not required
-maps_branch: 'maps7.3'
+
+#osm_repo_url: https://raw.githubusercontent.com/georgejhunt/maps
+#maps_branch: 'maps7.3'
+osm_repo_url: https://raw.githubusercontent.com/jvonau/maps
+maps_branch: github_host
```
holta: Should these 2 lines be revised to...something like?

```suggestion
osm_repo_url: https://raw.githubusercontent.com/iiab/maps
maps_branch: master
```

jvonau: Used in place of an override within local_vars, harder for everybody to test if you have to edit anything. Those are only in place for testing this PR and should be changed before merging once the iiab/maps repo is updated.

holta: Yup!  Hopefully @georgejhunt can weigh in this morning if he has time (on both these PR's) to help finalize this PR here and also:

- PR iiab/maps#54

jvonau: > Would need the last commit rolled back before committing to master, it's provided for testing convenience of the new github location scheme until the above changes are implemented in maps.

tim-moody:
I thought that our standard was to use the vars in a role for testing, but to make the var in vars/default_vars.yml the definitive value. Is the intent that this is not a user configurable value? I assume it will not be visible to the Admin Console install.

holta:
> I thought that our standard was to use the vars in a role for testing, but to make the var in vars/default_vars.yml the definitive value. Is the intent that this is not a user configurable value? I assume it will not be visible to the Admin Console install.

Sounds good to me!

jvonau:
> I thought that our standard was to use the vars in a role for testing, but to make the var in vars/default_vars.yml the definitive value. 

That has not changed, but #3165 didn't take care of default_vars, used defaults/main.yml to point to the testing repo that is still used today. Thought that value would of changed back to iiab/maps once https://github.com/iiab/maps/pull/45 came alive but I'm not the one who approves PRs 

> Is the intent that this is not a user configurable value? I assume it will not be visible to the Admin Console install.

Any entries in defaults/main.yml or default_vars.yml are still user configurable via local_vars.yml but may not have entries present to do so relying on the default value that is defined.

holta:
Either way.

Certainly as @tim-moody suggests, if values are changed very rarely every few years, they should likely be in vars/default_vars.yml primarily.

jvonau:
Guess to sum this up 'iiab_map_url' is going away, 'osm_repo_url' and 'maps_branch' are taking its place becoming the authoritative source pointer once iiab/maps have the required files added. Now 'map_catalog_url' is not quite accurate it's more 'map_mbtiles_url', I can change that while the hood is up or defer that to Adam. Just need to add one more commit to enable default_vars.yml over defaults/main.yml or I can roll back the last commit for the same effect before merging.

holta:
Either is fine, Thanks Jerry!

tim-moody:
should the names osm_repo and map_branch sync up, like osm_repo_branch
or even just allow for the branch as part of the url

georgejhunt:
My strategy is to test the values in osm-vector-maps/defaults/main.yml, which are used for testing massive changes to the map system. And then  change the variable name in the roles/install.yml to iiab_map_url (which is set in iiab/vars/default_vars.yml) as a separate, and later PR.

So I'm not in favor of iiab_map_url going away.

tim-moody:
@georgejhunt I think it's your call.

jvonau:
iiab_map_url is currently used to retrieve 2 files from http://download.iiab.io where do you propose these files be served from?

jvonau:
> should the names osm_repo and map_branch sync up, like osm_repo_branch or even just allow for the branch as part of the url

That discussion should of been part of PR 3165, I just went with what is already presently defined.

jvonau:
#3195

holta:
> @georgejhunt I think it's your call.

Thanks @georgejhunt if you can make a recommendation &mdash; so @jvonau can refine this PR per your recommendation?

jvonau:
https://github.com/iiab/maps/issues/55#issuecomment-1110136033

georgejhunt:
I recommend  that you take me out of the loop. Considering my diminished
capacity, I'm not good at analyzing the changes in
https://github.com/iiab/iiab/pull/3192/files, and figuring why the position
of installing packages was moved up in the playlist. Which changes matter,
and which do not? What are the unintended

I think this is a good time to let Jerry and Adam take over the development
of OSM.

I think master still functions, and as Jerry says, right now we are using a
development branch, which needs to be swapped. I have never been very good
at the back and forth interactive stuff. I don't monitor our skype channel.

I'm finding that if I don't enjoy it, I put it off. I've been putting off
IIAB a lot recently.

holta:
> good time to let Jerry and Adam take over the development of OSM

@georgejhunt you made all this map/satellite work happen (none of which could have happened without you!) and I apologize I've been quite busy this week with 3 unrelated hardware failures.

But yes we will help package this up as best we can in coming days &mdash; calling you when we get stuck.

Thanks to @jvonau's initiative overnight and @tim-moody who's helping a ton here &mdash; _truly a joint effort and tremendous joint accomplishment with the potential for Many Thousands of kids + teachers to use this over coming years!_

holta:
New PR:

- PR iiab/iiab-admin-console#484

Thanks everyone for these discussions, helping to clarify:

- iiab/maps#54
- iiab/maps#55

Issue #3199, opened 2022-04-29
remoteit role fails on Linux Mint ["Your OS or distribution is not supported by this install script."]


jvonau:
From @shanti-bhardwa 
>TASK [remoteit : Disable & Stop remote.it services {connectd, schannel}] *******
failed: [127.0.0.1] (item=connectd) => {"ansible_loop_var": "item", "changed": false, "item": "connectd", "msg": "Could not find the requested service connectd: host"}
failed: [127.0.0.1] (item=schannel) => {"ansible_loop_var": "item", "changed": false, "item": "schannel", "msg": "Could not find the requested service schannel: host"}
PLAY RECAP *********************************************************************
127.0.0.1                  : ok=42   changed=8    unreachable=0    failed=1    skipped=15   rescued=0    ignored=0

holta:
Thank you @shanti-bhardwa and @jvonau for this bug report.

shanti-bhardwa:
got this at the end. I rebooted then ran sudo iiab and it said already installed, nothing to do]
Traceback (most recent call last):
  File "/usr/lib/python3.8/urllib/request.py", line 1354, in do_open
    h.request(req.get_method(), req.selector, req.data, headers,
  File "/usr/lib/python3.8/http/client.py", line 1256, in request
    self._send_request(method, url, body, headers, encode_chunked)
  File "/usr/lib/python3.8/http/client.py", line 1302, in _send_request
    self.endheaders(body, encode_chunked=encode_chunked)
  File "/usr/lib/python3.8/http/client.py", line 1251, in endheaders
    self._send_output(message_body, encode_chunked=encode_chunked)
  File "/usr/lib/python3.8/http/client.py", line 1011, in _send_output
    self.send(msg)
  File "/usr/lib/python3.8/http/client.py", line 951, in send
    self.connect()
  File "/usr/lib/python3.8/http/client.py", line 1425, in connect
    self.sock = self._context.wrap_socket(self.sock,
  File "/usr/lib/python3.8/ssl.py", line 500, in wrap_socket
    return self.sslsocket_class._create(
  File "/usr/lib/python3.8/ssl.py", line 1040, in _create
    self.do_handshake()
  File "/usr/lib/python3.8/ssl.py", line 1309, in do_handshake
    self._sslobj.do_handshake()
TimeoutError: [Errno 110] Connection timed out

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
[...]

PR #3202, opened 2022-04-29
remoteit/tasks/install.yml: Refine PR #3201 using Ansible not bash

holta:
Bonus: using the Ansible approach also avoids duplicate `IIAB_LIKE=Ubuntu` lines in `/etc/os-release` e.g. in the rare case when someone reinstalls https://remote.it

Tested on Linux Mint 20.3 Una.

Building on:

- #3199
- PR #3201


shanti-bhardwa:
tested IIAB 8.0 on Linux Mint 20.3 and Medium Build with all the defaults works fine. See http://sprunge.us/Sfk3EG

PR #3207, opened 2022-05-02
add osm constants

jvonau:


tim-moody:
adm_const.py.j2 uses 
map_doc_root = '{{ vector_map_path }}'  

viewer_path is too generic, suggest osm_viewer_path
same for sat_dir, which is really tiles dir
osm_dir looks to b a replacement for vector_map_idx_dir, which is a better name, and shouldn't the value be vector_map_idx_dir = '/library/www/html/common/assets'?

tim-moody:
one other comment is that I originally put the addition of these constants and functions in 2-common so that they would be available to any subsequent role. that may no longer be the case.

jvonau:
> adm_const.py.j2 uses map_doc_root = '{{ vector_map_path }}'
>

That is better, I'll change this PR.
> viewer_path is too generic, suggest osm_viewer_path same for sat_dir, which is really tiles dir osm_dir looks to b a replacement for vector_map_idx_dir, which is a better name, 

I didn't try to improve on the names used, just the location where these are defined. The names used in [maps](https://github.com/iiab/maps/pull/61/files) could of been changed at anytime in the past but has not been but now it's an issue. Suppose the above files in maps could do the translation from the above suggestion to what is used in the scripts. I'll pass on the renaming for now, until others weigh in thanks.

>and shouldn't the value be vector_map_idx_dir = '/library/www/html/common/assets'?

Would that not be the same path as https://github.com/iiab/iiab/blob/master/roles/pylibs/templates/iiab_const.py.j2#L18?

holta:
Awesome:

A bit of consolidation after half a decade is a really great idea, especially if this tightens up disparate code for readability!

holta:
ASIDE: Overloading and abusing the acronym/name "osm" is probably ok, within limits!

(e.g. sometimes we talk about things like Sentinel satellite photos as if they are OSM, when they really have little to do with OpenStreetMap.  But it's hard to resist the 3-letter acronym "osm" even when it's not always entirely accurate...and maybe that's ok ;-)

jvonau:
What is the preference path or dir? For full context see https://github.com/iiab/maps/pull/61/files with the proposed names noted.

holta:
> What is the preference path or dir? For full context see https://github.com/iiab/maps/pull/61/files with the proposed names noted.

Either seems absolutely fine.  If it matters:

- `dir` is fewer letters
- `path` clarifies that it's a full path (not just a relative directory) if that's indeed the case?

Issue #3180, opened 2022-04-15
How should Admin Console log ansible runs?


tim-moody:
This will cause ansible to log to the standard iiab log even when run from Admin Console.

jvonau:
What is the standard iiab log? Note `export ANSIBLE_LOG_PATH=` is used within iiab-install, iiab-network, and runrole to direct the logging to different log files.

tim-moody:
I didn't realize runrole had a different log than iiab-install. In any event I am asking for iiab-install's log. The export variable takes precedence over the config file, so there will be no conflict with current scripts. I am unable to set the env variable.

holta:
> I am unable to set the env variable.

Why is that out of curiosity?

> Note export ANSIBLE_LOG_PATH= is used within iiab-install, iiab-network, and runrole to direct the logging to different log files.

Good point.

Some might argue that ~4 log files is too many.

On the other hand, mashing everything together into 1 single log file (e.g. /opt/iiab/iiab/iiab-install.log) is a bit much.

Certainly this would makes [iiab-diagnostics](https://github.com/iiab/iiab/blob/master/scripts/iiab-diagnostics.README.md) output less meaningful, as iiab-diagnostics takes the last 100 lines of each of these log files.

Ref:

- PR iiab/iiab-admin-console#475

jvonau:
log_path=/var/log/iiab-ansible.log maybe as the fallback? Just one more log file to check for.

jvonau:
To test this I broke the link of /opt/admin/cmdsrv/ansible.cfg and provided a new file with the above included, works nicely. The path and log file name is up for debate as this would be shared with runroles if the change is made in iiab without breaking the link.

holta:
Followup:

- PR iiab/iiab-admin-console#476
- #3181
- PR iiab/iiab-admin-console#477

holta:
We should think about Admin Console recording to a distinct log file separate from the initial install's `/opt/iiab/iiab/iiab-install.log` (regardless what it is named!) &mdash; so as to illuminate more insights via [iiab-diagnostics](https://github.com/iiab/iiab/blob/master/scripts/iiab-diagnostics.README.md), which only records the last 100 lines of each such file:

- PR https://github.com/iiab/iiab-admin-console/pull/477

holta:
[...]

PR #3219, opened 2022-05-12
iiab-diagnostics: journalctl -t IIAB-CMDSRV

holta:
@tim-moody can you confirm the `'Admin Console CMDSRV commands'` description is sufficient?

Related:

- #3180
- PR iiab/iiab-admin-console#488


tim-moody:
'log' might be better than 'commands' as there are status entries that are not strictly speaking commands.

PR #3223, opened 2022-05-19
Tolerate Ubuntu 22.10 Kinetic Kudu to help with testing

holta:
Both Microsoft (WSL2) and Ubuntu (https://Multipass.run) have made it dead easy to install daily builds of this pre-release OS.

To help with testing, long before its Thur Oct 20th release date late this year.


holta:
Supporting 4 different Ubuntu versions is too much.

_Hopefully by July we can make clear to the public that 20.04 LTS and 21.10 are no longer necessary._

PR #3226, opened 2022-05-19
wpa_supplicant.conf country code authoritative (for hostapd.conf) espec for RasPiOS

holta:
This is @jvonau's work, building on:

- PR #3179


jvonau:
Implemented in #3179

PR #3230, opened 2022-05-29
iiab-diagnostics: 4 fixes/cleanups

holta:
Four fixes/cleanups:

1. Show commit for both repos {iiab, iiab-admin-console} even if non-root
2. Cleaner "/CANONICAL-PATH/COMMAND PARAMS" output
3. Lint bash to `[[ $VAR == "" ]]`
4. `bash -c "$1"` instead of eval/exec to fix globbing, so commands like `ls -l /lib/firmware/brcm/*43455*` stop FAILING to output lines w/ filenames that contain spaces (ugly IFS issues!)

Refs:

- #3106
- PR #3209
- PR #3163
- PR #3219


holta:
Tested on Linux Mint 20.3, running as non-root.

Looks much better now, also when running `iiab-diagnostics` as root.

PR #3238, opened 2022-06-07
WIP: MongoDB Spring Cleaning (bug fixes) for #3236

holta:
See in-line comments explaining, to begin to resolve:

- #3236

Building on:

- PR #3214

Issue #3236, opened 2022-06-06
Sugarizer won't start due to "Illegal instruction /usr/bin/mongod --repair --dbpath /library/dbdata/mongodb" [also reveals flaws in /etc/mongod.conf, mongodb.service & iiab-mongodb-repair-if-no-lock]


holta:
Thanks to @nzola who uncovered this:

```
1958 Jun  6 13:40:18 raspberrypi iiab-mongodb-repair-if-no-lock[76019]: /usr/bin/iiab-mongodb-repair-if-no-lock: line 7: 76020 Illegal instruction     /usr/bin/mongod --repair --dbpath /library/dbdata/mongodb
1959 Jun  6 13:40:18 raspberrypi systemd[1]: mongodb.service: Control process exited, code=exited, status=132/n/a
1960 Jun  6 13:40:18 raspberrypi systemd[1]: mongodb.service: Failed with result 'exit-code'.
1961 Jun  6 13:40:18 raspberrypi systemd[1]: Failed to start High-performance, schema-free document-oriented database.
```

From http://sprunge.us/LbT5zf?en on Raspberry Pi OS on RPi 4, but I assume this applies to all OS's.

Likely as a result of MongoDB 5.x (released 11 months ago) which presumably no longer supports the legacy syntax / command used by [iiab-mongodb-repair-if-no-lock](https://github.com/iiab/iiab/blob/master/roles/mongodb/templates/iiab-mongodb-repair-if-no-lock.j2) :

- PR #3214

In any case, with MongoDB 6.0 now imminent, it's time for IIAB to catch up.  I'll investigate.

holta:
Just FYI IIAB's 241-line [template for /etc/mongod.conf](https://github.com/iiab/iiab/blob/master/roles/mongodb/templates/mongod.conf.j2) (IIAB currently used this for 32-bit Raspberry Pi OS only) appears archaic to the point where it's become highly misleading.

Whereas on 64-bit OS's, IIAB currently uses MongoDB 5.0's own default 42-line `/etc/mongod.conf` _which uses the wrong DB below i.e. `/var/lib/mongodb` instead of `/library/dbdata/mongodb` !_  (But still we might consider this as a basis for IIAB's future template?)

```
# mongod.conf

# for documentation of all options, see:
#   http://docs.mongodb.org/manual/reference/configuration-options/

# Where and how to store data.
storage:
  dbPath: /var/lib/mongodb
  journal:
    enabled: true
#  engine:
[...]

Issue #3190, opened 2022-04-22
MongoDB 4.4, 5.0 & 6.0RC1 don't run on Ubuntu 22.04 (preventing Sugarizer install) [missing apt package libssl1.1 -> libssl3 ?]


holta:
FYI @llaske & All,

Here's the reason Sugarizer won't install on Ubuntu 22.04 &mdash; regardless whether trying to install MongoDB 4.4 or 5.0 or 6.0RC1:

```
# apt-key list mongodb
Warning: apt-key is deprecated. Manage keyring files in trusted.gpg.d instead [of /etc/apt/trusted.gpg] (see apt-key(8)).
pub   rsa4096 2021-02-16 [SC] [expires: 2026-02-15]
      F567 9A22 2C64 7C87 527C  2F8C B00A 0BD1 E2C6 3C11
uid           [ unknown] MongoDB 5.0 Release Signing Key <packaging@mongodb.com>

# cat /etc/apt/sources.list.d/mongodb-org.list
deb [ arch=amd64,arm64 ] https://repo.mongodb.org/apt/ubuntu focal/mongodb-org/5.0 multiverse

# apt install mongodb-org mongodb-org-server
...
The following packages have unmet dependencies:
 mongodb-org-mongos : Depends: libssl1.1 (>= 1.1.1) but it is not installable
 mongodb-org-server : Depends: libssl1.1 (>= 1.1.1) but it is not installable
 mongodb-org-shell : Depends: libssl1.1 (>= 1.1.1) but it is not installable
E: Unable to correct problems, you have held broken packages.
```

( As mentioned 13 days ago at https://www.linuxshelltips.com/install-mongodb-ubuntu/#comment-9117 )

_Does this suggest we may need to wait for the release of MongoDB 5.4 or 6.0 or 6.1 ?_

If so, it might be many weeks away &mdash; given their latest "testing" and "development" branches (below) have the exact same `libssl1.1 (>= 1.1.1)` dependency problem:

- https://repo.mongodb.org/apt/ubuntu/dists/focal/mongodb-org/testing/multiverse/binary-amd64/mongodb-org-shell_6.0.0~rc1_amd64.deb
- https://repo.mongodb.org/apt/ubuntu/dists/focal/mongodb-org/development/multiverse/binary-amd64/mongodb-org-shell_6.0.0~latest_amd64.deb

Related:

- PR #3099
- PR #3189

holta:
Preliminary workaround ideas are being discussed here today:

[...]

PR #3099, opened 2022-01-04
sugarizer/defaults/main.yml: 1.5.0 -> 1.6.0

holta:
https://github.com/llaske/sugarizer/releases/tag/v1.6.0

Works with:

- PR #3040

PR #3241, opened 2022-06-08
WIP: Tighten up roles/mongodb — runs 'apt-mark hold mongodb-org mongodb-org-server' on 64-bit RPi if installing MongoDB >= 5.0

holta:
Towards making IIAB's MongoDB install routine more resilient on [all 4 of IIAB's mainline OS's](https://github.com/iiab/iiab/wiki/IIAB-Platforms) &mdash; across the most common CPU architectures:

- #3236
- PR #3238

Tested on RasPiOS Lite on Raspberry Pi 4.

This should be tested much more broadly to verify and further beef up where needed.

Related:

- PR #3040
- PR #3099
- #3190


jvonau:
Why not just remove the mongo apt repo definition to prevent future upgrades? The net result would be the same as using deb with url as is done with Kolibri.

holta:
> Why not just remove the mongo apt repo definition to prevent future upgrades? The net result would be the same as using deb with url as is done with Kolibri.

It's definitely an idea.

The OS's own apt repos might complicate things in some isolated cases hypothetically however &mdash; by suggesting diverse MongoDB package updates/alternatives that are not wanted hypothetically?

Some version of `apt-mark hold mongodb*` appears to lock MongoDB in place rather simply and also rather strictly (as desired on RasPiOS).

With different OS's deserving different treatment &mdash; trying hard not to allow all this complexity to mushroom beyond nec.

_Certainly I need to sleep on it._

Ref:

- #3217

holta:
Note to self... Ansible's [dpkg_selections](https://docs.ansible.com/ansible/latest/collections/ansible/builtin/dpkg_selections_module.html) module can hold apt packages if nec:

```
- name: Hold mongodb-org-server
  dpkg_selections:
    name: mongodb-org-server
    selection: hold
```

Per https://stackoverflow.com/a/63983134

jvonau:
Better question is what is missing from the 64-bit raspbian_mongodb_5.0.5.gz file that would require installing a binary incomparable version with apt? The new step is not needed with the (much) older 32-bit tar file version installed for RaspiOS.

holta:
> Better question is what is missing from the 64-bit raspbian_mongodb_5.0.5.gz file that would require installing a binary incomparable version with apt? The new step is not needed with the (much) older 32-bit tar file version installed for RaspiOS.

Testing shows there's only 1 file really needed from raspbian_mongodb_5.0.5.gz and that's `/usr/bin/mongod`

(The other 2 files added by raspbian_mongodb_5.0.5.gz don't seem to matter for Sugarizer... but apparently don't hurt...)

- /usr/bin/mongo
- /usr/bin/mongos

So the strategy here is: layer the 1 critical file on top of MongoDB's own 5.0.9 installer (to abstract away all the evolving/finer details across MongoDB's different files/directories, as much as possible — so IIAB intentionally does not try to micromanage...)

jvonau:
Or just install version 4.4.13 of mongodb from the repo, sidestepping the whole cpu issue that comes with 5.X.X as the same kind of cpu issue exists in Intel/AMD land not just on ARM64, mainly because you are targeting the reuse of older hardware. I'll further suggest that sugarizer be removed from auto installing within the medium and large vars files should you continue to go down the 5.X.X path to save pain and suffering for the newbies that may come along in the future with substandard Intel/AMD hardware.

holta:
This PR is now reworked to support softcoding the MongoDB version number to be installed on 64-bit OS's &mdash; set to `mongodb_64bit_version: 4.4` for now:

- With the unfortunate reality that MongoDB 4.0 has already been [EOL'd as of 2022-04-30](https://www.mongodb.com/support-policy/lifecycles).
- So `mongodb_64bit_version: 5.0` and/or `mongodb_64bit_version: 6.0` might potentially become useful over the coming year, among those with "post-2011" CPU's that support AVX instructions:
  - https://github.com/docker-library/mongo/issues/485#issuecomment-891991814

Finally: this PR was tested successfully on Ubuntu 20.04 (Server) x86_64 VM &mdash; http://box/sugarizer works and the PR would appear to be almost ready to merge.

holta:
Tested successfully on Mint 20.3 (graphical desktop) x86_64 VM — http://box/sugarizer works.

holta:
Tested successfully on 64-bit Raspberry Pi OS (Lite) on Raspberry Pi 4 — http://box/sugarizer works.

holta:
Future testing suggestion:

Test shut-down times on 32-bit Raspberry Pi OS e.g. on RPi 3 and 3 B+.

_(...if this older 32-bit OS still matters, now that 64-bit Raspberry Pi OS is becoming increasingly mainline!)_

Issue #3268, opened 2022-06-29
Workaround leads to need for cascading changes


jvonau:
### Expected behavior and actual behavior:
Error reported remoteit role fails on Linux Mint ["Your OS or distribution is not supported by this install script."] #3199
https://github.com/iiab/iiab/issues/3106#issuecomment-1146659973 https://github.com/iiab/iiab/issues/3106#issuecomment-1146663307 https://github.com/iiab/iiab/issues/3106#issuecomment-1146681454 https://github.com/iiab/iiab/issues/3106#issuecomment-1146683602

The workaround employed in #3201 and leaving the workaround in place led to requiring other changes in the ecosystem:
use iiab fact os instead of ansible distribution [#496](https://github.com/iiab/iiab-admin-console/pull/496)
using iiab facts means raspbian is its own os [#499]( https://github.com/iiab/iiab-admin-console/pull/499)

The takeaway is the 'fix' permanently altered what OS ansible detected in "ansible_distribution", this might come back to haunt the future in some unforeseen way, the line should be removed prior to leaving the install.yml task.

tim-moody:
>The takeaway is the 'fix' permanently altered what OS ansible detected in "ansible_distribution"

sorry, don't understand at all; how was ansible's os parsing changed?

jvonau:
The addition of "IIAB_LIKE=Ubuntu" in /etc/os-release by the remoteit role lead to the different behavior between Tim's and Andrew's [installs](https://github.com/iiab/iiab/issues/3106#issuecomment-1146663307), the difference being the addition of the one line leading to the change in reporting within "ansible_distribution" by ansible. This could confuse other installers that parse /etc/os-release also.

tim-moody:
iiab should not be modifying /etc/os-release. fyi, on my vm I have

NAME="Linux Mint"
VERSION="20.3 (Una)"
ID=linuxmint
ID_LIKE=ubuntu
PRETTY_NAME="Linux Mint 20.3"
VERSION_ID="20.3"
HOME_URL="https://www.linuxmint.com/"
SUPPORT_URL="https://forums.linuxmint.com/"
BUG_REPORT_URL="http://linuxmint-troubleshooting-guide.readthedocs.io/en/latest/"
PRIVACY_POLICY_URL="https://www.linuxmint.com/"
VERSION_CODENAME=una
UBUNTU_CODENAME=focal

on my nuc I have

NAME="Linux Mint"
VERSION="20.3 (Una)"
[...]

PR #3269, opened 2022-06-29
remoteit/tasks/install.yml: Remove "IIAB_LIKE=Ubuntu" after workaround, for Ansible OS detection

holta:
Resolves:

- #3268

I will test this to further verify.

Building on:

- PR #3201
- PR #3202


holta:
Ansible like remote.it are both very crude in using secret sauce to rather blindly scan wholesale for 'Ubuntu' instead of 'ubuntu' across the entire /etc/os-release (as their OS detection mechanism) but now we now.

Problem is now solved and tested.

Issue #3264, opened 2022-06-24
Networking refactoring & similar efforts


holta:
This is just a placemarker (pinned to the top of https://github.com/iiab/iiab/issues for now) to help draw more attention to PR's like:

- [x] #1536
  - [x] #2863
- [x] #3170
  - [x] PR #3281 
  - [x] PR #3299
  - [x] PR iiab/iiab-factory#228
- [x] PR #3173
  - [x] PR #3145
  - [x] #3286
  - [x] PR #3302
- [x] #3296
  - [x] PR #3312
  - [x] PR #3314
- [x] #3278
  - [x] #2585 
  - [x] PR #2645
  - [x] https://github.com/iiab/iiab/issues/3106#issuecomment-1140557739
  - [x] https://github.com/iiab/iiab/blob/46fad7936e3db81e05c8be9cc1d2874b9a9ca5ab/roles/network/tasks/netplan.yml#L62-L77
  - [x] PR #3314 addresses this + explains
- [x] PR #3308
  - [x] PR #3313
  - [x] PR #3318
  - [ ] An `/etc/profile.d/` version like /etc/local/sbin/netwarn for ssh sessions (across all OS's/distros/window managers) might also make sense in future?
- [x] # PR #3323
- [ ] #3309
- [ ] #3277
  - [x] PR #3272
  - [x] PR #3273
  - [x] PR #3274
- [x] #3293
- [x] PR #3361
- [ ] #3385 
  - [x] #3330
  - [x] PR #3390
- [x] PR #3392
  - [x] #3307
  - [x] #3393
  - [x] #3399
- [x] #3398
- [ ] #3403

holta:
@deldesir: everyone recognizes that you're more than busy with other things in Haiti :exploding_head: so ignore this as appropriate.

But *if* by some miracle you have a moment in July to review progress here with me and @jvonau and others (and/or offer suggestions) that'd be really great! :partying_face:

TOTAL ASIDE: Debian 12 Bookworm's initial freeze will be [2023-01-12](https://release.debian.org/testing/freeze_policy.html) so we might want to start testing IIAB installs on their [daily build](https://cdimage.debian.org/cdimage/daily-builds/daily/arch-latest/amd64/) later this year.

jvonau:
https://github.com/iiab/iiab/issues/3106#issuecomment-1140557739

jvonau:
[...]

PR #3173, opened 2022-04-07
Allow for network role to be fully optional

jvonau:
Alternate to #3145

PR #3272, opened 2022-06-29
Speed up roles/0-init by almost 50% (by moving 2 things to detected_network.yml) [and later remove them almost entirely]

holta:
This PR accelerates [roles/0-init](https://github.com/iiab/iiab/tree/master/roles/0-init) by almost 50% (~20 seconds instead of ~30 seconds) which is a real win to help new IIAB testers and old &mdash; to de-bureaucratize and focus.

Good enough for now.  More improvements will surely be possible over time &mdash; without sacrificing critical validation etc (the whole purpose of testing!)

Whether vars `internet_available` and `gw_active` are really needed at all (or whether they should evolve into something else, etc) can be resolved over time.  For now, they are moved from roles/0-init to [roles/network/tasks/detected_network.yml](https://github.com/iiab/iiab/blob/master/roles/network/tasks/detected_network.yml)

I want to get this merged quickly to help testers of all kind, who are tired of "hurry up and wait" workflows.

This PR is tested on Mint 20.

Please consider @jvonau's PR below which has many other good ideas to follow in July:

- PR #3173
- #3264

PS roles/1-prep/tasks/main.yml is also cleaned up, to communicate clearly when `/etc/iiab/iiab.env` is being populated.

Related:

- PR #2972


holta:
Testing a MEDIUM-sized IIAB install on Ubuntu Server 22.04 is underway (currently wrapping up Stage 7, looking good!)

jvonau:
So overall you lose 10 seconds by running the same code twice during iiab-install, spit on #2959 which would alert the user to the fact there is no internet to complete the install. Now in place of just recording there was no internet present to explain a potential failure of any other single role, there is no recorded feedback at all as the code moved to specific roles.

jvonau:
You broke the vnstat role [roles/vnstat/tasks/install.yml](https://github.com/iiab/iiab/blob/master/roles/vnstat/tasks/install.yml)

jvonau:
Don't toss more code on top, just back out of this idea.

holta:
We need to optimize for the **common case**:

- While I personally don't know anyone (ever) who's really depended on vnStat over the past decade, thanks for catching that, just it case.  There a strong argument that deadweight of this nature really needs to be removed from IIAB.  But we can evaluate that at some other time (silent users need to speak up, and demonstrate their need, if there are any!)
- What matters most is optimizing for the **common case** &mdash; which means things like roles/0-init used constantly by testers (I'll be talking to several potential new such people over coming months).  Keeping in mind that ~10 seconds makes a *massive* difference when you're waiting ~30 seconds every single time (now closer to 20 seconds) during rapid-testing cycles.  i.e. if just to state the obvious, an extra 10 seconds (temporary, in any case) during a 30 _minute_ install is completely irrelevant.
- Even though my profiling of roles/0-init was rudimentary earlier today, it quickly showed that setting both age-old and unnecessary networking variables were gobbling up far more time than I realized &mdash; removing them was indeed a big win.

jvonau:
> We need to optimize for the **common case**:
> 
>     * While I personally don't know anyone (ever) who's really depended on vnStat over the past decade, thanks for catching that, just it case.  There a strong argument that deadweight of this nature really needs to be removed from IIAB.  But we can evaluate that at some other time (silent users need to speak up, and demonstrate their need, if there are any!)
```
grep -r vnstat vars/
vars/local_vars_medical.yml:vnstat_install: True
vars/local_vars_medical.yml:vnstat_enabled: True
vars/local_vars_large.yml:# Network traffic monitor - from https://humdi.net/vnstat/
vars/local_vars_large.yml:vnstat_install: True
vars/local_vars_large.yml:vnstat_enabled: True
``` 
>     * What matters most is optimizing for the **common case** — which means things like roles/0-init used constantly by testers (I'll be talking to several potential new such people over coming months).  Keeping in mind that ~10 seconds makes a _massive_ difference when you're waiting ~30 seconds every single time (now closer to 20 seconds) during rapid-testing cycles.  i.e. if just to state the obvious, an extra 10 seconds (temporary, in any case) during a 30 _minute_ install is completely irrelevant.

Thought the common case is to catch the problem early like validate does, record the event to help explain why any role may fail and potentially do something other than run a role that is going to fail. Now with 'rapid-testing cycles' I think you are referring to using runrole repeatedly during testing or development.

> 
>     * Even though my profiling of roles/0-init was rudimentary earlier today, it quickly showed that setting both age-old and unnecessary networking variables was gobbling up far more time than I realized — removing them was indeed a big win.

The slowest part would be the 4 pings and the dns name lookup. This was part of the debugging routine to detect proxy issues that other users were having, if you can't get a heart beat then apt and friends are sure to fail in some cryptic fashion.

holta:
> with 'rapid-testing cycles' I think you are referring to using runrole repeatedly during testing or development.

Yes.  That's a big opportunity.

Possibly we can try to target getting roles/0-init under 10-15 seconds (for poor people too, eager to craft their own IIAB's out of junkier hardware) before end of year.

jvonau:
Other option prior to pulling out the detection out of 0-init would of been to make the detection optional on the iiab_stage > 9, new installs could run the detection and optionally react if something is funny while runrole is usually run with the install complete, iiab_stage at 9 when doing developmental work. The admin-console runs with stage being 9, so suppressed there and would cover runrole needing internet when debugging a single role that might of crashed during iiab-install. So where is the issue (how do we improve 0-init times) where this sort of stuff is discussed at?

tim-moody:
Internet detection was for getting packages and for roles that do downloads. Not sure if kiwix catalog is still downloaded. I guess Khan videos are not; not sure about language packs. cmdsrv has a function to detect internet that is pretty fast.

jvonau:
Almost all roles do downloads when installing, knowing if the internet was reachable when the roles didn't have the install,yml split required the detection. After the mass adoption of the install.yml layout within the roles the need for internet_available went away #2972 sans for 2 roles. Networking doesn't need the detection sans for the workaround for networkd-dispatcher and there are other ways to tell if the run is install mode but nobody asked what alternatives there might be, just dumped code into the network role. I know admin-console has it's own detection so it makes no sense to have the proxy detection within network, that is way too late for any kind of useful action or recording of the startup state that having the info in runtime section of the iiab.ini file provided.

I'll ask the question again where is the issue (how do we improve 0-init times) where this sort of stuff is discussed at? Given 3264 is specifically network related refactoring why no mention of this change there before it was implemented?

PR #3290, opened 2022-07-06
iiab apps to be installed

jvonau:
As iiab-apps-to-be-installed is a shared helper script, no real need for it it live in /usr/bin, as pointed out both parent scripts are resilient to missing commands. Always update parent scripts on the fly without a pass through 1-prep as these scripts are always in flux.

### DIFF
--- a/roles/0-init/tasks/main.yml
+++ b/roles/0-init/tasks/main.yml
@@ -32,11 +32,14 @@
 
 # Copies the latest/known version of iiab-diagnostics into /usr/bin (so it can
 # be run even if local source tree /opt/iiab/iiab is deleted to conserve disk).
-- name: Copy /opt/iiab/iiab/scripts/iiab-diagnostics to /usr/bin/
+- name: Copy iiab-summary & iiab-diagnostics from /opt/iiab/iiab/scripts/ to /usr/bin/
   copy:
-    src: "{{ iiab_dir }}/scripts/iiab-diagnostics"
+    src: "{{ iiab_dir }}/scripts/{{ item }}"
     dest: /usr/bin/
     mode: '0755'
+  with_items:
+    - iiab-summary
+    - iiab-diagnostics
 
 - name: Create globally-writable directory /etc/iiab/diag (0777) so non-root users can run 'iiab-diagnostics'
   file:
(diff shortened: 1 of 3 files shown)


holta:
_Apologies_ [iiab-diagnostics](https://github.com/iiab/iiab/blob/master/scripts/iiab-diagnostics) was being reworked at the same time as this (to include iiab-summary's output right on top of the pastebin that can be ~100X longer)

i.e. PR #3291 with its several iiab/iiab-factory companion PR's, to bring those all together (that's now done).

That said, the direction of this PR #3290 is on target, with the exception of `/usr/bin/iiab-apps-to-be-installed` (which should continue to be runnable on its own, at the command-line, as it's also very useful that way).

With all 3 programs substantially reworked now (iiab-apps-to-be-installed, iiab-summary, iiab-diagnostics) there will no doubt be other tidying up needed along the way (similar to PR 3290 here) when nettlesome UX-and-other corner cases arise.

jvonau:
Functions as a shared library when called within the other two and the path used is the same between the two files now. Still is run-able at the cmdline as `scripts/iiab-apps-to-be-installed`. Big gaps around the 'other requested files' now within iiab-diagnostics and the bold type doesn't look good when viewed online 
>    [1m"Merge pull request #3291 from holta/iiab_remote_url"[0m

holta:
> Big gaps around the 'other requested files'

It's true there are about ~5 lines of spacing around each of the 6 sections of iiab-diagnostics' output.
Definitely not perfect (some other visual spacer technique might very well be better).
Just as background this was done about 2 years ago, to try to create a strong visual separation between the 6 sections &mdash; for people browsing the ~2000 lines very quickly:

1. FILES SPECIALLY REQUESTED (FROM "iiab-diagnostics PATH/FILE1 PATH/FILE2")
2. REGULAR FILES
3. CONTENT OF DIRECTORIES (1-LEVEL DEEP)
4. OUTPUT OF COMMANDS
5. FIREWALL RULES
6. LOG FILES (e.g. LAST 100 LINES OF EACH)

> bold type doesn't look good when viewed online

Good point.

I'll work on cleaning that up.

Issue #3283, opened 2022-07-03
IIAB install froze at "pip install 7 packages into virtual environment: /opt/iiab/jupyterhub (~217 MB)" [on 64-bit RasPiOS — whereas Ubuntu 22.04 on x86_64 is not affected]


holta:
@georgejhunt would you have any recommendations to diagnose/mitigate ?

It's been completely stuck for more than 30 minutes now:

```
root@box:~# du -bsc /opt/iiab/jupyterhub/*
18904   /opt/iiab/jupyterhub/bin
16384   /opt/iiab/jupyterhub/etc
4096    /opt/iiab/jupyterhub/include
4793374 /opt/iiab/jupyterhub/lib
3       /opt/iiab/jupyterhub/lib64
69      /opt/iiab/jupyterhub/pyvenv.cfg
2277704 /opt/iiab/jupyterhub/share
7110534 total
```

iiab-summary (below) shows temperature is 75C, but everything else is running fine:

```
root@box:~# uptime
 19:04:12 up 56 min,  3 users,  load average: 1.07, 1.24, 1.22
```

```
root@box:~# iiab-summary
install_date = 2022-07-03T18:13:38Z   Current TZ: UTC

iiab: f2f4afb3, 19 PR's / 70 commits since tag 8.0-preview-2
    "Merge pull request #3282 from holta/iiab-summary_early_on"
    https://github.com/iiab/iiab   branch: master

iiab-admin-console: c6c6b9d, 43 PR's / 123 commits since tag v0.5.5.9
    "Merge pull request #505 from tim-moody/0.6.0.4-enh2"
    https://github.com/iiab/iiab-admin-console   branch: master

Raspberry Pi reference 2022-04-04
Generated using pi-gen, https://github.com/RPi-Distro/pi-gen, 27a8050c3c06e567c794620394a8c2d74262a516, stage2
/etc/debian-version: 11.3
display-manager? inactive   Arch1: arm64   Arch2: armhf
5.15.32-v8+ #1538 SMP PREEMPT Thu Mar 31 19:40:39 BST 2022 unknown
Cortex-A72   CPU(s): 4   Mem: 1849 Swap: 1023
Raspberry Pi 4 Model B Rev 1.4
Temperature(s): 74984

/opt/iiab/iiab/scripts/iiab-apps-to-be-installed: line 18: /tmp/iiab-apps-list: Permission denied
Filesystem      Size  Used Avail Use% Mounted on
[...]

PR #3294, opened 2022-07-07
break pip install into two plays

georgejhunt:
I did not do a completely new install of 64bit raspi OS.  But I deleted the jupyterhub tree, and it runs to completion

### DIFF
--- a/roles/jupyterhub/tasks/install.yml
+++ b/roles/jupyterhub/tasks/install.yml
@@ -48,11 +48,11 @@
 - name: "pip break apart jupyterlab install into two parts - must have conflicting mutual dependency (4 packages)"
   pip:
     name:
-        - jupyterlab
-        - jupyterhub_firstuseauthenticator
-        - jupyterhub-systemdspawner
-        - ipywidgets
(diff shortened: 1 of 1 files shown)


holta:
@georgejhunt

Can you clarify why you dropped `extra_args: "--no-cache-dir --pre"` in the 2nd (lower) Ansible paragraph ?

https://github.com/iiab/iiab/blob/cfe9a37b7ad4f2f4a4140b4f7b3254367504e63d/roles/jupyterhub/tasks/install.yml#L36-L61

Should that also be considered for removal in the 1st (upper) Ansible paragraph (if not now then in future?)

- #3283

georgejhunt:
I did not know what --pre or --no-cache did. So I thought I'd experiment
with removing. The "man" page fails to illuminate. I think you or Jerry
must have written the comment.

holta:
I think we **do** want `--no-cache-dir` to save space:

https://stackoverflow.com/questions/45594707/what-is-pips-no-cache-dir-good-for/45594808

Whereas I think we **do not** want `--pre` so as to avoid beta-like test releases of JupyterLab and JupyterHub — that we needed in the past, but no longer.

PR #3295, opened 2022-07-08
jupyterhub/tasks/install.yml: Use --no-cache-dir w/o --pre [AND stop patching firstuseauthenticator.py]

holta:
Implements:

- https://github.com/iiab/iiab/pull/3294#issuecomment-1178456164

Arising from:

- #3283


holta:
Tested on 64-bit RasPiOS Lite on RPi 4.

PR #3253, opened 2022-06-15
WIP: Begin integrating Matomo web analytics with IIAB

holta:
Huge thanks to @cwivagg whose work here I hope he can further enhance if he has time in the coming week or so!

DRAFT Instructions: https://github.com/holta/iiab/tree/matomo/roles/matomo#readme

(...don't yet explain that this is mainly about giving communities better tools to understand their own NGINX web server logs, but we can and should add a few field/operator tips soon!)

Background &mdash; Carl's work is (roughly) based on these 2018-2019 outlines:

- #1268 
- #1762

PR #3296, opened 2022-07-08
Remove named dhcpd

jvonau:
Builds on: #3173 Ref: #3264


holta:
I strongly support this.

If something like this is safe + tested it should be merged within the coming week.

Unless there's some reason to prioritize this instead:

- PR #3173

Also Matomo (PR #3253) and others should be merged well before the end of July — if all goes well!

jvonau:
You broke this PR on purpose, couldn't just leave named/dhcpd alone knowing they're not being used and I discussed removing the support during the call.

PR #3311, opened 2022-07-16
Download kiwix.apk ("redirect to latest") instead of kiwix-3.5.0.apk

holta:
Hoping that Kiwix keeps URL https://download.kiwix.org/release/kiwix-android/kiwix.apk updated in future.

As discussed on:

- PR #3310


holta:
CLARIF:

Most (but not all!) downloaders navigate the redirect induced by https://download.kiwix.org/release/kiwix-android/kiwix.apk and save it to (current, latest) filename:

```
kiwix-3.5.0.apk
```

tim-moody:
it worked before and it works again.

>Most (but not all!) downloaders navigate the redirect

what matters is whether the one used to install it works

PR #3312, opened 2022-07-16
Remove named dhcpd

jvonau:
Replacement for #3296 with [net_tweeks](https://github.com/iiab/iiab/compare/master...jvonau:net_tweeks?expand=1) #3264

PR #3313, opened 2022-07-16
Net warn

jvonau:
New features from #3308, bug fix for #3278. continuation of #3312 #3264

PR #3314, opened 2022-07-18
Remove named dhcpd + roles/firmware refinements + clean networkd-dispatcher for Mint

holta:
Building on @jvonau's:

- PR #3312


holta:
Tested on Ubuntu Server 22.04

holta:
@jvonau can you look over this PR quickly?

Prior to it being merged today if possible?

holta:
Tested on Ubuntu Server 22.04 VM: http://sprunge.us/1uZ8wI?en

jvonau:
Think this looks fine, that would just leave #3313 to hash through.

holta:
> Think this looks fine, that would just leave #3313 to hash through.

Thanks for weighing in.

This is a large PR with 54 files changes, but I agree it's time to move forward so broad community testing can begin across other HW/OS combinations, so let's do it.

PR #3316, opened 2022-07-19
roles/osm-vector-maps: Download three 25-48MB files from new repo iiab/maps-assets

holta:
Thanks to phone input from @georgejhunt earlier this morning, who recommended organizing/consolidating in this way.

This PR works with new repo: https://github.com/iiab/maps-assets

To try to reduce the git bloat on repo https://github.com/iiab/maps that should only contain code and smaller configuration/catalog files.

(To that end, the 26MB https://github.com/iiab/maps/blob/master/2020/cities1000.sqlite will be deleted in coming days, after this transition is complete.)

### DIFF
--- a/roles/osm-vector-maps/defaults/main.yml
+++ b/roles/osm-vector-maps/defaults/main.yml
@@ -8,18 +8,20 @@
 # If nec, change them by editing /etc/iiab/local_vars.yml prior to installing!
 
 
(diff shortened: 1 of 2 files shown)


holta:
Tested on Ubuntu Server 22.04 VM.  Works well.

PR #3318, opened 2022-07-21
/usr/local/sbin/netwarn pop-ups (if iiab-network needs to be run) on MATE & RasPiOS with desktop + roles/network refactoring

holta:
1. Building on @jvonau's PR #3313 with network/tasks/install.yml syntax error fixed, cleaner pop-ups, ~and an in-line explanation for `#!/bin/sh` on the top line of scripts/netwarn.sh (despite the fact this 1st line is ignored on certain OS's!)~

2. Boot pop-ups tested on 'Raspberry Pi OS with desktop' on RPi 4.

3. I learned that `./iiab-network` is sometimes no longer enough &mdash; in such QA-like scenarios, consider running:

   ```
   ./runrole --reinstall network
   ```

4. Adding the same pop-ups to Mint and stock Ubuntu should hopefully now be extremely easy &mdash; in [roles/network/tasks/netwarn.yml](https://github.com/holta/iiab/blob/net_warn/roles/network/tasks/netwarn.yml)

5. _What testing or refinements are needed before merging this PR?_

Related:

- PR #2609
- PR #3308
- PR #3313

Issue #3303, opened 2022-07-11
Call for testing: IIAB pre-installed on ubuntu-mate-22.04-desktop-arm64+raspi


jvonau:
2 New images for evaluation based on ubuntu-mate-22.04-desktop-arm64+raspi.img
[base](http://104.193.111.252/local_content/uploads/iiab-base-0711-ubuntu-mate-22.04-desktop-arm64+raspi.img.zip)
[small](http://104.193.111.252/local_content/uploads/iiab-min-0711-ubuntu-mate-22.04-desktop-arm64+raspi.img.zip)

Exact same user feel as the stock ubuntu-mate images, geared for Desktop therefore use of a monitor/keyboard/mouse is highly recommended. Tested on RPi-400 without connecting to the internet yet. Feedback welcome under this issue, don't file new issues, keep it here.

The image shipped with /etc/iiab/local_vars.yml renamed to local_vars.yml.builder to encourage a review before using, can/should be copied local_vars.yml. Also present is the chroot build script /etc/iiab/builder.sh employed by [mk-image](https://github.com/jvonau/mk-image) as (mate-)runme.sh

https://github.com/iiab/iiab-factory/pull/231

jvonau:
First Issue: If the screensaver engages your user password will not unlock the session, hard power cycle. Disable the power management and screensaver before they cut in.

Should somebody boot the stock upstream image please check for this behavior since I never booted a stock version of the ubuntu-mate-22.04-desktop-arm64+raspi.img source image as my build system is automated. This is most likely an upstream issue that was inherited.

jvonau:
Disable the setting here:
![Screenshot at 2022-07-11 11-18-20](https://user-images.githubusercontent.com/5579710/178319181-2b5eb8d9-9c70-4d33-a665-06dad6113d88.png)

Looks like an upstream issue and might be resolved when connected to the internet for updates
![Screenshot at 2022-07-11 11-35-21](https://user-images.githubusercontent.com/5579710/178319183-6ad7ae32-96cb-48dd-9640-132a72410d29.png)

jvonau:
Taken after providing internet access via tethered cell phone prior to reboot, seems stable with the screensaver turned off for now. http://sprunge.us/NXQi57

holta:
I've guessing it might be a bad microSD card (maybe, in any case overnight it froze [still hanging now] while writing to the card, 32% of the way through).  So I need to try again flashing the 3GB image in another way.

holta:
[...]

PR #3319, opened 2022-07-23
Set default for use in netplan.yml and wording

jvonau:
### Fixes bug:
Default to False or perhaps move the `when: ansible_local.local_facts.systemd_networkd == "enabled" or ansible_local.local_facts.systemd_networkd == "enabled-runtime"` to replace `when: systemd_networkd_active`

### DIFF
--- a/roles/network/tasks/install.yml
+++ b/roles/network/tasks/install.yml
@@ -3,7 +3,7 @@
 - name: Install dnsmasq -- configure LATER in 'network', after Stage 9
   include_tasks: roles/network/tasks/dnsmasq.yml    # Invoked by 1-prep (so full path needed)
 
-- name: Install package networkd-dispatcher (OS's other than RasPiOS and LinuxMint)
+- name: Install package networkd-dispatcher (OS's other than RasPiOS and Linux Mint)
   package:
     name: networkd-dispatcher    # 15kB download: Dispatcher service for systemd-networkd connection status changes
     state: present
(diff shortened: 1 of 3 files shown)


jvonau:
For the record the reason the 2 conditional were 2 parts in the past within detected_network.yml was to tell which one applied while debugging.

holta:
Is this line no longer sufficient or no longer useful?

https://github.com/iiab/iiab/blob/b73567260afa6aafde1705baacaf9346e4e117a6/roles/network/defaults/main.yml#L86

Related:

- PR #3318

jvonau:
Not sufficient when called via 1-Prep and the service is not enabled.

jvonau:
The only fix noted when clean installing on Ubuntu desktops, in this case Mate, other Ubuntu variants should be the same as they all use NetworkManager as the backend for use with netplan and ship with systemd-networkd present but disabled.

jvonau:
For the record `./runrole --reinstall network` would not be enough to uncover this one as we enable and use systemd-networkd for br0 but out of the box all desktops would have this service disabled, thus I found this one building new images for #3303

jvonau:
Note: As of today 92 packages should be upgraded post install on the upstream Mate-arm64 image while my images would have these updates baked into the image.

holta:
Is this PR tested + safe?

jvonau:
Just build image with this commit in place.. or the builder would puke..

jvonau:
If that is the only change ship it..

PR #3222, opened 2022-05-19
Check hardware for AP and if so record 'can_be_ap = True' to /etc/iiab/iiab.ini

holta:
Tested on 5 different hardware scenarios.

Refines the following PR, which also needed a couple Ansible fixes:

- PR #3070

PR #3323, opened 2022-07-28
network/tasks/detected_network.yml: can_be_ap grep failed on absence of regex

holta:
Emergency patch to help those whose laptops don't even mention AP at all in the output of `iw list`

Building on:

- #3057
- PR #3070 
- PR #3222

### DIFF
--- a/roles/network/tasks/detected_network.yml
+++ b/roles/network/tasks/detected_network.yml
@@ -113,6 +113,7 @@
   shell: iw list | grep '^[[:space:]]*\* AP$'
   register: look_for_ap
   when: discovered_wireless_iface != "none"
+  failed_when: False    # Hides red errors (stronger than 'ignore_errors: yes') -- otherwise Ansible will fail if grep returns '1' on absence of regex!
 
 - name: Set can_be_ap if 'iw list' output contains suitable '* AP'
   set_fact:


holta:
Quickly tested and Ed in the UK will now further test after merge.

holta:
1) Thank you to Ed from Zimbabwe (@EMG70) for surfacing this bug!

2) There are related issues he'll be posting about on another ticket &mdash; as he tries to get IIAB working on his Dell Inspiron 1750 laptop &mdash; trying to get any kind of WiFi learning hotspot working (internal WiFi ideally, but USB WiFi dongle if necessary!?)

jvonau:
Now was iw preinstalled by the os when the first failure occurred? 
#3070 f55f4f3f63ffb48a21a596378bd3b078b3c1678f
4d5143e2de59c3926350e1f48cf9a938395f7506

holta:
> was iw preinstalled by the os when the first failure occurred?

Not by the OS per se, but I believe that yes `iw` was already pre-installed &mdash; by `roles/1-prep` here:

https://github.com/iiab/iiab/blob/2efd7c01982ccecfb1297ae2bc93087172b38c2a/roles/1-prep/tasks/main.yml#L38-L40

Which ran:

https://github.com/iiab/iiab/blob/2efd7c01982ccecfb1297ae2bc93087172b38c2a/roles/network/tasks/install.yml#L22-L38

holta:
Please move the discussion to his actual ticket if possible?

- #3324

jvonau:
> > was iw preinstalled by the os when the first failure occurred?
> 
> Not by the OS per se, but I believe that yes `iw` was already pre-installed — by `roles/1-prep` here:
> 

That doesn't answer the question of why the stanza failed on the first pass.. Then at some point recorded can_be_ap as True

holta:
> why the stanza failed on the first pass

Recap, of what's explained in this PR itself:

Our Ansible logic was just wrong (prior to this PR).

As it failed when attempted on any laptop (machine) like the one @EMG70 is attempting to get working with IIAB &mdash; e.g. when 'AP' (or anything like it) is not contained anywhere in the output of `iw list`

In short, the logic that we had in place was just wrong, failing to account for the above kind of situation &mdash; when the shell command fails &mdash; as a result of grep return code of 1.

As mentioned earlier today: this failure occurs regardless of the whatever regex grep is scanning for.  Specifically: when the regular expression (whatever regex) is not found in the output of `iw list`

jvonau:
That is why there was ' | wc -l' which was vetted against the live test of subject #3057 and later changed with f55f4f3f63ffb48a21a596378bd3b078b3c1678f but was not cherry-picked into the tree then 4d5143e2de59c3926350e1f48cf9a938395f7506 that I suspect was not tested against non-AP capable hardware... back over to you to fix..

holta:
Indeed the earlier regular expression had serious errors (counting the wrong kinds of lines until about May, which is now cleaned up).

I agree a simple simple existence test is what's really needed (no counting, no fancy corner cases, just a Dead Simple yes-or-no presence test which things like `grep -q` were designed for).

So the code is not the most elegant as is &mdash; but so be it at least it's finally fixed yay.

_Thanks go to @EMG70 for uncovering this._

jvonau:
> Indeed the earlier regular expression had serious errors (counting the wrong kinds of lines until about May, which is now cleaned up).
> 

You have never shown any supporting evidence or issues reported to support that claim.

holta:
Erroneous results were in fact extremely common, caused be strings that happen to contain the letters 'AP' within another word.

As documented here: https://github.com/iiab/iiab/blob/d99fcaddd6956391f2684c110f834ae5e152c431/roles/network/tasks/detected_network.yml#L112

This is the reason the regex was much improved around May (of this year).

(Apparently EAP = Extensible Authentication Protocol)

jvonau:
In the face of the lack of presented evidence that statement is hearsay and nothing more.

PR #3329, opened 2022-08-01
Live w/ red errors to set can_be_ap ('failed_when: False' was too strong) + WiFi firmware questions w/ Linux Mint (Broadcom or Realtek)

holta:
In short, this PR was overkill:

- PR #3323

As confirmed by @EMG70's:

- https://github.com/iiab/iiab/issues/3324#issuecomment-1200546459

This Ansible stanza regrettably introduces many different code paths for what should really be a dead simple `grep -q` existence test.

In any case, they're now fixed and tested, and we live with Ansible's artificially-added complexity in this case.

PR #3331, opened 2022-08-01
Explain iiab-network alert: "WiFi chipset/firmware NOT CAPABLE of AP Mode (details above)"

holta:
Let's give the implementer/operator more context (with 2 intentionally very bright red messages) that their WiFi chipset/firmware is not capable of AP Mode.

We can (and should!) add a more helpful + actionable explanation in future, with a link to site(s) like:

- http://FAQ.IIAB.IO
- https://github.com/iiab/iiab/wiki/IIAB-Networking#common-customizations

Related:

- #3057
- PR #3070 
- PR #3222
- PR #3318
- PR #3323
- #3324
- PR #3329

### DIFF
--- a/roles/network/tasks/detected_network.yml
+++ b/roles/network/tasks/detected_network.yml
@@ -117,10 +117,10 @@
     # failed_when: False    # Hides red errors and is too strong (renders useless the look_for_ap.failed test below!)
 
   rescue:    # Force another red error msg (to explain) then proceed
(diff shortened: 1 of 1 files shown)


holta:
This can and will be refined with cleaner + more actionable messaging in future.

To give implementers/operators the very best suggestions we can &mdash; e.g. what are their best options if indeed they want to set up an IIAB learning hotspot.

Issue #815, opened 2018-05-24
Future advanced networking improvements


m-anish:
I'm creating this as a placeholder for possible future networking upgrades to the IIAB.

1. In large IIAB setups, QoS is increasingly an issue - one user watching a video eats up all the available wifi bandwidth. Something like radius can limit this problem. Another option can be packet filtering/shaping as described in the cookbook here -- http://lartc.org/howto/lartc.cookbook.ultimate-tc.html

2. There are many instances where upstream network connectivity is spotty (I don't need to give examples I think), there might be use cases where having more than one upstream connection can really improve conditions. Thankfully, multipath-tcp (http://multipath-tcp.org/) has been making some good strides on this front. To have a user experience that is transparent to the user and provides better and more reliable upstream connectivity, one would need to have multipath-tcp run on the IIAB (seems trivial to do on a recent debian/ubuntu machine), and have a multipath-tcp proxy hosted somewhere. The downside is that additional network bandwidth on the proxy/bridge end will be consumed -- but if some university/group is willing to support this, it can really improve things. We could additionally also choose for what kind of packets (data) to enable this for.

3. Multipath-tcp could potentially be useful on the LAN side as well, if there is IIAB and multiple routers involved. It could smartly route traffic between user-user and user-server. However, at this point, I am not very sure if it offers any tangible improvements over layer 2/3 meshing protocols such as batman-adv et. al.

holta:
@m-anish hope it's Ok I'm moving this to [IIAB 6.7 Milestone](https://github.com/iiab/iiab/milestone/4) as our [IIAB 6.6](http://download.iiab.io/6.6/) final release is now approaching?

m-anish:
Cool. I hope to start experimenting on this soon.

jvonau:
Update: advanced options now in play
'second_gateway_found' function will exclude the associated interface from the available lan device string parsing by setting 'exclude_devices' which is also available to be used within local_vars.yml to tune excludes for a custom network layout.
'iiab_wired_lan_iface' is computed but available as an override in local_vars for use as slave under br0

Since I'm in a writing mood other options available but less known
'iiab_lan_enabled: False' disables creation of iiab manged dhcp network
'nobridge: True' turns off auto br0 creation.
'user_lan_iface: device' would trump and override iiab_lan_iface detection for iiab's dhcp clients.
'user_wan_iface' would trump and override iaib_wan_detection.

tim-moody:
I take it these have precedence over everything else, for example

[...]

PR #3334, opened 2022-08-04
detected_network.yml: Disallow multi-line var second_gateway_found ~= exclude_devices (to avoid polluting /etc/iiab/iiab.ini)

holta:
To keep `/etc/iiab/iiab.ini` from being polluted with multi-line variable artifacts like the following...

```
exclude_devices = none
wlp12s0
num_lan_interfaces = 1
```

...which prevented @EMG70 from running Admin Console earlier today, when he had 3 gateways (no doubt without realizing it) apparently including (1) his laptop's Ethernet (2) his laptop's internal WiFi (3) a USB WiFi dongle.  As confirmed by his Dell laptop's iiab-network.log (Line 9374) that is broken into 2 lines here:

```
2022-08-01 07:22:05,597 p=3352 u=root n=ansible | TASK [network : Set exclude_devices if default gateway has been detected for wlx1ca7705e063c
wlp12s0] ***
```

So this PR should help intercept problematic network attempts of this sort, e.g. when IIAB is being pushed too far into complex networking scenarios we're not ready to support at this time!

A bit of background:

- https://github.com/iiab/iiab/issues/815#issuecomment-884497515
- #3324
- https://github.com/iiab/iiab/pull/3329#issuecomment-1201162096
- https://github.com/iiab/iiab/pull/3329#issuecomment-1201163944

### DIFF
--- a/roles/network/tasks/detected_network.yml
+++ b/roles/network/tasks/detected_network.yml
@@ -142,7 +142,7 @@
   register: second_gateway_found
   changed_when: False
 
-- name: Fail intentionally and explain, if multiple secondary gateways are detected
+- name: If multiple secondary gateways are detected, fail intentionally and explain
   fail:
     msg: "IIAB currently DOES NOT SUPPORT multiple secondary gateways: {{ second_gateway_found.stdout }}"
   when: second_gateway_found.stdout_lines is defined and second_gateway_found.stdout_lines | length > 1
--- a/roles/network/tasks/detected_network.yml
+++ b/roles/network/tasks/detected_network.yml
@@ -142,6 +142,10 @@
   register: second_gateway_found
   changed_when: False
 
+- assert:
+    that: second_gateway_found.stdout_lines | length == 1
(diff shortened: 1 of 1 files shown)


holta:
It took hours to get to the bottom of this issue (thank you to @shanti-bhardwa, @jvonau and @tim-moody) but now the root cause is understood.

So this PR should reduce the agony, when this happens to the next person &mdash; who inevitably likewise adds a USB WiFi dongle(s) when their laptop already has an Ethernet and WiFi connection to their home router etc &mdash; and then runs `iiab-network`

holta:
Good enough.

Can be further refined later if @jvonau / others want to refactor / enhance.

jvonau:
Adding a usb-dongle is not the problem, using the same dongle at an upstream connection before running iiab-network is the problem as there would be not a device available to assign to for use with hostapd. This leave the new device unused within the GUI needs to be better documented, for better user education.

PR #2522, opened 2020-09-15
strict ordering of resize service

jvonau:
Strict ordering to preform the resizing while mounted read only.
>pi@box:~ $ sudo systemctl status fake-hwclock.service 
● fake-hwclock.service - Restore / save the current clock
   Loaded: loaded (/lib/systemd/system/fake-hwclock.service; enabled; vendor preset: enabled
   Active: active (exited) since Tue 2020-09-15 10:07:37 BST; 15min ago
     Docs: man:fake-hwclock(8)
  Process: 124 ExecStart=/sbin/fake-hwclock load $FORCE (code=exited, status=0/SUCCESS)
 Main PID: 124 (code=exited, status=0/SUCCESS)

124
> pi@box:~ $ sudo systemctl status iiab-rpi-root-resize.service 
● iiab-rpi-root-resize.service - Root Filesystem Auto-Resizer
   Loaded: loaded (/etc/systemd/system/iiab-rpi-root-resize.service; enabled; vendor preset:
   Active: active (exited) since Tue 2020-09-15 10:07:37 BST; 16min ago
  Process: 136 ExecStart=/usr/sbin/iiab-rpi-max-rootfs.sh (code=exited, status=0/SUCCESS)
 Main PID: 136 (code=exited, status=0/SUCCESS)
Sep 15 10:07:37 box.lan iiab-rpi-max-rootfs.sh[136]: Checking for .resize-rootfs
Sep 15 10:07:37 box.lan iiab-rpi-max-rootfs.sh[136]: + echo 'Checking for .resize-rootfs'
Sep 15 10:07:37 box.lan iiab-rpi-max-rootfs.sh[136]: + '[' -f /.resize-rootfs ']'
Sep 15 10:07:37 box.lan iiab-rpi-max-rootfs.sh[136]: + exit 0

136
> pi@box:~ $ sudo systemctl status systemd-fsck-root.service 
● systemd-fsck-root.service - File System Check on Root Device
   Loaded: loaded (/lib/systemd/system/systemd-fsck-root.service; enabled-runtime; vendor pr
   Active: active (exited) since Tue 2020-09-15 10:07:37 BST; 17min ago
     Docs: man:systemd-fsck-root.service(8)
  Process: 137 ExecStart=/lib/systemd/systemd-fsck (code=exited, status=0/SUCCESS)
 Main PID: 137 (code=exited, status=0/SUCCESS)
[...]

PR #3336, opened 2022-08-07
Revert "1-prep/templates/iiab-expand-rootfs: Avoid reboot"

holta:
Reverts PR #3160

Might be superseded in future by something like:

- PR #2522

Related:

- #723
- #2517
- PR #3137
- #3147
- #3325

### DIFF
--- a/roles/1-prep/templates/iiab-expand-rootfs
+++ b/roles/1-prep/templates/iiab-expand-rootfs
@@ -14,8 +14,8 @@ if [ -f /.expand-rootfs ] || [ -f /.resize-rootfs ]; then
     if [ -x /usr/bin/raspi-config ]; then    # Raspberry Pi OS
         # 2022-02-17: Uses do_expand_rootfs() from:
         # https://github.com/RPi-Distro/raspi-config/blob/master/raspi-config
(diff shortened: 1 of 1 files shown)


holta:
See also @jvonau's 2022-04-02 summary:

> A reboot is not that bad and is what RaspiOS does when booting a 'fresh' image out of the box, just a change in behavior when using the builtin tools. Using growpart is way more flexible, and is the standard on Ubuntu based arm64 images. Now the question becomes should growpart be installed on all installations[*] to take advantage of the revised root device detection that should work on all `/dev/sd*` root filesystems? [PR #3137] That would be a bonus if cloning an intel based machine could be achieved with the tools within iiab-factory.

- https://github.com/iiab/iiab/issues/3159#issuecomment-1086679575

[*] In fact [growpart](https://manpages.ubuntu.com/manpages/bionic/man1/growpart.1.html) is installed as part of all IIAB's since 2022-03-15:

https://github.com/iiab/iiab/blob/4d41c1233804ef058198d242627ee94e8ec927b3/roles/1-prep/tasks/install-expand-rootfs.yml#L5

PR #3235, opened 2022-06-06
kiwix-tools 3.3.0 is coming very soon (with "opensearch")

holta:
Opensearch has apparently been included:

- PR kiwix/libkiwix#731

Testing the latest nightly version of kiwix-tools may very be helpful here &mdash; feedback very much appreciated from anybody who sees improvements (or problems!) with title search and full-text search right here:

- http://iiab.me/kiwix/wikipedia_en_all_maxi_2022-05/

(Currently the above site is running [kiwix-tools_linux-x86_64-2022-06-06.tar.gz](https://download.kiwix.org/nightly/2022-06-06/kiwix-tools_linux-x86_64-2022-06-06.tar.gz))

All this builds on:

- kiwix/kiwix-tools#554
  - https://github.com/kiwix/kiwix-tools/milestone/2
- PR #3221
- https://twitter.com/internet_in_box/status/1526948631135264770

Note this will be included as part of IIAB 8.0 Preview 2 shortly:

- https://twitter.com/internet_in_box/status/1533815764683194371

PR #3344, opened 2022-08-17
kiwix/defaults/main.yml: New kiwix-tools version 3.3.0-1

holta:
Building on:

- PR #3235


holta:
Tested on Raspberry Pi OS.

Might or might not be related to this intermittent bug with kiwix-tools's search drop-down:

- kiwix/libkiwix#809

kelson42:
This PR contains binary code/files. This should never happen. Please remove then and rewrite the git log before changing anything else. It's important this binary files never make it to the `master` branch (even in the git log).

PR #3304, opened 2022-07-12
Cwivagg/matomo 20220618

cwivagg:
### Matomo Role

Instructions: https://github.com/cwivagg/iiab/tree/cwivagg/matomo_20220618/roles/matomo#readme [revised from June]

Updates Adam's pull request https://github.com/iiab/iiab/pull/3253.

Context links: 

1. https://github.com/iiab/iiab/issues/1268

2. https://github.com/iiab/iiab/issues/1762

### Testing instructions

```
curl iiab.io/fast.txt | sudo bash -s 3304
cd /opt/iiab/iiab
sudo ./runrole matomo
```

PR #3354, opened 2022-08-31
Cwivagg timezone patch [Matomo]

cwivagg:
1. Fixes the bug by providing a fallback timezone, which is the original source of the problem.
2. Also fixes a problem where Matomo is having trouble accessing its own directory. I think this may be related to the Ansible updates after my commit and will root cause it at the same time as I am retroing how the above bug got past my testing.

Ubuntu 22 VirtualBox and Multipass Ubuntu 22. Per offline conversation with @holta , as long as Ansible is consistent across systems, this change should be fine. We capture both branches of the forking logic with these two systems (VirtualBox for the status code 200 case and Multipass for the status code 302 case).

@holta


holta:
1) When should this be merged?  If it's smoke-tested & you trust it, that's sufficient!

2) Out of curiosity why is `"{{ apache_user }}"` hard-coded to `www-data` ?

Related:

- PR #3304

cwivagg:
I've removed the www-data change. There's something weird going on with the directory ownership, but I'll need to look at this a little bit more. It's not related to the main bug fix, and I shouldn't have included it.

Issue #3343, opened 2022-08-17
Test IIAB on Ubuntu 22.10 "Kinetic Kudu" [with ppa:learningequality/kolibri-proposed] [keyring /etc/apt/trusted.gpg DEPRECATED]


holta:
The OS should be released 2022-10-20.

Thanks @pickypet if you & others can help here in coming weeks!

Somewhat related:

- #3182
- #3190
- #3289
- #3307
- #3345

FYI Ubuntu 22.10 should contain PHP 8.1, PostgreSQL 14, and Python 3.10:

https://github.com/iiab/iiab/blob/master/vars/ubuntu-2210.yml

Coincidentally this is identical to [Ubuntu 22.04](https://github.com/iiab/iiab/blob/master/vars/ubuntu-2204.yml) so very few problems are expected!

AISDE: Python 3.11 itself (expected to be quite a bit faster) should be released 2022-10-03.

holta:
To help @pickypete & others who want to help with **IIAB quality**, here's my recommendation for rapid testing on any Ubuntu machine:

```
multipass launch devel --bridged -n MY_VM -c 2 -m 2G -d 25G
```

[1] instead of `devel` you can equivalently write `kinetic` or `daily:22.10`
[2] skip `--bridged` (equivalent to `--network enp1s0` or similar) if your Ubuntu machine was not earlier configured with `multipass set local.driver=lxd` (which allows you to access to VM's IP address from anywhere in your home, e.g. using 192.168.0.158 as you can see below)

Then (optionally) if you want details of your new VM, run `multipass info MY_VM` to see:

```
Name:           MY_VM
State:          Running
IPv4:           10.88.67.47
                192.168.0.158
Release:        Ubuntu Kinetic Kudu (development branch)
Image hash:     29bdfe284056 (Ubuntu 22.10)
CPU(s):         2
Load:           0.07 0.02 0.00
Disk usage:     2.3GiB out of 24.8GiB
Memory usage:   187.1MiB out of 1.9GiB
Mounts:         --
```

Then open a shell (command-line) on the VM:

```
multipass shell MY_VM
```

Finally, install IIAB within the Ubuntu 22.10 (pre-release) VM:

```
curl iiab.io/fast.txt | sudo bash
```

holta:
WARNING!  @pickypet (& all) please remember that pre-release OS's (like Ubuntu 22.10) almost always need you to first run the following (in the VM of course!) prior to starting IIAB's install:

```
[...]

PR #3327, opened 2022-07-30
Install Kolibri using PPA, which promises auto-upgrades (when you apt update?)

holta:
1) This PR also lists 31 (not just 25) Kolibri languages for variable `kolibri_language` within those [local_vars.yml](https://wiki.iiab.io/go/FAQ#What_is_local_vars.yml_and_how_do_I_customize_it?) files that are designed to be human-readable documentation.  Building on;

   - PR #2826

2) And preserves Kolibri's legacy [.deb install method](https://kolibri.readthedocs.io/en/latest/install/ubuntu-debian.html#install-from-a-deb-file) for both testing and emergency override purposes (if you manually specify `kolibri_deb_url` e.g. in roles/kolibri/defaults/main.yml) as discussed here:

   - #3217

3) Unfortunately different PPA logic was necessitated on Debian-like OS's (`codename: focal`) &mdash; as compared to Ubuntu-like OS's.  But both should work.  As outlined in official guidelines: https://kolibri.readthedocs.io/en/latest/install/ubuntu-debian.html#install-from-ppa-repository

4) This PR was tested on 64-bit RasPiOS Lite on Raspberry Pi 4.

Tangentially related is Kolibri 0.15.6's recent fix of peer-importing of channels e.g. from nearby IIAB machines with URL's like http://box/kolibri :

- #3227

PR #3356, opened 2022-09-01
kolibri/tasks/install.yml: Modern apt keyring approach [& apt upgrading Kolibri very risky!]

holta:
Based on @jredrejo recommendations here:

- https://github.com/iiab/iiab/issues/3343#issuecomment-1233163573

Test installed on a Ubuntu 22.10 pre-release.

Building on:

- PR #3327
- PR #3347

### DIFF
--- a/roles/kolibri/tasks/install.yml
+++ b/roles/kolibri/tasks/install.yml
@@ -45,8 +45,8 @@
   # "When you use the PPA installation method, upgrades to newer versions
   # will be automatic, provided there is internet access available."
   #
-  # IN REALITY: Kolibri upgrades are still tricky, as 3 pink screens prompt you
-  # and PEOPLE WON'T KNOW TO TYPE IN things like Linux username 'kolibri' :/
+  # IN REALITY: apt upgrading Kolibri is risky, as 3 pink/blue screens prompt
+  # PPL WHO DON'T KNOW TO TYPE IN things like Linux username 'kolibri' :/ #3356
 
   # 2022-08-31: keyring /etc/apt/trusted.gpg DEPRECATED as detailed on #3343
   - name: Download Kolibri's apt key to /usr/share/keyrings/learningequality-kolibri.gpg
--- a/roles/kolibri/tasks/install.yml
+++ b/roles/kolibri/tasks/install.yml
@@ -44,11 +44,30 @@
   # https://kolibri.readthedocs.io/en/latest/install/ubuntu-debian.html says:
   # "When you use the PPA installation method, upgrades to newer versions
   # will be automatic, provided there is internet access available."
+  #
+  # IN REALITY: Kolibri upgrades are still tricky, as 3 pink screens prompt you
+  # and PEOPLE WON'T KNOW TO TYPE IN things like Linux username 'kolibri' :/
 
-  - name: Add Kolibri PPA repo 'ppa:learningequality/kolibri' (if is_ubuntu and not is_linuxmint)
+  # 2022-08-31: keyring /etc/apt/trusted.gpg DEPRECATED as detailed on #3343
+  - name: Download Kolibri's apt key to /usr/share/keyrings/learningequality-kolibri.gpg
+    shell: |
+      gpg --keyserver hkp://keyserver.ubuntu.com:80 --recv-keys DC5BAA93F9E4AE4F0411F97C74F88ADB3194DD81
+      gpg --yes --output /usr/share/keyrings/learningequality-kolibri.gpg --export DC5BAA93F9E4AE4F0411F97C74F88ADB3194DD81
+
+  - name: Add signed Kolibri PPA 'jammy' (if Ubuntu 22.04 or 22.10 or Mint 21 or Debian 12)
     apt_repository:
-      repo: ppa:learningequality/kolibri
-    when: is_ubuntu and not is_linuxmint
+      repo: "deb [signed-by=/usr/share/keyrings/learningequality-kolibri.gpg] http://ppa.launchpad.net/learningequality/kolibri/ubuntu jammy main"
+    when: is_ubuntu_2204 or is_ubuntu_2210 or is_debian_12    # MINT 21 COVERED BY is_ubuntu_2204
+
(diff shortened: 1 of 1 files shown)


holta:
> Test installed on a Ubuntu 22.10 pre-release.

The apt configuration file auto-generated by Ansible...has a very long filename here: (but that's Ok!)

```
root@box:~# cat /etc/apt/sources.list.d/ppa_launchpad_net_learningequality_kolibri_ubuntu.list
deb [signed-by=/usr/share/keyrings/learningequality-kolibri.gpg] http://ppa.launchpad.net/learningequality/kolibri/ubuntu jammy main
```

Just FYI this test install of Kolibri on IIAB was done on the latest daily build of Ubuntu 22.10 [pre-release] "Kinetic Kudu".

holta:
FYI here are some of the ~3 pink/blue screens that arise if one `apt` upgrades from Kolibri 0.15.6 to 0.15.7:

_(unfortunately these will lead to quite serious problems, as everyday IIAB operators will accidentally override pre-existing Kolibri systemd unit file & Linux username, something to think about!!)_

```
  ┌────────────────────────┤ Kolibri configuration ├────────────────────────┐
  │                                                                         │
  │ This will automatically call the system service to start Kolibri when   │
  │ your system boots.                                                      │
  │                                                                         │
  │ Do you wish to run Kolibri on start-up?                                 │
  │                                                                         │
  │                    <Yes>                       <No>                     │
  │                                                                         │
  └─────────────────────────────────────────────────────────────────────────┘
```

If one overrides the above default `<Yes>` to correctly select `<No>` (to preserve [IIAB's systemd file for Kolibri](https://github.com/iiab/iiab/blob/master/roles/kolibri/templates/kolibri.service.j2)) the next screen is:

```
  ┌────────────────────────┤ Kolibri configuration ├────────────────────────┐
  │                                                                         │
  │ Running Kolibri manually:                                               │
  │                                                                         │
  │ You have chosen not to run Kolibri as a system service and that means   │
  │ you should run it manually. Starting Kolibri with your current user     │
  │ account:                                                                │
  │                                                                         │
  │   kolibri start                                                         │
  │                                                                         │
  │                                                                         │
  │ Starting Kolibri as a system-wide service:                              │
  │                                                                         │
  │   sudo service kolibri start                                            │
  │                                                                         │
  │                                                                         │
  │ Start Kolibri when the system boots:                                    │
  │                                                                         │
  │   sudo systemctl enable kolibri                                         │
  │                                                                         │
  │                                 <Ok>                                    │
  │                                                                         │
  └─────────────────────────────────────────────────────────────────────────┘
```

Finally this 3rd screen is generally pre-populated with an incorrect username: (IIAB mandates Linux username `kolibri` not `iiab-admin`, Ouch !)

```
  ┌────────────────────────┤ Kolibri configuration ├────────────────────────┐
  │ The default is to choose your preferred desktop user account, for       │
  │ instance to ensure access to importing data from external USB devices.  │
  │                                                                         │
  │ Entering a username that doesn't exist will create a new system user    │
  │ with home directory /var/<username>.                                    │
  │                                                                         │
  │ Which user account should own the Kolibri server?                       │
  │                                                                         │
  │ iiab-admin_____________________________________________________________ │
  │                                                                         │
  │                                 <Ok>                                    │
  │                                                                         │
  └─────────────────────────────────────────────────────────────────────────┘
```

I don't have answers; am simply posting the above (Linux Terminal) screenshots, so we can all think about it :thinking:

holta:
> If one overrides the above default `<Yes>` to correctly select `<No>` (to preserve [IIAB's systemd file for Kolibri](https://github.com/iiab/iiab/blob/master/roles/kolibri/templates/kolibri.service.j2)) the next screen is:

Strangely the above Kolibri prompt neither disables nor clobbers [`/etc/systemd/system/kolibri.service`](https://github.com/iiab/iiab/blob/master/roles/kolibri/templates/kolibri.service.j2) no matter whether you answer `<Yes>` or `<No>`

'Yes' was tested on Ubuntu 20.04, 'No' was tested on Mint 21.  Again this is in response to "Do you wish to run Kolibri on start-up?"  _(So this definitely appears to be a Kolibri bug &mdash; a bug that ironically seems to prevent further problems for the moment &mdash; when apt upgrading Kolibri as part of IIAB !)_

In any case, if one goes ahead with the default `<Yes>`, the next screen (a bit different than above) is...just FYI:

```
    ┌──────────────────────┤ Kolibri configuration ├───────────────────────┐
    │                                                                      │
    │ The Kolibri system service                                           │
    │                                                                      │
    │ You have chosen to run Kolibri as a system service and it has been   │
    │ started automatically. You can start/stop/restart Kolibri with the   │
    │ following command                                                    │
    │                                                                      │
    │   sudo service kolibri [start, stop, restart]                        │
    │                                                                      │
    │                                <Ok>                                  │
    │                                                                      │
    └──────────────────────────────────────────────────────────────────────┘
```

holta:
I guess I'll go ahead and merge this PR.  So wider community testing validates this across "all" OS's.

(The issues of apt upgrading Kolibri being very risky and extremely confusing to IIAB operators &mdash; are indeed a real issue but this is a separate question in the end &mdash; we should definitely assess pragmatic options there, but somewhere separately!)

holta:
> (The issues of apt upgrading Kolibri being very risky and extremely confusing to IIAB operators — are indeed a real issue but this is a separate question in the end — we should definitely assess pragmatic options there, but somewhere separately!)

@jredrejo new ticket is opened here, just for discussion: (even if there are not easy answers!)

- #3357

holta:
Recap: Of course Ansible's [command](https://docs.ansible.com/ansible/latest/collections/ansible/builtin/command_module.html) module is occasionally safer than Ansible [shell](https://docs.ansible.com/ansible/latest/collections/ansible/builtin/shell_module.html) module.

Still, in this case I opted for to shorten the code (into 1 single stanza/paragraph instead of 2 stanzas/paragraphs) as follows:

https://github.com/iiab/iiab/blob/fa17e5ff83aefc6ce0d51ffd4ed28931de134552/roles/kolibri/tasks/install.yml#L51-L55

PR #3365, opened 2022-09-15
plocate: Faster & smaller than locate & mlocate

holta:
plocate is a drop-in replacement for mlocate.

If some distros or our favorite "Linux fundamentalist obsessives" (of every possible kind, friendly & non-friendly ;) have already installed locate or mlocate, this should still work (generally...)

Of course I've not tested every single OS variant &mdash; but if important corners cases arise later we can address those in due course.

### DIFF
--- a/roles/2-common/tasks/packages.yml
+++ b/roles/2-common/tasks/packages.yml
@@ -15,7 +15,8 @@
       - logrotate          #   67kB download: RasPiOS installs this regardless
       #- lynx              #  505kB download: Installed by 1-prep's roles/iiab-admin/tasks/main.yml
       #- make              #  376kB download: 2021-07-27: Currently used by roles/pbx and no other roles
-      - mlocate            #   92kB download
+      #- mlocate           #   92kB download
+      - plocate            #   97kB download: Faster & smaller than locate & mlocate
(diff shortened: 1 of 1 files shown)


holta:
FYI plocate is not available on Ubuntu 20.04 and Mint 20.

It's time to begin the transition away from supporting these OS's, to move the focus to modern OS's built on/around Debian 11 Bullseye as a new baseline.

Thank you to @turingtrust who ran with the manual workaround here (1 character change, reverting plocate to mlocate for 20.04) while experimenting with Microsoft's new WSL with systemd:

- #3377

holta:
Background, for those who read documentation ;)

Ubuntu 20.04 was "moved below the line" in early August &mdash; to begin moving mainline IIAB support away from Ubuntu 20.04 towards 22.04 (while this is not strictly enforced just yet) i.e. "may require extensive babysitting to get right" as @turingtrust confirmed:

https://github.com/iiab/iiab/wiki/IIAB-Platforms

holta:
Manual instructions (workaround notes) have been posted ("below the line") for communities that remain attached to Ubuntu 20.04 and/or Mint 20:

https://github.com/iiab/iiab/wiki/IIAB-Platforms

PR #3368, opened 2022-09-21
Published Password warnings work better with 'sudo -n'

holta:
This fix was tested to restore both IIAB's graphical popup and the command-line reminders.

Thanks to @jvonau who uncovered this silent omission/regression &mdash; exacerbated by increasingly strict security in recent versions of Raspberry Pi OS.

### DIFF
--- a/roles/iiab-admin/templates/sshpwd-lxde-iiab.sh.j2
+++ b/roles/iiab-admin/templates/sshpwd-lxde-iiab.sh.j2
@@ -26,7 +26,7 @@ check_user_pwd() {
 
     # 2021-08-28: New OS's use 'yescrypt' so use Perl instead of Python (#2949)
     # This also helps avoid parsing the (NEW) 4th sub-field in $y$j9T$SALT$HASH
-    field2=$(sudo -n grep "^$1:" /etc/shadow | cut -d: -f2)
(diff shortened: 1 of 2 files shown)


holta:
Also tested to work satisfactorily (sufficiently) in cases where the username (e.g. iiab-admin) running the job...does not have sudoer [NOPASSWD](https://askubuntu.com/questions/334318/sudoers-file-enable-nopasswd-for-user-all-commands) privileges.

So this IIAB password warning functionality is now made more resilient (in both cases) using `sudo -n` which is explained in sudo's man page here:

> -n, --non-interactive
Avoid prompting the user for input of any kind. If a password is required for the command to run, sudo will display an error message and exit.

holta:
RECAP this PR is both...

1) An IIAB improvement, so that colorful Published Password security warnings appear at the command-line in more situations than in the past (e.g. when run from a less-privileged Linux user account, that has partial access to `/etc/shadow` password hashes via `sudo -n`).
2) A bug fix so that the Published Password pop-up security warning is now restored for operators booting "Raspberry Pi OS with desktop" with an attached screen.

holta:
Improvements can be added later if nec.

e.g. if futher Graphical Desktop OS's become highly popular (Mint 21 as the leading contender?) and prove to need an equivalent "Published Password" graphical pop-up warning in future.

PR #3371, opened 2022-09-22
rectification of names

tim-moody:


jvonau:
Seem more suitable for iiab-factory, feels a rather bit unorthodox to move the directory while in the directory used to launch the utility. Side note leaves the filesystem a bit cluttered with a new iiab-installed directory, might cause a bit of confusion.
```
iiab-admin@box:/opt/iiab $ ls 
captiveportal  downloads  iiab  iiab-admin-console  iiab-factory  iiab-installed  kiwix  pip-packages  yum-packages
```

tim-moody:
>feels a rather bit unorthodox to move the directory while in the directory used to launch the utility

good point

PR #3373, opened 2022-09-22
Better description of current configuration when networking is disabled

jvonau:
### Fixes bug:
Correct message displayed to user when network_enabled is set to False
Reflect true state of configuration
RasPiOS


holta:
Looks very safe.  Great to have actionable docs integrated within the UX like this.

PR #3265, opened 2022-06-24
22.04 & Mint 21: Avoid Sugarizer+Moodle initially (TEMPORARY)

holta:
Less confusion for newcomers chomping at the bit for Ubuntu 22.04+ and (imminent) Mint 21 Beta.

This responds to @jvonau's suggestion (below) and also means we're better prepared for the masses when Ubuntu 22.04.1 is released around early August (at which point Ubuntu will officially recommend upgrades from 20.04).

Ref:

- PR #3189
- PR https://github.com/iiab/iiab/pull/3262#issuecomment-1165810271

PR #3189, opened 2022-04-21
Stage 6/7 bypass of apps not yet compatible w/ Ubuntu 22.04 (Kolibri, Nextcloud, Moodle, Sugarizer) [ASIDE: Node.js 18.x works manually or using deb.nodesource.com install script]

holta:
This is just an interim shim patch for the coming weeks &mdash; especially for newcomers embracing Ubuntu 22.04 "Jammy Jellyfish" being released in not so many hours today (2022-04-21) after 2 years!

More advanced IIAB implementers wanting to keep up day-by-day &mdash; with the very latest feasibility of (Nextcloud, Kolibri, Moodle, Sugarizer) on Ubuntu 22.04 &mdash; should keep an eye on Section 2. of:

https://github.com/iiab/iiab/wiki/IIAB-Platforms

RECAP: Anybody is free to attempt `cd /opt/iiab/iiab` then `sudo ./runrole <role>` to later try to install any of these 4 apps/services.

But most important, this PR serves to keep implementers from drowning in confusion when MEDIUM-sized IIAB installs fail to install {Kolibri, Nextcloud, Sugarizer} and LARGE-sized IIAB installs fail to install {Moodle} &mdash; all 4 of which aren't yet/quite compatible with Ubuntu 22.04's + Python 3.10 and/or PHP 8.1 and/or MongoDB.

[Related:]

- #3182
- PR #3184
- PR #3186
- #3190
- PR #3208

PR #3380, opened 2022-09-27
Force-install libssl1.1 if MongoDB needed on Ubuntu >= 22.04 (for Sugarizer)

holta:
Resolves:

- #3190 

For those communities wanting Sugarizer on Ubuntu 22.04+

Tested on the latest daily build of Ubuntu 22.10 (server version not desktop version).

Refs:

- PR #3099
- PR #3189
- PR #3241
- PR #3262
- PR #3265

### DIFF
--- a/roles/0-init/tasks/main.yml
+++ b/roles/0-init/tasks/main.yml
@@ -9,9 +9,10 @@
 # https://github.com/iiab/iiab/blob/master/vars/default_vars.yml
 - name: "Ansible just ran /etc/ansible/facts.d/local_facts.fact to set 11 vars -- here we extract 3 of those -- rpi_model: {{ ansible_local.local_facts.rpi_model }}, devicetree_model: {{ ansible_local.local_facts.devicetree_model }}, iiab_stage: {{ ansible_local.local_facts.stage }}"
   set_fact:
+    iiab_stage: "{{ ansible_local.local_facts.stage }}"
     rpi_model: "{{ ansible_local.local_facts.rpi_model }}"
     devicetree_model: "{{ ansible_local.local_facts.devicetree_model }}"
-    iiab_stage: "{{ ansible_local.local_facts.stage }}"
+    os_ver: "{{ ansible_local.local_facts.os_ver }}"
 
 # Initialize /etc/iiab/iiab.ini writing the 'location' and 'version' sections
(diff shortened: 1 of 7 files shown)


holta:
Also tested on Ubuntu 22.04 LTS (Jammy Jellyfish) itself.

Looks great :heavy_check_mark:

roles/0-init/tasks/main.yml
```diff
@@ -92,6 +93,8 @@
       value: "{{ rpi_model }}"
     - option: devicetree_model
       value: "{{ devicetree_model }}"
+    - option: os_ver
+      value: "{{ os_ver }}"
```
jvonau, line 97: This is sort of useless info for iiab,ini, only used to load the derived OS.yml file for ansible. Would not even get to this point if incorrect but might be better suited in [create_iiab_ini.yml](https://github.com/iiab/iiab/blob/master/roles/0-init/tasks/create_iiab_ini.yml) as the underling distro doesn't really change at runtime.

jvonau:
What is with the "is_ubuntu and os_ver is version('ubuntu-2204', '>=')"  in place of 'is_ubuntu_2204 or is_ubuntu_2210'?

https://docs.ansible.com/ansible/latest/user_guide/playbooks_tests.html

holta:
> What is with the "is_ubuntu and os_ver is version('ubuntu-2204', '>=')"

Ubuntu releases often.

e.g. Ubuntu 23.04 pre-release (weeks away) and 23.10 pre-releases will be circulating before we know it, and comparison operators (e.g. `>=`) provide a failsafe in the interim.

jvonau:
Don't need os_ver for that use {{ ansible_distribution_version }} to future proof against ansible.module_utils.compat.version.StrictVersion becoming 'true' at some point in the future. Looks like version_type= is a interim step towards a full [semver](https://devhints.io/semver) where mixed numeric+alpha characters are not used.

holta:
Interesting.  Ansible might clean up their ability to make very clear what exact OS/distro version is running in future.  That'd be great.

Meantime IIAB's `os_ver` is an active part of IIAB, giving us exactly what's needed, arising from: https://github.com/iiab/iiab/blob/e46e34679ec98eeae0a3ed95b85b1f63846469c4/scripts/local_facts.fact#L35-L54

PR #3379, opened 2022-09-26
Fix hostapd for Raspberry Pi Zero W (systemd unit files, so it doesn't start too early!)

holta:
This was extensively tested to fix hostapd on Raspberry Pi Zero W's running 32-bit Raspberry Pi OS Lite on and after 2022-08-22:

- #3352

Earlier today @jvonau recommended this change be merged without delay.

Before we go ahead however, this PR should really be re-tested using new "2022-09-06" version(s) of Raspberry Pi OS &mdash; for example:

- 32-bit RasPiOS Lite on Zero W
- 64-bit RasPiOS Lite on Zero 2 W

Tangentially related:

- PR #3179

PR #3388, opened 2022-10-03
0-init: Skip validate_vars.yml on Zero W & Zero 2 W (slow Raspberry Pi's)

holta:
Personally I would never run Ansible on a Raspberry Pi Zero W.

As moving the microSD card to an RPi 4 is more than 10X faster.

But for those who must, this PR (needs testing) is an option.

Related:

- https://github.com/iiab/iiab/pull/3379#issuecomment-1264773836


jvonau:
Original speed comment noted in https://github.com/iiab/iiab/pull/3379#issuecomment-1262604581

holta:
Fixed & Tested!

CAVEAT: I really don't encourage the use of Ansible on Zero W.  But I do understand that a few people have no other options (:

PR #3394, opened 2022-10-09
Refine remote.it operations, now that connectd system service is gone, and considering 3 valid states for /etc/remoteit/registration

holta:
remote.it Device Packages removed their **connectd** "parent" systemd service as of 2022-09-07.  So instead of bouncing their connectd service, going forward, we'll instead run `/usr/share/remoteit/refresh.sh` which should generally do the exact same thing.

This PR aims for a cleaner all-around solution to remote.it frailties.

Recognizing that `/etc/remoteit/config.json` may or may not exist a priori (with or without a claim code) and that `/etc/remoteif/registration` can be in any one of 3 states {doesn't exist, is empty to block registration handshake, OR contains a [presumably valid!] remoteit_license_key} ...until remote.it is enabled by either...

- `sudo /usr/bin/iiab-remoteit`

Or...

- `./runrole remoteit` which runs `roles/remoteit/tasks/enable-or-disable.yml` (after you set `remoteit_enabled: True` in [/etc/iiab/local_vars.yml](https://wiki.iiab.io/go/FAQ#What_is_local_vars.yml_and_how_do_I_customize_it?))

Needs testing in coming days!

Building on:

- PR #3161
- PR #3363

### DIFF
--- a/roles/remoteit/defaults/main.yml
+++ b/roles/remoteit/defaults/main.yml
@@ -7,7 +7,7 @@
 # If nec, change them by editing /etc/iiab/local_vars.yml prior to installing!
 
 
-# # 2022-03-31: https://remote.it/download/ offers 4 relevant "Device Packages"
+# # 2022-03-31: https://remote.it/download/ offered 4 relevant "Device Packages"
 # # 1) Raspberry Pi (ARM)    = armhf.rpi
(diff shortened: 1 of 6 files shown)


holta:
Oops.  Please see this PR instead:

- PR #3395

Issue #3399, opened 2022-10-13
IIAB installs on Debian 12 Bookworm pre-releases IF (1) KA Lite bypassed (2) PHP 8.2 requires Nextcloud 26? (3) Munin bypassed? [Ubuntu 23.04 issues quite similar!]


holta:
Testing on Debian 12 Bookworm pre-releases is easy if you create an instant VM:

```
multipass launch -n deb12 -c 2 -d 25G -m 2G https://cloud.debian.org/images/cloud/bookworm/daily/latest/debian-12-generic-amd64-daily.qcow2
```

Shell into the VM using:

```
multipass shell deb12
```

Then within the VM, run:

```
echo 'VERSION_ID="12"' >> /etc/os-release
curl iiab.io/install.txt | bash
```

A LARGE-sized IIAB install just about worked &mdash; here are the quick+dirty workarounds I used for now:

1) The 1st error was:

   > TASK [1-prep : Install 12 network packages: avahi-daemon, hostapd, iproute2, iptables-persistent, iw, libnss-mdns, netmask, net-tools, networkd-dispatcher, rfkill, wireless-tools, wpasupplicant -- later used by https://github.com/iiab/iiab/tree/master/roles/network] ***
fatal: [127.0.0.1]: FAILED! => {"changed": false, "msg": "No package matching 'networkd-dispatcher' is available"}

   Any suggested workarounds, to explore how IIAB will work on Debian 12 Bookworm (the basis for Raspberry Pi OS 12) in 2023?

   For now, I commented out [Line 29 of network/tasks/install.yml](https://github.com/iiab/iiab/blob/e46e34679ec98eeae0a3ed95b85b1f63846469c4/roles/network/tasks/install.yml#L29) to continue.

2) The 2nd error was:

   > TASK [kiwix : Make a crontab entry to restart kiwix-serve at 4AM (debuntu)] ********************************************************************************************
fatal: [127.0.0.1]: FAILED! => {"changed": false, "msg": "Destination /etc/crontab does not exist !", "rc": 257}

   I commented out [Lines 26-31 of kiwix/tasks/enable-or-disable.yml](https://github.com/iiab/iiab/blob/e46e34679ec98eeae0a3ed95b85b1f63846469c4/roles/kiwix/tasks/enable-or-disable.yml#L26-L31) to continue.

3) The 3rd error was entirely expected, as Moodle's 4.1 LTS's PHP 8.1 support is converging right now week-to-week:

[...]

PR #3400, opened 2022-10-13
2-common/tasks/packages.yml: Debian 12 needs 'apt install cron'

holta:
Resolves several issues mentioned here:

- #3399


holta:
If Debian 12 (Bookworm) still doesn't install [cron](https://packages.debian.org/bookworm/cron) by default in a week from now, this PR should be merged &mdash; so ongoing testing can proceed without headache after headache.

holta:
> If Debian 12 (Bookworm) still doesn't install [cron](https://packages.debian.org/bookworm/cron) by default in a week from now, this PR should be merged — so ongoing testing can proceed without headache after headache.

Same story almost a week later (Debian 12's daily build does not include cron by default).

PR #3426, opened 2022-11-16
Upgrade MineTest to only use package on all platforms

tim-moody:
Not yet tested


holta:
1. Can these 9 final lines be removed from `vars/raspbian-11.yml` if possible?

   ```

   # Minetest for RPi
   minetest_server_bin: /library/games/minetest/bin/minetestserver
   minetest_working_dir: /library/games/minetest
   minetest_game_dir: /library/games/minetest/games/minetest_game
   minetest_rpi_src_tar: minetest.5.1.1.tar.gz
   #minetest_rpi_src_url: "https://www.nathansalapat.com/downloads/{{ minetest_rpi_src_tar }}"
   minetest_rpi_src_url: "{{ iiab_download_url }}/{{ minetest_rpi_src_tar }}"
   minetest_rpi_src_untarred: Minetest
   ```

2. Where should var `minetest_game_dir` be set, prior to running `minetest_install_mods.yml` ?  (Or is this variable possibly less needed than in the past ?)

3. Both unused templates should be suffixed with`.unused` or similar, to make that clear:

   - roles/minetest/templates/minetest.conf.j2
   - roles/minetest/templates/minetest-server.service.j2

4. This default_vars.yml line needs to be revised:

   ```
   minetest_default_game: carbone-ng    # only carbone-ng and minetest are supported
   ```

5. Is overriding this default_vars.yml line something that we still want to promote?

   ```
   minetest_flat_world: False
   ```

6. `roles/README.rst` needs a few quick revisions, just to summarize the above.

tim-moody:
1 don't recall why all this is here but shouldn't be needed now.
2. & 4 think my default vars was not committed
5 was a TK request
6 in due time as indicated

vars/default_vars.yml
```diff
@@ -631,8 +631,13 @@ internetarchive_port: 4244    # for http://box:4244
 minetest_install: False
 minetest_enabled: False
 minetest_port: 30000
+minetest_working_dir: /usr/share/games/minetest
+
 minetest_server_admin: Admin
-minetest_default_game: carbone-ng    # only carbone-ng and minetest are supported
+# minetest_default_game: carbone-ng    # carbone-ng is not longer compatible so not supported
+minetest_default_game: minetest
+# minetest_default_game: dreambuilder - after 5.6
```
holta, line 639:
```suggestion
# minetest_default_game: dreambuilder - requires minetest-server >= 5.6
```

holta:
FYI Debian 12 (Bookworm) and Ubuntu 23.04 (Lunar Lobster) both today offer minetest-server 5.6.1 via apt.

So those wanting https://github.com/mt-mods/dreambuilder_game (which apparently requires minetest-server >= 5.6) should benefit in 2023.

Issue #3444, opened 2022-12-25
IIAB 8.0 fails to install on Pi 4 running  Desktop version 2022-09-22-raspios-bullseye-arm64.img on PI 4


shanti-bhardwa:
Get this error during install over WIFI only on PI4 with 2022-09-22-raspios-bullseye-arm64.img O/S.

TASK [iiab-admin : If so, install from template: /etc/xdg/lxsession/LXDE-pi/sshpwd-lxde-iiab.sh] *********************************************************************************************************************************************
An exception occurred during task execution. To see the full traceback, use -vvv. The error was: ansible.errors.AnsibleUndefinedVariable: 'sshd_service' is undefined. 'sshd_service' is undefined
fatal: [127.0.0.1]: FAILED! => {"changed": false, "msg": "AnsibleUndefinedVariable: 'sshd_service' is undefined. 'sshd_service' is undefined"}

PLAY RECAP

Cannot do iiab-diagnostics as the install did not complete. Here is the copy of the output.

TASK [iiab-admin : Configure user iiab-admin / password and its group(s), if iiab_admin_user_install] ****************************************************************************************************************************************
included: /opt/iiab/iiab/roles/iiab-admin/tasks/admin-user.yml for 127.0.0.1

TASK [iiab-admin : Does user 'iiab-admin' (iiab_admin_user) exist?] **************************************************************************************************************************************************************************
changed: [127.0.0.1]

TASK [iiab-admin : Establish Linux group 'iiab-admin' group, for login to Admin Console] *****************************************************************************************************************************************************
ok: [127.0.0.1]

TASK [iiab-admin : Configure user 'iiab-admin' with group 'iiab-admin' for login to IIAB's Admin Console (http://box.lan/admin) AND for IIAB community support commands (/usr/bin/iiab-*) at the command-line] *******************************
ok: [127.0.0.1]

TASK [iiab-admin : If user didn't exist, set password to 'g0adm1n'] **************************************************************************************************************************************************************************
skipping: [127.0.0.1]

TASK [iiab-admin : Add user iiab-admin to group sudo, for IIAB community support commands in /usr/bin like {iiab-diagnostics, iiab-hotspot-on, iiab-check-firmware}, if iiab_admin_can_sudo] *************************************************
ok: [127.0.0.1]

TASK [iiab-admin : Remove user iiab-admin from group sudo, if not iiab_admin_can_sudo] *******************************************************************************************************************************************************
skipping: [127.0.0.1]

[...]

PR #3446, opened 2022-12-26
Put Kiwix var {{ systemctl_program }} to rest — simple is almost always better

holta:
Unclear that systemctl really needs a full path but it can't hurt.

Tested on Ubuntu 23.04's latest daily, with this PR using canonical path `/usr/bin/systemctl` to be consistent with command-line usage.

Building on:

- PR #3440
- #3444

### DIFF
--- a/roles/kiwix/defaults/main.yml
+++ b/roles/kiwix/defaults/main.yml
@@ -57,6 +57,5 @@ kiwix_url: /kiwix
 kiwix_url_plus_slash: "{{ kiwix_url }}/"    # /kiwix/
 kiwix_path: "{{ iiab_base }}/kiwix"         # /opt/iiab/kiwix
 
-systemctl_program: /bin/systemctl
 kiwix_nginx_timeout: 600
 kiwix_threads: 4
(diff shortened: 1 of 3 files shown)


jvonau:
Or just alter the crontab stanza to actually use systemctl_program, touches one file instead of 3 files. Might also want to update to the cron ansble plugin in place of lineinfile like [matomo uses](https://github.com/iiab/iiab/blob/master/roles/matomo/tasks/install.yml#L183-L198).

jvonau:
Guess I should of asked if the 4am restart of kiwix is even still required as this was originally a cheap workaround to keep kiwix running from years ago. #484 #485 https://github.com/XSCE/xsce/commit/2a62885158a868bb9170e6121f822d635b5e9eda

jvonau:
Perhaps there could be a new variable that could enable the cron routine on demand in place of just restarting as I think that wipes out the cache of search history that was built up within kiwix-serve.

holta:
> Guess I should of asked if the 4am restart of kiwix is even still required as this was originally a cheap workaround to keep kiwix running from years ago. #484 #485 [XSCE/xsce@2a62885](https://github.com/XSCE/xsce/commit/2a62885158a868bb9170e6121f822d635b5e9eda)

kiwix-serve has very regular bugs as it make strides evolving + improving, as part of a continual community process, which cannot be fully predicted one month to the next.  As such the 4AM reboot serves to help protect communities/schools/users on a valuable ongoing basis.

PR #3443, opened 2022-12-24
Fix for Issue #3441

cwivagg:
### Fixes bug:

[Going to Matomo takes user to error page.](https://github.com/iiab/iiab/issues/3441)

### Description of changes proposed in this pull request:

Add a file and a directory that aren't created until someone browses to Matomo (apparently).

### Smoke-tested on which OS or OS's:

Just Ubuntu 22.04. These changes should cause no new issues. There are a couple known issues with Ubuntu 23.04 and Debian 12 (?) that will be fixed by future changes.

### Mention a team member @username e.g. to help with code review:

@holta

Issue #3441, opened 2022-12-23
matomo is the most privacy invasive pile of code I have ever came across bar none


jvonau:
### Expected behavior and actual behavior:
See some sort of meaningful webpage when accessing /matomo

![Screenshot from 2022-12-23 14-20-59](https://user-images.githubusercontent.com/5579710/209402452-d732abac-23ab-4b24-b4b3-aede2b239800.png)

### Steps to reproduce the problem:
navigate to /matomo within a web browser

holta:
Coincidentally this week @cwivagg is right now looking into it.

There appears to be different Matomo problems on different OS's {Debian 12, Ubuntu 23.04, ETC}.

Hopefully this will all be understood and/or resolved within a week or so.

- PR #3304

jvonau:
Should just make Matomo install optional, like we need another php service that could be abused as a backdoor.

jvonau:
On a side note @tim-moody iiab-admin-console does not offer the ability to enable/disable Matomo. I would of expected Matomo to appear within the 'Configuration' -> 'Services Enabled' -> 'For Monitoring and Administration' section.

![Screenshot from 2022-12-23 14-50-45](https://user-images.githubusercontent.com/5579710/209404539-cf564a1b-b0f2-4a62-b0c1-bb3a277083c5.png)

holta:
The ability to (programmatically) disable Matomo after it's installed with IIAB is not implemented (to my knowledge).
And is probably not needed, to be honest.
Still I will ask @cwivagg about such things when I speak to him, hopefully in coming days.

jvonau:
> The ability to (programmatically) disable Matomo after it's installed with IIAB is not implemented (to my knowledge). And is probably not needed, to be honest. Still I will ask @cwivagg about such things when I speak to him, hopefully in coming days.

The service must work out of the box or be dropped from the mandatory default install.

holta:
@cwivagg just pinged me and he's working on it :heavy_check_mark:

jvonau:
What URI is matomo expected to respond to? http://box.lan/matomo, http://WANIP/matomo, and/or http://hostname.domain/matemo?

jvonau:
Accessing the website with just http://WANIP/matomo results in a 404 error but if I append index.php I'm presented with:
![Screenshot from 2022-12-26 21-52-17](https://user-images.githubusercontent.com/5579710/209609647-06991312-8cf9-45c3-994e-f3d758760979.png)

[...]

Issue #1762, opened 2019-06-22
IIAB Community Analytics framework, above & beyond AWStats (likely Matomo?)


holta:
Some Internet-in-a-Box (IIAB) field projects want deeper community insights, to understand important/subtle usage patterns in ways that AWStats is not able to offer.

One of the most common questions is "how many unique phones/devices took advantage of this IIAB over the past day/week/month?"

However we seek to go about answering these kinds of questions, there are huge/tricky privacy-versus-surveillance questions if HW GUID's or cookies are tracked to make this happen.  And what Data Minimization / Data Deletion policy recommendations are appropriate to advance ethical best practices.

Let's discuss what's possible in 2019 &mdash; to give IIAB community leaders stronger insights into the ever-evolving patterns as to how their communities are using IIAB &mdash; while taking unintended antiprivacy consequences very seriously (so that disadvantaged populations are never abused/persecuted, e.g. based on their library reading/viewing/listening habits).

P.S. the fact that most Raspberry Pi computers lack an internal RTC (real-time clock & associated battery) means the challenge here is especially complicated, both for tabulating meaningful stats/patterns, and also to implement data deletion policies e.g. if mobile devices GUID's / cookies records (or similar) should be deleted completely from the server monthly or so?

Prior Work:

- http://FAQ.IIAB.IO "How do I view usage statistics?"
- #657
- #1268 "AWStats community patterns/analytics/statistics/metrics"
- #1332 "Enhance "How do I view usage statistics?" in FAQ for IIAB 7.0 or 7.1 release"

tim-moody:
what awstats offers is substantially more sophisticated than this request. if we assume that dnsmasq routinely gives the same device the same ip, then it even supports the count the devices request.

holta:
> what awstats offers is substantially more sophisticated than this request. if we assume that dnsmasq routinely gives the same device the same ip, then it even supports the count the devices request.

What if there are more than 256 users/devices?  (Even if the above assumption were to be validated.)

holta:
It's very likely that IIAB should drive to include [Matomo Analytics](https://matomo.org/) for IIAB 7.1 — for local operators, teachers, implementers,  content developers — and yes also kids/patients/citizens wanting to better understand their own community.

A hosted version should also be offered, offering a modern dashboard for NGO's, content partners and similar.

[...]

PR #3447, opened 2022-12-27
Somewhat higher PHP limits for Matomo — similar to Nextcloud, PBX, WordPress

holta:
This PR is almost entirely documentation, to better acknowledge that Matomo depends on PHP.

@cwivagg with create a follow-up PR solving deeper issues relating to:

- #3441
- PR #3443

This PR was tested on Ubuntu 22.04 using all 3 "code paths" :

- `matomo_install: False`
- `matomo_install: True` with `nginx_high_php_limits: False`
- `matomo_install: True` with `nginx_high_php_limits: True`

Building on the original:

- #1762
- PR #3304

### DIFF
--- a/roles/www_options/tasks/main.yml
+++ b/roles/www_options/tasks/main.yml
@@ -52,13 +52,13 @@
 
 
 - debug:
-    msg: 'THE 5 ANSIBLE STANZAS BELOW ONLY RUN... when: moodle_install or nextcloud_install or pbx_install or wordpress_install'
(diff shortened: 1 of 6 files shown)


jvonau:
The www_options role might be a good candidate to have lineinfile replaced with ini_file like matomo uses.

holta:
> The www_options role might be a good candidate to have lineinfile replaced with ini_file like matomo uses.

In general a great idea, yes.

In this narrow case however (with comments to help implementers/operators changing both php.ini files after the fact) lineinfile is still best in this particular situation.  As refactored (and fixed, really) here:

- PR #3449

Issue #1147, opened 2018-09-19
WordPress Umbrella Tkt, e.g. PHP blocks uploading of graphics > 2MB? (WP defaults not Ok?)


holta:
@ericnitschke asks if IIAB can consider new defaults to /etc/php/7.0/apache2/php.ini (and similar, across all OS's) to something like this?
```
upload_max_filesize 64M    ; default on RPi is 2M
post_max_size 128M    ; default on RPi is 8M
memory_limit 256M    ; default on RPi is 128M 
max_execution_time 300    ; default on RPi is 30
max_input_time 300    ; default on RPi is 60
```
(At a glance I'm concerned about modifying 5 defaults on such a RAM-constrained platform, and hope @kananigit and others can help investigate this, so occasional WordPress users still have the choice to keep their RPi machines responsive w/o too many resources grabbed by WordPress?)

holta:
@ericnitschke also has questions about:
- WordPress plugins being blocked (not installable if they are too large?)
- Wordpress' content export/import functions not fully working.

I've asked him to create separate issues at https://github.com/iiab/iiab/issues so others can understand what exactly he's trying to do in each case &mdash; on exactly which versions of {WordPress, IIAB, underlying OS, HW, etc} so others can fully reproduce this &mdash; which he will do in coming days.

holta:
Please also see:
- https://github.com/iiab/iiab/issues/201 Automate re-homing of http://HOSTNAME/wordpress? (instead of RELOCATE flag)
- https://github.com/iiab/iiab/issues/992 " http://box/wordpress + http://OTHER-DOMAIN/wordpress + http://OTHER-IP/wordpress "
- #986 Validate WordPress 4.9.9 on "2018-11-05"

holta:
@ericnitschke asks if https://en.wikipedia.org/wiki/BuddyPress (https://buddypress.org) works to create a social network / community-of-practice on WordPress on RPi, @kananigit do you know anything about this?

COMPARE...these 3 other "social" platforms/offerings that can be enabled within IIAB...FWIW:
- [Elgg 2.3.8](https://github.com/iiab/iiab/tree/master/roles/elgg) (used with IIAB in Malaysia)
- [Moodle 3.5.2 LTS](https://github.com/iiab/iiab/tree/master/roles/moodle) (being consider with IIAB in Kenya)
[...]

PR #3448, opened 2022-12-27
Several updates to make Matomo experience cleaner

cwivagg:
Several updates to make Matomo experience cleaner

### Description of changes proposed in this pull request:

* Avoid trusted host check so users can get to the Matomo website more easily.
* Remove documentation for users to circumvent the trusted host check since it is no longer necessary.
* Add documentation about the time zone and how we got here.
* Add PHP dependencies.
* Fix the <host>/matomo URL (as opposed to <host>/matomo/ which already worked).

### Smoke-tested on which OS or OS's:

Ubuntu 22.04 size 0 and 1 installs (with Matomo enabled).

### Mention a team member @username e.g. to help with code review:

@holta

PR #3449, opened 2022-12-29
PHP Limits Dependency Cleanup: www_options/tasks/set-php-limits.yml run on demand (by 5 relevant roles)

holta:
Tested on Ubuntu 23.04 (latest daily build) with many different `X_install: True` and `Y_install: False` configurations of the 5 IIAB Apps than depend on PHP...

- Matomo
- Moodle
- Nextcloud
- PBX
- WordPress

...as set in [/etc/iiab/local_vars.yml](https://wiki.iiab.io/go/FAQ#What_is_local_vars.yml_and_how_do_I_customize_it%3F)

This makes ad hoc install of individual roles far more sane & reliable (PHP limits are set as required, instead of being forgotten and set incorrectly whenever the implementer forgets to run `./runrole www_options` &mdash; which they thankfully no longer need to do &mdash; as a result of this PR).

Documentation is also simplified, e.g. to make clear that Moodle effectively sets `nginx_high_php_limits: True` automatically.

There are no functional changes in the end [as compared to earlier IIAB's, where the implementer/operator ran `./runrole www_options` correctly, when changing PHP limits].  Still, this PR rearranges things as will be documented (links fixed!) within places like:

- https://github.com/iiab/iiab/wiki
- http://FAQ.IIAB.IO

This PR builds on:

- #1147
- PR #3447

Tangentially related:

- PR #3304
- https://github.com/iiab/iiab/pull/3174#issuecomment-1366961444


holta:
> This [PR] makes ad hoc install of individual roles far more sane & reliable (PHP limits are set as required, instead of being forgotten and set incorrectly whenever the implementer forgets to run `./runrole www_options` — which they thankfully no longer need to do — as a result of this PR).

In that sense, this PR is arguably more of a bug fix than an enhancement.

jvonau:
Ok, I can see how 'set_php_limits' is used to limit the running of the task once per ansible run. What sets 'nginx_high_php_limits' to be 'True'  in order to gain the larger values requested by the roles? Post Coffee Edit: True is enforced with moodle, the rest will use the value that is defined in local_vars.yml.

holta:
> What sets 'nginx_high_php_limits' to be 'True' in order to gain the larger values requested by the roles?

Not really requested by the roles.  This (`nginx_high_php_limits: True`) is intended to be set in `/etc/iiab/local_vars.yml` typically on Line 213 here:

https://github.com/iiab/iiab/blob/7882dcb7f858a03ce701b56bb71ec0960ae2dae5/vars/local_vars_medium.yml#L212-L219

In the case of communities needing Moodle (and quite possibly Nextcloud at some future date) this setting (`nginx_high_php_limits: True`) is mandated automatically on Lines 70 and 84 here:

- https://github.com/holta/iiab/blob/set-php-limits/roles/www_options/tasks/set-php-limits.yml#L70
- https://github.com/holta/iiab/blob/set-php-limits/roles/www_options/tasks/set-php-limits.yml#L84

(As you mention, variable `set_php_limits` [defined or undefined] allows many roles' installs to be installed at once, so that `set-php-limits.yml` is run at most once per Ansible run.)

jvonau:
Perhaps the roles' main.yml in place of install.yml, might be a better place to call set-php-limits, the values can be adjusted with runrole or admin-console by just running the play/role.

holta:
> Perhaps the roles' main.yml in place of install.yml, might be a better place to call set-php-limits, the values can be adjusted with runrole or admin-console by just running the play/role.

It's an idea, not unlike `tasks/enable-or-disable.yml` in each of these PHP-enabled roles.  I definitely considered this.

For now I've decided against both above, keeping PHP settings as part of the core/initial (almost idempotent!) `tasks/install.yml` &mdash; that can almost always be re-run (in most cases anyway, during machine provisioning early days) using:

`./runrole --reinstall <ROLE NAME>`

At a higher level, establishing normative PHP settings is certainly an install-like action.  Somewhat related to my perspective that more manual fine-tuning of both `php.ini` files also remains very important to emphasize &mdash; and definitely to be encouraged among communities that go down the road of using these larger roles/apps, e.g.:

- `/etc/php/7.4/cli/php.ini`
- `/etc/php/7.4/fpm/php.ini`

Or...

- `/etc/php/8.1/cli/php.ini`
- `/etc/php/8.1/fpm/php.ini`

holta:
> There are no functional changes in the end [as compared to earlier IIAB's, where the implementer/operator ran ./runrole www_options correctly, when changing PHP limits]. Still, this PR rearranges things as will be documented (links fixed!) within places like:
>
> - https://github.com/iiab/iiab/wiki
> - http://faq.iiab.io/

Note To Self - docs discussing `./runrole www_options` need to be revised here:

- https://wiki.iiab.io/go/FAQ#Is_a_%22Rapid_Power_Off%22_button_possible_for_low-electricity_environments%3F
- https://wiki.iiab.io/go/FAQ#WordPress_&_Moodle_Administration:_What_tips_&_tricks_exist%3F

jvonau:
> > Perhaps the roles' main.yml in place of install.yml, might be a better place to call set-php-limits, the values can be adjusted with runrole or admin-console by just running the play/role.
> 
> It's an idea, not unlike `tasks/enable-or-disable.yml` in each of these PHP-enabled roles. I definitely considered this.
>

That is the idea, dynamic soft changes on the fly.
 
> For now I've decided against both above, keeping PHP settings as part of the core/initial (almost idempotent!) `tasks/install.yml` — that can almost always be re-run (in most cases anyway, during machine provisioning early days) using:
> 
> `./runrole --reinstall <ROLE NAME>`
> 
> My perspective being that manual fine-tuning of both `php.ini` is more important, certainly to be encouraged among communities go down the road of using these larger roles/apps, e.g.:
> 
>     * `/etc/php/8.1/cli/php.ini`
> 
>     * `/etc/php/8.1/fpm/php.ini`

Using --reinstall on the role with runrole introduces a needless pass though the install process when all that is required is adjusting of the php values. If the goal is to prevent altering of what might be user changes in the future then I would suggest making use of iiab_state.yml to record the pass that was made through set-php-limits for future ansible runs.

holta:
A crazy option would be to re-introduce execution of tasks/set-php-limits.yml within www_options/tasks/main.yml for (misguided!) folks really want to use Ansible to regularly tweak PHP settings.

Certainly a bulldozer (Ansible) should be avoided as much as possible when threading a needle (php.ini tweaks).

In any case, for now I'd like to go forward removing `./runrole www_options` from FAQ.IIAB.IO to simplify documentation (as much as possible anyway) and experiment with PHP settings being (intentionally) fully contained within each role's tasks/install.yml

jvonau:
If the task set-php-limits was turned into a role then one could call 'runrole set-php-limits' at will without calling the parent role at all but still could have install.yml chained call in place. Kind of like how yarn/nodejs/mongodb are rigged up.

tim-moody:
correct me if I'm wrong, but all this does is change an ini file, so requires ms to execute.

holta:
> If the task set-php-limits was turned into a role then one could call 'runrole set-php-limits' at will without calling the parent role at all but still could have install.yml chained call in place. Kind of like how yarn/nodejs/mongodb are rigged up.

It's an idea.  But I'd almost rather keep execution of tasks/set-php-limits.yml within www_options/tasks/main.yml if this is absolutely necessary.

Certainly I'd very much rather not break the higher-level notion that `install.yml` and `enable-or-disable.yml` are simply all you need (in most all cases, with some exceptions) just for the sake of very obscure php.ini settings that few people care about.

> all this does is change an ini file, so requires ms to execute.

True.  But execution time is not the driving concern here.  Simplifying overall complexity (documentation, maintainability, etc) is the primary goal here from my perspective.

tim-moody:
simplest would be to have each role that needs limits changed just do it and not worry if it happens more than once.

jvonau:
> > If the task set-php-limits was turned into a role then one could call 'runrole set-php-limits' at will without calling the parent role at all but still could have install.yml chained call in place. Kind of like how yarn/nodejs/mongodb are rigged up.
> 
> It's an idea. But I'd almost rather keep execution of tasks/set-php-limits.yml within www_options/tasks/main.yml if this is absolutely necessary.
>

That would be preferred to me over forcing a reinstall of the calling parent role, less likely to damage the existing install. 
 > Certainly I'd very much rather not break the higher-level notion that `install.yml` and `enable-or-disable.yml` are simply all you need (in most all cases, with some exceptions) just for the sake of very obscure php.ini settings that few people care about.
> 

One could just drop the automatic edits and just dump the editing of php.ini onto the lap of the implementer, after all it is not the end-user that would make use of the setting rather the organization that is putting the hardware/image out into the wild.

> > all this does is change an ini file, so requires ms to execute.
> 
> True. But execution time is not the driving concern here. Simplifying overall complexity (documentation, maintainability, etc) is the primary goal here from my perspective.

Perhaps add not overwriting user edits to php.ini done as post installation fine tuning out in the field.

holta:
> Perhaps add not overwriting user edits to php.ini done as post installation fine tuning out in the field.

Indeed!

This is one of several reasons I moved (all messing around with php.ini) from each tasks/main.yml to each tasks/install.yml &mdash; so that it's philosophically thought of as part of the install process (right now this happens right after `<ANY PHP ROLE>/tasks/install.yml` installs its own PHP + PHP extensions/requirements).

The one exception (nagging use case, as always) is of course folks racing for their flights at the very last minute before deployment, suddenly deciding they want `nginx_high_php_limits: True` and not knowing entirely what they're doing.

I'll think about it.  Of course there's only so far we can go (and only so far we should go) protecting people from themselves / who actively choose to be careless.

holta:
Rebased on top of the just-merged:

- PR #3448

holta:
`set-php-limits.yml` is ALSO now executed by `www_options/tasks/main.yml` (not just by IIAB's ~5 PHP-based roles).

This allows post-install toggling of variable `nginx_high_php_limits` in [/etc/iiab/local_vars.yml](https://wiki.iiab.io/go/FAQ#What_is_local_vars.yml_and_how_do_I_customize_it?) &mdash; e.g. if you then run...

```
cd /opt/iiab/iiab
sudo ./runrole www_options
```

(...as you're running to the airport en route to a deployment.)

holta:
Tested using `./runrole www_options` with all 3 of the [aforementioned](https://github.com/iiab/iiab/pull/3447) "code paths" on Ubuntu 23.04

Good enough.

Time to merge this PR for broader community testing.

holta:
> > There are no functional changes in the end [as compared to earlier IIAB's, where the implementer/operator ran `./runrole www_options` correctly, when changing PHP limits]. Still, this PR rearranges things as will be documented (links fixed!) within places like:

> * https://wiki.iiab.io/go/FAQ#Is_a_%22Rapid_Power_Off%22_button_possible_for_low-electricity_environments%3F -- _no changes required_
> * https://wiki.iiab.io/go/FAQ#WordPress_&_Moodle_Administration:_What_tips_&_tricks_exist%3F
> - default_vars.yml & relevant local_vars.yml files -- _already changed by this PR, in https://github.com/iiab/iiab/tree/master/vars)_

Hopefully these FAQ.IIAB.IO changes are good enough:

https://wiki.iiab.io/index.php?title=FAQ&type=revision&diff=505&oldid=502

jvonau:
See, we can have our cake and eat it too. :+1:

PR #3452, opened 2022-12-30
Php timezone

jvonau:
http://sprunge.us/kXs9wb

### DIFF
--- a/roles/0-init/tasks/main.yml
+++ b/roles/0-init/tasks/main.yml
@@ -54,8 +54,8 @@
   include_tasks: validate_vars.yml
   when: not (rpi_model | regex_search('\\bW\\b'))    # Ansible require double backslashes, e.g. with \b "word boundary" anchors: https://www.regular-expressions.info/wordboundaries.html https://stackoverflow.com/questions/56869119/ansible-regular-expression-to-match-a-string-and-extract-the-line/56869801#56869801
 
-- name: "Time Zone / TZ: Set symlink /etc/localtime to UTC if it doesn't exist?"
-  include_tasks: tz.yml
+#- name: "Time Zone / TZ: Set symlink /etc/localtime to UTC if it doesn't exist?"
+#  include_tasks: tz.yml
 
 - name: Set hostname / domain (etc) in various places
   include_tasks: hostname.yml
--- a/roles/www_options/tasks/main.yml
+++ b/roles/www_options/tasks/main.yml
@@ -51,6 +51,15 @@
   when: lxde_pi_autostart_present.stat.exists and chromium_present.stat.exists
 
 
+- name: Install timezone templates for php-cli and php-fpm
+  template:
+    src: "{{ item.src }}"
+    dest: "{{ item.dest }}"
+  with_items:
+    - { src: 'local_timezone.j2', dest: '/etc/php/{{ php_version }}/cli/conf.d/99-timezone.ini' }
+    - { src: 'local_timezone.j2', dest: '/etc/php/{{ php_version }}/fpm/conf.d/99-timezone.ini' }
+  when: not (ansible_local.local_facts.local_timezone == "none")
+
 # 2022-12-29: set-php-limits.yml is ALSO attempted (on demand) by every
 # <ROLE>/tasks/install.yml that needs it (Matomo, Moodle, Nextcloud, PBX,
 # WordPress) so './runrole <ROLE>' and similar are fully self-sufficient!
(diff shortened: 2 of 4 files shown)


jvonau:
#69 #1147

holta:
I personally haven't seen evidence that `{{ ansible_date_time.tz }}` is unreliable in recent years.  Of course (A) there are many transliterated synonyms for certain Time Zones, and Ansible's string output can (B) oscillate (e.g. between EDT and EST) depending on summer and winter.  These are more feature than bug (but some may not want these features...)

If a fixed TZ string is truly needed (?) other options include:

- `timedatectl show -p Timezone --value` is rather authoritative.  It uses (1) `/etc/localtime` primarily, or if that's missing it falls back to (2) `/etc/timezone` (if it exists).

- `readlink /etc/localtime | awk -F "zoneinfo/" '{print $2}'` provides the TZ string derived from the symlink.

- `cat /etc/timezone` is no longer authoritative, but can be used as a fallback if absolutely nec.

jvonau:
ansible_date_time.tz is not unreliable just wrong for the purpose.
timedatectl requires the system to be booted, and may not be running. 
readlink works if the system is configured/provisioned.
/etc/timezone is authoritative given timedatectl sets the value or was set by the OS installer at install time.

holta:
> I personally haven't seen evidence that `{{ ansible_date_time.tz }}` is unreliable in recent years.

Let me revise that: Ansible's output above muddles Asia/Tel_Aviv and Asia/Kolkata as IST.  While each is correct, these TZ abbreviations makes it impossible to distinguish e.g. Israel Standard Time (UTC +2) from India Standard Time (UTC +5:30).

And while it's true you can use `{{ ansible_date_time.tz_offset }}` to disambiguate between +2 and + 5:30 &mdash; and you can also use `{{ ansible_date_time.tz_dst }}` to disambiguate between "summer" and "winter" time changes &mdash; this is all very unwieldy.

Conclusion: The original Linux system's TZ (e.g. /etc/localtime) is far more informative, providing all TZ info in 1 single very authoritative place (instead of having to fuss over interpreting 3 separate variables!)

> (A) there are many transliterated synonyms for certain Time Zones, and Ansible's string output can (B) oscillate (e.g. between EDT and EST) depending on summer and winter. These are more feature than bug (but some may not want these features...)

Just FYI it seems likely Ansible is using the same TZ "abbreviator" mechanism as Linux command:

```
date +%Z
```

In short, the above command suffers the exact same problem as Ansible's TZ abbreviation function (both are 1-way functions, that are not reversible, i.e. they produce inconclusive and ambiguous output).

holta:
> /etc/timezone is authoritative given timedatectl sets the value or was set by the OS installer at install time.

In some circumstances.  Those assumptions are not universal.  Some mainline OS installers do not set up /etc/timezone in recent years, ever since /etc/localtime became more authoritative.

jvonau:
> > /etc/timezone is authoritative given timedatectl sets the value or was set by the OS installer at install time.
> 
> In some circumstances. Those assumptions are not universal. Some mainline OS installers do not set up /etc/timezone in recent years, ever since /etc/localtime became more authoritative.

Point one of these OSs out please. I might be open to have local_facts read /etc/timezone first and fallback to using readlink if not found.

jvonau:
timedatectl updates /etc/timezone if the file exists but does not create it if missing, so I will add the fall back using readlink

holta:
I don't recall which Ubuntu/Debian derivative.  But this did happen to me several times in recent years (no /etc/timezone existed).

More important for Matomo: `timedatectl show -p Timezone --value` does the job authoritatively, interpreting the time zone without the need for guesswork or maintenance, regardless whether OS/industry/global standards change.  e.g. it even does the right thing when primary /etc/localtime and secondary /etc/timezone are both missing (in fact this means UTC, according to the latest standard).  With all these low-level grungy corner cases successfully abstracted away by `timedatectl show -p Timezone --value`, I personally would avoid using anything else.

jvonau:
required reading: http://0pointer.de/public/systemd-man/timezone.html https://wiki.archlinux.org/title/System_time
timedatectl does not work without being booted, and would fail in a chrooted environment like pristine image building.

holta:
> timedatectl does not work without being booted, and would fail in a chrooted environment like pristine image building.

Be that as it may, that might not be relevant to @cwivagg (his matomo/tasks/install.yml already relies on systemd / systemctl).

e.g. dynamically invoking `timedatectl show -p Timezone --value` right there from matomo/tasks/install.yml gets the very latest System TZ into PHP for safety (even survives implementers poking around with Time Zones in another window 1min beforehand, on their brand new graphical OS!)

Certainly a substantially friendlier Matomo UX with more meaningful TZ would indeed be great, whether or not squeezed into IIAB 8.0's imminent release :clock11:

jvonau:
You are so full of yourself.

PR #3459, opened 2023-01-07
Closed Issue: revert to using pip for ansible on 32bit RasPiOS [& ongoing issues w/ wheels, cryptography, rust]

jvonau:
### Fixes bug:
Stops polluting /etc/apt/sources.list.d/ and /usr/share/keyrings/ with files that are never used to install ansible while using "pip3 install". ie Ubuntu and RasPiOS-64bit
### Description of changes proposed in this pull request:
Restores "pip3 install" for RasPiOS-32-bit
### Smoke-tested on which OS or OS's:
Ref: Slimmed down version of #3456

PR #3493, opened 2023-03-17
Fix PR #3491 to install ansible-core on all OS's as Python 3.11+ proliferates — accounting for older versions of pip that lack flag `--break-system-packages`

holta:
Tightens up and fixes up:

- PR #3491

Also upgrades cryptography 37.0.4 to 39.0.2 on 32-bit OS's as mentioned here:

- https://github.com/iiab/iiab/pull/3459#issuecomment-1473000053

Tested on 64-bit Ubuntu 23.04 pre-release and the latest 32-bit Raspberry Pi OS Lite.

### DIFF
--- a/scripts/ansible
+++ b/scripts/ansible
@@ -193,7 +193,7 @@ if ! uname -m | grep -q 64; then
 fi
 
 echo -e "\n\n'pip3 install --upgrade ansible-core' will now run:\n"                                   # REMINDER: ansible-core 2.12 (released 2021-11-08) requires Python >= 3.8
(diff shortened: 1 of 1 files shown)


jvonau:
Now that the rust tool chain is [corrected upstream](https://github.com/piwheels/piwheels/pull/328) and [now builds correctly](https://www.piwheels.org/project/cryptography/) for 32bit RasPiOS there in no need to 'pin' the release to a specific release version, the latest version available should be used going forward.

Issue #3516, opened 2023-03-29
Strange switch to 64bit kernel on 32bit RasPiOS [on 64-bit CPU architectures]


jvonau:
This is just a heads up that something changed in the behavior on 32bit RasPiOS that might cause some weird errors going forward.
http://sprunge.us/xLDTSv?en

Highlights below
  18 Raspberry Pi reference 2022-09-22
  19 Generated using pi-gen, https://github.com/RPi-Distro/pi-gen, 8a42abcd1dbd8c9c1fdfca4e0c3778255b2f9cc4, stage2
  20 /etc/debian-version: 11.6
  21 display-manager? inactive   Arch1: armhf   Arch2: 
  22 6.1.19-v8+ #1637 SMP PREEMPT Tue Mar 14 11:11:47 GMT 2023 unknown
  23 Cortex-A72   CPU(s): 4   Mem: 3794 Swap: 99
  24 Raspberry Pi 4 Model B Rev 1.1
  98 install_date = 2023-01-03T18:10:51Z
109 runtime_date = 2023-03-17T12:17:00Z
110 ansible_version = 2.14.3
111 kernel = 5.15.84-v7l+

Note line '22 6.1.19-v8+' that is the 64bit kernel now running but as of 2023-03-17 the machine was running '111 kernel = 5.15.84-v7l+' There is no 'arm_64bit=1' present in config.txt as the docs suggest that is needed to boot the 64bit kernel.
```
jerryv@box:~ $ cat /boot/config.txt 
# For more options and information see
# http://rpf.io/configtxt
# Some settings may impact device functionality. See link above for details

# uncomment if you get no picture on HDMI for a default "safe" mode
#hdmi_safe=1

# uncomment the following to adjust overscan. Use positive numbers if console
# goes off screen, and negative if there is too much border
#overscan_left=16
#overscan_right=16
#overscan_top=16
#overscan_bottom=16

# uncomment to force a console size. By default it will be display's size minus
# overscan.
#framebuffer_width=1280
#framebuffer_height=720

# uncomment if hdmi display is not detected and composite is being output
#hdmi_force_hotplug=1

# uncomment to force a specific HDMI mode (this will force VGA)
#hdmi_group=1
#hdmi_mode=1

# uncomment to force a HDMI mode rather than DVI. This can make audio work in
# DMT (computer monitor) modes
#hdmi_drive=2

# uncomment to increase signal to HDMI, if you have interference, blanking, or
# no display
#config_hdmi_boost=4

[...]

PR #3478, opened 2023-02-15
sugarizer-server-1.5.0

holta:
MongoDB minimal version is now 3.2+

Further detail:
https://github.com/llaske/sugarizer-server/releases/tag/v1.5.0

Related:

- PR #3099
- PR #3214
- PE #3238
- PR #3241
- PR #3469

PR #3518, opened 2023-03-31
MongoDB block on RasPiOS-32bit

jvonau:
### Fixes bug:
https://github.com/iiab/iiab/issues/3516#issuecomment-1490656423 will effect medium installs going forward
https://github.com/iiab/iiab/pull/3478#issuecomment-1444395170 won't fix
Ref: https://github.com/iiab/iiab/issues/2873#issuecomment-884362675  #2880 
adds dpkg_arch that could be used in other roles
Don't install MongoDB on 32bit userspace
RasPiOS 32bit with all updates running latest 64bit kernel tested with runrole
http://sprunge.us/nyhaCL?en

### DIFF
--- a/roles/0-init/tasks/create_iiab_ini.yml
+++ b/roles/0-init/tasks/create_iiab_ini.yml
@@ -43,3 +43,5 @@
       value: "{{ rpi_model }}"
     - option: devicetree_model
       value: "{{ devicetree_model }}"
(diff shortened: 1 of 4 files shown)


jvonau:
Better installers use [dpkg](https://downloads.remote.it/remoteit/install_agent.sh) to get the userspace

PR #3439, opened 2022-12-22
Dynamically extract php_version & python_version

holta:
Uses `apt list php` to auto-determine PHP version, allowing IIAB installs to be less bureaucratic & more reliable, to help with testing and pre-release OS's like Debian 12 Bookworm and Ubuntu 23.04 Lunar Lobster.

Also renames `php_ver` to `php_version`

This PR tested on the latest Ubuntu 23.04

Testing is ongoing on Ubuntu 22.04

Building on:

- PR #3438

PR #3237, opened 2022-06-07
test.yml: Document ansible_local.local_facts, ansible_architecture etc

holta:
In support of Ansible literacy for IIAB in general.

As affects situations similar to:

- iiab/iiab-admin-console#496
- #3236

PR #3520, opened 2023-03-31
Record arch

jvonau:
>Avoid expanding [scripts/local_facts.fact](https://github.com/iiab/iiab/blob/master/scripts/local_facts.fact) which should shrink in coming years, rather than expanding — to reduce the number of global variables (in favor of modular-not-monolithic code, whenever possible).

Should meet the above requirement, just record the value so it can be easily displayed in the admin-console

### DIFF
--- a/roles/0-init/tasks/create_iiab_ini.yml
+++ b/roles/0-init/tasks/create_iiab_ini.yml
@@ -4,6 +4,10 @@
     path: "{{ iiab_ini_file }}"
     state: touch
 
+- name: Run command 'dpkg --print-architecture' to identify OS architecture (CPU arch as revealed by ansible_archicture ~= ansible_machine is NOT enough!)
+  command: dpkg --print-architecture
+  register: dpkg_arch
+
 - name: Add 'location' variable values to {{ iiab_ini_file }}
   ini_file:
     path: "{{ iiab_ini_file }}"
@@ -44,4 +48,4 @@
     - option: devicetree_model
       value: "{{ devicetree_model }}"
     - option: dpkg_arch
-      value: "{{ dpkg_arch }}"
+      value: "{{ dpkg_arch.stdout }}"
--- a/roles/0-init/tasks/create_iiab_ini.yml
+++ b/roles/0-init/tasks/create_iiab_ini.yml
@@ -4,14 +4,18 @@
     path: "{{ iiab_ini_file }}"
     state: touch
 
-- name: Run command 'dpkg --print-architecture' to identify OS architecture (CPU arch as revealed by ansible_archicture ~= ansible_machine is NOT enough!)
+- name: Run command 'dpkg --print-architecture' to identify OS architecture (CPU arch as revealed by ansible_architecture ~= ansible_machine is NOT enough!)
   command: dpkg --print-architecture
   register: dpkg_arch
+
+- name: Run command 'dpkg --print-foreign-architectures' (secondary OS arch, if available)
+  command: dpkg --print-foreign-architectures
+  register: dpkg_foreign_arch
 
 - name: Add 'location' variable values to {{ iiab_ini_file }}
   ini_file:
     path: "{{ iiab_ini_file }}"
-    section: location
+    section: initial-location
     option: "{{ item.option }}"
     value: "{{ item.value | string }}"
   with_items:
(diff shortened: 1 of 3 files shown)


holta:
Both `dpkg --print-architecture` and `dpkg --print-foreign-architectures` are already recorded in iiab-summary and iiab-diagnostics output.

Are they also needing to be recorded during IIAB's initial Ansible install, is that the idea?

If so, that seems reasonable.

(`0-init/tasks/create_iiab_ini.yml` might then spell that out clarifying a bit in `/etc/iiab/iiab.ini` ...maybe something like `option: original_dpkg_arch` or `option: initial_dpkg_arch` or some such?)

holta:
> (`0-init/tasks/create_iiab_ini.yml` might then spell that out clarifying a bit in `/etc/iiab/iiab.ini` ...maybe something like `option: original_dpkg_arch` or `option: initial_dpkg_arch` or some such?)

Or maybe the `section: version` and `section: location` sections within `/etc/iiab/iiab.ini` are both sufficiently static already (implying these 2 sections created by create_iiab_ini.yml are not supposed to ever change ?)

jvonau:
> > (`0-init/tasks/create_iiab_ini.yml` might then spell that out clarifying a bit in `/etc/iiab/iiab.ini` ...maybe something like `option: original_dpkg_arch` or `option: initial_dpkg_arch` or some such?)
> 
> Or maybe the `section: version` and `section: location` sections within `/etc/iiab/iiab.ini` are both sufficiently static already (implying these 2 sections created by create_iiab_ini.yml are not supposed to ever change ?)

You should know the purpose and reasons for the layout that was **'borrowed'** from [xsce](https://github.com/XSCE/xsce/blame/release-6.1/roles/0-once/tasks/xsce_ini.yml) but to spell it out for you the reason is to capture the install time variables that were in play at install time. Remember 'runtime' is updated every time ansible runs, it looks like a duplication but as you now realize create_iiab_ini.yml only runs once. Using 'original' or 'initial' really doesn't gain anything other that being a nit picker, should never change without willful alteration and that is the purpose of recording the value. (Thinking towards apt repo additions that are not correct for the arch in use and how that came into being) Now with this 'Avoid expanding scripts/local_facts.fact', all #3518 was doing was the same procedure that was used with #3439 as I felt this variable would of come in handy for other roles that might need attention going forward. What I find amusing is this 'to reduce the number of global variables' cope out, default_vars.yml and local_vars.yml are full of 'global variables'.

holta:
Sure.

But those 2 section titles should really be...something like `section: initial-location` and `section: initial-version` if we want other people to understand our work.

jvonau:
Now if you want to add --print-foreign-architectures also that would be cool too. While partially a duplication of iiab-summary but that is a post-install tool that would not capture the 'out of the box' pre-iiab environment that might be altered in some unknown way. Just recording the known starting point, just like any other values that are recorded in the same stanza that should not change without willful intervention.

> Sure.
> 
> But those 2 section titles should really be...something like `section: initial-location` and `section: initial-version` if we want other people to understand our work.

Just a header change feel free.

holta:
> Now if you want to add --print-foreign-architectures also that would be cool too. While partially a duplication of iiab-summary but that is a post-install tool that would not capture the 'out of the box'

Ok

jvonau:
['ansible_distribution'](https://github.com/iiab/iiab/blob/master/roles/0-init/tasks/create_iiab_ini.yml#L27) might be better with ['os_ver'](https://github.com/iiab/iiab/blob/master/roles/0-init/tasks/main.yml#L15) ?

holta:
> ['ansible_distribution'](https://github.com/iiab/iiab/blob/master/roles/0-init/tasks/create_iiab_ini.yml#L27) might be better with ['os_ver'](https://github.com/iiab/iiab/blob/master/roles/0-init/tasks/main.yml#L15) ?

FWIW after Ansible 2.6+ they ask us not to use `ansible_distribution` — for whatever reason, one is supposed to use `ansible_facts['distribution']` instead:

https://docs.ansible.com/ansible/latest/playbook_guide/playbooks_conditionals.html#ansible-facts-distribution

- https://github.com/iiab/iiab/pull/3237#issuecomment-1148810125

jvonau:
os_ver matches the form of other entries in create_iiab_ini.yml, this is internal information, should match the detection in use and helps explain OS.yml usage.

jvonau:
Grouping: [Move](https://github.com/iiab/iiab/pull/3520/files#diff-2ea56d1f6e703ae998eacb1c2ad5dfdb282c05a94ec11108ba61190832578954L42-L45) to be under dpkg_foreign_arch?

jvonau:
Looks fine to me, run with it.

PR #3521, opened 2023-04-01
Modernize Yarn install w/ signed apt/PPA key (prereq for Internet Archive)

holta:
This avoids the ugly warnings whenever running `apt update` that had been blocking IIAB installs earlier:

- PR iiab/iiab-factory#236

Modernized apt/PPA key-signing inspired by:

- #3356
- [roles/mongodb/tasks/install.yml](https://github.com/iiab/iiab/blob/master/roles/mongodb/tasks/install.yml)

Tested on Ubuntu 22.04

cc: @mitra42 (hope this is OK!)


mitra42:
If it installs fine then sounds good - I'm traveling and mostly offline so can't test.

Issue #3526, opened 2023-04-04
How to pre-test RasPiOS 12 [cffi, cryptography, lxml "wheels" for Python 3.11 now exist!]


jvonau:
Install the latest arm64 image of RasPiOS and boot
```
sudo apt update
sudo apt -y upgrade
sudo apt -y autoremove
sudo reboot 
```
```
sudo sed -i 's/bullseye/bookworm/g' /etc/apt/sources.list

sudo apt update
sudo apt -y dist-upgrade
# Don't go too far away there a couple of confirmations to confirm just press enter when prompted.
# Any changed files should be a result of the image creation process I would presume. 
# I hate suggesting that method but you can stop and read the actions and decide for yourself.

sudo apt -y autoremove
sudo reboot
```
sudo raspi-config and set the country code for the WiFi and other settings. 
Might want to try the NetworkManager option at this time.
Edit: Make the required changes in local_facts.fact and provide a OS.yml file is covered with #3525 not needed now.
Install IIAB as per normal.

jvonau:
Just a how to note, not for general consumption but I thought I would share.

holta:
> Install IIAB as per normal.

Did an IIAB install largely complete?

(Either way, instructions like this are indeed extremely useful for newcomers.)

Related:

- #3399

jvonau:
I went with the 'none' version of local_vars as a starting point, installed clean.

jvonau:
FYI

jerryv@box:~ $ apt list python2*
Listing... Done

jvonau:
https://github.com/iiab/iiab/issues/3289#issuecomment-1497728119
> perhaps using the last deb package(s) released from an older install.

http://ftp.debian.org/debian/pool/main/p/python2.7/?C=M;O=D
```
wget http://ftp.debian.org/debian/pool/main/p/python2.7/libpython2.7-minimal_2.7.18-8_arm64.deb
sudo apt install ./libpython2.7-minimal_2.7.18-8_arm64.deb
wget http://ftp.debian.org/debian/pool/main/p/python2.7/python2.7-minimal_2.7.18-8_arm64.deb
sudo apt install ./python2.7-minimal_2.7.18-8_arm64.deb
```
The order is important and are 'dpkg_arch' dependent, that was the reasoning behind my original #3518 having dpkg_arch available for other roles.

holta:
@jvonau mentions at https://github.com/iiab/iiab/pull/3535#issuecomment-1503771152 :

[...]

PR #3527, opened 2023-04-06
Python2 fix [for KA Lite on Debian 12 or Ubuntu 23.04 / 23.10]

jvonau:
### Fixes bug:
#3450 #3399 #3289 
### Description of changes proposed in this pull request:
see new file
### Smoke-tested on which OS or OS's:
http://sprunge.us/rtK5U5 My home-brew RasPiOS-12arm64 test box #3526 Should work on Ubuntu-23.04 and Debian-12 also
>1478  108343 ?        00:00:01   kalite
```
1994 2023-04-05 20:53:33,009 p=106685 u=root n=ansible | TASK [0-init : STAGE 0 HAS COMPLETED ======================================] ******************************************************
1995 2023-04-05 20:53:33,096 p=106685 u=root n=ansible | TASK [kalite : Assert that "kalite_install is sameas true" (boolean not string etc)] **********************************************
1996 2023-04-05 20:53:33,184 p=106685 u=root n=ansible | ok: [127.0.0.1]
1997 2023-04-05 20:53:33,237 p=106685 u=root n=ansible | TASK [kalite : Assert that "kalite_enabled | type_debug == 'bool'" (boolean not string etc)] **************************************
1998 2023-04-05 20:53:33,303 p=106685 u=root n=ansible | ok: [127.0.0.1]
1999 2023-04-05 20:53:33,355 p=106685 u=root n=ansible | TASK [kalite : Install KA Lite if 'kalite_installed' not defined, e.g. in /etc/iiab/iiab_state.yml] *******************************
2000 2023-04-05 20:53:33,510 p=106685 u=root n=ansible | included: /opt/iiab/iiab/roles/kalite/tasks/install.yml for 127.0.0.1
2001 2023-04-05 20:53:33,585 p=106685 u=root n=ansible | TASK [kalite : Download https://raw.githubusercontent.com/learningequality/ka-lite/master/requirements.txt to /opt/iiab/pip-packages/kalite.txt] ***
2002 2023-04-05 20:53:35,535 p=106685 u=root n=ansible | ok: [127.0.0.1]
2003 2023-04-05 20:53:35,586 p=106685 u=root n=ansible | TASK [kalite : Install packages: python2, python-setuptools, virtualenv (for Python 2)] *******************************************
2004 2023-04-05 20:53:35,641 p=106685 u=root n=ansible | skipping: [127.0.0.1]
[...]

PR #3529, opened 2023-04-09
ka-lite libssl1.1 use raspi repo

jvonau:
### Fixes bug:
On the off chance **deb11u4+rpt1_arm64** is already installed **deb11u4_arm64** would refuse to install as apt sees the '+rpt1' as a later version. The only situation I can think where this might come in to play of is in an upgrade of RasPiOS-11 to 12 #3526 but at least cover that possibility.

Change to use archive.raspberrypi.org repo for arm64
3526


holta:
Refines:

- PR #3527

Issue #3512, opened 2023-03-27
Linphone sip extensions cannot be reached when iphone or desktop is locked.It seems this is a known issue.Online solutions advise to enable push notifications but still doesn’t work.Apparently a battery saving trick from Apple. I am trying to source android phones to test in meantime. I have tried Zoiper but push notification is a paid for feature.All solutions welcome. https://github.com/balusreekanth/ios-asterisk-push [Asterisk, FreePBX]


EMG70:

holta:
Thank you @EMG70 for reporting ✅

If possible let's link to the most pragmatic Android workaround (or more real solution) if you can suggest one for IIAB's Guide below, in the coming week or so?

https://github.com/iiab/iiab/tree/master/roles/pbx#readme

cc: @lemueldsouza

EMG70:
> Thank you @EMG70 for reporting ✅
> 
> If possible let's link to the most pragmatic Android workaround (or more real solution) if you can suggest one for IIAB's Guide below, in the coming week or so?
> 
> https://github.com/iiab/iiab/tree/master/roles/pbx#readme
> 
> cc: @lemueldsouza

OK I will look into Android solutions in the coming week.

holta:
ASIDE: @EMG70 worked hard today trying a few different things to get Asterisk 19 or 20 working on 64-bit Raspberry Pi OS:

- #3489

Sorry it appears not to have worked out 🕊️

holta:
Workaround / solution posted to pbx/README.adoc thanks to @EMG70 !

- PR #3533

PR #3533, opened 2023-04-10
Pbx readme [Wave Lite app option, so Android phones actually ring when screen's locked etc]

EMG70:
Linphone does not have push notifications (cannot be reached when screen is locked) so I have added Wave lite app as an option.


holta:
@lemueldsouza please take a look if you can in coming days, Thanks!

This PR is a workaround / solution for:

- #3512

lemueldsouza:
Looks good to me! 
Thanks @EMG70 for reporting and helping fix the issue, keep up the good work!

Issue #3434, opened 2022-12-13
munin-node.service doesn't start until reboot [if IPv6 has just been disabled by sysctl] on Ubuntu 23.04 (Lunar Lobster) final release [similar on Debian 12 (Bookworm)]


holta:
Hardly surprising, given Ubuntu 23.04 is in rapid flux 4 months prior to release (i.e. this worked last month, and might well fix itself in coming months or weeks).

Still, here's the current error just for the record:

> TASK [munin : Enable & Start 'munin-node' systemd service] *************************************************************************************************************
fatal: [127.0.0.1]: FAILED! => {"changed": false, "msg": "Unable to start service munin-node: Job for munin-node.service failed because the control process exited with error code.\nSee \"systemctl status munin-node.service\" and \"journalctl -xeu munin-node.service\" for details.\n"}

jvonau:
> See "systemctl status munin-node.service" and "journalctl -xeu munin-node.service" for details.

And those details are what?

holta:
Unclear it's worth spending any time at all on this.

Nevertheless, here are details just in case:

```
# systemctl status munin-node.service
× munin-node.service - Munin Node
     Loaded: loaded (/lib/systemd/system/munin-node.service; enabled; preset: enabled)
     Active: failed (Result: exit-code) since Thu 2022-12-15 09:41:35 EST; 8min ago
       Docs: man:munin-node(1)
             http://guide.munin-monitoring.org/en/stable-2.0/reference/munin-node.html
    Process: 25949 ExecStartPre=/usr/bin/install -o munin -g munin -d /run/munin (code=exited, status=0/SUCCESS)
    Process: 25950 ExecStart=/usr/sbin/munin-node --foreground $DAEMON_ARGS (code=exited, status=1/FAILURE)
   Main PID: 25950 (code=exited, status=1/FAILURE)
        CPU: 104ms

Dec 15 09:41:35 box systemd[1]: munin-node.service: Scheduled restart job, restart counter is at 5.
Dec 15 09:41:35 box systemd[1]: Stopped Munin Node.
Dec 15 09:41:35 box systemd[1]: munin-node.service: Start request repeated too quickly.
Dec 15 09:41:35 box systemd[1]: munin-node.service: Failed with result 'exit-code'.
Dec 15 09:41:35 box systemd[1]: Failed to start Munin Node.
```

```
# journalctl -xeu munin-node.service | head -80
[...]

PR #3560, opened 2023-04-30
Temporarily Enable sysctl's IPv6 during Munin install/enable, so munin-node.service starts

holta:
Resolves:

- #3434

Tested on Ubuntu Server 23.04


holta:
Also tested with a LARGE-sized IIAB install on Debian 12 RC2+ here:

iiab-diagnostics: http://sprunge.us/IiIYRG?en

Issue #3539, opened 2023-04-17
vnStat error when building iiab [on Debian 11]


EMG70:
### Steps to reproduce the problem:
Tried running iiab many times and been failing in past few days at same stage.
### IIAB version, OS version, hardware details:
Iiab 8.1 ,Debian 11

### Please run the `iiab-diagnostics` 
command, to generate a URL summarizing your IIAB configuration, for volunteers seeking to help you: (Include that short URL here, Thanks!)

http://sprunge.us/wEnNC5?en

holta:
Thanks @EMG70 for reporting!
Highlighting the error message here:

>  
1910 2023-04-17 15:09:31,923 p=3937 u=root n=ansible | TASK [vnstat : Create database for WAN to collect vnStat data] *****************
1911 2023-04-17 15:09:32,159 p=3937 u=root n=ansible | fatal: [127.0.0.1]: FAILED! => {"changed": true, "cmd": "/usr/bin/vnstat -i wlp2s0", "delta": "0:00:00.004179", "end": "2023-04-17 15:09:32.135351", "msg": "non-zero return code", "rc": 1, "start": "2023-04-17 15:09:32.131172", "stderr": "", "stderr_lines": [], "stdout": "Error: Failed to open database \"/var/lib/vnstat/vnstat.db\" in read-only mode.\nThe vnStat daemon should have created the database when started.\nCheck that it is configured and running. See also \"man vnstatd\".", "stdout_lines": ["Error: Failed to open database \"/var/lib/vnstat/vnstat.db\" in read-only mode.", "The vnStat daemon should have created the database when started.", "Check that it is configured and running. See also \"man vnstatd\"."]}

holta:
FYI vnStat ("a network traffic monitor for Linux and BSD") may not be commonly used with IIAB, but at least it's actively maintained:

https://github.com/vergoh/vnstat/releases

holta:
ASIDE: IIAB currently uses this `/etc/vnstat.conf` template derived from 2011-06-01's vnStat 1.11

https://github.com/iiab/iiab/blob/master/roles/vnstat/templates/vnstat.conf.j2

If vnStat is genuinely useful to certain IIAB implementers/communities, this should probably be updated to vnStat 2.10 (released 2022-10-22) or similar?

holta:
vnStat's changelog, covering more than 20 years:

https://humdi.net/vnstat/CHANGES

holta:
[...]

PR #3562, opened 2023-05-01
Toughen up vnStat: Mitigate #3539 db errors, stop using 2011's /etc/vnstat.conf, etc

holta:
Tested on Debian 11, trying to reproduce the conditions of:

- #3539


holta:
Good enough for now!

- IIAB's customizing of apt's version of `/etc/vnstat.conf` (i.e. customizing line `Interface ""` using `iiab_wan_iface: "{{ ansible_default_ipv4.alias }}"`) is probably no longer necessary — given that vnStat makes clear that it automatically selects the default interface.
- We can tighten up in a future PR if the much simpler strategy (i.e. using `/etc/vnstat.conf` as it evolves from apt, without any customizations!) is tested as Solid Enough on a couple different OS's.

PR #3482, opened 2023-02-25
Set WiFi firmware in /lib/firmware/cypress due to RasPiOS & Ubuntu changes

holta:
This PR restores IIAB implementers' ability to select WiFi firmware on most Raspberry Pi's.

This is necessary because Raspberry Pi OS (and Ubuntu now too) have recently changed how WiFi firmware is selected for 43430 and 43455 chips (i.e. the Broadcom/Cypress/Infineon WiFi/Bluetooth chip used in most Raspberry Pi's).

Specifically, selecting a preferred WiFi firmware in directory `/lib/firmware/brcm` is no longer workable — with Raspberry Pi's that use these 2 chips — as can be seen from these newer (OS-provided) symlinks here:

```
root@box:/lib/firmware/brcm# ls -l brcm*rasp* | grep '\.\.' | cut -d' ' -f 9-
brcmfmac43430-sdio.raspberrypi,3-model-b.bin -> ../cypress/cyfmac43430-sdio.bin
brcmfmac43430-sdio.raspberrypi,3-model-b.clm_blob -> ../cypress/cyfmac43430-sdio.clm_blob
brcmfmac43430-sdio.raspberrypi,model-zero-w.bin -> ../cypress/cyfmac43430-sdio.bin
brcmfmac43430-sdio.raspberrypi,model-zero-w.clm_blob -> ../cypress/cyfmac43430-sdio.clm_blob
brcmfmac43455-sdio.raspberrypi,3-model-a-plus.bin -> ../cypress/cyfmac43455-sdio.bin
brcmfmac43455-sdio.raspberrypi,3-model-a-plus.clm_blob -> ../cypress/cyfmac43455-sdio.clm_blob
brcmfmac43455-sdio.raspberrypi,3-model-b-plus.bin -> ../cypress/cyfmac43455-sdio.bin
brcmfmac43455-sdio.raspberrypi,3-model-b-plus.clm_blob -> ../cypress/cyfmac43455-sdio.clm_blob
brcmfmac43455-sdio.raspberrypi,4-model-b.bin -> ../cypress/cyfmac43455-sdio.bin
brcmfmac43455-sdio.raspberrypi,4-model-b.clm_blob -> ../cypress/cyfmac43455-sdio.clm_blob
```

Distinct from the "10 above" symlinks (that lead to directory `/lib/firmware/cypress`) — most of the rest (of the other 18) still use `/lib/firmware/brcm` interestingly — at least for now:

```
[...]

PR #3565, opened 2023-05-02
Clarify PR #3564 (pre-download WiFi firmware backups)

holta:
Refines this PR to be more readable & understandable:

- PR #3564

Building on:

- PR #3482


holta:
Tested on 64-bit RasPiOS Lite with a diversity of different setups in `/lib/firmware/cypress`

Issue #3574, opened 2023-05-13
armhf version of kiwix-tools 3.5.0 fails on all Raspberry Pi's — 32-bit OR 64-bit — regardless of OS [WAS: Fresh Install. Home page links not working] [502 Bad Gateway, medical preset]


stebbo63:
Hello,
I am completely new to iiab.  please excuse any errors on my part.

I have installed iiab on a new raspberry pi.  I selected "m" during the initial install process.  I then went to "Install Content" and selected "Medical Content for English Speakers".

Both the initial install and the install content seemed to go well.

However, many of the home screen menu options are not working.  I keep getting "502 Bad Gateway" on many of the options.

For example, the following options (not a complete list) produce a 502 Bad Gateway

- MDWiki Medical Encyclopedia with Videos
- StatsPearls Article Library
- NIH Genetic and Rare Diseases
and more.

Some menu options do work, for example

- The Global Emergency Medicine Wiki
- Healthphone Medical Video Collection in English

This is the URL for the MDWiki option which doesn't work 
http://box.lan/kiwix/viewer#mdwiki_en_all_2023-03/

Any assistance would be greatly appreciated.

Regards,
Chris.

[stebbo.txt](https://github.com/iiab/iiab/files/11469261/stebbo.txt)

holta:
Thanks @stebbo63 for the report.

We'll look into it as soon as possible in coming days trying to reproduce what you're describing.

> the install content seemed to go well.

Have enough hours gone by, for the content downloads to complete?

Are there any progress/problem indications if you look around **Admin Console (http://box.lan/admin) > Utilities > Display Job Status** ?

tim-moody:
What happens if you go to url http://box.lan/kiwix/ Let's make sure kiwix is installed and working.

tim-moody:
just noticed your stebbo.txt above which is iiab-diagnostics and shows kiwix running, but still good to check.

holta:
Indeed /etc/iiab/iiab_state.yml shows:

`kiwix_installed: True`

Just FYI kiwix-tools 3.5.0 is very new as of recent days, so this line of inquiry is a good one, towards verifying that.

tim-moody:
not sure what to make of 

```
May 13 18:42:44 box systemd[1]: Stopped Provides the Kiwix ZIM File Server.
May 13 18:42:44 box systemd[1]: Starting Provides the Kiwix ZIM File Server...
[...]

PR #3422, opened 2022-11-10
Block iiab-install of JupyterHub & Calibre-Web on 32-bit OS's, til RasPiOS fixes Rust, wheels / cryptography

holta:
These 2 apps should probably not be running on ancient 32-bit "Zero W" machines in the first place!

(But Calibre-Web is part of  the MEDIUM-sized IIAB install, so this will help newcomers from getting tripped up installing something that they shouldn't...)

See:

- "Known Issues" at: https://github.com/iiab/iiab/wiki/IIAB-8.0-Release-Notes#known-issues
- @jvonau's suggestion at: https://github.com/iiab/iiab/pull/3421#issuecomment-1310597316
- https://github.com/iiab/iiab/pull/3459#issuecomment-1378928044

PR #3575, opened 2023-05-13
Try aarch64 version of kiwix-tools 3.5.0+ [WARNING: such microSD cards will no longer be insertable in 32-bit Raspberry Pi Zero W]

holta:
The downside (even if this does work well on 64-bit Raspberry Pis!) is that microSD cards will no longer be insertable into older 32-bit Raspberry Pi "Zero W" machines 😢

Tangentially related:

- #3516
  - Explained at: https://github.com/iiab/iiab/pull/3422#issuecomment-1533441463
- #3574

### DIFF
--- a/roles/kiwix/defaults/main.yml
+++ b/roles/kiwix/defaults/main.yml
@@ -8,11 +8,11 @@
 # If nec, change them by editing /etc/iiab/local_vars.yml prior to installing!
 
 
-# INSTRUCTIONS TO REINSTALL Kiwix:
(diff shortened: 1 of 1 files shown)


holta:
CLARIF that this PR is _also_ a partial workaround for the very serious/sudden failures of the `armhf` version of kiwix-tools 3.5.0 across all Raspberry Pi's:

- #3574

IN SHORT: This PR appears to work as a workaround in most 64-bit Raspberry Pi scenarios.

PR #3578, opened 2023-05-15
Fix PR #3576, allowing regular Kiwix installs too!

holta:
Oops: once again bitten by Ansible overwriting a register even when a stanza's `when:` clause is not triggered.

Tested on 32-bit Raspberry Pi OS, and 64-bit Raspberry Pi OS testing is ongoing...

Patches:

- PR #3576

 Building on:

- #3574
- PR #3575


holta:
> 64-bit Raspberry Pi OS testing is ongoing...

Also confirmed ✅

Issue #3566, opened 2023-05-03
Testing IIAB with Node.js 20, instead of Node.js 18


holta:
FYI Node.js 20 was released on 2023-04-18 and will become LTS on 2023-10-24.

6 days before that, on 2023-10-18, Node.js 18 will cease to be "Active LTS" as it becomes "Maintenance LTS":

- https://nodejs.dev/en/about/releases/
- https://nodejs.org/en/download/releases
- https://nodejs.org/en/blog

ASIDE: Node.js 19 reaches End-of-Life in a few weeks from now, on 2023-06-01.

Related:

- PR #3466

holta:
Node.js 20 retested will all IIAB apps (LARGE-sized).

PR #3584, opened 2023-05-20
Begin transition to 'nodejs_version: 20.x' [& internetarchive upstream patch]

holta:
Others have successfully transitioned from Node.js 18 to 20 in the past month, so likely no need to wait for October when it more officially becomes LTS: 

- #3566

### DIFF
--- a/roles/internetarchive/tasks/install.yml
+++ b/roles/internetarchive/tasks/install.yml
@@ -9,10 +9,10 @@
   include_role:
     name: nodejs
 
-- name: Assert that 10.x <= nodejs_version ({{ nodejs_version }}) <= 19.x
+- name: Assert that 10.x <= nodejs_version ({{ nodejs_version }}) <= 20.x
   assert:
-    that: nodejs_version is version('10.x', '>=') and nodejs_version is version('19.x', '<=')
-    fail_msg: "Internet Archive install cannot proceed, as it currently requires Node.js 10.x - 19.x, and your nodejs_version is set to {{ nodejs_version }}.  Please check the value of nodejs_version in /opt/iiab/iiab/vars/default_vars.yml and possibly also /etc/iiab/local_vars.yml"
+    that: nodejs_version is version('10.x', '>=') and nodejs_version is version('20.x', '<=')
+    fail_msg: "Internet Archive install cannot proceed, as it currently requires Node.js 10.x - 20.x, and your nodejs_version is set to {{ nodejs_version }}.  Please check the value of nodejs_version in /opt/iiab/iiab/vars/default_vars.yml and possibly also /etc/iiab/local_vars.yml"
     quiet: yes
 
 - name: "Set 'yarn_install: True' and 'yarn_enabled: True'"
--- a/roles/nodejs/README.md
+++ b/roles/nodejs/README.md
@@ -8,15 +8,15 @@ Nodesource.com often supports Debian long before each Debian release, whereas fo
 For late-breaking details on Nodesource.com support for your particular Linux OS, keep an eye on:
 
 - https://github.com/nodesource/distributions#deb
-- https://deb.nodesource.com/node_18.x/dists/
-  - https://deb.nodesource.com/node_18.x/pool/main/n/nodejs/
-  - https://nodejs.org/dist/latest-v18.x/
+- https://deb.nodesource.com/node_20.x/dists/
+  - https://deb.nodesource.com/node_20.x/pool/main/n/nodejs/
+  - https://nodejs.org/dist/latest-v20.x/
 - https://deb.nodesource.com/node_19.x/dists/
   - https://deb.nodesource.com/node_19.x/pool/main/n/nodejs/
   - https://nodejs.org/dist/latest-v19.x/
 - _ETC!_
 
-If Nodesource.com does not yet support your Linux OS and IIAB's asked to install Node.js &mdash; IIAB will then [fall back](https://github.com/iiab/iiab/blob/15a689e30b4eea325e4bb8d35e19990dd8062fbc/roles/nodejs/tasks/install.yml#L103-L107) to: (running the equivalent of)
+If Nodesource.com does not yet support your Linux OS and IIAB's asked to install Node.js &mdash; IIAB will then [fall back](https://github.com/iiab/iiab/blob/91a5cd33f34d5d2a55e75bf0cdc85bcd9d7b4821/roles/nodejs/tasks/install.yml#L103-L107) to: (running the equivalent of)
 
 ```
 sudo apt install nodejs npm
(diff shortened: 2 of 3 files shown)


holta:
@mitra42 as we prepare for Node.js 20, can you take quick look at this error?

> 2023-05-20 11:58:03,282 p=2824 u=root n=ansible | TASK [internetarchive : Run 'yarn add https://github.com/internetarchive/dweb-mirror' to download/populate /opt/iiab/internetarchive/node_modules (CAN TAKE ~5 MINUTES)] ***
2023-05-20 11:58:10,427 p=2824 u=root n=ansible | fatal: [127.0.0.1]: FAILED! => {"changed": true, "cmd": "yarn config set child-concurrency 1 && yarn add https://github.com/internetarchive/dweb-mirror", "delta": "0:00:06.977257", "end": "2023-05-20 11:58:10.405128", "msg": "non-zero return code", "rc": 128, "start": "2023-05-20 11:58:03.427871", "stderr": "warning https://github.com/internetarchive/dweb-mirror > multihashes > multibase@4.0.6: This module has been superseded by the multiformats module\nwarning https://github.com/internetarchive/dweb-mirror > @internetarchive/dweb-archive-dist > react > fbjs > core-js@1.2.7: core-js@<3.23.3 is no longer maintained and not recommended for usage due to the number of issues. Because of the V8 engine whims, feature detection in old core-js versions could cause a slowdown up to 100x even if nothing is polyfilled. Some versions have web compatibility issues. Please, upgrade your dependencies to the actual version of core-js.\nwarning https://github.com/internetarchive/dweb-mirror > @futurepress/epubjs-reader > epubjs > @types/localforage@0.0.34: This is a stub types definition for localforage (https://github.com/localForage/localForage). localforage provides its own type definitions, so you don't need @types/localforage installed!\nerror Command failed.\nExit code: 128\nCommand: git\nArguments: ls-remote --tags --heads https://github.com/mitra42/readable-stream-clone.git\nDirectory: /opt/iiab/internetarchive\nOutput:\nfatal: could not read Username for 'https://github.com': terminal prompts disabled", "stderr_lines": ["warning https://github.com/internetarchive/dweb-mirror > multihashes > multibase@4.0.6: This module has been superseded by the multiformats module", "warning https://github.com/internetarchive/dweb-mirror > @internetarchive/dweb-archive-dist > react > fbjs > core-js@1.2.7: core-js@<3.23.3 is no longer maintained and not recommended for usage due to the number of issues. Because of the V8 engine whims, feature detection in old core-js versions could cause a slowdown up to 100x even if nothing is polyfilled. Some versions have web compatibility issues. Please, upgrade your dependencies to the actual version of core-js.", "warning https://github.com/internetarchive/dweb-mirror > @futurepress/epubjs-reader > epubjs > @types/localforage@0.0.34: This is a stub types definition for localforage (https://github.com/localForage/localForage). localforage provides its own type definitions, so you don't need @types/localforage installed!", "error Command failed.", "Exit code: 128", "Command: git", "Arguments: ls-remote --tags --heads https://github.com/mitra42/readable-stream-clone.git", "Directory: /opt/iiab/internetarchive", "Output:", "fatal: could not read Username for 'https://github.com': terminal prompts disabled"], "stdout": "yarn config v1.22.19\nsuccess Set \"child-concurrency\" to \"1\".\nDone in 0.03s.\nyarn add v1.22.19\ninfo No lockfile found.\n[1/4] Resolving packages...\ninfo Visit https://yarnpkg.com/en/docs/cli/add for documentation about this command.", "stdout_lines": ["yarn config v1.22.19", "success Set \"child-concurrency\" to \"1\".", "Done in 0.03s.", "yarn add v1.22.19", "info No lockfile found.", "[1/4] Resolving packages...", "info Visit https://yarnpkg.com/en/docs/cli/add for documentation about this command."]}

FYI the above is a stock IIAB install (with this PR preparing for Node.js 20) on Ubuntu 22.04 LTS.

tim-moody:
Where possible I think it is right to move to the latest version of apps, so moodle, kolibri, sugarizer. But php, python, nodejs, database software are in another category, as app dependencies. So just as with python we have moved to venv, perhaps we should move to nvm for nodejs. The version we run really depends on the needs of apps that use it. In my observation, apps lag behind as upgrading node forces upgrades to packages and dependency hell.

holta:
> perhaps we should move to nvm for nodejs

It's an idea.  At the same time, for what it's worth, IIAB had no issues or problems transitioning from Node.js 10.x to 12.x to 14.x to 16.x to 18.x over many years now.

Possibly this current version 20 question is just a very rare case (@mitra42 might be able to confirm?)

Certainly Python and PHP compatibility problems are much more common, that's very clear !  (In any case, just to point out we haven't run into Node.js compatibility problems over recent years — so for whatever reason such issues are extremely rare among IIAB apps/roles to date.)

tim-moody:
which apps do we support that use nodejs?

tim-moody:
https://nodejs.org/en/blog/announcements/nodejs16-eoll

Changing the End-of-Life Date for Node.js 16 to September 11th, 2023

18 has security support until 2025

holta:
> which apps do we support that use nodejs?

- Internet Archive (roles/internetarchive)
- JupyterHub (roles/jupyterhub)
- Node-RED (roles/nodered)
- Sugarizer (roles/sugarizer)
- FreePBX (roles/pbx)

> perhaps we should move to nvm for nodejs

Another question is how much work that would be.  My above observation is just that there was lot of Node compatibility turmoil prior to Node.js 10.x, among several IIAB apps/roles, Sugarizer especially.

Whereas that turmoil almost entirely disappeared since then, for whatever reasons.  Presumably as Node.js became more stable & standardized & understood since then?

But hopefully @mitra42 can help shed a bit more light here.

mitra42:
The problem as @tim-moody points out is dependency hell. dweb-mirror probably doesn't do anything specific to node18 or node20, but it depends on things that may or may not have been upgraded, and may themselves depend on older versions of things -

Its hard to read that error message above - please use the triple-backquote to include formatted text but 
```
https://github.com/internetarchive/dweb-mirror > @internetarchive/dweb-archive-dist > react > fbjs > core-js@1.2.7
```
suggests that dweb-archive-dist depends on an older version of react that depends on a no-longer-supported version of core-js. In general that will be OK, because we've tested with that combination, though its possible but unusual that a node version change could break something. Mostly this means an increase of build-time error messages that can be ignored till all the package owners (including me) get around to upgrading and figuring out which version to depend on. 

I think there is a fatal error there due to a dependency of dweb-mirror on my fork of readable-stream-clone (which contained a bug-fix which the owner took a long time to merge back). My fork was deleted recently and it looks like while I had fixed a dependency on it, that fixed hadn't been published on npm - my bad, and I fixed that.  I can confirm that dweb-mirror installs on node18, but I don't have a node20 environment to check this in. 

The main problem you might run into is that while, as @holta says, node dependency problems aren't that common any more, the big issue is packages switching to ESM modules instead of common-js - and using them requires changing code and build processes etc. I've seen this break all kinds of things in dependency chains, for example upgrading to current version of webtorrent took me hours recently since their docs were buggy as well. 

The added complexity is that ESM works much better in a browser than on servers/node, and many of the modules that have switched to ESM don't seem to have tested their module in a server/node environment. 

In addition one way to get around the dependency problems with ESM is to import dependencies from esm.js - that can (often) work fine in browsers, but I'm guessing is going to cause problems in offline usage, and not just for the 5 apps that you list depending on node, many more use browser-side Javascript in their UI.

tim-moody:
I rest my case. I wasn't going to mention react specifically, but I have spent time on react apps where the dependency on the dependency that changed its call signature or is obsolete and disappeared because node was upgraded is a major headache.

I think many node developers now use nvm and pm2 to try to isolate the environment.

holta:
@mitra42 thanks for the very conscientious explanations.  

And for patching things upstream.  IIAB App "internetarchive" (http://box:4244) is now tested to work well on Node.js 20: (whether or not the issue had anything to do with 18 versus 20)

```
root@u2204:~# node -v
v20.2.0
```
```
root@u2204:~# grep internetarchive /etc/iiab/iiab_state.yml
internetarchive_installed: True
```

> I think many node developers now use nvm and pm2 to try to isolate the environment.

Very good to know.

Of course a negative side of "throwing everything over the fence and ignoring it for a few years" is that code + minds get very rusty, as technical debt accumulates sight unseen, and nobody can fix it (nevermind vouch for its security!) ~5 years later.

So all approaches have costs + risks, inevitably — neither software nor security are self-maintaining by any deity above.😉

holta:
@mitra42 thanks again for patching internetarchive yesterday:

People in London and in Zimbabwe just now told me their [roles/internetarchive](https://github.com/iiab/iiab/tree/master/roles/internetarchive) install failed early yesterday, on Node.js 18.

(So I told them to try again today!)

image

mitra42:
Yes - @holta this wasn't a Node18-20 issue, it was an issue with not pushing my fix to dweb-mirror after deleting my fork of readable-stream-clone.

PR #3583, opened 2023-05-20
pin virtualenv to version in U22.04

jvonau:
### Fixes bug:
#3573 
### Description of changes proposed in this pull request:
use version of virtualenv from U22.04
### Smoke-tested on which OS or OS's:
U23-04 VM

PR #3582, opened 2023-05-19
'pip install virtualenv==20.21.1' for Python 2 required by KA Lite on Ubuntu 23.10

holta:
### Fixes bug:

- #3573

### Description of changes proposed in this pull request:

apt install of http://launchpadlibrarian.net/651276954/virtualenv_20.19.0+ds-1_all.deb does not work, as it drags in 20.23+ version of python3-virtualenv, as explained in the PR.  So pip (with all its risks) is used instead.

### Smoke-tested on which OS or OS's:

Ubuntu 23.10

@EMG70 may be able to test on Debian 12 and Ubuntu 23.04 next week.

### Mention a team member @username e.g. to help with code review:

@neomatrixcode figured this out.

@jvonau @tim-moody should take a look if they have time.

Related:

- #3535
- PR #3583

PR #3535, opened 2023-04-11
notes and dependencies [for Python 2 on new OS's on {amd64, arm64, armhf}. Tested on Debian 12 & Ubuntu 23.04]

jvonau:
### Fixes bug:
Pulled the trigger a little fast on #3534 guess I should of tagged it as WIP or Draft
### Description of changes proposed in this pull request:
notes and dependencies

PR #3587, opened 2023-05-21
Pin python3-platformdirs & python3-virtualenv to 22.04 versions, for Python 2 on Ubuntu 23.10+ ? [try apt instead of pip, for KA Lite]

holta:
This PR is essentially what @jvonau outlined at:

- https://github.com/iiab/iiab/pull/3583#issuecomment-1556048916

I suppose only time will tell (with Ubuntu 24.04 LTS about 11 months away) if this more long-winded `apt` approach is either (A) safer... or (B) more fragile... than the `pip` approach implemented by:

- PR #3582

Related:

- PR #3535

### DIFF
--- a/scripts/install_python2.sh
+++ b/scripts/install_python2.sh
@@ -62,14 +62,18 @@ EOF
 esac
 
 apt update
-# 2023-05-20: 4 lines below borrow from Ubuntu 22.04: (Is this really less
-# fragile than the pip approach ~40 lines above, in preparing for 24.04 ?)
-apt -y install python3-platformdirs=2.5.1-1
-apt-mark hold python3-platformdirs
-apt -y install python3-virtualenv=20.13.0+ds-2
-apt-mark hold virtualenv
-# 2023-05-21 PR #3587: Above 4 lines should really install a more recent
-# version of virtualenv, probably from 'lunar' (Ubuntu 23.04) ?
+if grep -qi ubuntu /etc/os-release; then   # Ubuntu 23.10+ (and Mint 22+ ?) need this.  Ubuntu 23.04 tolerates it.
+    # 2023-05-20: 4 lines below borrow from Ubuntu 22.04: (Is this really less
+    # fragile than the pip approach ~40 lines above, in preparing for 24.04 ?)
+    apt -y install python3-platformdirs=2.5.1-1
+    apt-mark hold python3-platformdirs
+    apt -y install python3-virtualenv=20.13.0+ds-2
+    apt-mark hold virtualenv
+    # 2023-05-21 PR #3587: Above 4 lines should really install a more recent
+    # version of virtualenv, probably from 'lunar' (Ubuntu 23.04) ?
+else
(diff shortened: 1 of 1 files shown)


holta:
FYI this approach works with KA Lite, when tested on Ubuntu 23.10:

```
root@u2310:~# virtualenv --version
virtualenv 20.13.0+ds from /usr/lib/python3/dist-packages/virtualenv/__init__.py

root@u2310:~# apt list virtualenv
Listing... Done
virtualenv/mantic 20.23.0+ds-1 all

root@u2310:~# apt list -a python3-platformdirs
Listing... Done
python3-platformdirs/mantic 3.5.0-1 all [upgradable from: 2.5.1-1]
python3-platformdirs/now 2.5.1-1 all [installed,upgradable to: 3.5.0-1]

root@u2310:~# apt list -a python3-virtualenv
Listing... Done
python3-virtualenv/mantic 20.23.0+ds-1 all [upgradable from: 20.13.0+ds-2]
python3-virtualenv/now 20.13.0+ds-2 all [installed,upgradable to: 20.23.0+ds-1]
```

This PR should also be tested on Debian 12 and Ubuntu 23.04, if this `apt` approach (whose version of virtualenv is about 16 months older than the `pip` approach) is strongly preferred.

holta:
This PR is tested on Debian 12 Bookworm's near-final pre-release (KA Lite works).

Raspberry Pi OS 12 will be much more important of course — when that's released in a few short weeks:

- #3526

Finally: if this `apt` approach has legs, we might later improve it using a much more recent version of `virtualenv` from Ubuntu 23.04

holta:
This PR and/or the PR #3582 approach seem to make most sense in preparing for Ubuntu 24.04 LTS.

_Whereas either approach should really be simplified — to use virtualenv 20.17.1 (already works with Python 2) built into Debian 12 and Raspberry Pi OS 12 — for people running those "Bookworm" OS's:_

https://packages.debian.org/bookworm/virtualenv

holta:
Let's give this a shot.

All these rickety mechanisms to make KA Lite work for a couple more years in 2023-2025 will surely need to evolve.

On the positive side it  does work, and lays a decent framework (including in-line docs) for the imminent crop of "Bookworm" OS's and hopefully Ubuntu 24.04 LTS after that.

@jvonau can these 2 warnings be cleaned up on Debian 12 Bookworm, or should we live with them for now?

```
W: GPG error: http://archive.ubuntu.com/ubuntu jammy InRelease: The following signatures couldn't be verified because the public key is not available: NO_PUBKEY 871920D1991BC93C
W: GPG error: http://archive.ubuntu.com/ubuntu jammy-updates InRelease: The following signatures couldn't be verified because the public key is not available: NO_PUBKEY 871920D1991BC93C
```

holta:
> @jvonau can these 2 warnings be cleaned up on Debian 12 Bookworm, or should we live with them for now?

These probably can be ignored.  Either way, I'm happy to merge this in the coming ~24h, regardless what the future holds.

Issue #3473, opened 2023-02-08
Validate Transmission 4.x as Debian / Ubuntu / RasPiOS adopt it in coming months? [4.1.0 ETA 2024 should be more solid]


holta:
Big news ~33 months later...hopefully becoming a part of Debian 12 & many other distros...as Torrenting allows much faster downloading of ZIM files + other Content Packs too:

https://github.com/transmission/transmission/releases/tag/4.0.0

IIAB's Ansible role (that installs & configures Transmission) may need tweaking here:

https://github.com/iiab/iiab/blob/c74b8c751fe831140c58b7c3f876fc54993d7972/roles/transmission/tasks/install.yml#L22-L53

LIkewise Transmission's central settings.json file may need updating or replacing here:

https://github.com/iiab/iiab/blob/master/roles/transmission/templates/settings.json.j2

Related:

- #1078

holta:
FYI Transmission 4.0.1 is ~4 days old and moving fast, with more changes expected as part of Transmission 4.1:

https://github.com/transmission/transmission/releases/tag/4.0.1

holta:
Transmission 4.0.2 (February 15) and 4.0.3 (April 13) show this important BitTorrent client (built into many Linux distributions) is now "back in business" 😃

https://github.com/transmission/transmission/releases/tag/4.0.2
https://github.com/transmission/transmission/releases/tag/4.0.3

holta:
@neomatrixcode just in case you have time to compile Transmission 4.1.0-dev on Raspberry Pi?

I guess there's still some debate as to whether 4.0.3 is yet satisfyingly stable for royalty and commoners both — but regardless of that debate — either way I'd be in favor of experimenting with compiling their `main` branch to evaluate the future not the past:

- transmission/transmission#5585

Tangentially Related:

- #1078
- #3526

holta:
Transmission 4.0.6 works but has experienced several problems, beyond just this Ubuntu packaging bug:

- #3756

And Transmission 4.1 has been delayed (originally expected in 2023) perhaps b/c the rework / refactoring was over-ambitious.

So new community strategies are being investigated — some of which will hopefully help people pull together: (beyond the usual flood of toxic finger-pointing)

- https://github.com/transmission/transmission/issues/7058

PR #3599, opened 2023-06-12
WIP: Try 'transmission_compile_latest: True' BitTorrent software (e.g. Transmission 4.1.0-dev, instead of 3.0 from May 2020)

holta:
@neomatrixcode this is just a proof-of-concept for now.

But if this-or-similar proves reliable/promising we can optimize later... for both reliability and faster install (compile time is about 10 min on a recent PC, but likely to be painful on all Raspberry Pi's, possibly an hour or even worse!?)

Related:

- #1078
- PR #2572
- #3473

### DIFF
--- a/roles/transmission/README.rst
+++ b/roles/transmission/README.rst
@@ -111,10 +111,13 @@ Docs
 As of June 2023, these docs appear to be the most up-to-date:
 
 - https://github.com/transmission/transmission/tree/main/docs
+   - https://github.com/transmission/transmission/blob/main/docs/Building-Transmission.md
    - https://github.com/transmission/transmission/blob/main/docs/Configuration-Files.md
+   - https://github.com/transmission/transmission/blob/main/docs/Editing-Configuration-Files.md
    - https://github.com/transmission/transmission/blob/main/docs/Headless-Usage.md
-- https://wiki.archlinux.org/title/transmission (updated regularly)
 - https://cli-ck.io/transmission-cli-user-guide/ (2016 but still useful)
+   - https://github.com/transmission/transmission#command-line-interface-notes ("``transmission-cli`` is deprecated and exists primarily to support older hardware dependent upon it. In almost all instances, ``transmission-remote`` should be used instead.")
+- https://wiki.archlinux.org/title/transmission (updated regularly)
 - https://trac.transmissionbt.com/wiki (2006-2019)
 
 Logging
--- a/roles/transmission/README.rst
+++ b/roles/transmission/README.rst
@@ -115,6 +115,8 @@ As of June 2023, these docs appear to be the most up-to-date:
    - https://github.com/transmission/transmission/blob/main/docs/Configuration-Files.md
    - https://github.com/transmission/transmission/blob/main/docs/Editing-Configuration-Files.md
    - https://github.com/transmission/transmission/blob/main/docs/Headless-Usage.md
+   - https://github.com/transmission/transmission/blob/main/docs/rpc-spec.md
+      - https://transmission-rpc.readthedocs.io
 - https://cli-ck.io/transmission-cli-user-guide/ (2016 but still useful)
    - https://github.com/transmission/transmission#command-line-interface-notes ("``transmission-cli`` is deprecated and exists primarily to support older hardware dependent upon it. In almost all instances, ``transmission-remote`` should be used instead.")
 - https://wiki.archlinux.org/title/transmission (updated regularly)
--- a/roles/transmission/README.rst
+++ b/roles/transmission/README.rst
@@ -125,14 +125,14 @@ As of June 2023, these docs appear to be the most up-to-date:
 Logging
 -------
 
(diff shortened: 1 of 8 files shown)


holta:
Currently being tested on multiple OS's using `iiab-pr3599.yml` which contains:

```
runcmd:
  - mkdir /etc/iiab
  - curl https://raw.githubusercontent.com/iiab/iiab/master/vars/local_vars_unittest.yml > /etc/iiab/local_vars.yml
  - [sed, -i, 's/^transmission_install:.*/transmission_install: True/', /etc/iiab/local_vars.yml]
  - [sed, -i, 's/^transmission_enabled:.*/transmission_enabled: True/', /etc/iiab/local_vars.yml]
  - curl iiab.io/risky.txt | bash -s 3599 &
```

OS can then be selected using Multipass commands (that run everything!) like:

```
multipass launch 22.04 -n u2204 -c 2 -d 25G -m 2G --cloud-init iiab-pr3599.yml && multipass shell u2204
```

holta:
FYI recent versions of `transmission-daemon` 4.1.0-dev appear to fail after ~90 seconds with IIAB (http://box:9091 works for about 90 seconds only!) due to an IPv6 configuration dilemma / glitch detailed here:

- https://github.com/transmission/transmission/issues/5362#issuecomment-1586552109

It would seem (?) that Transmission is getting confused (at the OS/kernel/networking level) while trying to detect whether or not IPv6 is truly turned off of not (on IIAB machines).

holta:
> IPv6 is truly turned off

Just for the record, here's where IIAB turns off IPv6:

https://github.com/iiab/iiab/blob/ef804c25345622f8b37da130855e8462fd27b2c1/roles/network/tasks/install.yml#L57-L69

holta:
This PR is tested on several Ubuntu 22.04, 23.10 (latest pre-release) and also on the just-released Debian 12 "Bookworm" ✅

Testing on Raspberry Pi OS on Raspberry Pi 4 is now underway — to see if that compile might also work?  (After a long, long time presumably, if so!)

holta:
> Testing on Raspberry Pi OS on Raspberry Pi 4 is now underway — to see if that compile might also work?

It all works (functional testing too!)  Two data points:

* (True) 32-bit RasPiOS on RPi 4:
46min compile time
1.1GB /opt/iiab/transmission (should really be auto-deleted right after `make install` !)

* 64-bit RasPiOS on RPi 4:
52min compile time
1.5GB /opt/iiab/transmission (should really be auto-deleted right after `make install` !)

holta:
Imposing a ~50min compile on all who choose to install MEDIUM-sized and LARGE-sized software installs onto Raspberry Pi 4's is not the end-of-the-world, but still it's NOT my first choice!

Still I think we should go ahead with this over coming weeks/months, for now, to broaden community testing & evaluation of this far better BitTorrent infra (Transmission 4.1.0-dev as compared to 3.0 from 2020).

Then [transmission-rpc](https://transmission-rpc.readthedocs.io/) might also be needed, via pip install to a venv, possibly even rather soon?

In any case IIAB should optimize/refactor/simplify all these Transmission advances + options in coming months.  One option might be to post the 5 needed binaries (for `/usr/local/bin`) to https://iiab.io/packages or somewhere similar — e.g. especially if luck is on our side and Transmission 4.1.0 is released shortly in coming months — as more discover its tremendous progress over the past 3 years :pray:

neomatrixcode:
looks good to me

neomatrixcode:
I'm all for the idea of publishing the binaries at https://iiab.io/packages. I find it quite practical

holta:
> I'm all for the idea of publishing the binaries at https://iiab.io/packages. I find it quite practical

Ok.

And a more short-term PR might free up 100's of MB simply by uninstalling (purging if we dare) these 5 apt packages... that are not really needed after the compile's done:

- build-essential
- cmake
- libcurl4-openssl-dev
- libssl-dev
- libsystemd-dev

As evidenced by...

```
root@box:/usr/local/bin# ls -lSh
total 292M
-rwxr-xr-x 1 root root 72M Jun 12 13:20 transmission-daemon
-rwxr-xr-x 1 root root 71M Jun 12 13:21 transmission-create
-rwxr-xr-x 1 root root 70M Jun 12 13:21 transmission-remote
-rwxr-xr-x 1 root root 70M Jun 12 13:21 transmission-show
-rwxr-xr-x 1 root root 11M Jun 12 13:21 transmission-edit
...
```

```
root@box:~# grep disk /etc/iiab/iiab.ini | sort -nrk3 | column -t
disk_used_a_priori        =  2637901824
iiab_software_disk_usage  =  1375277056
transmission_disk_usage   =  636248064
mysql_disk_usage          =  352161792
www_options_disk_usage    =  72245248
nginx_disk_usage          =  41005056
openvpn_disk_usage        =  39845888
network_disk_usage        =  13516800
iiab_admin_disk_usage     =  8065024
remoteit_disk_usage       =  4513792
www_base_disk_usage       =  3198976
usb_lib_disk_usage        =  368640
sshd_disk_usage           =  180224
pylibs_disk_usage         =  20480
```

From a fresh install of the latest version of this PR onto Ubuntu 23.10 with iiab-diagnostics here: http://sprunge.us/MgjYkV?en

PR #3391, opened 2022-10-04
Auto-install latest/official kiwix-tools + tighten up code

holta:
Instead of IIAB having to hard-code and host a new version of [kiwix-tools](https://github.com/kiwix/kiwix-tools) every month or so:

1. This makes installing the lastest/official kiwix-tools release automatic, based on Kiwix's permalink redirects:
   - https://www.kiwix.org/en/downloads/kiwix-serve/
   - kiwix/k8s#50
2. Bonus it's now easier to quickly update to the very latest official kiwix-tools release (if any IIAB implementer wants that pre-flight / pre-deployment).  It's now as easy as running:
   ```
   cd /opt/iiab/iiab
   sudo ./runrole --reinstall kiwix
   ```
3. Specific versions of kiwix-tools can now be hard-coded even more easily than in the past, by hand editing `/opt/iiab/iiab/roles/kiwix/defaults/main.yml` that explains your options much more clearly &mdash; e.g. during regression testing (progress testing!) comparing different versions of kiwix-tools, etc.

RECAP: As usual, this installs kiwix-tools' 3 executables {kiwix-manage, kiwix-search, kiwix-serve} into /opt/iiab/kiwix/bin but does it in a more compact and automated way than in the past.

Tested on Ubuntu 22.10 (pre-release).

Not a big deal, but a Caveat: downloading of kiwix-tools will occasionally now take a bit longer (e.g. sometimes a full minute) as Kiwix download mirrors can be quite scattershot.  These inconsistent results appear to be luck of the draw (very random) from my experience, probably based Kiwix's "round robin" mirror server rotation that's quite a familiar issue over recent years.

Building on:

- PR #3235 
- PR #3344

PR #3600, opened 2023-06-12
WIP: Kiwix arch changes

holta:
In keeping with upstream changes as they evolve here:

- kiwix/kiwix-tools#623

For the moment, both above redirects [are] live.

If however Kiwix chooses to drop armv8 (which is presumably 32-bit, possibly @mgautierfr can confirm?) that can be dropped (i.e. if indeed those are 32-bit builds, the armv6 builds would then be used in their place).

Many thanks to @mgautierfr for all his hard work making all this possible ✅

- PR #3391 
- PR #3576

### DIFF
--- a/roles/kiwix/defaults/main.yml
+++ b/roles/kiwix/defaults/main.yml
@@ -29,8 +29,8 @@ kiwix_arch_dict:      # 'dpkg --print-architecture' key would be: (to mitigate #
   #i386:              # ?
   i686: i586          # ?
   x86_64: x86_64      # amd64
-  armv6l: armhf       # armhf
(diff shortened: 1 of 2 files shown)


holta:
Building on 2023-05-14 work here:

- #3574
- PR #3576

Thanks again to @mgautierfr for reworking kiwix-tools's compile targets/toolchain over recent months.

mgautierfr:
> which is presumably 32-bit, possibly @mgautierfr can confirm?

Yes, I confirm. Both armv6 and armv8 are 32-bit. aarch64 is 64-bit.

rgaudin:
Versionless redirects to `armhf` versions now points to `armv6`.

holta:
> Versionless redirects to `armhf` versions now points to `armv6`.

Good idea, to cover legacy operations!

PR #3554, opened 2023-04-26
JupyterHub tightened up w/ python3-psutil, --system-site-packages [python3-venv also removed from Calibre-Web & Lokole] [Node.js apt source pollution detox]

jvonau:
### Fixes bug:
#3553 https://github.com/iiab/iiab/pull/3549#issuecomment-1523231127
### Description of changes proposed in this pull request:
same fix as calibre-web
### Smoke-tested on which OS or OS's:
Pending

PR #3503, opened 2023-03-22
use --system-site-packages and install package to omit building 'netifaces' wheel [Calibre-Web]

jvonau:
Smaller subset of #3500 without optional-requirements.txt

This works because the wheel is provided by apt, the OS package manager, and with --system-site-packages has access to the system's preinstalled wheels when creating the virtual environment. Normally --system-site-packages would be omitted and the packages would be pulled down from pypi but 'netifaces' doesn't have the latest wheels for python3.10+ but the OS does.

PR #3607, opened 2023-07-15
calibre-web/tasks/install.yml: Install ffmpeg not just imagemagick

holta:
@deldesir requests this and I agree with his decision.

Building on:

- PR #3503
- PR #3554
- PR #3606


deldesir (approved):
It's used to generate the covers [thumbnails] from videos

PR #3608, opened 2023-07-16
New Gitea release 1.20

holta:
New Gitea 1.20 release:
https://github.com/go-gitea/gitea/releases/tag/v1.20.0

ANNC:
https://blog.gitea.com/release-of-1.20.0/

Building on:
- PR #3499

PR #3609, opened 2023-07-16
Gitea 1.20 patches, e.g.  app.ini.j2

holta:
At least resolves the many `/etc/gitea/app.ini` problems and updates URL's etc for more recent versions of Gitea:

- PR #3608

Gitea's PGP signature checking of their downloaded binaries is still a grand mess years later (https://docs.gitea.com/installation/install-from-binary#verify-gpg-signature doesn't actually work with 1.20) but this is their fault.  Hopefully they'll finally fix that soon.

### DIFF
--- a/roles/gitea/defaults/main.yml
+++ b/roles/gitea/defaults/main.yml
@@ -19,7 +19,7 @@ iset_suffixes:
 
 gitea_iset_suffix: "{{ iset_suffixes[ansible_machine] | default('unknown') }}"    # A bit safer than ansible_architecture (see kiwix/defaults/main.yml)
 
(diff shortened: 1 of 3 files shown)


holta:
Smoke-tested and very basic functional testing done on a pre-release on Ubuntu 23.10

holta:
> Gitea's PGP signature checking of their downloaded binaries is still a grand mess years later (https://docs.gitea.com/installation/install-from-binary#verify-gpg-signature doesn't actually work with 1.20) but this is their fault. Hopefully they'll finally fix that soon.

CLARIF: Their PGP signature is verifiable with 1.19 (latest binary for amd64) but they're dropped the ball again with 1.20

PR #3601, opened 2023-06-13
remote.it workaround for incoming ICMP firewalls e.g. GitHub Actions [at least allowing IIAB installs to proceed, when service schannel does not exist]

holta:
Thanks to @neomatrixcode for surfacing this problem!

Bulding on:

- #3394
  - PR #3395

PR #3611, opened 2023-07-26
Remote.it CLI can no longer coexist with remoteit Device Package

holta:
As a result of their sudden upstream change in recent weeks :/

Background:

- PR #3161
- PR #3395
- PR #3601


holta:
Smoke-testing & basic functionality testing underway on Ubuntu 23.10 pre-release.

PR #3618, opened 2023-08-11
iiab-diagnostics: Change pastebin service (sprunge.us -> paste2.org)

holta:
As pastebin service https://sprunge.us stopped working several weeks ago:

- kevr/sprunge#8


holta:
Reverting to https://dpaste.com/about (as IIAB used years ago) might be another option — if that proves superior.

PR #3619, opened 2023-08-12
iiab-diagnostics: Change pastebin service (paste2.org -> dpaste.com)

holta:
[DPASTE.COM](https://dpaste.com/help) may be limited to 30 days by default, but appears substantially more solid/reliable than [PASTE2.ORG](https://paste2.org) whose pastebin URL's (dangerously?) redirect to phishing-like spam.

So this PR supersedes:

- PR #3618


holta:
New size limit is ["1,000,000 characters"](https://dpaste.com/help)

pastebins sent from CLI are currently deleted after 30-days by default — we can raise this later using their [dpaste.com HTTP API](https://dpaste.com/api/) if nec?

Example iiab-diagnostics output: https://dpaste.com/G9EVYNFWJ

PR #3634, opened 2023-09-08
scripts/ansible: Evaluate 'apt install ansible-core' on "32-bit" RasPiOS 12 Bookworm

holta:
Possible replacement for PR #3632 if @EMG70 has time to try this approach?

(NOT elegant... basically forcing the install of _old & overweight ansible-core 2.14.3..._ but might be easier than fighting "endless dependency hell" !)

Related:

- #3501
- #3516
- #3526
- PR #3617
- PR #3632

PR #3632, opened 2023-09-07
scripts/ansible: Evaluate upstream piwheels / cryptography fix on "32-bit" RasPiOS 12 Lite [pinning to cryptography 40.0.1 no longer nec!] [create /etc/pip.conf missing from RasPiOS as nec, to avoid cffi glitch]

holta:
Let's see if `cryptography==40.0.1` is hopefully no longer necessary!

Thank you @EMG70 for testing on 32-bit RasPiOS 12 Lite (Bookworm) — after you've applied all the OS updates from RasPiOS 11 Lite (Bullseye) and rebooted as explained in #3526 — you can then run a fresh IIAB install including this PR using:

```
curl iiab.io/install.txt | bash -s 3632
```

Related:

- #3526
- #3547
- PR #3604

PR #3617, opened 2023-08-11
Brute Force 'apt install ansible-core' (e.g. 2.14.3 on Debian Bookworm) on 32-bit legacy i386, to avoid rust/wheels/cryptography compiling mess

holta:
Not an elegant solution but this is a workaround for some corner cases arising with:

- PR #3613 
- PR #3615 

Background:

- #3547
- PR #3604
- PR #3610

Issue #3501, opened 2023-03-20
question: why does ansible need version 2.11.12 ?


tim-moody:
this seems hard coded in the script

holta:
This is out-of-date and will be raised.

Ansible 2.10 and 2.11 are both [end-of-life'd](https://docs.ansible.com/ansible/latest/reference_appendices/release_and_maintenance.html#ansible-core-support-matrix) and full of bugs that would make IIAB QA a literal and absolute nightmare — testing IIAB code or any code against dozens of different versions of Ansible from years past is neither plausible, realistic, nor wise 😨

tim-moody:
on Ub 22.04 VM:
installed ansible with apt (got 2.10.8)
modified iiab-install to accept 2.10.1 and higher
ran medium install
completed with no complaints

tim-moody:
you're sure right about EOL. amazing how out of date the apt packages are.

jvonau:
> on Ub 22.04 VM: installed ansible with apt (got 2.10.8) modified iiab-install to accept 2.10.1 and higher ran medium install completed with no complaints

In https://github.com/iiab/iiab/pull/3494#issuecomment-1475398577 I listed what is available to install, what should of been tried was installing ansible-core, on U-22.04 that should of installed 2.12.0

tim-moody:
I'm sure you're right, but in any event iiab medium and adm cons both installed with 2.10.8 with no complaints.

holta:
FWIW Ansible 2.12 will be EOL'd shortly (2023-05-22).

The larger point (in my opinion anyway) is that QA becomes impractical when testing against many different historical versions of Ansible.  Some of which might complete.  But deliver different  functional results.  Somewhat like this famous cartoon.

![image](https://user-images.githubusercontent.com/2458907/226678195-94357fe3-ffd6-458d-bf56-ea7fa2786dd6.png)

tim-moody:
every version will be historical at some point and bleeding edge versions are more likely to be buggy that ones that have aged.

holta:
We'll have to politely disagree.  Obviously I have a very different view.  FWIW my experience is that:

- Recent versions of Ansible will always provide a far more solid better basis for CI/CD-centered QA, saving time for many along the way.
- Older versions of Ansible will always contain many more bugs according to published documentation (admittedly a tautology, but also a reality that bugs by definition surface and are solved over time, which in the end is why recent versions provide a better basis for CI/CD-centered QA).
[...]

PR #3637, opened 2023-09-10
"32-bit" OS's more resilient: Verify that apt package 'ansible-core' is in fact available (equivalent to testing for Debian 12+ but safer!)

holta:
Building on:

- PR #3634

Related:

- #3501
- #3516
- #3526
- PR #3617
- PR #3632


holta:
Clarifs/Reminder:

- Locking in this hard-coded old ansible-core version 2.14.3 (will soon be a year out-of-date, and contains many known bugs) is hardly a long-term solution.

- But it might prove good enough for more recent "32-bit" OS's, until a better solution is found.

- ansible core 2.13 reaches [EOL on 2023-11-06](https://docs.ansible.com/ansible/latest/reference_appendices/release_and_maintenance.html#ansible-core-support-matrix), the same day that ansible-core 2.16 is [expected](https://docs.ansible.com/ansible/devel//roadmap/ROADMAP_2_16.html).

Issue #3547, opened 2023-04-24
2 bits of bad news just for the record [cryptography, piwheels, rust, Ansible]


jvonau:
2 bits of bad news just for the record:

1. Latest screenshot from https://www.piwheels.org/project/cryptography/ indicates cryptography 40.0.2 (released 2023-04-14) fails on both Debian 10 (Buster) and 11 (Bullseye):

   ![image](https://user-images.githubusercontent.com/2458907/233806443-fd7f00bd-a25e-4eeb-95b9-99e45526df70.png)

2. "any wheels built with Rust will only work on Pi 2/3/4 not 1/Zero" for now... possibly 32-bit ARM issues might never be fully resolved in a stable way... as the world increasingly focuses on 64-bit:

   - https://github.com/piwheels/piwheels/pull/328#issuecomment-1518606329
   - https://github.com/piwheels/packages/issues/348

_Originally posted by @holta in https://github.com/iiab/iiab/issues/3459#issuecomment-1518810705_

jvonau:
I suggest that an install be booted on a rpi-w and see if ansible runs.

holta:
> I suggest that an install be booted on a rpi-w and see if ansible runs.

Running Ansible on 32-bit Raspberry Pi Zero W might possibly be too much to ask for and not really needed.

In 2023 especially.  But in any case, here's a test, in case there's something to learn:

https://github.com/iiab/iiab/pull/3550#issuecomment-1520567117

holta:
Just FYI `cryptography` 41.0.2 was released yesterday and still fails on Bullseye / Python 3.9, according to https://www.piwheels.org/project/cryptography/ :

![image](https://github.com/iiab/iiab/assets/2458907/ee0632d7-d02f-4bf8-897f-3e2fde6b07ad)

Background:

- #3459
- https://github.com/iiab/iiab/blob/975ce2f54e1fee82ba6412641b47a51ac078921d/scripts/ansible#L215-L217

holta:
Same story with `cryptography` 41.0.3 just [released](https://www.piwheels.org/project/cryptography/) earlier this week:

![image](https://github.com/iiab/iiab/assets/2458907/227a8e10-35f0-42d5-9664-70d7d23757ac)

holta:
Great news [piwheels](https://www.piwheels.org/project/cryptography/)  appears fixed across the board:

[...]

PR #3652, opened 2023-10-01
scripts/ansible: 'pip install --prefer-binary ansible-core' is simpler/faster & more future-proof

holta:
More future-proof solution than even:

- PR #3651 

Huge thanks to those who provided essential, breakthrough tips here:

- #3650

Tangentially related:

- #3547
- PR #3615 
- PR #3617
- PR #3632


holta:
Re-tested with another smoke test on fully 32-bit RasPiOS 11 (Bullseye), just as another reconfirmation of these manual steps:

- https://github.com/iiab/iiab/issues/3650#issuecomment-1742146219

Issue #3657, opened 2023-10-11
On RasPiOS Bookworm, just released: "Could not find the requested service systemd-resolved: host"


holta:
@tim-moody, @EMG70 and I confirm IIAB install attempts on this [brand new OS](https://www.raspberrypi.com/news/bookworm-the-new-version-of-raspberry-pi-os/) complete the 9 stages of `/opt/iiab/iiab/iiab-install` — but then get stuck after that:

```
TASK [network : Enable & Restart systemd-networkd.service] ***************************************************************************
changed: [127.0.0.1]

TASK [network : Enable & Restart systemd-resolved.service] ***************************************************************************
fatal: [127.0.0.1]: FAILED! => {"changed": false, "msg": "Could not find the requested service systemd-resolved: host"}

PLAY RECAP ***************************************************************************************************************************
127.0.0.1                  : ok=329  changed=187  unreachable=0    failed=1    skipped=137  rescued=0    ignored=1
```

Related:

- https://github.com/iiab/iiab/blob/60763fab8e41ee126a21af250cb6639203cbaed3/roles/network/tasks/sysd-netd-debian.yml#L97-L102
- #3399
- https://github.com/search?q=repo%3Aiiab%2Fiiab%20systemd-resolved&type=code
- `Error: NetworkManager is not running.` repeatedly last week appeared on screen 4 times, at the bottom of `sudo iiab-network` output for @EMG70, as shown by screenshot in [#3526](https://github.com/iiab/iiab/issues/3526#issuecomment-1749740700)

holta:
Reconfirmed with a SMALL-sized IIAB install on 64-bit RasPiOS Bookworm Lite on RPi 4 here, in color to give a bit more context:

![image](https://github.com/iiab/iiab/assets/2458907/c461f823-1756-4588-944c-cd9d79360a0f)

holta:
@jvonau would know much more, if he has time:

- Lines 75-77 already having run, is it possible Line 82 should be `when: systemd_networkd_active and not network_manager_active` ? (Or possibly just `when: not network_manager_active` ❓)

- Any chance Lines 85-87 might be going away — e.g. if `rpi_debian.yml ` perhaps might no longer be necessary in future?

https://github.com/iiab/iiab/blob/60763fab8e41ee126a21af250cb6639203cbaed3/roles/network/tasks/main.yml#L75-L91

holta:
FWIW Debian 12 release notes clarify:

[...]

PR #3658, opened 2023-10-11
PR For Discussion: "Just install systemd-resolved at the same place as systemd-networkd is installed"

holta:
[Former ticket title: Don't run sysd-netd-debian.yml if NetworkManager active?]

@jvonau would know much better (good chance he has a much better approach).

So this PR is just a stub towards getting things right for the new RasPiOS Bookworm, towards better understand and then resolving this issue:

- #3657

Related:

- #3399

### DIFF
--- a/roles/network/tasks/install.yml
+++ b/roles/network/tasks/install.yml
@@ -32,6 +32,7 @@
       - netmask                #   25kB download: Handy utility -- helps determine network masks
       - net-tools              #  248kB download: RasPiOS installs this regardless -- @jvonau suggests possibly deleting this...unless oldtimers really want these older commands in iiab-diagnostics output?
       - networkd-dispatcher    #   15kB download: Dispatcher service for systemd-networkd connection status changes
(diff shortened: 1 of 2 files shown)


holta:
Clarifications from @jvonau:

> Just install systemd-resolved at the same place as systemd-networkd is installed..

Done: [60a14c5](https://github.com/iiab/iiab/pull/3658/commits/60a14c5a6869d2b4f1557f93d41f11038e69d2aa)

(Does that look reasonable?)

> Trouble with [NetworkManager] is in some cases the network is not enabled till a user is logged in..
> So br0 is unavailable till then resulting in no hotspot

> > Don't run sysd-netd-debian.yml if NetworkManager active?

> Your proposed change will [damage] Ubuntu and Mint

Ok.  That's now reverted: [b64737b](https://github.com/iiab/iiab/pull/3658/commits/b64737bca6deac87142c02e1825cb321a00fcb81)

holta:
These PR's were merged instead:

- PR #3659 
- PR #3660

PR #3659, opened 2023-10-12
Patch 8

jvonau:
### Fixes bug:
RasPiOS
Ubuntu's Netplan has both NM and sysd preinstalled, account for what RasPiOS might be missing
not yet

### DIFF
--- a/roles/network/tasks/sysd-netd-debian.yml
+++ b/roles/network/tasks/sysd-netd-debian.yml
@@ -4,10 +4,10 @@
     name:
       - networkd-dispatcher    #   15kB download: Dispatcher service for systemd-networkd connection status changes
 
-- name: 'Install networkd-resolved
+- name: 'Install systemd-resolved for RasPiOS
   package:
     name:
-      - networkd-resolved
+      - systemd-resolved
   when:  is_raspbian and os_ver is version('raspbian-12', '>=')   
 
 - name: Copy the bridge script - Creates br0
--- a/roles/network/tasks/sysd-netd-debian.yml
+++ b/roles/network/tasks/sysd-netd-debian.yml
@@ -4,6 +4,12 @@
     name:
       - networkd-dispatcher    #   15kB download: Dispatcher service for systemd-networkd connection status changes
 
+- name: 'Install networkd-resolved
+  package:
+    name:
+      - networkd-resolved
+  when:  is_raspbian and os_ver is version('raspbian-12', '>=')   
+
 - name: Copy the bridge script - Creates br0
   template:
     dest: /etc/systemd/network/IIAB-Bridge.netdev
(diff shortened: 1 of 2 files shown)


holta:
Related:

- #3657
- PR #3658

roles/network/tasks/sysd-netd-debian.yml
```diff
@@ -1,4 +1,15 @@
 # sysd-netd-debian.yml
+- name: 'Install networkd-dispatcher
```
holta, line 2:
```suggestion
- name: Install networkd-dispatcher
```

roles/network/tasks/sysd-netd-debian.yml
```diff
@@ -1,4 +1,15 @@
...
+  package:
+    name:
+      - networkd-dispatcher    #   15kB download: Dispatcher service for systemd-networkd connection status changes
+
+- name: 'Install systemd-resolved for RasPiOS
```
holta, line 7:
```suggestion
- name: Install systemd-resolved for RasPiOS 12+
```

holta (commented):
Dangling quotes can be removed.

roles/network/tasks/sysd-netd-debian.yml
holta: Ansible (via GitHub Actions) chokes on dangling quotes:

https://github.com/jvonau/iiab/actions/runs/6490464817/job/17626277491

holta:
Should `sysd-netd-debian.yml` [install apt package `systemd-resolved`](https://github.com/jvonau/iiab/blob/cdd88d20c55e5c7c9d45b5d2b045c7ff60ff09a7/roles/network/tasks/sysd-netd-debian.yml#L7-L11) onto... _every OS that offers that package?_

Along the lines of this bash code?

```
if apt-cache show systemd-resolved > /dev/null; then
    apt -y install systemd-resolved
fi
```

_Is something like that useful?_  (If so, here are 2 quick-and-dirty Ansible stubs doing the same thing...)

```
- name: Install systemd-resolved (if apt package available)
  shell: apt -y install systemd-resolved || true
```

```
- name: Install systemd-resolved (if apt package available)
  package:
    name: systemd-resolved
  ignore_errors: yes
```

tim-moody:
I take it there is no case where systemd-resolved is both a separate package and installed by systemd.

tim-moody:
After I removed the two syntax errors this ran to completion on rpi4 with 2023-10-10-raspios-bookworm-armhf-lite.img.

hostapd is running.

tim-moody:
my laptop connected over hotspot and box.lan was resolved
not sure what other tests are needed

tim-moody:
using AdmCons added password to hotspot. laptop connected securely

PR #3676, opened 2023-11-29
Calibre-Web prereqs for audio/video + set app.db logging to /var/log/calibre-web.log w/ log level "DEBUG"

holta:
For @deldesir's ongoing efforts to make Calibre-Web more useful to IIAB communities.

PR #3680, opened 2023-12-05
enable-or-disable.yml: Don't create symlink /library/www/html/calibre-web -> /library/calibre-web (NGINX can go directly if/as nec!)

holta:
In keeping with the latest:

- PR iiab/calibre-web#57

Does this possibly require adjustments to... ?

- PR iiab/calibre-web#51

Thanks @deldesir for reviewing!

FYI this PR builds on:

- PR #3676


deldesir (commented):
Ok, so we'll read /library/calibre-web directly then.

deldesir (approved):
Look good to me. I'll do necessary changes to align with this.

tim-moody:
in line 29 do you mean cat or cp?

PR #3699, opened 2024-01-19
Update (approx 10!) Calibre-Web app.db settings / configuration defaults

holta:
Calibre-Web has evolved a lot in recent years!  This probably should have been done a year ago or more (-:

In any case, this builds upon:

- PR #3676

Tangentially related:

- PR iiab/calibre-web#103


roles/calibre-web/files/app.db
deldesir, line 1: Looks good.

holta:
A few examples of this PR's newly proposed Calibre-Web defaults... (if `/library/calibre-web/config/app.db` does not exist!)

image

Admin user settings would default to...

image

Guest user settings would default to...

image

"Default Settings for New Users" would be...

image

"Default Visibilities for New Users" would be...

image

PR #3697, opened 2024-01-11
iiab-diagnostics: Show Calibre-Web "version" (~50 recent commits) + systemd service status + 3 "log files"

holta:
@deldesir is this sufficient, to facilitate rapid + informed debugging?

(Anything else that's conducive to rapid + informed debugging?)

Related:

- PR #3694
- PR #3695
- PR #3696

PR #3700, opened 2024-01-20
iiab-diagnostics: Pipe everything thru `cat -v` so ASCII control characters become visible

holta:
Helps make hidden characters in log files and other diagnostics files more visible, partially solving situations like:

- iiab/calibre-web#105

Related:

- PR #3697
- PR #3698


holta:
Example OLD result:

image
...from: https://sprunge.us/XZeTGV?en#n-1094

Example NEW result...

image
...from: https://sprunge.us/FdZLKw?en#n-1095

CC: @deldesir @chapmanjacobd

PR #3714, opened 2024-03-04
Py2

jvonau:
### Fixes bug:
#3711
### Description of changes proposed in this pull request:
newer workflow for U2404
### Smoke-tested on which OS or OS's:
U2404 VM at cmdline only ansible run pending

Issue #3711, opened 2024-02-24
roles/kalite fails to pin setuptools to 44 on emerging OS's (e.g. Ubuntu 24.04) with Python 3.12 — "Wheel for pip for Python 2.7 is unavailable. apt install python2-pip-whl"


holta:
I don't know yet if this will be easily fixable or not?

@jvonau might know:

iiab-diagnostics: http://sprunge.us/L2kogP?en

```
root@u24:~# python3 --version
Python 3.12.2
```

```
root@u24:~# cat /etc/lsb-release
DISTRIB_ID=Ubuntu
DISTRIB_RELEASE=24.04
DISTRIB_CODENAME=noble
DISTRIB_DESCRIPTION="Ubuntu Noble Numbat (development branch)"
```

```
root@u24:~# hostnamectl
 Static hostname: box
       Icon name: computer-vm
         Chassis: vm 🖴
      Machine ID: 7a3d1f28a7fd4c97a0f1e93bf8ae69d8
         Boot ID: e711e5e242344d81a0cce0ae2680f4f5
  Virtualization: kvm
Operating System: Ubuntu Noble Numbat (development branch)
          Kernel: Linux 6.6.0-14-generic
    Architecture: x86-64
 Hardware Vendor: QEMU
  Hardware Model: Standard PC _Q35 + ICH9, 2009_
Firmware Version: unknown
   Firmware Date: Wed 2022-02-02
    Firmware Age: 2y 3w 1d
```

Error from /opt/iiab/iiab/iiab-install.log:

> TASK [kalite : Use pip to pin setuptools to 44 in /usr/local/kalite/venv] *******************************************************************************************************************************************************************
[...]

PR #3715, opened 2024-03-06
Touch-ups for: scripts/install_python2_kalite-venv_u2404.sh

holta:
Quick maintainability / readability touch-ups on top of @jvonau's work at:

- PR #3714 

As @jvonau explains above, we can extend this in various directions later if/as necessary.

Recap: this is primarily about KA Lite on Ubuntu 24.04 as outline here...

- #3711


holta:
@emg70 @jvonau

This PR with KA Lite works on Ubuntu 24.04: http://sprunge.us/b1AH6f?en

(Problematic error "Connecting to server. Please wait..." below seems like it can be ignored — as it likely applies to all OS's...)

image

PR #3719, opened 2024-03-08
no-pipx [possible alternative approach to installing yt-dlp for Calibre-Web]

jvonau:
Would be nice to see the provided share/doc/yt_dlp/README.txt file without being root.

### DIFF
--- a/roles/calibre-web/tasks/install.yml
+++ b/roles/calibre-web/tasks/install.yml
@@ -68,13 +68,16 @@
 - name: If Calibre-Web is being enhanced with audio/video "books" too, install/upgrade additional prereqs -- SEE https://github.com/iiab/calibre-web/wiki
   shell: |
     if [ -f {{ calibreweb_venv_path }}/scripts/lb-wrapper ]; then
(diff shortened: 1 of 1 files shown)


holta:
Personally I like `pipx` as it attempts to encapsulate detail, i.e. simplify venv detail.

Calibre-Web documentation has been written with that in mind: https://github.com/iiab/calibre-web/wiki#xklb-and-yt-dlp-back-end-integration

So the two competing realities below are each valid, each needing consideration:

1) I'm not excited about reworking docs & testers' expectations.  And would prefer a long-term path with pipx-like succinct simplicity.

2) At the same time pipx is not yet perfect, so this PR should also be considered as one possibility.

holta:
Either way, instructions for upgrading `yt-dlp` will need to be crystal clear in places like:

- https://github.com/iiab/calibre-web/wiki#xklb-and-yt-dlp-back-end-integration

jvonau:
Given the grossness of the logging with mentions of failed imports leads me to not wanting to support any part of this sloppiness.

Issue #3709, opened 2024-02-10
Recent Kolibri installs show "502 Bad Gateway" — "Conflicts with the installed package 'python3-markdown'" (e.g. when samba or pipx are in use on Debian 12 or RasPiOS Bookworm)


holta:
@nzola wrote:

> Kolibri gives error "502 Bad Gateway" when trying to access it on the browser?
 I did 3 iiab install on rpi4. All of them give the same error

@nzola trying a recent Beta release of Kolibri 0.16 will probably work around this?

https://github.com/learningequality/kolibri/releases

In other words, try setting the following in your [/etc/iiab/local_vars.yml](https://wiki.iiab.io/go/FAQ#What_is_local_vars.yml_and_how_do_I_customize_it%3F) prior to installing IIAB:

```
kolibri_deb_url: https://github.com/learningequality/kolibri/releases/download/v0.16.0-beta13/kolibri_0.16.0b13-0ubuntu1_all.deb
```

jvonau:
> 2024-02-10 17:46:35,213 p=44617 u=root n=ansible | TASK [kolibri : apt install https://github.com/learningequality/kolibri/releases/download/v0.16.0-beta13/kolibri_0.16.0b13-0ubuntu1_all.deb (if kolibri_deb_url IS defined)] ***
2024-02-10 17:46:48,968 p=44617 u=root n=ansible | fatal: [127.0.0.1]: FAILED! => {"changed": false, "msg": "Conflicts with the installed package 'python3-markdown'"}
2024-02-10 17:46:49,014 p=44617 u=root n=ansible | TASK [kolibri : SEE ERROR ABOVE (skip_role_on_error: False)]

jvonau:
control file

> Package: kolibri
Source: kolibri-source
Version: 0.16.0b13-0ubuntu1
Architecture: all
Maintainer: José L. Redrejo Rodríguez <jredrejo@debian.org>
Installed-Size: 219964
Depends: python3 (>= 3.6), python3-pkg-resources, adduser
Recommends: python3-cryptography (>= 1.2.3)
Conflicts: python3-markdown (>= 3.4.1)
Section: education
Priority: optional
Homepage: https://learningequality.org/kolibri
Description: The offline app for universal education
 An educational platform for learners of all ages.
 .
 Retrieve and Distribute Open Educational Resources (OER) from online
 resources or from device to device.
 .
 Resources are shared as 'channels', amongst which can be found
 Khan Academy, PhET and CK-12.
[...]

PR #3726, opened 2024-04-08
Restore apt install of Kolibri 0.16.1+, thx to upstream changes

holta:
While Kolibri 0.17.x pre-releases are not quite yet available (for OS with Python 3.12 like Ubuntu 24.04), the ability to install recent release of Kolibri via apt should be restored very shortly, hopefully in coming hours, as explained at:

- https://github.com/learningequality/kolibri/issues/11892#issuecomment-2043073998

Related:

- PR #3343
- PR #3709
- PR #3722


holta:
Confirmed with a LARGE-sized fresh install of IIAB on Ubuntu Server 22.04:

http://sprunge.us/EMoyEa?en

PR #3732, opened 2024-04-30
Safer /usr/local/bin/yt-dlp symlink creation

holta:
### Fixes bug:

yt-dlp is not runnable at the CLI on some OS's like Ubuntu 23.10, whereas it works on Ubuntu 24.04

As outlined here:

- https://github.com/iiab/calibre-web/issues/158#issuecomment-2087086538

bash code is reinforced to be much more resilient, handling both kinds of OS/environments, so that we tolerate both common paths used by pipx, thereby refining this PR from 2023-12-30:

- PR #3689

Debian 12.  Further testing will be useful after merging!  On OS's like:

- Ubuntu 24.04
- Ubuntu 22.04 and/or Mint 21
- Raspberry Pi OS 12

@EMG70 @deldesir

### DIFF
--- a/roles/calibre-web/tasks/install.yml
+++ b/roles/calibre-web/tasks/install.yml
@@ -74,7 +74,13 @@
         else
             pipx install xklb
             ln -sf /root/.local/bin/lb /usr/local/bin/lb
-            ln -sf /root/.local/share/pipx/venvs/xklb/bin/yt-dlp /usr/local/bin/yt-dlp
+            if [ -f /root/.local/share/pipx/venvs/xklb/bin/yt-dlp ]; then
+                ln -sf /root/.local/share/pipx/venvs/xklb/bin/yt-dlp /usr/local/bin/yt-dlp
+            elif [ -f /root/.local/pipx/venvs/xklb/bin/yt-dlp ]; then
+                ln -sf /root/.local/pipx/venvs/xklb/bin/yt-dlp /usr/local/bin/yt-dlp
+            else
+                echo "ERROR: yt-dlp NOT FOUND"
+            fi
         fi
         cp {{ calibreweb_venv_path }}/scripts/lb-wrapper /usr/local/bin/
         chmod a+x /usr/local/bin/lb-wrapper


holta:
Tested on Ubuntu 23.10 ✅

> ### Smoke-tested on which OS or OS's:
> Debian 12. Further testing will be useful after merging! On OS's like:
> 
> * Ubuntu 24.04
> * Ubuntu 22.04 and/or Mint 21
> * Raspberry Pi OS 12

holta:
Documentation https://github.com/iiab/calibre-web/wiki#gear-back-end clarified for @EMG70 and implementers/contributors:

image

tim-moody:
are you saying /usr/local/bin points to /root/... venv? shouldn't it be the other way around?

holta:
Here's an example from Debian 12, to clarify:

```
root@box:~# ls -l /usr/local/bin/yt-dlp
lrwxrwxrwx 1 root root 39 Apr 30 18:29 /usr/local/bin/yt-dlp -> /root/.local/pipx/venvs/xklb/bin/yt-dlp
```

Likewise here's an example from Ubuntu 24.04:

```
root@box:~# ls -l /usr/local/bin/yt-dlp
lrwxrwxrwx 1 root root 45 May  1 11:36 /usr/local/bin/yt-dlp -> /root/.local/share/pipx/venvs/xklb/bin/yt-dlp
```

I think this is good enough to cover most OS's for now (I hope!)

Recap seems to be:
pipx changed the way it sets up venv paths, now using /root/.local/share by default, starting in ~2023.

tim-moody:
what if an app uses yt-dlp and doesn't install calibre-web?

holta:
> what if an app uses yt-dlp and doesn't install calibre-web?

Calibre-Web's use of XKLB is designed to simplify video-centric metadata at many different levels.

That means respecting XKLB's choice to install a very recent yt-dlp (not an old version as provided by apt) within XKLB's venv.

(Of course if someone really also wants to install another version of yt-dlp elsewhere, they are free to do that, via apt or any other method, and arrange paths accordingly.  Hasn't happened yet to my knowledge?)

Issue #3745, opened 2024-05-27
Sadly, the sprunge.us pastebin service appears offline or broken, affecting essential iiab-diagnostics communications


holta:
To be investigated in coming days, even if there's no immediately convenient alternative (we might need to dig deeper in June and/or wait for sprunge.us to patch up their service!?)

Related:

- https://github.com/iiab/iiab/blob/master/scripts/iiab-diagnostics.README.md
- https://github.com/iiab/iiab/blob/master/scripts/iiab-diagnostics
- PR #3633 
  - https://github.com/kevr/sprunge/issues/8

holta:
Not perfect[*] but this workaround is now implemented:

- PR #3748

[*] Reason: dpaste.com (1) limits pastebins to 30 days, and (2) allegedly limits pastebins ~1,000,000 characters.

Reverting to sprunge.us might be an option in future, if the sprunge.us service comes back to life?

holta:
Resolved by changing to dpaste.com:

- PR #3748
- PR #3749

PR #3749, opened 2024-06-05
dpaste.com pastebin disallows null char — so make iiab-summary & iiab-diagnostics output safer (and force odd chars to be visible)

holta:
This is a bug fix thanks to @nzola:

It's necessary on RPi due to the null char at the end of `/proc/device-tree/model` — so this PR:

- removes all null chars from that line, in iiab-summary output
- additionally, forces odd chars to be visible when iiab-diagnostics receives iiab-summary output (just in case, as OS's evolve!)

Building on:

- #3745
- PR #3748

### DIFF
--- a/scripts/iiab-diagnostics
+++ b/scripts/iiab-diagnostics
@@ -132,7 +132,7 @@ echo "This is: $outfile" >> $outfile
 echo >> $outfile
 echo -e "\n\n\n0. HW + SW Quick Summary" >> $outfile
 echo >> $outfile
-/opt/iiab/iiab/scripts/iiab-summary >> $outfile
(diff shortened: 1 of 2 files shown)


holta:
RECAP: [iiab-diagnostics](https://github.com/iiab/iiab/blob/master/scripts/iiab-diagnostics.README.md) pastebins should now be substantially more reliable — across all HW & OS's...

- dpaste.com is better supported (and less flaky!)

- But rather different than sprunge.us so a few small output sanity changes are in place (e.g. this PR3749)

- I can wire in the dpaste.com API or CLI later if absolutely/truly later necessary, e.g. so pastebins don't expire after 30 days

holta:
PS hopefully dpaste.com is solid + sufficient for years to come!

But just FYI here's another option — that works even without `pastebinit` being installed: 😀

```
<ls-or-ps-or-any-command> | nc termbin.com 9999
```

CAVEAT: https://termbin.com pastebins are strictly limited to 1 week 🫤

PR #3752, opened 2024-06-15
Strengthen iiab-diagnostics for xklb & yt-dlp version number visibility

holta:
So debugging becomes much more efficient, in cases like:

- iiab/calibre-web#187


deldesir (commented):
Looks good.

PR #3771, opened 2024-07-18
iiab-update: minor refinements for safety + usability

holta:
1) Running the bash script with `bash -e` seems prudent during initial testing/validation especially.
2) When run in fast mode (`iiab-update -f` or `iiab-update --fast`), copying with `cp -u` (--update, i.e. files are replaced if they're older than the corresponding source file) is more appropriate — when overwriting /usr/bin/iiab-update, /usr/bin/iiab-summary and /usr/bin/iiab-diagnostics.

Related:

- iiab/calibre-web#31
- PR #3768
- PR #3769
- PR #3770

PR #3770, opened 2024-07-18
Preload all of iiab-update (bash script) into RAM — serious/intermittent errors were arising from the script effectively overwriting itself [GOAL: Beta-not-Alpha quality!]

holta:
Work in progress, building on:

- PR #3768
- PR #3769

PR #3768, opened 2024-07-18
Very basic iiab-update command, that also upgrades IIAB Calibre-Web

holta:
- This PR is introduces an extremely simple command-line tool (iiab-update) to "git pull" from IIAB's 2 main repos, while offering a few basic extras, e.g. to validate that these very basic upgrades of IIAB and Admin Console appear sane.
- Also assists with "git pull" upgrading of IIAB Calibre-Web (i.e. sync'ing from GitHub to local repo `/usr/local/calibre-web-py3`, and using pipx to upgrade xklb and yt-dlp).
- Safeguards are built in, in all 3 cases — e.g. to avoid damage if local repo is being customized (e.g. if any of the 3 local repos have files modified or have staged changes; or if any of these 3 local repos {iiab, iiab-admin-console, calibre-web} are not currently set to the "master" branch).

Related:

- iiab/calibre-web#31

PR #3774, opened 2024-07-19
`iiab-update -f` installs Calibre-Web's latest requirements.txt + mentions alternative `iiab-update` w/o flags

holta:
Thanks @avni for having tested, surfacing the problem arising from upstream changes to requirements.txt like `flask-httpAuth` suddenly appearing in https://github.com/janeczku/calibre-web/commit/2d470e0ce134339782108b4590113147e288b7f8#diff-4d7c51b1efe9043e44439a949dfd92e5827321b34082903477fd04876edb7552

So...

- `iiab-update -f` should now be substantially more resilient (e.g. when upstream changes in https://github.com/janeczku/calibre-web/commits/master/requirements.txt are merged into IIAB Calibre-Web).  All in all, it's now extremely similar to [IIAB's Ansible role](https://github.com/iiab/iiab/blob/master/roles/calibre-web/tasks/install.yml) (but much faster!)
- `iiab-update -f` should now be easier to understand, with this prominent blue warning displayed at the bottom as it finishes, reminding folks to try the alternative:

image

Tested on Ubuntu 24.10

Related:

- iiab/calibre-web#31
- PR #3768
- PR #3769
- PR #3770
- PR #3771
- PR #3772
- PR #3773

### DIFF
--- a/scripts/iiab-update
+++ b/scripts/iiab-update
@@ -21,7 +21,7 @@
         echo -e "\n\n\e[44;1mAttempting a FAST upgrade of IIAB Calibre-Web...\e[0m\n"
         echo -e "\n\e[33m'iiab-update -f' DOES NOT apply apt updates.\e[0m"
     else
-        echo -e "\n\n\e[44;1mUpgrade IIAB core software: (apt updates, Ansible, Admin Console, etc)\e[0m\n"
+        echo -e "\n\n\e[44;1mUpgrading IIAB core software: (apt updates, Ansible, Admin Console, etc)\e[0m\n"
         echo -e "\n\e[44;1mOr try 'iiab-update -f' for a FAST upgrade of IIAB Calibre-Web!\e[0m\n\n"
         echo -e "\e[4mNow running: apt update\e[0m\n"
         apt update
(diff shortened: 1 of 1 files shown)


holta:
Also tested on Ubuntu 24.04: it works (the Calibre-Web service runs) — _even if the results aren't always pretty:_
- 2 such warnings appear (every time) on each of two different Ubuntu 24.04 machines
- 3 such warnings appear (every time) on a 3rd Ubuntu 24.04 machine

image

In other words:

> Now running: bin/pip install -r requirements.txt --prefer-binary
>
> WARNING: Skipping /usr/lib/python3.12/dist-packages/argcomplete-3.1.4.dist-info                                                                                                                                                              due to invalid metadata entry 'name'
WARNING: Skipping /usr/lib/python3.12/dist-packages/argcomplete-3.1.4.dist-info                                                                                                                                                              due to invalid metadata entry 'name'
> Now running: systemctl restart calibre-web

holta:
@EMG70 asks if the following warning needs to be heeded❓ (Or can this warning possibly be ignored, do you know @EMG70 if IIAB Calibre-Web worked right after you saw this??)

IMG-20240719-WA0001

> Warning: The unit file, source configuration file or drop-ins of calibre-web.service changed on disk. Run 'systemctl daemon-reload' to reload units.

holta:
@EMG70 clarified that Calibre-Web worked — he did not need to run `systemctl daemon-reload` (so far!!)

🙏

tim-moody:
@jvonau could say more, but I believe systemd caches unit files, so if you change one and don't reload you are still using the old one until you reboot.

I think standard practice for changing unit files includes daemon-reload prior to restart.

holta:
> systemd caches unit files, so if you change one and don't reload you are still using the old one until you reboot

Yes, definitely.
I personally haven't experienced this systemd warning.
So far it appears just 1 person has encountered this:

> unit file, source configuration file or drop-ins of calibre-web.service changed on disk

I don't yet understand why... let's keep an eye on it... as the code in question (below) _should not_ be messing with systemd unit file (calibre-web.service) etc:

> cd /usr/local/calibre-web-py3
> bin/pip install -r [requirements.txt](https://github.com/iiab/calibre-web/blob/master/requirements.txt) --prefer-binary

tim-moody:
change of date might be enough to trigger it

PR #3775, opened 2024-07-20
Improved upgrading of Calibre-Web via BOTH approaches: (1) `iiab-update -f` fast way, via bash (2) `iiab-update` slow/forcible/thorough way, via Ansible

holta:
The bash version of Calibre-Web's upgrader (within `/usr/bin/iiab-update`) is increasingly now very solid, very explanatory — and in fact more resilient + more useful than the Ansible version in almost every way. 😄

So the [Ansible version](https://github.com/iiab/iiab/blob/master/roles/calibre-web/tasks/install.yml) will need to be tightened up very soon.

Hopefully as part of this PR in coming days, or coming hours.

This PR (WIP) was tested on Ubuntu 24.10.  It builds on:

- iiab/calibre-web#31
- PR #3768
- PR #3769 
- PR #3770 
- PR #3771
- PR #3772 
- PR #3773 
- PR #3774

### DIFF
--- a/roles/calibre-web/tasks/install.yml
+++ b/roles/calibre-web/tasks/install.yml
@@ -72,7 +72,7 @@
 # https://github.com/janeczku/calibre-web/pull/1459
 
 #- name: "Clone i.e. download Calibre-Web ({{ calibreweb_version }}) from {{ calibreweb_repo_url }} to {{ calibreweb_venv_path }} (~120 MB initially, ~203+ MB later) -- if {{ calibreweb_venv_path }} created just above"
-- name: "Clone (or 'git pull' to update, forcibly!) Calibre-Web ({{ calibreweb_version }}) from {{ calibreweb_repo_url }} to {{ calibreweb_venv_path }} (~122 MB initially, ~191+ MB later)"
+- name: "Clone (or 'git pull' to update, forcibly!) Calibre-Web ({{ calibreweb_version }}) from {{ calibreweb_repo_url }} to {{ calibreweb_venv_path }} (~122 MB initially, ~191+ or ~203+ MB later)"
   git:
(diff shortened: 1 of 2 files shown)


holta:
> So the [Ansible version](https://github.com/iiab/iiab/blob/master/roles/calibre-web/tasks/install.yml) will need to be tightened up very soon.

RECAP / CONTEXT: Ideally the slow/thorough Ansible approach and fast/friendly bash approach [ `iiab-update -f` ] would neither duplicate each other's install/reinstall/upgrade Calibre-Web functionalities... nor deviate too much from one another...  respecting [DRY](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself) ~= [SSOT](https://en.wikipedia.org/wiki/Single_source_of_truth ) principles as much as is feasible.

_...but for pragmatic/UX/community reasons — both approaches probably remain necessary — so I'll work on better aligning the 2 approaches, at the very bare minimum!_

🎯

holta:
Fresh install and 2 different kinds of Calibre-Web upgrades tested with Ubuntu 24.10

nzola:
> Fresh install and 2 different kinds of Calibre-Web upgrades tested with Ubuntu 24.10

Great news!

holta:
Uh Oh, Ansible's git module destructively reclones (removing commits from any test branch commits etc!) _even when `force: no` is set._ In short, Ansible has sloppy documentation, implying the opposite. As confirmed by others:

- https://stackoverflow.com/questions/59500228/git-pull-command-using-ansible#comment105175779_59500700
- https://medium.com/@vankhoa011/doing-a-git-pull-without-typing-username-password-and-doesnt-save-any-secrets-in-the-system-by-ad556c6658f0
- https://serverfault.com/questions/666564/how-to-do-a-git-pull-push-using-ansible

holta:
> Uh Oh, Ansible's git module destructively reclones (removing commits from any test branch commits etc!) _even when `force: no`_

Should now be fixed by:

- PR #3777

Issue #3856, opened 2024-12-10
hostapd would not start until I ran `rfkill unblock 1` on the latest Raspberry Pi OS (why?!)


holta:
FYI hostapd's systemd service would not start until I ran `rfkill unblock 1` to remove the **softblock of wlan** presumably arising from the latest version of Raspberry Pi OS?

Originally, `rfkill` output appeared as follows:

```
root@box:~# rfkill
ID TYPE      DEVICE      SOFT      HARD
 0 bluetooth hci0   unblocked unblocked
 1 wlan      phy0     blocked unblocked
```

In other words, hostapd started working as soon as I got `rfkill` output to appear as:

```
root@box:~# rfkill
ID TYPE      DEVICE      SOFT      HARD
 0 bluetooth hci0   unblocked unblocked
 1 wlan      phy0   unblocked unblocked
```

Suggests the following line(s) may need to be reworked?

https://github.com/iiab/iiab/blob/adfc1cc6962e2d91ab8ecc943fadb6d3dd72b394/roles/network/tasks/rpi_debian.yml#L56-L59

Others (Nov 25, 2024 thread below) would seem to be running into very similar?

https://forums.raspberrypi.com/viewtopic.php?t=379975

BACKGROUND:

- I fresh installed MEDIUM-sized IIAB.
- Onto the latest (graphical version of) 64-bit Raspberry Pi OS.
- All apt updates were applied & bootloader upgraded before beginning.
- Raspberry Pi 4 Model B Rev 1.5

```
root@box:~# rpi-eeprom-update
BOOTLOADER: up to date
   CURRENT: Mon 15 Apr 13:12:14 UTC 2024 (1713186734)
    LATEST: Mon 15 Apr 13:12:14 UTC 2024 (1713186734)
   RELEASE: default (/lib/firmware/raspberrypi/bootloader-2711/default)
            Use raspi-config to change the release.

  VL805_FW: Using bootloader EEPROM
     VL805: up to date
   CURRENT: 000138c0
    LATEST: 000138c0
```

tim-moody:
I noticed this occasionally with the previous release, but found that it resolved the second time I restarted hostapd.

holta:
> resolved the second time I restarted hostapd.

FYI in my case, rebooting and things like `systemctl restart hostapd` did not work.

holta:
Either way, we should probably update IIAB's code to run `rfkill unblock wifi` or `rfkill unblock wlan`

(Much as raspi-config does!)

[...]

PR #3866, opened 2024-12-18
deal with unblocking wifi while NM is active on newer RaspiOS

jvonau:
### Fixes bug:
#3856
as discussed 
no plans to test

### DIFF
--- a/roles/network/tasks/main.yml
+++ b/roles/network/tasks/main.yml
@@ -68,6 +68,11 @@
   #### End services
 
   #### Start network layout
+  - name: Unblock WiFi on RaspiOS
+    shell: raspi-config nonint do_wifi_country {{ host_country_code }}
+    ignore_errors: True
+    when: is_raspbian
+
   #- name: Redhat networking
   #  include_tasks: ifcfg_mods.yml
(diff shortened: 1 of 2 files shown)


holta:
Thanks for this PR which looks very promising.  Two suggestions:

- Minor Change: Let's go with `command:` which is safer that `shell:`

- I want to think about whether `ignore_errors: True` is too strong in this case — surfacing the error (e.g. if `{{ host_country_code }}` is not set acceptably) might actually be wise/best 🤔

jvonau:
> Thanks for this PR which looks very promising. Two suggestions:
> 
>     * Minor Change: Let's go with `command:` which is safer that `shell:`

You have had ample time over the years to update the removed stanza but you choose now to change the preferred syntax
 
>     * I want to think about whether `ignore_errors: True` is too strong in this case — surfacing the error (e.g. if `{{ host_country_code }}` is not set acceptably) might actually be wise/best 🤔

There was no validation of host_country_code in the past and that didn't bother you. Guess you are at your own whim on where ignore_errors is used. Good luck keeping the dynamic templates updated for the new behavior, not one more hint from me now.

PR #3872, opened 2024-12-28
Update iiab-vpn to fix ANSI color codes on macOS and change text color on blue bg

avni:
### Fixes bug:

ANSI codes show up on macOS terminal when running iiab-vpn. Screenshots available upon request. 

- Update iiab-vpn to fix ANSI color codes on macOS by using \033 instead of \e
- Also change text color to light gray when using blue bg so it's more legible. 

Tested on Ubuntu 24.04.1 LTS and macOS 15.1.1

@holta , @deldesir


holta:
Plz do test on RasPiOS (where text & emoji rendering can be flaky at times!) in the coming week-or-so after this is merged, if you can, Ok?

holta:
Thanks @avni!

> Update iiab-vpn to fix ANSI color codes on macOS by using \033 instead of \e

Just a bit of CONTEXT explaining this PR's fix for macOS:

image

- https://stackoverflow.com/questions/28782394/how-to-get-osx-shell-script-to-show-colors-in-echo
- https://unix.stackexchange.com/questions/730403/echo-e-e-does-not-print-an-escape-from-bash-script-on-macos
- https://en.wikipedia.org/wiki/ANSI_escape_code

avni:
> Plz do test on RasPiOS (where text & emoji rendering can be flaky at times!) in the coming week-or-so after this is merged, if you can, Ok?

Tested on RPi 500 with Debian 12 (bookworm) and rendering looks good!

PR #3869, opened 2024-12-19
rename apache_allow_sudo [to allow_www_data_sudo] and make True default

tim-moody:

Issue #3868, opened 2024-12-19
Retest of update_server_time


tim-moody:
Tested on W2 installed with none local vars. Neither a connection to the home page by my Android phone nor my Win11 laptop updates the system time.

In order order to be activated the iiab parameter apache_allow_sudo must be True, but it is False in default vars and all localvars except medicine. The menu parameter allow_server_time_update, set via Admin Console, must also be True.

Once I changed apache_allow_sudo and ran a reinstall on ansible and checked allow_server_time_update, both my phone and laptop updated the system time on connection to the home page.

The name apache_allow_sudo is rather stale and could be changed to allow_www_data_sudo.

holta:
Valuable insights, thanks @tim-moody !

> The name apache_allow_sudo is rather stale and could be changed to allow_www_data_sudo.

I was tempted to do something just like that ~4 years ago:

- This is not the only situation where our legacy code confusingly uses the word "apache" to mean {nginx, webserver}.

- If some or all of these can be updated to be much more accurate, that'd be very positive (if there's time, this month or soon if possible!)

tim-moody:
My question is should this be turned off by default.

holta:
> My question is should this be turned off by default.

Should be on by default in my opinion (even if yes it's not bullet-proof, obviously!)

PS @IoshuaCastillejos did the early testing a month ago, noticing that system clock is not sync'ing from nearby (WiFi-connected) phones/laptops to IIAB — and he might be able to help again in coming weeks possibly if we make progress here?

tim-moody:
>Should be on by default in my opinion

It used to be on by default I think, but was turned off to eliminate the poweroff button. I think it should be on in iiab and then either or both set time and poweroff can be turned on or off in Admin Console

tim-moody:
Unless testing was done with medicine local vars it is turned off, so of course does not set the time.

tim-moody:
#3869

jvonau:
[...]

PR #3867, opened 2024-12-19
Refine PR #3866 to unblock WiFi hotspot on newer RasPiOS with NM (NetworkManager)

holta:
### Fixes bug:

At the very least, this PR improves the implementer UX situation (no clear solution with Raspberry Pi Imager) around:

- #3856

By refining @jvonau's PR #3866 with a couple safeguards, and explanations.

Thanks to @tim-moody for verifying key facts making this possible.

### Description of changes proposed in this pull request:

WIth this PR, execution of roles/network/tasks/main.yml (equivalent to running iiab-network) intentionally stops if variable `host_country_code` was set to a bogus country value (or was not set to any value at all!)  In conjunction with raspi-config's underlying command [iw reg set "$REGDOMAIN"](https://github.com/RPi-Distro/raspi-config/blob/8a1fad85fdbdc5fc006aee8585ddccdb8eecf295/raspi-config#L889) which fails when $REGDOMAIN is not a valid country code (FYI current country code setting can be read back using `iw reg get`).

Example of execution intentionally failing with this PR, so that the operator is notified: (using intentional bogus country code `XX` in this example!)

> TASK [network : Run 'raspi-config nonint do_wifi_country XX' (using var host_country_code) to unblock WiFi, if RasPiOS] ****
fatal: [127.0.0.1]: FAILED! => {"changed": true, "cmd": ["raspi-config", "nonint", "do_wifi_country", "XX"], "delta": "0:00:00.060765", "end": "2024-12-18 19:50:12.751273", "msg": "non-zero return code", "rc": 1, "start": "2024-12-18 19:50:12.690508", "stderr": "", "stderr_lines": [], "stdout": "", "stdout_lines": []}
>
> PLAY RECAP *****************************************************************************************************************
127.0.0.1                  : ok=69   changed=10   unreachable=0    failed=1    skipped=47   rescued=0    ignored=0

### Smoke-tested on which OS or OS's:

Latest 64-bit Raspberry Pi OS (graphical version) on RPi 4.

Other country code logic pasted in here to help us keep track of it all...

```
root@raspberrypi:/opt/iiab/iiab/roles/network# grep -rni country
templates/hostapd/iiab-test-wifi.j2:40:# WiFi country code progress on arm64 OS's discussed on #3078
templates/hostapd/iiab-test-wifi.j2:46:    REG_DOM=$(grep country /run/netplan/wpa-$IFACE.conf | awk -F = '{ print $2 }')
[...]

Issue #3105, opened 2022-01-17
Files from USB Drive Fail to Appear [double mounting on Graphical Desktop OS's e.g. Ubuntu MATE]


jackrodonnell:
I have a USB stick which contains html files.  I've mounted the USB on my Lenovo T440p laptop which is running with UbuntuMate 20.04.3 and IIAB 8.0  I select "Insert a USB key/drive to publish files here" from the menu.  I see my mounted Parent Directory and my mounted USB as USB0.  However none of the HTML files appear when I select USB0

 http://sprunge.us/2zzXQZ
Thanks

jackrodonnell:
The HTML files that failed to display were on a USB with ext4.  I moved the files to a USB with Fat32.  Same problem. I removed the HTML files and added a PDF file on each of these USBs.  Both PDF files on these 2 USBs were visible via IIAB.  I then placed one HTML file and one PDF file on each of these 2 USBs.  IIAB was not able to display any of these files

tim-moody:
>I select "Insert a USB key/drive to publish files here" from the menu

what menu is that?

what url are you viewing when you say the files do not display? what when they do? http://box/USB?

what does ll /library/www/html/local_content/ show?

>I then placed one HTML file and one PDF file on each of these 2 USBs

was the html file named index.html? something else?

jackrodonnell:
Tim
See the screenshot for the menu.  
![Screenshot at 2022-01-17 10-18-33](https://user-images.githubusercontent.com/16327335/149805757-422a495c-81d8-4561-b324-2f082fc8ea3b.png)

Concerning the URL: On my local laptop that is running IIAB, I have tried both http://172.18.96.1/usb and box/usb.  Neither one results in a display of the HTML files. However I just tried accessing IIAB from my phone.  When I enter http://172.18.96.1/usb the HTML files DO NOT  appear on my phone.  When I access using box/usb the HTML files DO appear on my phone

Concerning /library/www/html/local_content/  I'm not sure where this directory is at??

Concerning "was the html file named index.html?"   YES.  When I accessed IIAB on my phone I could open index.html and all of the subsequent links functioned

Thanks
Jack

tim-moody:
what happens if you use box/local_content as the url on the two devices?

jackrodonnell:
When I access box/local_content/USB0 on my local device with IIAB or remotely on my  iPhone, the HTML files are not visible

tim-moody:
[...]

PR #3875, opened 2025-01-03
[Upload2USB] Initial commits of upload2usb app for homework assignments for students and teachers worldwide! [and, upload of photos/projects to community's USB stick]

avni:
**Initial commits of upload2usb app for homework assignments for students and teachers worldwide!**

**🏅[v0.1] P0 Features _COMPLETE!!!_**
🏁 👍 🌈 🚀 🌟 ✅ 

- [x] App design and basic functionality of uploading files to USB stick
- [x] Modularized code so that one file, hw.php, has all USB-related functions that are included on other pages. 
- [x] Create today's folder if it doesn't exist (only on upload page)
- [x] Count and display number of files uploaded on all pages
- [x] Visually Strong & Visually Distinctive "receipt" appears as soon as HW was successfully uploaded to USB stick.
- [x] Detects first removable USB drive returned by “lsblk” and gets mounted path for use (superseded) 
- [x] Make sure "sync" is one of the mount options within UM_MOUNTOPTIONS
- [x] KISS: If "teacher(s)" have plugged in multiple USB sticks, the UI should refuse to upload homework, and explain that only 1 single USB stick is permitted (when "student" wants to upload homework!)
- [x] Allow same file name but different content (use incremental numbering)
- [x] Reject duplicate content (same or different file name) 
- [x] Restrict what file types can be uploaded, using mime_content_type
- [x] Add upload2usb button from box/usb that links to app
- [x] Get usb/upload URL working (in lieu of local_content/upload)

**🍒 QA:** 
- Basic functional testing done on RPI 4 

====
**NEXT STEPS** 

**🔍 Testing:**
- Testing by others! 

**🐞 Bugs:** 
- Support ALL file names with spaces (cp and mv to USB mount point locations seems to have trouble with this for certain files. Some files with spaces work just fine.)

**🥇[v0.2] P1 Features**
- Public/private uploads  (Prevent homework copying) 

**🎷[v0.3] P2 Features**
- To protect IIABs (those based on microSD cards especially!) we definitely need to look into uploading directly to the USB stick in coming weeks. (And test Cheap USB Sticks intensively, to understand that side better too!). See [here](https://github.com/iiab/iiab/pull/3875#issuecomment-2608056864) for more info.
- Three USB Sticks: Support more than one USB drive plugged in, e.g., w/ Argon40 SSD
- Uploading 100 photos

**💡 Considerations / Functionality we are thinking through (pulled from posts below)**
- How often do FAT32 USB sticks inevitably fail — e.g., losing their filesystem (homework, photos, videos, student projects, etc) — and under what conditions (or most common with which brands of USB sticks!)
- A somewhat secret/hidden HTML button ("for teachers only") that unmounts the USB stick's filesystem. 🎩 🪄 
- File Size Constraints to prevent prevent "students" from uploading/exchanging gigantic HD movie files. 
- Support late assignment uploads
- **Scenarios to Test (concerns that it may fail)** 
    - Simultaneous uploads 
    - USB drives getting unmounted after server reboot. See [#3906](https://github.com/iiab/iiab/issues/3906)


holta:
> Three USB Sticks: Support more than one USB drive plugged in, e.g., Argon40 SSD

"KISS" Recommendation:

If "teacher(s)" have plugged in multiple USB sticks, the UI should refuse to upload homework, and explain that only 1 single USB stick is permitted (when "student" wants to upload homework!)

holta:
Beautiful PR to kick off this Central Design Challenge I'll help with in coming days: (as these 2 competing requirements below, are BOTH critical!)

1. When citizens (not always students!) upload creative & artistic projects, these SHOULD be visible as much as humanly possible (within http://box/usb) — to foster a collaborative atmosphere, within tiny / high-trust IIAB communities.

   Very much like an early BBS, to bring people together, rather than to separate them with rigid ACL regimentation.  And then of course later on, yes: there will need to be occasional policing of malicious content, allowing "teachers" to quickly pull the plug, when hostile or damaging uploading behaviors inevitably emerge.

2. Whereas when uploading multiple-choice homework worksheets and other homeworks where cheating is too high risk, "students" SHOULD NOT be able to copy other students' homework (such homeworks will be uploaded to a more hidden folder in the "teacher's" USB stick as was inserted into IIAB — so that "cheaters" cannot copy those homeworks from http://box/usb).

tim-moody:
Any restrictions on what can be uploaded, .php, .js, .exe, etc.?

holta:
> Any restrictions on what can be uploaded, .php, .js, .exe, etc.?

Definitely a worthwhile idea to pursue!

As "teachers" are certainly not all literate as to what files they should not be clicking on, when they later insert the USB stick into their Windows laptop typically.

avni:
> Any restrictions on what can be uploaded, .php, .js, .exe, etc.?

> As "teachers" are certainly not all literate as to what files they should not be clicking on, when they later insert the USB stick into their Windows laptop typically.


Not yet, but agree, it's a good idea to pursue! We can either not allow specific file types OR only allow specific formats. PHP has a _[mime_content_type](https://www.php.net/manual/en/function.mime-content-type.php)_ function that makes implementation for this easy once we decide on approach. Added to the list!

holta:
This PR is fascinating in that it democratizes the purpose of a "teacher's" USB sticks.  What was a relatively "didactic" READ-ONLY USB stick or USB drive (already a very agile and powerful device in the hands of an inspired & creative teacher!) — suddenly becomes Much More Of A READ/WRITE living/collaborative community basic BBS (bulletin board system) of sorts — that will as a result experience much more wear & tear:

- So as these new READ/WRITE community workflows are tested in coming weeks, a fascinating question will be **how often do cheap FAT32 USB sticks inevitably fail** — e.g. losing their filesystem (homeworks, photos, videos, student projects, etc) — and under what conditions (or most common with which brands of USB sticks!)

- As we gather Ground Truth here, some tricks/tweaks might well extend the life of cheap FAT32 USB sticks (that teachers in very poor communities have no choice but to depend on!)  For example, if necessary we can add a somewhat secret/hidden HTML button ("for teachers only") that unmounts the USB stick's filesystem.  Of course no such tricks are "bulletproof" magic wands!  🪄 🎩   Still, we'll explore along the way what really works ( and what doesn't 😉  ) starting with the obvious, e.g.:

  - **Linux sync Command Explained with 7 Real Life Examples ("save important files and prevent data loss")**
    https://phoenixnap.com/kb/linux-sync

ASIDE: One unintentional-but-real benefit of FAT32-formatted USB sticks is that they prevent "students" from uploading/exchanging gigantic HD movie files larger than 4GB.  In other words, hinting at all sorts of community-positive heuristics that could easily be added later on, e.g. limiting the size of uploads to 1GB or some such, to reduce the very worst abuses WHILE actively inviting the uploading of meaningful local community voices / videos! 🗳️

avni:
> If "teacher(s)" have plugged in multiple USB sticks, the UI should refuse to upload homework, and explain that only 1 single USB stick is permitted (when "student" wants to upload homework!)

Agree! Added to P0 list!

avni:
> This PR is fascinating in that it democratizes the purpose of a "teacher's" USB sticks. What was a relatively "didactic" READ-ONLY USB stick or USB drive (already a very agile and powerful device in the hands of an inspired & creative teacher!) — suddenly becomes Much More Of A READ/WRITE living/collaborative community basic BBS (bulletin board system) of sorts — that will as a result experience much more wear & tear

💟 💡❤️  

> ASIDE: One unintentional-but-real benefit of FAT32-formatted USB sticks is that they prevent "students" from uploading/exchanging gigantic HD movie files larger than 4GB. In other words, hinting at all sorts of community-positive heuristics that could easily be added later on, e.g. limiting the size of uploads to 1GB or some such, to reduce the very worst abuses WHILE actively inviting the uploading of meaningful local community voices / videos! 🗳️

Leaving aside adding file size constraints for now, but definitely something we need to think through. I'd vote to first observe usage and then see what is needed in terms of constraints, if any.

chapmanjacobd:
just a quick superficial feedback:

it may be worth using something higher level like udisksctl to manage removable devices.

also, it may be worth checking what is working well for other institutions like Canvas LMS... etc. If we can't use that off the shelf maybe check how people are using that (established patterns and existing practices) and if there are any well known pain points before reinventing the wheel

other than that, the design docs look pretty straightforward!

holta:
> Canvas LMS... etc. If we can't use that off the shelf maybe check how people are using that (established patterns and existing practices) and if there are any well known pain points before reinventing the wheel

Right!

And at the very same time building on the huge success of cheapo/disposable USB sticks as an incredibly tactile/expressive "Poor Man's LMS" in many countries over recent years, with our very basic documentation written up here about 5 years ago:

https://wiki.iiab.io/go/FAQ#Can_teachers_display_their_own_content%3F

CLARIF: One of the reasons the USB stick has worked so well (teachers typically put PDFs/videos/docs onto the USB stick using their own Windows laptop, which are then browsable by students/anyone nearby the IIAB WiFi hotspot using URL http://box/usb) is that it's almost 100% BUREAUCRACY-FREE — without any usernames/passwords or buttons/menus whatsoever.

So the goal here (I hope!) is to stick to (pun intended) that very minimal bare bones feature set, if we possibly can, thereby allowing USB Sticks to become an even more vibrant 2-WAY homework submission & VERY Basic "Poor Man's" BBS (a.k.a. Power To The People :-)

avni:
> it may be worth using something higher level like udisksctl to manage removable devices.

Thanks for this suggestion @chapmanjacobd ! Will look at udisksctl along with looking at usbmount and devmon/udevil further.

holta:
> * **Linux sync Command Explained with 7 Real Life Examples ("save important files and prevent data loss")**
>   https://phoenixnap.com/kb/linux-sync

Instead of blindly accepting ancient folklore, let's verify firsthand the claim that running Linux's `sync` command after each file is written to the USB stick (MIGHT!) rapidly accelerate the wearing out of certain USB sticks:

- Linux Sync Buffer Progress Monitoring (e.g. For Copy To USB) ('I always type “sync” in a shell command line to make sure all buffers are flushed to the USB stick before I eject the USB stick')
  https://www.kaper.com/notes/linux-sync-buffer-progress-monitoring-e-g-for-copy-to-usb/

- Force writing to USB drives... ('things like the "sync" command are needed before actually ejecting the USB drive')
  https://bbs.archlinux.org/viewtopic.php?id=263115

- Alarmist OLD NEWS from 2005, but still interesting... ('Under the right circumstances, even copying a single file to a flash
drive mounted with the "sync" option can destroy the entire drive!')
  https://groups.google.com/g/linux.kernel/c/vJ218uhISIg/m/XuEKbBmRFY0J

chapmanjacobd:
Interesting 🤔 I would think that calling `fsync` is similar to the shell's `wait` command. But if you are fsync-ing between each file for a lot of small files (relative to the size of the write buffer) I could see how that might increase ware

holta:
> let's verify firsthand the claim that running Linux's `sync` command after each file is written to the USB stick (MIGHT!) rapidly accelerate the wearing out of certain USB sticks:

Very useful summary:

https://github.com/rbrito/usbmount#generic-comments-about-flash-drives

In other words, we should test this PR as it converges by (1) verifying that USB stick is mounted with the `sync` option, and then (2) trying to wear out a few cheapo USB sticks, by saving "many hundreds" (or many thousands) of files onto these sticks, to see what really happens!!

chapmanjacobd:
ahh yeah that makes sense. Mounting with the `sync` mount option would be different from calling `sync` periodically or just before unmounting--it would be more akin to running `fsync` after each file was changed

There is some doubt in my mind about whether there will be much difference in real world scenarios--especially since cheap flash memory today is a bit more robust than it was even ten years ago (but definitely more robust than twenty years ago)

> Our chip-level measurements show endurance far in excess of nominal values quoted by manufacturers, by a factor of as much as 100. 
> https://www.usenix.org/legacy/event/fast10/tech/full_papers/boboila.pdf

> many of them may not even use 100 cycles before the stick is lost or damaged, so even 600 cycles is plenty
> https://goughlui.com/2017/05/30/experiment-8gb-usb-flash-drive-endurance-test/

It's probably fine to mount with the sync option by default (but if you're mounting read-only it's not really needed anyway)

The only issue that I would understand is if the application that is writing files relies on the write buffer to rewrite portions of the file as it is written--this will cause more ware on the drive. I think older office productivity software used to do this (Lotus123, Word, etc). I don't know how common this is today but I would assume that it is not common in the post-19xx world of high-level languages.

In my personal experience flash drives usually fail by being physically damaged--especially at the connection point--but this is also becoming less common as they've gotten smaller... more likely to be lost now

holta:
Not final, but here's the design thinking so far around 2 distinct scenarios that overseas educator types are asking us for here: _(basically their visionary idea, that @avni is now implementing, all extremely promising!)_

- ONE SCENARIO: Homeworks uploaded by students to the USB stick inside Internet-in-a-Box will sometimes lead to cheating, as students copy each others' homework.

- ANOTHER VERY DIFFERENT SCENARIO: Students / citizens upload photos or videos or projects to the USB stick inside Internet-in-a-Box, and cheating is NOT a problem — quite the opposite in fact: we should WELCOME students looking at each other's artistic / creative works — as a kind of "Poor Man's BBS" (Bulletin Board System) community meeting-of-the-minds.

In all cases, **student homeworks / citizen journalist projects would be uploaded into USB stick root-level folder: `/UPLOADS.YYYY-MM-DD`**

So the key idea is to have _TWO different kinds of USB sticks_ — in order to help teachers embrace and support BOTH above (very different!) community dynamics / scenarios:

- Teachers will simply create a root-level folder named `/SHARE` on their USB stick, if they want to stop uploads from circulating among students (a.k.a. the "too many cheaters" scenario!)  _The mere existence of this folder will lead to 2 consequences, upon inserting the USB stick into IIAB:_

- Any files INside the USB stick's `/SHARE` folder will continue to be visible (and downloadable) by other students browsing http://box/usb 👏

- Whereas any files OUTside the USB stick's `/SHARE` folder will be hidden from (and not downloadable) by other students browsing http://box/usb 🗝️

RECAP / recommendation is that **every USB stick would have 1 of 2 different "personalities..."**

Teachers can change a USB stick anytime, using their own laptop, to/from either personality:

1. Stick Personality 1 ~= "PUBLIC BULLETIN BOARD SYSTEM" (encouraging community voices & community expression), or...

2. Stick Personality 2 ~= "HIDDEN HOMEWORKS" (alongside a folder for public sharing)

COROLLARIES... ?

- Both kinds of USB sticks can be inserted into IIAB at the same time, without any problem (some sticks will hide a few things, whereas the other sticks will be radically public!)

- Just FYI upload-to-USB-drive will always be blocked when multiple USB sticks are inserted into an IIAB.  (Students and Teachers who forget will be reminded... that the "upload homework" button will only work for students... when exactly 1 single USB stick [maximum!] is inserted into that IIAB device.)

- The variable `iiab_usb_lib_show_all` will almost definitely go away, if this PR demonstrably proves to be on a good track in coming weeks.

avni:
**Jan 5 updates:** 

- The app can now be installed via the `usb_lib` role. We may change this later,  but for now, the app is available at `http://box/local_content/upload2usb/` when installed. If IIAB is already installed, you can do `./runrole --reinstall usb_lib` to get the files into the `/library/www/html/local_content` folder. 
- I also removed all instances of "homework" and the app is now renamed to 'Upload to USB' on all pages so that it's more broadly applicable to the 2 primary user needs we are solving for: 1/ public BBS and community sharing; and 2/ private HW submission.

avni:
> ahh yeah that makes sense. Mounting with the sync mount option would be different from calling sync periodically or just before unmounting--it would be more akin to running fsync after each file was changed
> It's probably fine to mount with the sync option by default (but if you're mounting read-only it's not really needed anyway)

Thanks for these specifics, @chapmanjacobd, we are going to go with this approach and then test (and ask others to test with us too!)! Thank you!

holta:
> probably fine to mount with the sync option by default (but if you're mounting read-only it's not really needed anyway)

Right!  Just for the record, here's a `/etc/usbmount/usbmount.conf` pasted in from a typical IIAB — Line 38 below confims the `sync` option is already enabled during usbmount mounts — allowing testing / refinements to begin shortly:

```console
# Configuration file for the usbmount package, which mounts removable
# storage devices when they are plugged in and unmounts them when they
# are removed.

# Change to zero to disable usbmount
ENABLED=1

# Mountpoints: These directories are eligible as mointpoints for
# removable storage devices.  A newly plugged in device is mounted on
# the first directory in this list that exists and on which nothing is
# mounted yet.
MOUNTPOINTS="/media/usb0 /media/usb1 /media/usb2 /media/usb3
             /media/usb4 /media/usb5 /media/usb6 /media/usb7"

# Filesystem types: removable storage devices are only mounted if they
# contain a filesystem type which is in this list.
FILESYSTEMS="vfat ext2 ext3 ext4 hfsplus exfat fuseblk ntfs"

#############################################################################
# WARNING!                                                                  #
#                                                                           #
# The "sync" option may not be a good choice to use with flash drives, as   #
# it forces a greater amount of writing operating on the drive. This makes  #
# the writing speed considerably lower and also leads to a faster wear out  #
# of the disk.                                                              #
#                                                                           #
# If you omit it, don't forget to use the command "sync" to synchronize the #
# data on your disk before removing the drive or you may experience data    #
# loss.                                                                     #
#                                                                           #
# It is highly recommended that you use the pumount command (as a regular   #
# user) before unplugging the device. It makes calling the "sync" command   #
# and mounting with the sync option unnecessary---this is similar to other  #
# operating system's "safely disconnect the device" option.                 #
#############################################################################
# Mount options: Options passed to the mount command with the -o flag.
# See the warning above regarding removing "sync" from the options.
MOUNTOPTIONS="sync,noexec,nodev,noatime,nodiratime"

# Filesystem type specific mount options: This variable contains a space
# separated list of strings, each which the form "-fstype=TYPE,OPTIONS".
#
# If a filesystem with a type listed here is mounted, the corresponding
# options are appended to those specificed in the MOUNTOPTIONS variable.
#
# For example, "-fstype=vfat,gid=floppy,dmask=0007,fmask=0117" would add
# the options "gid=floppy,dmask=0007,fmask=0117" when a vfat filesystem
# is mounted.
FS_MOUNTOPTIONS="-fstype=vfat,umask=0000 -fstype=ntfs,umask=0000 -fstype=exfat,umask=0000"

# If set to "yes", more information will be logged via the syslog
# facility.
VERBOSE=no
```

tim-moody:
>Whereas any files OUTside the USB stick's /SHARE folder 

I believe there are already several reserved names for the folders to share.

I would restrict the homework folder, not the sharable folder. Could call it /Private if /Homework is not general enough.

holta:
> I would restrict the homework folder, not the sharable folder. Could call it /Private if /Homework is not general enough.

This idea was considered.  It has problematic drawbacks: (1) Uploads should NOT always be private; (2) The "uploading personality" (browsable uploads versus non-browsable, but not both simultaneously to avoid student confusion!) should be **dictated in advance** by the community person or teacher that controls that stick; (3) Just FWIW NGINX folder listing functionality does not make hiding individual / nested (sub)folders very practical — in any case it seems much better to steer away from that altogether.

So a very practical way to deal with all 3 above needs, is shaping up along these lines:

- Embrace the existence of (or NON-existence of) folder `/SHARE` to dictate the "uploading personality" of each stick — each teacher-or-similar Community Builder nurtures in-person project submissions (i.e. uploads, BBS-style or not, public flavor or more private flavor) around the USB stick s/he controls.

- To keep things extremely simple, students/citizens upload to `/UPLOADS.YYYY-MM-DD` every single time — no matter which of the two "uploading personalities" any stick has.

tim-moody:
Suggest looking at the readme https://github.com/iiab/iiab/tree/master/roles/usb_lib which lists specific folder names

deldesir:
I'd like to offer some feedback on this PR.  I intended to manually test it by reviewing the design draft, but I admit I need some guidance because pulling this branch seems not adding the expecting feature for me.

I have a "SHARE" directory on a USB stick that I inserted. This directory contains files, along with some additional files located at the root of the USB drive. However, when I access http://box/usb/, no files are displayed.
For reference, here's the log: https://paste.centos.org/view/9bb2e75d

holta:
> https://paste.centos.org/view/9bb2e75d

@deldesir separately, we should understand & solve your `hostapd` failure to start, in iiab-install.log here:

> 2025-01-06 16:01:54,010 p=1996 u=root n=ansible INFO| TASK [network : Restart hostapd when WiFi is present but not when using WiFi as gateway with wifi_up_down False] *******

> 2025-01-06 16:01:56,947 p=1996 u=root n=ansible INFO| fatal: [127.0.0.1]: FAILED! => {"changed": false, "msg": "Unable to start service hostapd: Job for hostapd.service failed because the control process exited with error code.\nSee \"systemctl status hostapd.service\" and \"journalctl -xeu hostapd.service\" for details.\n"}

(QUESTION: did you set a set country code using `raspi-config` — ideally before installing IIAB?)

Related?

- #3856

deldesir:
> (QUESTION: did you set a set country code using `raspi-config` — ideally before installing IIAB?)

No, I forgot to set it prior the installation. Is it ok to set it now? (Follow up on https://github.com/iiab/iiab/issues/3856?)

holta:
> Is it ok to set it now? (Follow up on #3856?)

If #3856 is really the issue:

0. Become root: `sudo -i`
1. Try setting Wi-Fi country code using: `raspi-config`
2. Run: `iiab-network`

Related:

- PR #3867

holta:
@deldesir could this be a conflict between 2+ different country codes — FWIW all **four Wi-Fi country codes {98, 99, US, HT}** are mentioned in your iiab-diagnostics below!

- [country code] "value 98 represents a synthesized regulatory domain, based on the intersection of the available source of regulatory information (which can include the EEPROM, the userland setting, and a country IE [information elements] from your AP)." according to http://linuxwireless.sipsolutions.net/en/developers/Regulatory/CRDA/__v4.html

And/or could this be a conflict between 2 different Wi-Fi channels (1 and 6) in your arrangement?  Did you install IIAB via Wi-Fi using an AP [access point] with **channel 1** possibly?  Thereby blocking hostapd from enabling IIAB's Wi-Fi hotspot on **channel 6** ?  NOTE that IIAB's [upstream and downstream Wi-Fi channels must use the same channel](https://github.com/iiab/iiab/blob/1da4d45509868e5fe0fc5c6bf0ed71bf1fd9b8a1/vars/default_vars.yml#L10-L12) !

_(In any case, let's move this issue of `hostapd` failing to start, to a new ticket please, so this is solved/explained methodically for everyone!)_

Excerpts below from your iiab-diagnotics: (https://paste.centos.org/view/9bb2e75d)

/etc/iiab/iiab.ini contains:
```
hostapd_enabled = True
host_ssid = Konex
host_wifi_mode = g
wifi_up_down = True
host_country_code_applied = 98
host_channel = 6
client_wifi_channel = 36
```

/etc/iiab/local_vars.yml contains:
```
host_country_code: US
```

/etc/hostapd/hostapd.conf contains:
```
# Basic configuration

interface=ap0
ctrl_interface=/var/run/hostapd
ctrl_interface_group=0

ssid=Konex
channel=1
bridge=br0

# Hardware configuration
driver=nl80211
country_code=98
# limit emissions to what is legal in country
ieee80211d=1
hw_mode=g
```

COMMAND: /usr/sbin/iw reg get    # Detected WiFi country code / legal frequencies
```
global
country HT: DFS-FCC
        (2402 - 2472 @ 40), (N/A, 30), (N/A)
        (5170 - 5250 @ 80), (N/A, 24), (N/A), AUTO-BW
        (5250 - 5330 @ 80), (N/A, 24), (0 ms), DFS, AUTO-BW
        (5490 - 5730 @ 160), (N/A, 24), (0 ms), DFS
        (5735 - 5835 @ 80), (N/A, 30), (N/A)

phy#0
country 99: DFS-UNSET
        (2402 - 2482 @ 40), (6, 20), (N/A)
        (2474 - 2494 @ 20), (6, 20), (N/A)
        (5140 - 5360 @ 160), (6, 20), (N/A)
        (5460 - 5860 @ 160), (6, 20), (N/A)
```

COMMAND: /usr/bin/systemctl status hostapd    # Downstream Wi-Fi: Is hostapd running?
```
× hostapd.service - Hostapd IEEE 802.11 AP, IEEE 802.1X/WPA/WPA2/EAP/RADIUS Authenticator
     Loaded: loaded (/etc/systemd/system/hostapd.service; enabled; preset: enabled)
     Active: failed (Result: exit-code) since Mon 2025-01-06 20:13:43 EST; 47min ago
    Process: 1496 ExecStart=/usr/sbin/hostapd -B -P /run/hostapd.pid /etc/hostapd/hostapd.conf (code=exited, status=1/FAILURE)
        CPU: 12ms

Jan 06 20:13:43 box systemd[1]: Starting hostapd.service - Hostapd IEEE 802.11 AP, IEEE 802.1X/WPA/WPA2/EAP/RADIUS Authenticator...
Jan 06 20:13:43 box hostapd[1496]: Line 13: Invalid country_code '98'
Jan 06 20:13:43 box hostapd[1496]: Cannot enable IEEE 802.11d without setting the country_code
Jan 06 20:13:43 box hostapd[1496]: 2 errors found in configuration file '/etc/hostapd/hostapd.conf'
Jan 06 20:13:43 box hostapd[1496]: Failed to set up interface with /etc/hostapd/hostapd.conf
Jan 06 20:13:43 box hostapd[1496]: Failed to initialize interface
Jan 06 20:13:43 box systemd[1]: hostapd.service: Control process exited, code=exited, status=1/FAILURE
Jan 06 20:13:43 box systemd[1]: hostapd.service: Failed with result 'exit-code'.
Jan 06 20:13:43 box systemd[1]: Failed to start hostapd.service - Hostapd IEEE 802.11 AP, IEEE 802.1X/WPA/WPA2/EAP/RADIUS Authenticator.
```

```
2025-01-06 16:01:48,495 p=1996 u=root n=ansible INFO| changed: [127.0.0.1] => (item={'option': 'host_country_code_applied', 'value': '98'})
```

jvonau:
>host_channel = 6
client_wifi_channel = 36

>Jan 06 20:13:21 box networkd-dispatcher[872]: NET-DISP-WiFi wlan0 routable
Jan 06 20:13:21 box networkd-dispatcher[872]: Using channel 44 for carrier
Jan 06 20:13:21 box networkd-dispatcher[872]: Hostapd set for 6
Jan 06 20:13:21 box networkd-dispatcher[872]: Upstream Channel greater than 13 or is the same - not changing hostapd.conf

Hotspot only does 2.4Ghz channels 1-13 while NetworkManager connected to a 5Ghz network, the great guru can point out where in the docs that is spelled out.

jvonau:
> host_country_code_applied = 98

That is a [detected value](https://github.com/iiab/iiab/blob/master/roles/network/tasks/hostapd.yml#L12C1-L15C22), the desktop the using NetworkManager, so the question becomes "Where did the bogus country code come from?"

avni:
**Tuesday, 1/7 update:** 

1. Moved app from upload2usb/ to upload/ (URL to access is now: http://box/local_content/upload/) and minor cleanup. [This is committed]
2. I read up on about nginx, nested locations, alias, root, and precedence. I'm close to having http://box/usb/upload/ working though it's not quite there yet. I'm testing in: /opt/iiab/iiab/roles/nginx/templates/iiab.conf.j2 (role: www_options). References:
    - https://www.uptimia.com/learn/nginx-location-priority-explained
    - https://nginx.org/en/docs/http/ngx_http_core_module.html#alias
    - https://artfulrobot.uk/blog/untangling-nginx-location-block-matching-algorithm
3. I read up on usbmount, specifically remounting unmounted drives (as happens w/ the Argon40 periodically). 
    - Files worked in: /opt/iiab/iiab/roles/usb_lib/templates/, /usr/share/usbmount/usbmount, /opt/iiab/iiab/roles/usb_lib/templates/mount.d/70-usb-library 
    - References: https://github.com/rbrito/usbmount#hook-scripts.  
4. There's also still an issue with nginx and the webserver having permission to write to USB drives. I played around with the nginx config, file permissions, and restarting nginx, phpfhm to no avail. 
    - References: https://stackoverflow.com/questions/38632858/permission-denied-on-mkdir-php-with-nginx
    
===    
    
> I have a "SHARE" directory on a USB stick that I inserted. This directory contains files, along with some additional files located at the root of the USB drive. However, when I access http://box/usb/, no files are displayed.

To test the upload2usb app, based on the commit described in `#1` above, go to: 
http://box/local_content/upload/ to see the upload form that a student, community contributor will use.

jvonau:
@avni What is the filesystem type of the usb stick? (ie fat32 ExFat ext3) I'm going is assume some native MS format but is something that is needed to be known. Are you using a desktop based OS? Perhaps the usb stick is being mounted twice, once by usbmount and again by the desktop. #3105

Just curious about the setting of "allow_www_data_sudo" from #3869 in your local vars file being used.

avni:
> What is the filesystem type of the usb stick? (ie fat32 ExFat ext3)

FAT32
└─sdc1 vfat   FAT32 IIAB-1 84CE-1BEF                              58.6G     0% /media/usb3

> Just curious about the setting of "allow_www_data_sudo" from https://github.com/iiab/iiab/pull/3869 in your local vars file being used.

Thanks @jvonau, I think that gets us closer. I was using an older version of IIAB that had the variable set to FALSE by default, @holta updated the IIAB on this box, this evening.  After updating, I tried 3 approaches to create the **UPLOADS.YYYY-MM-DD** directory on the USB drive: 
- I tried using the php function mkdir(...) as I was using previously. 
- I changed the php to use `shell_exec ('sudo mkdir....')` in lieu of the php function `mkdir()`. 
- I also tried, `echo exec("sudo mkdir -m 755 $target_folder_path");` to match the  /www_base/files/html/services/power_off.php script. 

The `mkdir` command silently fails now and the folder is not created on the USB drive.  So not successful but closer. 

Another option is to determine the `/library/www/html/local_content/USB0|1|2...` mapping for each mountpoint and use that instead of the mountpoint directory as the parameter to `mkdir`, and see if that works.

====
> I read up on about nginx, nested locations, alias, root, and precedence. I'm close to having http://box/usb/upload/ working though it's not quite there yet. I'm testing in: /opt/iiab/iiab/roles/nginx/templates/iiab.conf.j2 (role: www_options). References:
https://www.uptimia.com/learn/nginx-location-priority-explained
https://nginx.org/en/docs/http/ngx_http_core_module.html#alias
https://artfulrobot.uk/blog/untangling-nginx-location-block-matching-algorithm

I thought I was close on this, but feel further away after trying again. Nothing too meaningful gleaned from the tests today.

More tomorrow.

jvonau:
Given usbmount mount the sticks as root and the webserver is running as www-data it does not surprise me any of the mkdir approaches tried are failing. That said there is php used within [set-server-time.php](https://github.com/iiab/iiab-admin-console/blob/master/roles/js-menu/files/menu-files/services/set-server-time.php) but has a support file placed within /etc/sudoers.d/, perhaps something similar is needed to make your php play nice. see https://github.com/iiab/iiab/issues/3868#issuecomment-2579440926

tim-moody:
In case it helps, my memory is that I had permissions problems with FAT USB sticks in that they don't really have file permissions, so they become root only when mounted in linux.

Also as I remember for Admin Console I needed the user to be able to upload image files for menu item icons and could only manage it in php. To minimize permission problems I created a www-data writeable folder outside of the web space to which the upload occurs and then used an escalation of privilege, via CMDSRV, to move the file to a folder owned by root. I also enforced some file type rules prior to moving. The big difference is that all this happened in Admin Console so we had an authenticated user.

jvonau:
Having /library/working/uploads owned by the webserver that would be a good place to store the uploads until they could be mass moved to the usbkey.

avni:
Thanks all, super helpful. I very much thought it would require modifying /etc/sudoers settings after seeing @jvonau's [post above](https://github.com/iiab/iiab/pull/3875#issuecomment-2580501177) and seeing the poweroff and set_time settings in there, BUT it didn't! I added a sudo setting to allow the apache user to execute mkdir and `shell_exec` still did not work. 

I reverted to using, the `mkdir()` php function, and it did work! And it works w/o modifying the sudo settings. 
I think what happened is that the root cause was what @jvonau  [pointed out ](https://github.com/iiab/iiab/pull/3875#issuecomment-2578161692) a few days ago with the `allow_www_data_sudo` setting being set to false on the IIAB instance I am testing on. 

After @holta updated the IIAB instance, I must have not tested mkdir() properly and only tested `shell_exec()` and `echo exec()`. 🤦‍♀️   

Oy! HUGE issue solved. Thanks all so much! 🙏

tim-moody:
I see that the name `allow_www_data_sudo`  is misleading in that it implies www-data is a sudoer. However, www-data's sudoer permission is restricted to a few commands. To do otherwise would be a huge security risk. If php running as www-data is able to execute mkdir on a directory on which www-data does not otherwise have permissions, that is a big surprise. To me it means either that the USB has no permissions due to its file system, which I had thought meant root only, or it got mounted in such a way that www-data has write permission.

avni:
That's a good point. I was curious why the /etc/sudoers settings were needed if allow_www_data_sudo made it possible for the commands to work. 

This is the permissions on the usb drive, it's owned by root as you hypothesized.
`drwxrwxrwx  5 root root 32768 Jan  1  1970 usb3`

To test, I can turn `allow_www_data_sudo` to false and see what happens. 

---
There was a day where I did mount the USB manually by using, `mount /dev/sdb1 /media/usb2` (I probably shouldn't have done that). The Argon40 does not auto mount or unmounts a disk at certain times (I haven't figured out exactly which use cases), so perhaps that was it?

jvonau:
https://github.com/iiab/iiab/blob/master/roles/usb_lib/tasks/main.yml#L35-L40

avni:
> https://github.com/iiab/iiab/blob/master/roles/usb_lib/tasks/main.yml#L35-L40

Ah! OK - I will test flipping usb_lib_umask0000_for_kolibri and allow_www_data_sudo to False one at a time and report back. Working on the actual app for a bit now that we've gotten through this bump. 

Thank you both!

jvonau:
> That's a good point. I was curious why the /etc/sudoers settings were needed if allow_www_data_sudo made it possible for the commands to work.
> 
> This is the permissions on the usb drive, it's owned by root as you hypothesized. `drwxrwxrwx 5 root root 32768 Jan 1 1970 usb3`
> 
> To test, I can turn `allow_www_data_sudo` to false and see what happens.

Well you would really need to rerun admin's install script for that change to be effective, just cheat and move the files out of that directory as root of course. I'd reboot just to be sure the new behavior is picked up.   


> There was a day where I did mount the USB manually by using, `mount /dev/sdb1 /media/usb2` (I probably shouldn't have done that). The Argon40 does not auto mount or unmounts a disk at certain times (I haven't figured out exactly which use cases), so perhaps that was it?

That would do it as the umask is not set automatically.

jvonau:
> > https://github.com/iiab/iiab/blob/master/roles/usb_lib/tasks/main.yml#L35-L40
> 
> Ah! OK - I will test flipping usb_lib_umask0000_for_kolibri and allow_www_data_sudo to False one at a time and report back. Working on the actual app for a bit now that we've gotten through this bump.
> 
> Thank you both!

The default I believe for the umask is True so that would be unnecessary

avni:
> That would do it as the umask is not set automatically.

Thanks, I think that's the root cause then. It's hard to trace all the implications of changes. I'd like to provide the user an actionable errormsg if they (for some reason) set `usb_lib_umask0000_for_kolibri` to False and things start breaking on the app for them. That's lower priority though given the default is True, as you say.

jvonau:
@avni Sent private email, too toxic here.

avni:
**1/11 Update** 

v0.1 of the app is ready for testing! Would appreciate if someone can test this PR before we merge it! Features include the following (today's commits are bolded): 

- App design and basic functionality of uploading files to USB stick
- Modularized code so that one file, hw.php, has all USB-related functions that are included on other pages. 
- Create today's folder if it doesn't exist (only on submission page)
- Count and display number of HW submitted on all pages
- Visually Strong & Visually Distinctive "receipt" appears as soon as HW was successfully uploaded to USB stick.
- Detects first removable USB drive returned by “lsblk” and gets mounted path for use (superseded)
- **Make sure "sync" is one of the mount options within UM_MOUNTOPTIONS**
- **KISS: If "teacher(s)" have plugged in multiple USB sticks, the UI should refuse to upload homework, and explain that only 1 single USB stick is permitted (when "student" wants to upload homework!)**
- **Allow same file name but different content (use incremental numbering)**
- **Reject duplicate content (same or different file name)** 
- **Restrict what file types can be uploaded, using mime_content_type**

avni:
>  I will test flipping usb_lib_umask0000_for_kolibri and allow_www_data_sudo to False one at a time and report back. 

Tested all possible combinations of the two variables: 
1. usb_lib_umask0000_for_kolibri = False, allow_www_data_sudo = False
2. usb_lib_umask0000_for_kolibri = False, allow_www_data_sudo = True
3. usb_lib_umask0000_for_kolibri = True, allow_www_data_sudo = False 
4. usb_lib_umask0000_for_kolibri = True, allow_www_data_sudo = True

_Finding/Resolution_: `mkdir` on a USB stick works when `usb_lib_umask0000_for_kolibri` is True, it doesn't work otherwise. 

Added renaming `usb_lib_umask0000_for_kolibri` to `usb_lib_umask0000_for_writable_sticks` to my running todo list.

holta:
> renaming `usb_lib_umask0000_for_kolibri` to `usb_lib_umask0000_for_writable_sticks`

Also consider shortening the variable name to `umask0000_for_writable_sticks` or `usb_lib_writable_sticks`

- There's no rule that every Ansible var name must begin with a common prefix `usb_lib_` on the left.
- Unless you're sure that you will need a wide array of `usb_lib_*` vars? (I hope not, as control should be in the hands of teachers/parents/librarians, much more than implementers!)

tim-moody:
@avni I don't think allow_www_data_sudo exists any more unless you have added it.

holta:
> @avni I don't think allow_www_data_sudo exists any more unless you have added it.

@avni there's a new variable `allow_www_data_poweroff` as of 3 days ago, which should not affect you:

- PR #3889

avni:
@EMG70 Here are the test cases for testing. Thank you! 🙏 

### TESTING SETUP (all test cases require that IIAB installation is already complete)
- sudo -i
- cd /opt/iiab/iiab
- git switch -c upload2usb
- git pull --no-rebase --no-edit https://github.com/avni/iiab master
- ./runrole --reinstall usb_lib 
- ./runrole --reinstall www_options
- Go to: http://box/usb/upload/ to test. 

### TEST CASES
**Test Case 1 (Success test case)** 
	1.  Have 1 USB Stick inserted into your IIAB
	2. The teacher asks students to upload homework assignment.
	3. Student uploads a file. 
	4. Expected Result: Return page shows success. File is visible in USB stick with the name the student gave it. 

**Test Case 2: (Sabotaging Test Case aka same file content, diff filename)**
	1.  Have 1 USB Stick inserted into your IIAB
	2. The teacher asks students to upload homework assignment.
	3. Student uploads a file, with say, a name of ‘swingingsummer.pdf’, which is successful. 
	4. Then student decides to rename same file to a different name, say, ‘swingingsummerin5countries.pdf’
	5. Expected Result: Return page shows success. File is visible in USB stick with the name the student gave it. The old file exists as well. 

****Test Case 3: (Different content, existing filename)****
	1.  Have 1 USB Stick inserted into your IIAB
	2. The teacher asks students to upload homework assignment.
	3. Student uploads a file, with say, a name of ‘whirlingdervish.docx’, which is successful. 
	4. Student then decides to name a file with different content with the same name and upload that! 
	5. Expected Result: Return page shows success. File is visible in USB stick with name whirlingdervish-2.docx. Subsequent files of the same name will be appended with -3, -4, and so on. 

**Test Case 4: Same filename, same content** 
	1.  Have 1 USB Stick inserted into your IIAB
	2. The teacher asks students to upload homework assignment.
	3. Student uploads a file, with say, a name of ‘whirlingdervish.docx’, which is successful. 
	4. Student then uploads the same file again. 
	5. Expected Result: Return page returns error, let user know file already exists. 

**Test Case 5: Test File Types we cannot allow uploaded (e.g., exe, xml)**
	1.  Have 1 USB Stick inserted into your IIAB
	2. The teacher asks students to upload homework assignment.
	3. Student uploads an xml or exe file. 
	4. Expected Result: Return page returns error, lets user know they cannot upload high-risk files. . 

**Test Case 6: 1 <> USB Stick**
	1. Have 0 USB Sticks inserted into your IIAB
	2. Go to http://box/usb/upload/
	3. Expected Result: Return page returns error, let user know one and only one USB stick must be plugged in. 
	4. Insert 2 or more USB sticks into your IIAB. 
	5. Go to http://box/usb/upload/
	6. Expected Result: Return page returns error, let user know one and only one USB stick must be plugged in. 

**Test Case 7: Simultaneous Testing (Test uploading >1 file simultaneously)**
	1. Have 1 USB Stick inserted into your IIAB
	2. Find 1 or more large files (~1 GB) to test with and put them on 2-3 devices that all can access your IIAB. 
	3. Upload the large files simultaneously from those devices via http://box/usb/upload/
	4. Expected Result: Return page shows success. Files are all visible in USB stick.

**Test Case 8: Confirm "Upload to USB" button shows up on all http://box/usb and http://box/local_content directory listing**
        1. Go to http://box/local_content/
        2. Confirm that a button that says, "Upload to USB" shows up in upper right. Click on it. 
        3. Expected Result: User is taken to page with App. 
        4. Go to http://box/usb/
        5. Confirm that a button that says, "Upload to USB" shows up in upper right. Click on it. 
        6.  Expected Result: User is taken to page with App.

holta:
> Test Case 4: Same filename, same content

Personally I wouldn't worry much about minor corner case optimizations like Test Case 4:

- Much more urgently, almost every teacher says they need homeworks w/o copying/cheating, on Day 1.

- i.e. the teacher's option to turn off "Bulletin Board" mode (and turn it back on for PBL = Project Based Learning, e.g. upload of Science Project posters & photos/videos etc!)

EMG70:
Thank you @avni .I will give it a try and let you know.

avni:
> Personally I wouldn't worry much about minor corner case optimizations like Test Case 4

A simple non-optimized implementation is in place now. Only took a few minutes.

holta:
> > it may be worth using something higher level like udisksctl to manage removable devices.
> 
> Thanks for this suggestion @chapmanjacobd ! Will look at udisksctl along with looking at usbmount and devmon/udevil further.

Not important for us, at this time especially, but just FYI 'New "mountinfo" Program Set To Be Bundled With Linux 6.14' [expected in March, 2025]

> Linux 6.14 is set to add a new "mountinfo" program to the Linux kernel source tree... The intent with mountinfo is on being a simple user-space program to demonstrate how to make use of the Linux kernel's statmount() and listmount() interfaces. In turn with this information and the output of mountinfo is the ability to get similar information to what is exposed by /proc/[pid]/mountinfo.

https://phoronix.com/news/Linux-6.14-mountinfo

EMG70:
Apologies for the poor screenshots taken from my phone.Please note,two attempts were made in uploading after remounting USB when i made some errors in uploads ,so you may find many files uploaded.

Test Case 1: ✅

Screenshot From 2025-01-21 20-59-42
Screenshot From 2025-01-21 21-00-25

------------------------------------------------------------------


Test Case 2: ✅

Screenshot From 2025-01-21 21-02-14



--------------------------------------------------------------------------------

Test Case 3:  ✅
Screenshot From 2025-01-21 22-19-01





-----------------------------------------------------------------------------------

Test Case 4: ✅

Screenshot From 2025-01-21 21-03-50

----------------------------------------------------------------------------------------------

Test Case 5: ✅

Screenshot From 2025-01-21 21-04-19
Screenshot From 2025-01-21 21-05-08

--------------------------------------------------------------------------------------------------

Test Case 6: ✅

Screenshot From 2025-01-21 21-05-25

Screenshot From 2025-01-21 21-06-02


Test Case 7: ✅ Retest
The biggest pdf 10MB ,I will retest with a larger file tomorrow.
Screenshot From 2025-01-21 22-21-28

Screenshot From 2025-01-21 22-21-49

avni:
Thanks so much @EMG70, for `Test Case #7`,  I only see one copy of the file uploaded. Did you try uploading the file from multiple devices at the same time?

roles/usb_lib/tasks/install.yml
```diff
@@ -7,6 +7,7 @@
 # https://github.com/rbrito/usbmount/blob/master/README.md (2018-08-10)
 # https://github.com/rbrito/usbmount/blob/master/usbmount.conf (2010-04-25)
 
+# usb_lib_umask0000_for_kolibri (in /etc/iiab/local_vars.yml) must be set to true in order to be able to write to mounted USB sticks
```
holta: Ansible vars can be set in many different places.

(Saying that a var must be set in local_vars.yml ...is a bit of an exaggeration.)

So I prefer using "e.g." to keep this incredibly simple — so as to avoid overwhelming newbies:

```suggestion
# usb_lib_umask0000_for_kolibri (e.g. in /etc/iiab/local_vars.yml) must be set to true in order to be able to write to mounted USB sticks
```

Full Story... is outlined here: https://github.com/iiab/iiab/wiki/IIAB-Variables#order-of-execution-and-precedence

EMG70:
> Thanks so much @EMG70, for `Test Case #7`, I only see one copy of the file uploaded. Did you try uploading the file from multiple devices at the same time?

Yes, Test case 7 was originally done on Pi400 but forgot to take screenshots.
I then tried it on the 24.04 server hence files looking a bit different.Sorry about that.

avni:
Thanks for confirming @EMG70. Can you also try `Test Case #7` uploading different files from different devices at the same time and see if they are all successful?

avni:
upload2usb button is now available on every box/usb listing page in the upper right. The button links to http://box/local_content/upload for now. Added `Test Case #8` above to test for this: [LINK](https://github.com/iiab/iiab/pull/3875#issuecomment-2599628892).

Screenshot 2025-01-22 at 4 26 45 AM

holta:
> ** Bugs:**
> 
> * Support file names with spaces (cp and mv to USB mount point locations seems to have trouble with this)

Things like `cp` and `mv` are very easy to fix, to support filenames with spaces (but "fixing these" is likely the wrong way forward! :-)

Deeper Challenge: IIAB microSD cards will be permanently damaged — during every homework/project uploaded — as a result of files (large file especially, but small files too!) constantly being written temporarily to the microSD card, yep even with modern wear-leveling. 😁 😟 Photographers know that microSD cards wear down / wear out (pro cameras have 2 slots ;) — with too many write cycles, when regularly writing files that are too large, and with temperature issues.

SO: To protect IIABs (those based on microSD cards especially!) we definitely need to look into **uploading directly to the USB** stick in coming weeks.  _(And also test Cheap USB Sticks intensively, to understand that side better too! 😄)_

EMG70:
Test results after re-testing on PI400 and students device/s.
case 1.  ✅
Screenshot 2025-01-22 190124

Screenshot 2025-01-22 190259

**********************************************************************************************


Case 2  ✅
Screenshot 2025-01-22 190522

Screenshot 2025-01-22 190820

Screenshot 2025-01-22 190538

Screenshot 2025-01-22 190858

*****************************************************************************************

Case 3:✅
Screenshot 2025-01-22 191623

Screenshot 2025-01-22 191808
Screenshot 2025-01-22 191828


******************************************************************************************


Case: 4 ✅
Screenshot 2025-01-22 192058
Screenshot 2025-01-22 193405
Screenshot 2025-01-22 192109

*********************************************************************************************

Case: 5 ✅
Screenshot 2025-01-22 193623
Screenshot 2025-01-22 193631

*********************************************************************************************

Case : 6 ✅
Screenshot 2025-01-22 193818
Screenshot 2025-01-22 193920

**********************************************************************************************

Case : 7 - To be tested with a few larger files ✅
Screenshot 2025-01-22 194825

**********************************************************************************************


Case : 8 
Screenshot From 2025-01-22 21-32-58

avni:
> Things like cp and mv are very easy to fix, to support filenames with spaces (but "fixing these" is likely the wrong way forward! :-)

Some files with spaces work fine, but some don't. Need to investigate the pattern. I updated the task in [Next Steps](https://github.com/iiab/iiab/pull/3875#issue-2767153695) to make that clear. 

> SO: To protect IIABs (those based on microSD cards especially!) we definitely need to look into uploading directly to the USB stick in coming weeks. (And also test Cheap USB Sticks intensively, to understand that side better too! 😄)

Added as a Task to 0.3

> http://box/local_content/upload 

http://box/usb/upload is now working and the App can be accessed from this URL. [Test cases](https://github.com/iiab/iiab/pull/3875#issuecomment-2599628892) are updated.

holta:
Thanks @EMG70 for all your testing in recent days, and thanks @avni for making this all happen!

A) These are the 2 highest remaining priorities from my perspective — let's try to solve these in January:

- Prevent Cheating (Copying) of homeworks, when teacher creates folder on their stick (folder "PUBLIC" might be even more clear than SHARE, etc).
- Rename usb_lib_umask0000_for_kolibri: #3907

B) These 2 might be tougher — but hopefully strong progress can be made in February:

- Write uploads directly to USB stick w/o tmp file wearing down IIAB's microSD card! #3924
- Fix "USB stick is not mounted after reboot" ? #3906

In addition to Avni's great list of UX battle-testing priorities, that she's kept updated on the top of this/original ticket: https://github.com/iiab/iiab/pull/3875#issue-2767153695

holta:
Three menu descriptions ([en-usb.json](https://github.com/iiab-share/js-menu-files/blob/master/menu-defs/en-usb.json), [fr-usb.json](https://github.com/iiab-share/js-menu-files/blob/master/menu-defs/fr-usb.json), [es-usb.json](https://github.com/iiab-share/js-menu-files/blob/master/menu-defs/es-usb.json)) are now available to people using this new IIAB capability:

image

Related:

- PR #3923

EMG70:
New IIAB installed on Raspberry Pi400
SUDO IIAB-DIAGNOSTICS - https://paste.centos.org/view/60030825


CASE 1 :✅
Screenshot From 2025-02-17 18-17-50
Screenshot From 2025-02-17 19-34-51




CASE 2 : ✅
Screenshot From 2025-02-17 18-21-56
Screenshot From 2025-02-17 18-20-57
Screenshot From 2025-02-17 19-42-27





CASE 3 :  ✅
Screenshot From 2025-02-17 18-29-50
Screenshot From 2025-02-17 18-26-41
Screenshot From 2025-02-17 19-41-13






CASE 4 :  ✅

Screenshot From 2025-02-17 18-32-13
Screenshot From 2025-02-17 18-32-25





CASE 5 :  ✅

Screenshot From 2025-02-17 18-34-44

Screenshot From 2025-02-17 18-34-56




CASE 6 :  ✅


Screenshot From 2025-02-17 18-38-07
Screenshot From 2025-02-17 18-38-51







CASE 7 : ❌    The large files 400MB and 800MB files do not show successful upload on student's computer although the files are uploaded to the server.

Screenshot From 2025-02-17 20-10-25
Screenshot From 2025-02-17 20-03-43



CASE 8:  ✅ 
Screenshot From 2025-02-17 19-54-57

avni:
> New IIAB installed on Raspberry Pi400
> SUDO IIAB-DIAGNOSTICS - https://paste.centos.org/view/60030825

NB: @EMG70 is testing again to validate that the usbmount changes in #3944 did not cause regressions.

> CASE 7 : ❌ The large files 400MB and 800MB files do not show successful upload on student's computer although the files are uploaded to the server.

@EMG70 Can you share the value of `upload_max_filesize`  in /etc/php/8.2/fpm/php.ini?

PR #3884, opened 2025-01-09
disable 020_apache_poweroff as it is unneed now

jvonau:


holta:
CLARIF on the topic of `020_apache_poweroff` from @tim-moody:

- https://github.com/iiab/iiab/issues/3868#issuecomment-2580329935

jvonau:
https://github.com/iiab/iiab-admin-console/pull/594 cleaner keeps the fiddling in one place and doesn't use the raw shell

holta:
@tim-moody review this PR if you can?

tim-moody:
Here's my take on this and https://github.com/iiab/iiab-admin-console/pull/594

I prefer setting sudo permissions in iiab, not in iiab-admin-console. Just seems more consistent with our division of functionality.

I agree that set time and poweroff permissions should be grouped into a single file (along with any future www-data permissions)

I would prefer its name to be www-data-permissions as iiab-admin is a different user. Could also use httpd-user or some such, but please don't softcode in ansible as it is harder to read and will likely never change unless we move to a new web server.

In terms of invoking poweroff through the cmdsrv script instead of directly, the only downside is the 3 second delay, which is meant to give cmdsrv time to respond to the request from the front end. If we were the maintainers of poweroff and the cmdsrv script then consolidating would be a good choice. As is it seems less important to me, but I see the point and don't really object.

jvonau:
> Here's my take on this and [iiab/iiab-admin-console#594](https://github.com/iiab/iiab-admin-console/pull/594)
> 
> I prefer setting sudo permissions in iiab, not in iiab-admin-console. Just seems more consistent with our division of functionality.

Currently there is a split that should be unified in one place and nothing in IIAB uses this change to sudo that I know of. Given the web based usage and the consumer of the sudo change I would argue that iiab-admin-console would be the preferred location. Now with the path to the script it is very clear why and what the change to sudo is used for.

> I agree that set time and poweroff permissions should be grouped into a single file (along with any future www-data permissions)
> 

Good common ground.
> I would prefer its name to be www-data-permissions as iiab-admin is a different user. Could also use httpd-user or some such, but please don't softcode in ansible as it is harder to read and will likely never change unless we move to a new web server.
> 

Done the rename but why it is only one less file to change when apache_user is renamed. lol However there may be some unforeseen change in the future so I'm not going there right now.

> In terms of invoking poweroff through the cmdsrv script instead of directly, the only downside is the 3 second delay, which is meant to give cmdsrv time to respond to the request from the front end. If we were the maintainers of poweroff and the cmdsrv script then consolidating would be a good choice. As is it seems less important to me, but I see the point and don't really object.

holta:
I don't have an opinion:

Except that refactoring (and cleaner naming) can be incredibly healthy at times, as new realities inevitably come to light, whatever is mutually decided!

tim-moody:
>Currently there is a split that should be unified in one place

what is the split, the permission and the scripts? the permissions were all for system utilities before the change to poweroff.

I think all the sudoer config belongs in one place. The sudoer config is in iiab; the sudoer password check is in iiab; the iiab-admin user is in iiab. I think this one belongs there too.

jvonau:
> > Currently there is a split that should be unified in one place
> 
> what is the split, the permission and the scripts? the permissions were all for system utilities before the change to poweroff.

Well you have the drop in file in iiab and the consumer both shutdown and set-time are provided by  iiab-admin-console.
> I think all the sudoer config belongs in one place. The sudoer config is in iiab; the sudoer password check is in iiab; the iiab-admin user is in iiab. I think this one belongs there too.

All related to the creation of the iiab-admin user and grouped under roles/iiab-admin.

tim-moody:
In that case move the whole lot, assuming you're OK to have iiab-admin-console own sudoer

jvonau:
> In that case move the whole lot, assuming you're OK to have iiab-admin-console own sudoer

You have the keys to the kingdom, I could care less as the creation of iiab-admin is disjointed from the curl routine that calls [iiab](https://github.com/iiab/iiab-factory/blob/master/iiab) and ansible in this repo anyway.

jvonau:
Never mind [power_off.php](https://github.com/iiab/iiab/blob/master/roles/www_base/files/html/services/power_off.php) is supplied from here

PR #3892, opened 2025-01-16
don't restart wpa_supplicant when network_manager_active is set

jvonau:
### Fixes bug:
Noted traceback while reviewing

PR #3893, opened 2025-01-16
don't mess with the filesystem when firmware is going to be stock

jvonau:
https://github.com/iiab/iiab/pull/3892#issuecomment-2596211157

### DIFF
--- a/roles/network/tasks/main.yml
+++ b/roles/network/tasks/main.yml
@@ -106,7 +106,7 @@
 - name: Select RPi firmware mode
   include_role:
     name: firmware
-  when: rpi_model != "none"
+  when: rpi_model != "none" and rpizerow_rpi3_wifi_firmware != "os" and rpi3bplus_rpi4_wifi_firmware != "os"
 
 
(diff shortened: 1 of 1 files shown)


holta:
I appreciate that this is well-intended.

But I'm personally not in favor as this would make it much harder to restore "os" (i.e. original, stock Raspberry Pi OS) WiFi firmware, among implementers who want that:

- Original design of these 2 variables goes to great lengths to keep as much control in the hands of implementers as possible.

- Setting these 2 variables to "os" during initial IIAB install does little more than back up stock/OS firmware, leaving essentially everything else untouched, just in case the RasPiOS default firmware evolves over time.  So as to give implementer full control among all options.

jvonau:
You admitted you were incorrect in your assumptions about rpi5 firmware, perhaps there are other assumptions are incorrect. This would go towards helping with https://github.com/iiab/iiab/pull/3892#issuecomment-2596042644

jvonau:
#3895

PR #3899, opened 2025-01-17
iiab-diagnostics: A bit more logging around wpa_supplicant, NetworkManager, dmesg | grep Firmware, rpi-eeprom-update to show bootloader version

holta:
If some of these logs prove overweight / unneeded, they can later be removed!

Related:

- https://github.com/iiab/iiab/issues/3897#issuecomment-2598171506


holta:
@jvonau I hope 100 lines are enough, from `journalctl -b | grep NetworkManager | head -100` ?

(But if not, let me know, and we can readjust — keeping iiab-diagnostics's pastebin bloat in mind if possible.)

PR #3908, opened 2025-01-20
detect Firmware rejected country setting [e.g. for DRC Congo's `CD` WiFi country code]

jvonau:

PR #3909, opened 2025-01-22
iiab-diagnostics: Refine 3+ 'sudo dmesg | grep ...' cmds

holta:
Seeks to uncover a few key/missing details as discussed within:

- https://github.com/iiab/iiab/pull/3908#issuecomment-2608240104

Quickly tested on Ubuntu and Raspberry Pi OS.

Building on:

- PR #3899
- PR #3901


jvonau:
Not sure if sudo is needed, think iiab-diagnostics was run with sudo as  
>-rw------- 1 root root 302 Nov 19 14:44 /etc/NetworkManager/system-connections/preconfigured.nmconnection

has the contents listed and is 0600.

holta:
Indeed sudo's not always needed.

Conversely, in some cases that I tested (e.g. raw Ubuntu) adding sudo helps reveal much more, for certain commands.

PR #3915, opened 2025-01-25
usb_lib/*, vars/*: prevent cheating (copying) of homework

avni:
usb_lib/*, vars/*: prevent cheating (copying) of homework when teacher creates 'PUBLIC' folder on their USB stick

### Description of changes proposed in this pull request:
- Rework usb_lib to support public and private content on USB sticks to prevent cheating (copying) of homework. 
- When teacher creates a 'PUBLIC' folder on their USB stick, only content from that folder is visible in the http://box/usb/ directory listing; all other content is hidden in the browser. 
- If no 'PUBLIC' folder exists, all content on USB stick is visible. 
- Comment out the IIAB_USB_LIB_SHOW_ALL variable as it's no longer used
- Add j2 extension to mount.d/70-usb-library given it's a template
- Move umount.d/70-usb-library to unused as it's not used; iiab-clean-usb.sh is used for unmounting disks.
- Cleanup logging

### Smoke-tested on which OS or OS's:
Smoke tested on RPi4. 

### Mention a team member @username e.g. to help with code review:
@holta

PR #3913, opened 2025-01-25
vars/* and usb_lib/*: rename usb_lib_umask0000_for_kolibri

avni:
Addresses #3907 

### Description of changes proposed in this pull request:
renames usb_lib_umask0000_for_kolibri to usb_lib_writable_sticks so it is more generally applicable. modifies multiple files in vars/ and roles/usb_lib/

### Smoke-tested on which OS or OS's:
smoke-tested on RPi4 by exporting Kolibri channel and uploading a few files in upload2usb app. 

### Mention a team member @username e.g. to help with code review:
@holta

PR #3918, opened 2025-01-26
upload/upload-file.php: fix var name bug and add better error handling

avni:
### Fixes bug:
- Throw exception if file is not able to be copied from /tmp to USB stick
- Remove reference to non-existing variable

RPi4 

@holta


holta:
Ok to merge this ASAP?

Hopefully this will not create any merge conflicts with things like PR #3913 and #3915, right?

Related:

- PR #3875

holta:
@avni confirmed this appears very safe.

PR #3927, opened 2025-01-29
PHP's own defaults are Way Too Low, so run www_options/tasks/php-settings.yml every time

holta:
This simplification was probably long overdue, providing much more practical PHP defaults for smaller IIAB installs, that also helps streamline unit-testing of student uploads to teacher's USB stick, e.g. things like:

- PR #3875
- PR #3913
- PR #3914
- PR #3916
- PR #3918 
- PR #3919
- PR #3923

PR #3931, opened 2025-02-01
Restore both GitHub Actions workflows simulating Raspberry Pi

holta:
### Fixes bug:

While not absolutely necessary it can't hurt to run these two CI tests simulating RPi 3 and RPi Zero 2 W?

So this TZ (Time Zone) hack works around the PR #3927 problem in php-settings.yml, as originally outlined at: https://github.com/iiab/iiab/commit/4da759a84b9913a04186c97c4e30c73cbf4784a8

- Shell hack does the equivalent of `timedatectl show -p Timezone --value` to extract the actual TZ from all the correct places in /etc — so this can be populated into PHP
- Updates the description of USB_LIB for iiab.ini in roles/usb_lib/tasks/main.yml

All 3 GitHub Actions workflows (Ubuntu 24.04, debian12 on RPi 3, RasPiOS on Zero 2 W).

### DIFF
--- a/roles/usb_lib/tasks/main.yml
+++ b/roles/usb_lib/tasks/main.yml
@@ -77,7 +77,7 @@
     - option: name
       value: USB_LIB
     - option: description
-      value: '"USB_LIB automounts Teacher Content on USB drives to /library/www/html/local_content, so students can browse it almost immediately at http://box/usb"'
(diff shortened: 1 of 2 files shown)


holta:
Good enough!

GitHub Actions workflows can & should be improved later, to better simulate actual Raspberry Pis!

PR #3952, opened 2025-02-18
for discussion [Ansible Collections further into venv?]

jvonau:
For discussion

PR #3950, opened 2025-02-17
pip install {pymysql, psycopg, passlib} to Ansible venv, on demand — to fix 6 IIAB Apps with new `interpreter_python=/usr/local/ansible/bin/python3` in ansible.cfg

holta:
### Fixes bug:

Resolves 6 IIAB App regressions, arising from /opt/iiab/iiab/ansible.cfg's change from `interpreter_python=/usr/bin/python3` to `interpreter_python=/usr/local/ansible/bin/python3` as outlined here:

- https://github.com/iiab/iiab/pull/3948#issuecomment-2661845405

Originally arising from:

- #3945

### Description of changes proposed in this pull request:

pip install each of the 3 on-demand...

1. `pymysql` for MySQL for {mediawiki, nextcloud, wordpress, matomo, pbx}
2. `psycopg` for PostgreSQL for Moodle
3. `passlib` for Munin

...on demand, instead of apt/deb installing each of the 3 python3-XXX packages system-wide.

psycopg2 is replaced with `psycopg` — which is really the new psycopg3.

pip flag `--upgrade` is added to try to allow for (some!) after-the-fact upgrades of Ansible, e.g. so things _aren't quite as fragile (while not perfect!)_ if an IIAB field implementer later runs: `sudo /opt/iiab/iiab/scripts/ansible`

### Smoke-tested on which OS or OS's:

Ubuntu 25.04 but needs other tests...

### Mention a team member @username e.g. to help with code review:

@jvonau

PR #3948, opened 2025-02-16
JupHub [JupyterHub fix & cleanup for Debian 13 Trixie, using full path to Python 3.13 in /opt/iiab/iiab/ansible.cfg] 

jvonau:
alternate to #3947

PR #3947, opened 2025-02-16
(1) Debian 13 requires python3-packaging for JupyterHub (2) python3-psutil is currently not necessary

holta:
### Fixes bug:

- #3945

### Description of changes proposed in this pull request:

Mandate installation of apt package `python3-packaging` rather than `python3-psutil`.

### Smoke-tested on which OS or OS's:

Ubuntu 24.04, Debian 12, Debian 13 (latest pre-release).

Still needs testing on RasPiOS 12 (and eventually RasPiOS 13, later this year).

### Mention a team member @username e.g. to help with code review:

@jvonau with the original python3-venv and python3-psutil discussion from April 2023 here:

- PR #3554

PR #3954, opened 2025-02-19
Further experiment with Ansible Collections in venv

holta:
### Fixes bug:

Myself I don't yet see how moving Ansible Collections away from Ansible's default location (/usr/share/ansible/collections) to our Ansible venv genuinely helps.

However if there's solid/demonstrable need for moving Ansible Collections to venv, and this Ansible Collections in venv idea can somehow reduce code maintenance overall, then this tightened up PR could potentially be an alternate to @jvonau's:

- PR #3952

If Ansible Collections in venv prove wise/necessary[*] then this PR can help tighten up.

[*] So far I personally don't see overall value, as this idea has the side effect of increasing code complexity, leading to additional code maintenance.

Ubuntu 24.04 and 25.04, in VM and ~2 GitHub Actions emulating RPi.

### Recent / Related:

- #3945
- PR #3947
- PR #3948
- PR #3950

### DIFF
--- a/ansible.cfg
+++ b/ansible.cfg
@@ -6,3 +6,6 @@
 #squash_actions = apk, apt, dnf, homebrew, openbsd_pkg, pacman, pkgng, yum, zypper, package
 [defaults]
 interpreter_python=/usr/local/ansible/bin/python3
+collections_path=/usr/local/ansible/collections
+# ...overriding default shown by 'ansible --version' here:
+# ansible collection location = /root/.ansible/collections:/usr/share/ansible/collections
--- a/scripts/ansible
+++ b/scripts/ansible
@@ -221,6 +221,16 @@ if [[ $(dpkg --print-architecture) == "i386" ]] && apt-cache show ansible-core >
     # 2023-08-10 #3613/#3615/#3617: apt-not-pip kludge for legacy 32-bit i386
     # (DEBIAN 12+ ETC) avoids #3547 rust/wheels/cryptography compiling mess!
     $APT_PATH/apt -y install ansible-core    # Bookworm ~= ansible-core 2.14.3
+
+    echo -e "\n\nIIAB requires these ~4 Ansible Collections: (we upgrade them here if possible!)\n"
+    cat > /opt/iiab/iiab/ansible.cfg << EOF
+[defaults]
+interpreter_python=/usr/bin/python3
+EOF
+    git commit -am "Alter ansible.cfg interpreter_python for deb install of ansible-core"
+    ansible-galaxy collection install --force-with-deps \
+                                      -r /opt/iiab/iiab/collections.yml \
+                                      -p /usr/share/ansible/collections
 else
     # 2023-03-22: OS's like Ubuntu 23.04 and Debian 12 (e.g. with Python 3.11+) ask
     # that virtual environments (venv) be used to safely isolate pip installs:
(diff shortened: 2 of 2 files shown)


jvonau:
https://github.com/iiab/iiab/pull/3952#issuecomment-2669830371

jvonau:
For admin-console perhaps these two could be symlinks to the files in iiab?
[ansible.cfg](https://github.com/iiab/iiab-admin-console/blob/master/ansible.cfg)
[ansible_hosts](https://github.com/iiab/iiab/blob/master/ansible_hosts)

holta:
> For admin-console perhaps these two could be symlinks to the files in iiab? [ansible.cfg](https://github.com/iiab/iiab-admin-console/blob/master/ansible.cfg) [ansible_hosts](https://github.com/iiab/iiab/blob/master/ansible_hosts)

Very worthwhile possibility.

> /usr/share/ansible/collections or... /usr/local/ansible/collections/

So far my vote is to preserve Ansible norm `/usr/share/ansible/collections` (this also reduces code complexity, and keeps what's working well).

jvonau:
> > For admin-console perhaps these two could be symlinks to the files in iiab? [ansible.cfg](https://github.com/iiab/iiab-admin-console/blob/master/ansible.cfg) [ansible_hosts](https://github.com/iiab/iiab/blob/master/ansible_hosts)
> 
> Very worthwhile possibility.
> 
> > /usr/share/ansible/collections or... /usr/local/ansible/collections/
> 
> So far my vote is to preserve Ansible norm `/usr/share/ansible/collections` (this also reduces code complexity, and keeps what's working well).

The "norms" you speak of is based on flawed instructions to use 'pip install ansible-core' which you should know [is not is not possible with the latest os package managers.](https://github.com/iiab/iiab/blob/master/scripts/ansible#L225-L227) While pipx just creates the virtual environment in $HOME. 
[https://github.com/iiab/iiab/blob/master/scripts/ansible.md.unused](https://github.com/iiab/iiab/commit/eb9d4648868f3e1120491878e2e95780e6aea853) is kind of sad given I did the work in 2028
What has comeback is sort of like 2885 [just for pip instead of deb packaging](https://github.com/iiab/iiab/blob/master/scripts/ansible#L176-L186). 


background:
- https://github.com/iiab/iiab/pull/2028
- https://github.com/iiab/iiab/pull/2030
- https://github.com/iiab/iiab/pull/2885 
- https://github.com/iiab/iiab/pull/3494 
- https://github.com/iiab/iiab/pull/3496

The use of collections began when [os package installs](https://github.com/iiab/iiab/pull/2647/files#diff-372c898a6ec3422ed280fedc1ccef0d9aa7e810951cf0da88111dad4359a65d6L89-L93) were used and the os package manager used /usr/share/ansible before the base/core split.
- https://github.com/iiab/iiab/pull/2647
- https://github.com/iiab/iiab/pull/2655

What 3947 pointed out was the collections being in /usr/share/ansible needed to have the apt packages installed so installing the collections' dependencies into ansible's virtual environment is the right thing to do thus dropping the os packages needed.
- https://github.com/iiab/iiab/pull/3947
- https://github.com/iiab/iiab/pull/3948
- https://github.com/iiab/iiab/pull/3950

Having /usr/share/ansible hold the collections seems more for os manager packaging and upstream's official owner seems to suggest the collections should be installed into the [virtual environment](https://www.redhat.com/en/blog/python-venv-ansible) that would install the collections into $HOME/.ansible but we want to be able to cruse the collections as an ordinary user not hidden in /root. Having the collections stored in the virtual environment make it very clear that the file are ansible related. Just trying to being ahead of the curve before the best practices in the upstream docs are updated along with the default values being used.

holta:
That's a very useful writeup.  Thanks @jvonau for articulating several of these evolving tensions.

At the same time let's be careful not to get out ahead of our skis (premature generalization is the root of all evil ~ paraphrasing Knuth / 1974).  So Debian 13 will presumably be interesting and influential this year, wherever it goes in the end, but it and RasPiOS 13 won't be official for another few months.  We'll also know a lot more later this year when Ubuntu 25.10 and 26.04 LTS pre-releases make evolving norms even more clear here, one way or another, i.e. where this is all really headed in 2025.  And of course recent days have become messier than necessary, as such norms being in flux is to be expected in years like this.  Just a few notes for now:

- Python venv's can definitely mean incremental progress in some areas.  However, nor should we treat them as a Docker-like solution to all things (e.g. let's not overload Python venv's prematurely, especially when this is right now leading to complexity/issues/bugs).

- Now we know in hindsight that PR https://github.com/iiab/iiab/pull/3947 would have been a bureaucracy-free interim solution, avoiding several of the current "blowback" issues arising from https://github.com/iiab/iiab/pull/3948 (just one additional example is... Ansible install for [i386](https://github.com/iiab/iiab/blob/1308fb86708c00b4f7ac49c24e8cd1f515dfc4dd/scripts/ansible#L220-L223) used by a few legacy hardware folks, is still basically broken since Monday, as it's lacking what https://github.com/iiab/iiab/pull/3950 inject into our Ansible venv).

- I should have more time later this week to dig a bit more.  At the same time, we cannot expect a "Silver Bullet" solution (we will not find one!)  Certainly a "philosophically pure" approach can sometimes be attractive, if we don't get too far ahead of ourselves.  But conversely, our work lives in more-or-less-FOSS ecosystems shaped by many others — evolving organically at its own pace — and the release of Trixie OS's (Debian 13 or RasPiOS 13) followed by Ubuntu 25.10 and 26.04 pre-releases right after that, will all help confirm critical trends here over the course of 2025.  i.e. which approaches/norms will truly become more mainline than others (or at least broadly adopted & documented etc!)

jvonau:
Some better docs:
https://docs.ansible.com/ansible/latest/reference_appendices/config.html
https://docs.ansible.com/ansible/latest/cli/ansible-config.html
https://forum.ansible.com/t/location-of-collections/5724/7
https://docs.ansible.com/ansible/latest/collections_guide/index.html
https://docs.ansible.com/ansible/latest/collections_guide/collections_installing.html
https://docs.ansible.com/ansible/latest/collections_guide/collections_using_playbooks.html
https://docs.ansible.com/ansible/latest/collections/ansible/builtin/pip_module.html
https://docs.ansible.com/ansible/latest/collections/community/general/htpasswd_module.html
https://docs.ansible.com/ansible/latest/collections/community/general/ini_file_module.html

jvonau:
These would further help clarifying which collection is in use. 
https://github.com/iiab/iiab/pull/3952/commits/c2b634a6d2a3fc8c5e4df5c6b680ea3493e383f4
https://github.com/iiab/iiab/pull/3952/commits/6f0ceb78a4cb371027e7b096503178df6114be6d
https://github.com/iiab/iiab/pull/3952/commits/a706531c84514e87a71039e7d08ce271acebb3e0
https://github.com/iiab/iiab/pull/3952/commits/9fb521953c04ee40a2602f7faa4dcef8a8bb6af8
https://github.com/iiab/iiab/pull/3952/commits/1c1606cf52e838c641db6dbb91fe36a7adbc4a7d

#3955

PR #3926, opened 2025-01-29
iiab-diagnostics: 3 commands to record locale info

holta:
Towards helping everyone to quickly get to the bottom of locale issues like:

- #3882
- #3917

Issue #3917, opened 2025-01-26
Changing locale in raspi-os desktop breaks ansible


Dphilly-215:
I've installed iiab twice now using the pi imager. The first time I was able to download maps and change the default pw. After reboot I try to login to admin console locally from the pi and get this error get-adm-conf failed iiab-cmdsrv failed to initialize. Tried to restart the iiab cmdsrv service, it didn't help. Reinstalled using the pi imager again. Didn't do anything but change my location in pi settings from UK to US and again same error can't get into admin. On first boot I'm able to get into admin not sure why changing location breaks the ability to login. On my 3rd reinstall now not sure what the issue is. Plenty of space on SD, 128gb sd card.

holta:
Hi @Dphilly-215

Can you run `sudo iiab-diagnostics` and post the resulting URL here to help us diagnose?

tim-moody:
@Dphilly-215 you seem comfortable at the linux terminal. You could also help me diagnose what is happening by doing the following as root:

```
cd /opt/admin/cmdsrv/
systemctl stop iiab-cmdsrv
./iiab-cmdsrv3.py
```

You are now running cmdsrv manually and errors will not be trapped, so you should get more information about any error.

Please paste in any errors that appear on the terminal screen.

jvonau:
>Didn't do anything but change my location in pi settings from UK to US

How did you do that, with raspi-imager or with raspi-config, did you reboot afterwards?

Dphilly-215:
@jvonau i used the raspi config manager on pi image in preferences. I'm on my 3rd install and haven't touched anything in the pi config, no issues so far. The last install what I changed was location and wlan country. Changed wlan from unset to US. The other suggestions above I will try soon and provide a diag report. @tim-moody I'm doing another install on a different sd card so I don't mess up the work I've done already. I'll "break" this one and send information.

Dphilly-215:
[8.1_raspbian-11_2025-01-26_13%3A44%3A56_EST_dphilly.txt](https://github.com/user-attachments/files/18551256/8.1_raspbian-11_2025-01-26_13.3A44.3A56_EST_dphilly.txt)

Dphilly-215:
@tim-moody from how I understand this, changing the location changed the ISO type causing my issue?

root@box:/opt/admin/cmdsrv# systemctl stop iiab-cmdsrv
root@box:/opt/admin/cmdsrv# ./iiab-cmdsrv3.py
[...]

Issue #3882, opened 2025-01-09
Unable to access Admin Console ["Invalid Login" even with correct password & Ansible would not install — due to faulty /etc/default/locale -> /etc/locale.conf]


isuzumac:
I am all of a sudden experiencing being locked out of admin console, when I first set up iiab I changed the password and have been able to login with no issues but now it's telling me "invalid login"

Diagnostic Link
https://paste.centos.org/view/bc3ef7d3

holta:
@isuzumac, Does changing your password at the Linux command-line help?

(In other words, use the `passwd` command while logged in as the user in question!)

Refs:

- iiab/iiab-admin-console#593
- #3876

isuzumac:
> @isuzumac, Does changing your password at the Linux command-line help?
> 
> (In other words, use the `passwd` command while logged in as the user in question!)
> 
> * [Invalid Login Admin Console  iiab-admin-console#593](https://github.com/iiab/iiab-admin-console/issues/593)

It does not unfortunately.

holta:
What username are you using?  Screenshot(s) of failure please?

isuzumac:
@holta Using user name iiab-admin
![Screenshot_20250108_180727_Chrome](https://github.com/user-attachments/assets/ef2e9a5a-41b4-4956-8b0a-7aeff4acb7bf)

holta:
How exactly are you setting the password for `iiab-admin` ?

Screenshot if possible, including output of the `whoami` command?

isuzumac:
Screenshot shows how I am setting the password 
![20250108_190626](https://github.com/user-attachments/assets/1f087a9f-7bc7-4413-8583-04ee65426c4c)

holta:
@tim-moody any idea what's happening here?

holta:
@isuzumac if you don't figure this out on your own after rebooting etc — we can arrange a VPN-like connection to log into your IIAB directly, if you want us to diagnose in coming days?

isuzumac:
That would be awesome!! Please let me know what time works best for you guys!! 

I ran sudo iiab-update to try and resolve my issues and was met with this.
![20250108_205516](https://github.com/user-attachments/assets/4c6ebad6-53b6-4765-9452-7e6512429422)

holta:
@isuzumac are you running `iiab-update` as root?  (e.g. did you run `sudo iiab-update` ?)

isuzumac:
Yes I ran `sudo iiab-update`

[...]

PR #3958, opened 2025-03-01
iiab-diagnostics: Also record /etc/locale.conf (e.g. for Debian 13 Trixie), in addition to legacy file-or-symlink /etc/default/locale

holta:
Related:

- https://github.com/iiab/iiab/issues/3882#issuecomment-2692300227
- #3917
- PR #3926
- https://wiki.iiab.io/go/FAQ#What_to_do_if_Raspberry_Pi_OS_messes_up_language_locale?


holta:
CLARIF on the purpose of this PR:

While this PR is excessive (recording file or symlink /etc/default/locale was probably more than enough for coming years), this PR might help IIAB be a bit more distro-agnostic, and anticipates the possible case where /etc/locale.conf is its very own file WITHOUT the symlink from /etc/default/locale in future years.

Issue #3961, opened 2025-03-05
Can IIAB offer better Ansible messaging when hostapd fails to (re)start during IIAB installs e.g. on "RasPiOS with desktop" ? [should 'sudo iiab' try to rerun iiab-network when it fails to complete? enforce WiFi country code in certain conditions?]


holta:
Can we offer better Ansible messaging to implementers who get stuck during vanilla IIAB installs like this?

> TASK [network : Waiting 10 seconds for network to stabilize for ap0] ***********
changed: [127.0.0.1]
>
> TASK [network : Restart hostapd when WiFi is present but not when using WiFi as gateway] ***
fatal: [127.0.0.1]: FAILED! => {"changed": false, "msg": "Unable to start service hostapd: Job for hostapd.service failed because the control process exited with error code.\nSee \"systemctl status hostapd.service\" and \"journalctl -xeu hostapd.service\" for details.\n"}
>
> PLAY RECAP *********************************************************************
127.0.0.1                  : ok=787  changed=441  unreachable=0    failed=1    skipped=193  rescued=0    ignored=2

```
root@raspberrypi:~# systemctl status hostapd
× hostapd.service - Hostapd IEEE 802.11 AP, IEEE 802.1X/WPA/WPA2/EAP/RADIUS Authenticator
     Loaded: loaded (/etc/systemd/system/hostapd.service; enabled; preset: enabled)
     Active: failed (Result: exit-code) since Wed 2025-03-05 13:36:38 EST; 11min ago
    Process: 57593 ExecStart=/usr/sbin/hostapd -B -P /run/hostapd.pid /etc/hostapd/hostapd.conf (code=exited, status=1/FAILURE)
        CPU: 17ms

Mar 05 13:36:38 box hostapd[57593]: nl80211: deinit ifname=ap0 disabled_11b_rates=0
Mar 05 13:36:38 box hostapd[57593]: Could not read interface ap0 flags: No such device
Mar 05 13:36:38 box hostapd[57593]: nl80211 driver initialization failed.
Mar 05 13:36:38 box hostapd[57593]: ap0: interface state UNINITIALIZED->DISABLED
Mar 05 13:36:38 box hostapd[57593]: ap0: AP-DISABLED
Mar 05 13:36:38 box hostapd[57593]: ap0: CTRL-EVENT-TERMINATING
Mar 05 13:36:38 box hostapd[57593]: hostapd_free_hapd_data: Interface ap0 wasn't started
[...]

Issue #3912, opened 2025-01-24
Firmware rejected country setting issue remains


jvonau:
> @holta When are you opening issues around the bogus countries in raspi-imager and raspi-config that are not part of [wireless-regdb](https://git.kernel.org/pub/scm/linux/kernel/git/wens/wireless-regdb.git/)'s [db.txt](https://git.kernel.org/pub/scm/linux/kernel/git/wens/wireless-regdb.git/plain/db.txt)?
> 
> https://github.com/iiab/iiab/issues/3890#issuecomment-2601007839
> [real_clean.txt](https://github.com/user-attachments/files/18470294/real_clean.txt) Anything with a + are in iso  but not in db.txt.
> 
> https://github.com/iiab/iiab/issues/3890#issuecomment-2608227256 
> [imager-diff.txt](https://github.com/user-attachments/files/18511346/imager-diff.txt) Anything with a + are in raspi-imager's countries.txt but not in db.txt.
> 
> These are a bit of a problem as wpa_supplicant logging is getting spammed really bad and leaves 00 as an unset set country code. 

 _Originally posted by @jvonau in [#3890](https://github.com/iiab/iiab/issues/3890#issuecomment-2611256121)_

jvonau:
From 767bf738 snipped because it is overwhelming
>COMMAND: /usr/bin/sudo journalctl -b | grep wpa_supplicant    # wpa_supplicant log since boot
 Jan 23 11:36:21 iiab-test-cd systemd[1]: Starting wpa_supplicant.service - WPA supplicant...
Jan 23 11:36:21 iiab-test-cd wpa_supplicant[532]: Successfully initialized wpa_supplicant
Jan 23 11:36:21 iiab-test-cd systemd[1]: Started wpa_supplicant.service - WPA supplicant.
Jan 23 11:36:23 iiab-test-cd wpa_supplicant[532]: nl80211: kernel reports: Registration to specific type not supported
Jan 23 11:36:23 iiab-test-cd wpa_supplicant[532]: wlan0: CTRL-EVENT-REGDOM-CHANGE init=CORE type=WORLD
Jan 23 11:36:29 iiab-test-cd wpa_supplicant[532]: wlan0: CTRL-EVENT-REGDOM-CHANGE init=CORE type=WORLD
Jan 23 11:36:29 iiab-test-cd wpa_supplicant[532]: wlan0: WPS-PBC-ACTIVE
[...]

Issue #2638, opened 2020-11-16
What USB 3.0 stick/drive patterns degrade a Raspberry Pi's 2.4GHz WiFi?


holta:
Thanks to @jvonau and others who've uncovered some real problems here!  Let's try to identify a couple general guidelines for schools/clinics using USB 3.0 flash drives (etc!) with Internet-in-a-Box, so they can avoid accidentally killing their Raspberry Pi's internal WiFi hotspot.

A big Thank You to _everybody_ who can report their experiences here:

- Can USB 2.0 flash drives safely be used in USB 3.0 slots?
- Can USB 3.0 flash drives safely be used in USB 2.0 slots?
- What fails when a USB3 stick is put in a USB3 slot ?
  - Presumably WiFi is affected much more when data is being retrieved from the USB3 stick &/or drive?
  - Is it possible WiFi isn't affected at all when data is not being read from [or written to] the USB 3.0 stick &/or drive?
  - Is the "bottom" USB 3.0 port more susceptible, on Raspberry Pi 4's?  (FWIW many RPi "hats" seem to advertise use of this bottom port in their demonstration/marketing photos.)
- What _other_ tests should we run to better understand this?

Refs:

- #2497
- #2632
- raspberrypi/firmware#1430
- http://FAQ.IIAB.IO > "44 Can I install IIAB onto a USB drive?"

holta:
Does https://pibox.in/2020/08/29/usb-c-hubs-and-wifi-issues-fixed/ say it all?  Thanks to @jvonau for uncovering this!

Tin foil shielding of cable connectors[*] sounds like a JOKE — until you watch these videos.

_[*] and better cable connectors & EMI / RF interference ferrite chokes in general._

(What USB3 sticks &/or USB3 devices appear to be the highest risk?)

JamesM92:
Not sure if helpful, but I'm using a usb3 to sata m2 ssd to control my pi. I currently have it connected via hardwire ethernet.

It is able to broadcast its wifi. But anybody who tries to connect to it find that they cant authenticate their connection.

Also the pie is unable to connect to another wifi source. 

Would a diagnostics download be helpful? Or would somebody need to remote into it to be useful?

holta:
@JamesM92 can you send the URL resulting from running `sudo iiab-diagnostics` for starters?  Thanks !

jvonau:
[...]

PR #3965, opened 2025-03-12
Fix very flaky IIAB installs with: (1) deferred raspi-config do_wifi_country (2) refined iiab-clone-wifi.service.j2 to add ap0 to correct phyname {phy0, phy1, ...} (3) better WiFi firmware crash logging [e.g. to help when "official" Raspberry Pi OS WiFi firmware cyfmac43455-sdio-standard.bin OFTEN crashes]

holta:
### Fixes bug:

Many IIAB installs are currently failing (most IIAB installs on "RasPiOS with desktop" and very possibly others) as mentioned in:

- #3961

- Instead of assuming phy0 when using `iw phy` to add interface ap0, the correct phyname (very often it's phy1 rather than phy0) is determined dynamically using `ls /sys/class/ieee80211/ | head -1` within [/etc/systemd/system/iiab-clone-wifi.service](https://github.com/iiab/iiab/blob/master/roles/network/templates/hostapd/iiab-clone-wifi.service.j2)

- Instead of running `raspi-config nonint do_wifi_country XX` prior to NM-debian.yml restarting of NetworkManager (problem accidentally introduced [2024-12-18](https://github.com/iiab/iiab/commit/6a6d153034e90d05d817d5424873cf8035318b68)), that order is now reversed.

Both above are well-tested to solve the problem (many IIAB installs are currently failing).

- This PR also adds WiFi firmware crash monitoring, to better prepare everyone for the future — via roles/firmware/tasks/warn-fw-crash.yml, that's run near the bottom of roles/network/tasks/main.yml

- Inline comments and Ansible messaging are substantially cleaned up and clarified.  These can become increasingly and substantially more meaningful in 2025.  As mentioned on #3961.

LATER: Should `sudo iiab` [rerun roles/network](https://github.com/iiab/iiab/blob/e4a13ce7aec79299b1de7bdead541f8af3a6b6a7/iiab-stages.yml#L63-L65) when it fails to complete, thereby respecting and honoring [SKIP_ROLE_ON_ERROR=false](https://github.com/iiab/iiab/blob/423e21c52ca87aa780135d04a29eb627c49ea58d/iiab-install#L18) as with all other roles invoked by ./iiab-install ?  Probably.  But this PR chooses to defer that decision until a later date.

64-bit Raspberry Pi OS with desktop, on Raspberry Pi 4 ,5 and 500.

@jvonau

### Related:

- RPi-Distro/firmware-nonfree#43
- raspberrypi/firmware#1768
- raspberrypi/linux#6049

### Tangentially related:

- iiab/iiab#3912


holta:
@EMG70 would you have time to test this PR on a Raspberry Pi 4 or 5 or 500 especially — or on any other RPi?

1. If so, please start with 64-bit "Raspberry Pi OS with desktop" and run the following ETHERNET-based IIAB install:

   ```
   curl iiab.io/install.txt | bash -s 3965
   ```

2. Zero-sized unit-tests are fine (i.e. very fast unit tests!) if you set these two vars right near the top of /etc/iiab/local_vars.yml :

   ```
   network_install: True
   network_enabled: True
   ```

3. Testing "Raspberry Pi OS Lite" can't hurt too — if you happen to have time for that later too?

jvonau:
Seems like a change in kernel behavior for some reason
`ls -la /sys/class/ieee80211/`
> total 0
drwxr-xr-x  2 root root 0 Mar 12 03:16 .
drwxr-xr-x 66 root root 0 Dec 31  1969 ..
lrwxrwxrwx  1 root root 0 Mar  6 05:56 phy0 -> ../../devices/platform/soc/fe300000.mmcnr/mmc_host/mmc1/mmc1:0001/mmc1:0001:1/ieee80211/phy0

Mind posting the same results?

holta:
For @jvonau:

```
root@box:~# ls -la /sys/class/ieee80211/
total 0
drwxr-xr-x  2 root root 0 Mar 12 01:19 .
drwxr-xr-x 72 root root 0 Mar 12 01:19 ..
lrwxrwxrwx  1 root root 0 Mar 12 01:25 phy1 -> ../../devices/platform/axi/1001100000.mmc/mmc_host/mmc1/mmc1:0001/mmc1:0001:1/ieee80211/phy1
```

iiab-diagnostics: https://paste.centos.org/view/a75818b7

jvonau:
Wondering out loud if the firmware is crashing and gets reloaded. Trying to figure out how a second phy becomes visible to the kernel as that is a hardware path that the soft link points to.

jvonau:
Something is fishy
> COMMAND: /usr/bin/sudo dmesg | grep Firmware:    # (Wi-Fi) firmware boot diagnostics
 [    2.836993] brcmfmac: brcmf_c_preinit_dcmds: Firmware: BCM4345/6 wl0: Aug 29 2023 01:47:08 version 7.45.265 (28bca26 CY) FWID 01-b677b91b
[  384.310328] brcmfmac: brcmf_c_preinit_dcmds: Firmware: BCM4345/6 wl0: Aug 29 2023 01:47:08 version 7.45.265 (28bca26 CY) FWID 01-b677b91b

holta:
> Wondering out loud if the firmware is crashing

Indeed, it looks like "official" Raspberry Pi OS WiFi firmware cyfmac43455-sdio-standard.bin is (extremely) unreliable:

```
root@box:~# dmesg | grep -i firmware
[    0.000000] psci: PSCIv1.1 detected in firmware.
[    0.016551] raspberrypi-firmware soc:firmware: Attached to firmware from 2025-02-12T10:51:52, variant start_cd
[    0.020555] raspberrypi-firmware soc:firmware: Firmware hash is f788aab600000000000000000000000000000000
[    2.292285] rp1-firmware rp1_firmware: RP1 Firmware version eb39cfd516f8c90628aa9d91f52370aade5d0a55
[    2.836993] brcmfmac: brcmf_c_preinit_dcmds: Firmware: BCM4345/6 wl0: Aug 29 2023 01:47:08 version 7.45.265 (28bca26 CY) FWID 01-b677b91b
[  381.256514] ieee80211 phy0: brcmf_fw_crashed: Firmware has halted or crashed
[  383.868473] brcmfmac: brcmf_sdio_checkdied: firmware trap in dongle
[  384.310328] brcmfmac: brcmf_c_preinit_dcmds: Firmware: BCM4345/6 wl0: Aug 29 2023 01:47:08 version 7.45.265 (28bca26 CY) FWID 01-b677b91b

root@box:~# ls -l /etc/alternatives/cyfmac43455-sdio.bin
lrwxrwxrwx 1 root root 51 Aug 10  2024 /etc/alternatives/cyfmac43455-sdio.bin -> /lib/firmware/cypress/cyfmac43455-sdio-standard.bin
```

jvonau:
>[    2.557143] brcmfmac: F1 signature read @0x18000000=0x15264345
[    2.559698] brcmfmac: brcmf_fw_alloc_request: using brcm/brcmfmac43455-sdio for chip BCM4345/6
[    2.560817] usbcore: registered new interface driver brcmfmac
[    2.836762] brcmfmac: brcmf_c_process_txcap_blob: no txcap_blob available (err=-2)
[    2.836993] brcmfmac: brcmf_c_preinit_dcmds: Firmware: BCM4345/6 wl0: Aug 29 2023 01:47:08 version 7.45.265 (28bca26 CY) FWID 01-b677b91b

And we have a winner

>[  376.044605] brcmfmac: brcmf_cfg80211_set_power_mgmt: power save enabled
[  380.655848] brcmfmac: brcmf_set_channel: set chanspec 0xd022 fail, reason -52
[  380.763936] brcmfmac: brcmf_set_channel: set chanspec 0xd026 fail, reason -52
[  380.867877] brcmfmac: brcmf_set_channel: set chanspec 0xd02a fail, reason -52
[  380.975743] brcmfmac: brcmf_set_channel: set chanspec 0xd02e fail, reason -52
[  381.256514] ieee80211 phy0: brcmf_fw_crashed: Firmware has halted or crashed
[  383.868414] brcmfmac: brcmf_sdio_bus_rxctl: resumed on timeout
[  383.868473] brcmfmac: brcmf_sdio_checkdied: firmware trap in dongle
[  383.868500] brcmfmac: dongle trap info: type 0x4 @ epc 0x001bfc3a
[  383.868507] brcmfmac: brcmf_dump_obss: dump_obss error (-110)
[  383.869378] ieee80211 phy0: brcmf_fil_cmd_data: bus is down. we have nothing to do.
[  383.869383] ieee80211 phy0: brcmf_cfg80211_get_tx_power: error (-5)
[  383.869564] ieee80211 phy0: brcmf_fil_cmd_data: bus is down. we have nothing to do.
[  383.869566] ieee80211 phy0: brcmf_cfg80211_get_tx_power: error (-5)
[  383.869621] ieee80211 phy0: brcmf_fil_cmd_data: bus is down. we have nothing to do.
[  383.869624] ieee80211 phy0: brcmf_cfg80211_get_tx_power: error (-5)
[  384.165234] brcmfmac: F1 signature read @0x18000000=0x15264345
[  384.166392] brcmfmac: brcmf_fw_alloc_request: using brcm/brcmfmac43455-sdio for chip BCM4345/6
[  384.310074] brcmfmac: brcmf_c_process_txcap_blob: no txcap_blob available (err=-2)
[  384.310328] brcmfmac: brcmf_c_preinit_dcmds: Firmware: BCM4345/6 wl0: Aug 29 2023 01:47:08 version 7.45.265 (28bca26 CY) FWID 01-b677b91b
[  384.368062] brcmfmac: brcmf_cfg80211_set_power_mgmt: power save enabled

The above show the crash I'll assume long before the start of the installation. 

> [ 7052.873431] ieee80211 phy1: brcmf_cfg80211_get_channel: chanspec failed (-52)

The above is from having the WiFi on channel 165 when hostapd starts.

>=IIAB==========================================================================
COMMAND: /usr/sbin/iw dev    # List wireless interfaces
 phy#1
        Interface ap0
                ifindex 8
                wdev 0x100000004
                addr 02:b9:6d:d3:7a:95
                type AP
                channel 165 (5825 MHz), width: 20 MHz, center1: 5825 MHz
                txpower 31.00 dBm
        Unnamed/non-netdev interface
                wdev 0x100000002
                addr 2e:cf:67:2b:6d:33
                type P2P-device
                txpower 31.00 dBm
        Interface wlan0
                ifindex 5
                wdev 0x100000001
                addr 2c:cf:67:2b:6d:33
                type managed
                channel 165 (5825 MHz), width: 20 MHz, center1: 5825 MHz
                txpower 31.00 dBm
 

Nice try but nope, the dynamic detection should not be needed as the device should be static when the firmware doesn't crash.

jvonau:
And you bet the farm on this platform. SMH

holta:
Just a reminder that everyone's in great shape by continuing to ignore Raspberry Pi's "official" WiFi firmware after initial IIAB install.

(IIAB's [roles/firmware](https://github.com/iiab/iiab/tree/master/roles/firmware) not only expands the number of "student" WiFi connections; but also makes WiFi firmware significantly more reliable.)

jvonau:
Since this issue is during installation and firmware related here is how I would solve this issue to allow the install to continue to iiab-admin-console:  

- remove the network role from https://github.com/iiab/iiab/blob/master/iiab-stages.yml
- add `iiab-network` to be just before the photograph this message in `iiab`
- I'd suggest for 'ICO' to make the network role last for the same reason avoid disruptions during installation.

The keys are yours have at it.

jvonau:
@EMG70 Please boot a new vanilla image and before the curl routine to install IIAB run:
`/usr/bin/sudo dmesg | grep Firmware:` 
then post the output or a picture

When IIAB is not even in the picture and there are 2 lines present take this issue upstream. Have to file issues or nothing will get fixed.

holta:
> @EMG70 Please boot a new vanilla image and before the curl routine to install IIAB run:
> `/usr/bin/sudo dmesg | grep Firmware:`

@jvonau @EMG70 the answer here is known to be: (reconfirmed many times over the past week, i.e. with the 43455 WiFi chip on RPi 4, 5 and 500)

```
root@raspberrypi:~# dmesg | grep Firmware:
[    3.190782] brcmfmac: brcmf_c_preinit_dcmds: Firmware: BCM4345/6 wl0: Aug 29 2023 01:47:08 version 7.45.265 (28bca26 CY) FWID 01-b677b91b
```

jvonau:
> @jvonau @EMG70 the answer here is known to be: (reconfirmed many times over the past week, i.e. with the 43455 WiFi chip on RPi 4, 5 and 500)
> 
> ```
> root@raspberrypi:~# dmesg | grep Firmware:
> [    3.190782] brcmfmac: brcmf_c_preinit_dcmds: Firmware: BCM4345/6 wl0: Aug 29 2023 01:47:08 version 7.45.265 (28bca26 CY) FWID 01-b677b91b
> ```

Missing the point **again**, if there are 2 lines the firmware crashed.

holta:
Stop berating.

I'm very well aware of the issue.

I can do more tests in the coming week, but my rough sense (so far) is that "official" WiFi firmware cyfmac43455-sdio-standard.bin crashes most often on RPi 500 — whereas tests so far indicate it crashes on RPi 4 and 5 somewhat less often.

holta:
FYI here's just one way that _often but not always works_ to reproduce the crashing of the "official" WiFi firmware cyfmac43455-sdio-standard.bin on a freshly installed (vanilla, untouched except for apt updates) 64-bit "RasPiOS  with desktop" on Raspberry Pi 5:

```
nmcli radio wifi on
nmcli radio wifi off
```

jvonau:
File the issue upstream, NM is causing the crash of the firmware

jvonau:
Who is berating? Just sound forensic digging and engineering advice. 
Playing with `nmcli radio` needs more background information:
- Was this before unlocking the WiFi with the OS provided methods?
- Was there or not an active WiFi connection once unlocked?
- After providing Regulatory Domain Code information and rebooting can you get the firmware to crash when fiddling around with `nmcli radio`?

You have not proven there is anything really wrong with the firmware per say. Could be the way then desktop uses NM to drive the firmware.

> Mar 05 11:15:07 raspberrypi NetworkManager[603]: <info>  [1741191307.5069] device (wlan0): driver supports Access Point (AP) mode
Mar 05 11:15:07 raspberrypi NetworkManager[603]: <info>  [1741191307.5093] manager: (wlan0): new 802.11 Wi-Fi device (/org/freedesktop/NetworkManager/Devices/3)
Mar 05 11:15:07 raspberrypi NetworkManager[603]: <info>  [1741191307.5101] device (wlan0): state change: unmanaged -> unavailable (reason 'managed', sys-iface-state: 'external')
Mar 05 11:15:07 raspberrypi NetworkManager[603]: <info>  [1741191307.5137] failed to open /run/network/ifstate

The above would be checking for if systemd-networkd is present.
> Mar 05 11:15:19 raspberrypi NetworkManager[603]: <info>  [1741191319.4757] agent-manager: agent[01bfe9ef8284c102,:1.34/org.freedesktop.nm-applet/1000]: agent registered

The above would be the desktop agent used to populate the picklist.
> [  359.977435] warning: `cat' uses wireless extensions which will stop working for Wi-Fi 7 hardware; use nl80211
[  376.044605] brcmfmac: brcmf_cfg80211_set_power_mgmt: power save enabled
[  380.655848] brcmfmac: brcmf_set_channel: set chanspec 0xd022 fail, reason -52
[  380.763936] brcmfmac: brcmf_set_channel: set chanspec 0xd026 fail, reason -52
[  380.867877] brcmfmac: brcmf_set_channel: set chanspec 0xd02a fail, reason -52
[  380.975743] brcmfmac: brcmf_set_channel: set chanspec 0xd02e fail, reason -52
[  381.256514] ieee80211 phy0: brcmf_fw_crashed: Firmware has halted or crashed
[  383.868414] brcmfmac: brcmf_sdio_bus_rxctl: resumed on timeout

Without out having the Regulatory Domain Code there would no 5GHz coverage, I think the above is related to either NM scanning the WiFi spectrum or the desktop asking for NM to do that to put together the picklist. Why either would be scanning with rfkill in place is a question for upstream. 

If NM is doing the scanning the same behavior should be present on non-desktop stage2 images otherwise the behavior is related to the desktop implementation.

> Mar 05 11:15:06 raspberrypi systemd[1]: Starting wpa_supplicant.service - WPA supplicant...
Mar 05 11:15:06 raspberrypi wpa_supplicant[611]: Successfully initialized wpa_supplicant
Mar 05 11:15:06 raspberrypi systemd[1]: Started wpa_supplicant.service - WPA supplicant.
Mar 05 13:22:00 box wpa_supplicant[611]: nl80211: kernel reports: Registration to specific type not supported
Mar 05 13:22:15 box wpa_supplicant[611]: p2p-dev-wlan0: CTRL-EVENT-DSCP-POLICY clear_all
Mar 05 13:22:15 box wpa_supplicant[611]: p2p-dev-wlan0: CTRL-EVENT-DSCP-POLICY clear_all
Mar 05 13:22:15 box wpa_supplicant[611]: nl80211: deinit ifname=p2p-dev-wlan0 disabled_11b_rates=0
Mar 05 13:22:15 box wpa_supplicant[611]: Could not read interface p2p-dev-wlan0 flags: No such device
Mar 05 13:22:15 box wpa_supplicant[611]: wlan0: CTRL-EVENT-DSCP-POLICY clear_all
Mar 05 13:22:15 box wpa_supplicant[611]: wlan0: CTRL-EVENT-DSCP-POLICY clear_all
Mar 05 13:22:15 box wpa_supplicant[611]: nl80211: deinit ifname=wlan0 disabled_11b_rates=0
Mar 05 13:22:15 box wpa_supplicant[611]: nl80211: kernel reports: Registration to specific type not supported

Back over to you.

holta:
- FWIW the very same failure occurs on the latest (apt updated) 64-bit Raspberry Pi OS Lite.  Maybe be a good time to stop berating Raspberry Pi OS's mainline OS (their desktop version) after about a decade, sticking to actual facts instead going forward heh (-:

- Running `nmcli radio wifi on` then `nmcli radio wifi off` (on an apt updated but otherwise untouched OS's) are commands that tend to (sometimes but not always!) trigger the failure.

- I suspect (perhaps wrongly) that we likely may be able to find other commands that similarly cause WiFi firmware cyfmac43455-sdio-standard.bin to crash soon thereafter.

- Whether or not a WiFi Country Code has already been set or not set (a.k.a. Regulatory Domain Code) _would appear to be_ irrelevant.  In other words, the failure of WiFi firmware cyfmac43455-sdio-standard.bin (would appear to?) occur regardless.

- Hopefully one of us will spot something in the coming week or so — to help inform others (and upstream) illuminating more what on earth is going on here.

- A Big Thank You to @EMG70 for successfully testing this PR earlier today!  I'll probably merge this PR in the coming 24h to help everyone's RPi installs become much more reliable.  We can always improve on it later, if really necessary.

tim-moody:
>sticking to actual facts instead going forward heh (-:

I do lots of lite installs on rpi 5 and have never experienced this problem. What exactly is the evidence to look for?

holta:
> What exactly is the evidence to look for?

As stated on #3961 and above, start with an untouched (apt updated, rebooted) 64-bit RasPiOS connected by Ethernet.  For now we should probably stick to RPi 4, 5 and 500 as these contain the 43455 WiFi chip:

1) Then, here are a couple different ways to watch live... for WiFi firmware crashes:

   - `watch 'iw reg get; iw dev'`    # Look for phy#0 becoming phy#1 becoming phy#2 (and onwards, each time WiFi firmware fails!)
   - `watch 'dmesg | grep Firmware:'`    # Look for the list growing, each time WiFi firmware fails...

[ Or combine both above, running every 0.1 seconds: `watch -n.1 'iw reg get; iw dev; echo; dmesg | grep Firmware:'` ]

2) To accelerate the WiFi firmware failures (crashes), try to run things like `nmcli radio wifi on` and then `nmcli radio wifi off` (i.e. in another window, or in another shell).

3) Hopefully we can find other commands too, that hint at cause-and-effect patterns?  e.g. run `curl iiab.io/install.txt | bash` with local_vars.yml option ZERO (unit test) and then set `network_install: True` with `network_enabled: True` in /etc/iiab/local_vars.yml before running `sudo iiab` — this is a much slower way to reproduce it (e.g. ~10 minutes instead of ~10 seconds) but seems to fail almost every time (prior to applying this PR) _with certain machines_ FWIW.

holta:
> start with an untouched (apt updated, rebooted) 64-bit RasPiOS connected by Ethernet. For now we should probably stick to RPi 4, 5 and 500 as these contain the 43455 WiFi chip

CLARIF: WiFi firmware also fails if RasPiOS (from 2024-11-19) was not apt updated.

holta:
> Whether or not a WiFi Country Code has already been set or not set (a.k.a. Regulatory Domain Code) would appear to be irrelevant. In other words, the failure of WiFi firmware cyfmac43455-sdio-standard.bin (would appear to?) occur regardless.

JUST A THEORY... but FWIW here's something that I'm trying to test:

- WiFi firmware cyfmac43455-sdio-standard.bin still crashes intermittently after `raspi-config nonint do_wifi_country US && reboot` has been run... _but it seems to crash a lot less often??_

holta:
> * WiFi firmware cyfmac43455-sdio-standard.bin still crashes intermittently after `raspi-config nonint do_wifi_country US && reboot` has been run... _but it seems to crash a lot less often??_

An example with diagnostics from 64-bit RasPiOS Lite (2014-11-19, _without_ apt updates) posted below for the record.  OS was completely untouched except that I ran `raspi-config nonint do_wifi_country US && reboot` — and then I ran `nmcli radio wifi on` and `nmcli radio wifi off` intermittently to try to trigger the WiFi firmware failure — in this case WiFi firmware cyfmac43455-sdio-standard.bin seems to have crashed many seconds after I ran one of those `nmcli radio wifi on` commands:

```
root@raspberrypi:~# ls /sys/class/ieee80211/
phy1

root@raspberrypi:~# raspi-config nonint get_wifi_country
US

root@raspberrypi:~# tail -1 /proc/cpuinfo
Model           : Raspberry Pi 500 Rev 1.0

root@raspberrypi:~# cat /etc/rpi-issue
Raspberry Pi reference 2024-11-19
Generated using pi-gen, https://github.com/RPi-Distro/pi-gen, 891df1e21ed2b6099a2e6a13e26c91dea44b34d4, stage2

root@raspberrypi:~# dmesg | grep -B20 -A5 Firmware:
...
[    5.898219] brcmfmac: brcmf_c_preinit_dcmds: Firmware: BCM4345/6 wl0: Aug 29 2023 01:47:08 version 7.45.265 (28bca26 CY) FWID 01-b677b91b
...
[  879.923746] brcmfmac: brcmf_cfg80211_set_power_mgmt: power save enabled
[  894.571700] brcmfmac: brcmf_dump_obss: dump_obss error (-52)
[  910.923793] brcmfmac: brcmf_cfg80211_set_power_mgmt: power save enabled
[  952.458764] brcmfmac: brcmf_cfg80211_set_power_mgmt: power save enabled
[  985.444215] brcmfmac: brcmf_dump_obss: dump_obss error (-52)
[  987.255604] brcmfmac: brcmf_cfg80211_set_power_mgmt: power save enabled
[  995.463638] ieee80211 phy0: brcmf_fw_crashed: Firmware has halted or crashed
[  997.919470] brcmfmac: brcmf_sdio_bus_rxctl: resumed on timeout
[  997.919527] brcmfmac: brcmf_sdio_checkdied: firmware trap in dongle
[  997.919560] brcmfmac: dongle trap info: type 0x4 @ epc 0x001bfc3a
                 cpsr 0x2000019f spsr 0x200001bf sp 0x0025fe50
                 lr   0x00049bd7 pc   0x001bfc3a offset 0x25fdf8
                 r0   0x00256828 r1   0x00000001 r2 0x00010000 r3 0x18001000
                 r4   0x00256828 r5   0xffffffff r6 0x00000000 r7 0x00000001
[  997.919570] brcmfmac: brcmf_dump_obss: dump_obss error (-110)
[  997.999867] mmc1: card 0001 removed
[  998.219876] mmc1: new ultra high speed DDR50 SDIO card at address 0001
[  998.220169] brcmfmac: F1 signature read @0x18000000=0x15264345
[  998.221380] brcmfmac: brcmf_fw_alloc_request: using brcm/brcmfmac43455-sdio for chip BCM4345/6
[  998.370639] brcmfmac: brcmf_c_process_txcap_blob: no txcap_blob available (err=-2)
[  998.370874] brcmfmac: brcmf_c_preinit_dcmds: Firmware: BCM4345/6 wl0: Aug 29 2023 01:47:08 version 7.45.265 (28bca26 CY) FWID 01-b677b91b
[ 1004.968999] brcmfmac: brcmf_cfg80211_set_power_mgmt: power save enabled
[ 1095.148935] brcmfmac: brcmf_cfg80211_set_power_mgmt: power save enabled
[ 1098.435621] brcmfmac: brcmf_dump_obss: dump_obss error (-52)
[ 1288.601686] brcmfmac: brcmf_cfg80211_set_power_mgmt: power save enabled
[ 1291.905207] brcmfmac: brcmf_dump_obss: dump_obss error (-52)
```

jvonau:
> * FWIW the very same failure occurs on the latest (apt updated) 64-bit Raspberry Pi OS Lite.  Maybe be a good time to stop berating Raspberry Pi OS's mainline OS (their desktop version) after about a decade, sticking to actual facts instead going forward heh (-:
>

The entire desktop environment changed in the last year, so the counter starts at zero confidence. When I see the Lite dump I'll be more inclined to believe you until then it's just hearsay as Tim has done extensive testing and documenting the results. Perhaps what you are seeing is related to the RF environment around you. What you are suggesting points to NM and/or the way the images are prepared.
 >     * Running `nmcli radio wifi on && nmcli radio wifi off` (on an apt updated but otherwise untouched OS's) is one command that tends to (sometimes but not always!) trigger the failure.
> 

Was the Regulatory Domain Code set after `wifi on`? Might not be the only command needed to diagnose, should be checking `rfkill list` `iw reg get` `dmesg` before and after toggling those setting and checking which NM files are altered.
>     * I suspect (perhaps wrongly) that we likely may be able to find other commands that similarly cause WiFi firmware cyfmac43455-sdio-standard.bin to crash soon thereafter.
> 

Perhaps usb booted external storage is make the situation worse?  https://github.com/iiab/iiab/issues/2638
>     * Whether or not a WiFi Country Code has already been set or not set (a.k.a. Regulatory Domain Code) _would appear to be_ irrelevant.  In other words, the failure of WiFi firmware cyfmac43455-sdio-standard.bin (would appear to?) occur regardless.
> 

Perhaps the workflow needs refinement, apt upgrade and set the Regulatory Domain Code then reboot would make the most sense? Think that is Tim's workflow.
>     * Hopefully one of us will spot something in the coming week or so — to help inform others (and upstream) illuminating more what on earth is going on here.
> 
>     * A Big Thank You to @EMG70 for successfully testing this PR earlier today!  I'll probably merge this PR in the coming 24h to help everyone's RPi installs become much more reliable.  We can always improve on it later, if really necessary.

Much like putting coverup over a pimple the underling issue is still there, just defers the problem. The hotspot will shutdown and/or fail to start as soon as the firmware locks up again. My suggestion above would be about the same, pushing the problem to a different failure point during the install.

jvonau:
AI Overview
[Learn more](https://support.google.com/websearch?p=ai_overviews&hl=en-CA)
The error brcmf_dump_obss: dump_obss error (-52) indicates a problem during the process of dumping Open BSS (Basic Service Set) statistics by the Broadcom Wi-Fi firmware, likely due to a hardware clock issue or firmware trap. 
Here's a more detailed explanation: 

   What it means:

   The error message brcmf_dump_obss: dump_obss error (-52) suggests that the 
host (your device) is trying to call the dump_obss iovar 
(a firmware command) to get Open BSS statistics, but the firmware is failing to execute this command, resulting in an error code of -52.
    Root Cause:
    The root cause is likely related to the hardware clock within the Wi-Fi dongle (or chip) being off or failing to function correctly during the dump_obss process. This can lead to firmware traps or crashes.
    Solution:
    The suggested solution involves ensuring that the hardware clock in the dongle is active when the dump_obss command is executed. A patch was implemented in the Linux kernel to disable the MPC (Multi-Processing Capability) during dump_obss and re-enable it afterwards to address this issue.
    Context:
    This error is often seen in the context of Broadcom Wi-Fi drivers (brcmfmac) and can occur when using devices like Raspberry Pi 5 with specific firmware images.


https://forums.raspberrypi.com/viewtopic.php?t=368792
https://github.com/RPi-Distro/firmware-nonfree/issues/43
https://forum.armbian.com/topic/39348-no-wifi-on-rpi5/
https://www.reddit.com/r/AsahiLinux/comments/1f8p2i9/wifi_suddenly_broke/

Perhaps turning roaming and/or powersave off might help. 
Happy Testing.

holta:
Similar to https://github.com/iiab/iiab/pull/3965#issuecomment-2719573931 here's another 64-bit RasPiOS Lite data point in case it helps — this WiFi firmware cyfmac43455-sdio-standard.bin crash occurred _with apt updates applied_, seemingly as a direct result of my running `nmcli radio wifi off` — _on an RPi 4 here:_

```
root@raspberrypi:~# ls /sys/class/ieee80211/
phy1

root@raspberrypi:~# raspi-config nonint get_wifi_country
US

root@raspberrypi:~# tail -1 /proc/cpuinfo
Model           : Raspberry Pi 4 Model B Rev 1.2

root@raspberrypi:~# cat /etc/rpi-issue
Raspberry Pi reference 2024-11-19
Generated using pi-gen, https://github.com/RPi-Distro/pi-gen, 891df1e21ed2b6099a2e6a13e26c91dea44b34d4, stage2

root@raspberrypi:~# dmesg | grep -B20 -A5 Firmware:
...
[    4.406691] brcmfmac: brcmf_c_preinit_dcmds: Firmware: BCM4345/6 wl0: Aug 29 2023 01:47:08 version 7.45.265 (28bca26 CY) FWID 01-b677b91b
...
[ 8855.939288] brcmfmac: brcmf_cfg80211_set_power_mgmt: power save enabled
[ 8872.855414] brcmfmac: brcmf_cfg80211_set_power_mgmt: power save enabled
[ 8875.717475] brcmfmac: brcmf_dump_obss: dump_obss error (-52)
[ 9322.397927] brcmfmac: brcmf_cfg80211_set_power_mgmt: power save enabled
[12351.828968] brcmfmac: brcmf_cfg80211_set_power_mgmt: power save enabled
[12371.254708] brcmfmac: brcmf_cfg80211_set_power_mgmt: power save enabled
[12387.448882] ieee80211 phy0: brcmf_fw_crashed: Firmware has halted or crashed
[12390.067471] brcmfmac: brcmf_sdio_bus_rxctl: resumed on timeout
[12390.067774] brcmfmac: brcmf_sdio_checkdied: firmware trap in dongle
[12390.067936] brcmfmac: dongle trap info: type 0x4 @ epc 0x001bfc3a
                 cpsr 0x2000019f spsr 0x200001bf sp 0x0025fe50
                 lr   0x00049bd7 pc   0x001bfc3a offset 0x25fdf8
                 r0   0x002567bc r1   0x00000001 r2 0x00010000 r3 0x18001000
                 r4   0x002567bc r5   0xffffffff r6 0x00000000 r7 0x00000001
[12390.067968] brcmfmac: brcmf_dump_obss: dump_obss error (-110)
[12390.156644] mmc1: card 0001 removed
[12390.271894] mmc1: new high speed SDIO card at address 0001
[12390.273345] brcmfmac: F1 signature read @0x18000000=0x15264345
[12390.277539] brcmfmac: brcmf_fw_alloc_request: using brcm/brcmfmac43455-sdio for chip BCM4345/6
[12390.473964] brcmfmac: brcmf_c_process_txcap_blob: no txcap_blob available (err=-2)
[12390.474449] brcmfmac: brcmf_c_preinit_dcmds: Firmware: BCM4345/6 wl0: Aug 29 2023 01:47:08 version 7.45.265 (28bca26 CY) FWID 01-b677b91b
```

holta:
Two quick clarifications:

> Perhaps usb booted external storage is make the situation worse? #2638

All tests (that I know of, over the past week) have used internal microSD cards.  So thankfully a USB3-booted boot drive (or any other USB3 mounted device) is not in question here.

> Perhaps what you are seeing is related to the RF environment around you.

Tests were done in 3 completely different locations over the past week.  Equivalent failures occurred in all 3 locations.  Which is just to say that every one of these 3 locations had a handful of nearby SSIDs visible, as is normal in any urban environment.

jvonau:
Being able to induce the failures on 'Lite' while diking around with commands that would normally never be run in ordinary circumstances just points to the desktop or firmware. File the issues upstream.

holta:
> RPi-Distro/firmware-nonfree#43

Thanks: I posted there ("upstream") on the hopes that might help us dig deeper and/or in a better direction?

jvonau:
HDMI interference could be messing with the WiFi, think there are references in the issues. Not wasting much more of my time, have grandkids I'd rather enjoy.

jvonau:
also related:
https://github.com/RPi-Distro/firmware-nonfree/issues/49
https://github.com/home-assistant/operating-system/issues/3367
https://github.com/raspberrypi/linux/issues/6049

jvonau:
> Two quick clarifications:
> 
> > Perhaps usb booted external storage is make the situation worse? #2638
> 
> All tests (that I know of, over the past week) have used internal microSD cards. So thankfully a USB3-booted boot drive (or any other USB3 mounted device) is not in question here.
>

Guess I should of not even mentioned usbboot, just the use of usb3 slots with any usb based storage solution could trigger 2638. 
> > Perhaps what you are seeing is related to the RF environment around you.
> 
> Tests were done in 3 completely different locations over the past week. Equivalent failures occurred in all 3 locations. Which is just to say that every one of these 3 locations had a handful of nearby SSIDs visible, as is normal in any urban environment.

As pointed out HDMI cables can contribute to RF interference 
https://forums.raspberrypi.com/viewtopic.php?f=28&t=254640&start=25

holta:
I'd like better logging of WiFi firmware crashes:

- To help everyone better document very flaky WiFi firmwares crashing (arising from upstream of upstream ;)

- So FYI I'll instrument some very basic logging (in this PR) and likely merge it very shortly, so that IIAB installs[*] finally get back on track.  (If necessary, this WiFi firmware crash logging can evolve later with another PR, if crash context needs to improve there in one direction or another!)

[*] Very unreliable at the moment — something like 80% of IIAB installs fail on Raspberry Pi OS with desktop on RPi 500, across very intensive testing in 4 locations over recent weeks.  Four completely different urban neighborhoods were tested just in case nearby WiFi congestion might be related to WIFi firmware crashing so often (neighborhood location doesn't seem to make much difference).  Finally, RPi 4 and RPi 5 don't seem much better, appearing to fail in the exact same way as RPi 500 (but I haven't tested those as much, to have as broad a statistical sample).

holta:
For starters, logging things like `dmesg | grep -B10 -A25 'Firmware has halted or crashed'` (and circulating such to a wider demographic, via [iiab-diagnostics](https://github.com/iiab/iiab/blob/master/scripts/iiab-diagnostics.README.md) etc) is probably enough?

- As we know, two-to-three seconds later `dmesg` shows `firmware trap in dongle`
- RPi-Distro/firmware-nonfree#43
- Logging can definitely evolve later, if upstream (of upstream) hopefully gives indications as to what they need 🙏

Tangentially related WiFi firmware crashing issue on Raspberry Pi Zero 2 W might shine a light on how to proceed:

- raspberrypi/firmware#1768

tim-moody:
I have never been aware of this when installing on rpi 5 or 2 W. I assume this is because:

1. I typically install using ethernet
2. I use the stock default vars that installs and configures the legacy firmware that does not crash.

So anyone who does the same will likely not be affected by this problem nor will anyone who uses one of our prebuilt images?

holta:
> I assume this is because [...] I typically install using ethernet

No.  WiFi firmware crashes happen often during Ethernet installs of IIAB, as mentioned near the top of this ticket, and as shown in #3961 iiab-diagnostics.  Who knows why WiFi firmware **/lib/firmware/cypress/[cyfmac43455-sdio-standard.bin](https://github.com/RPi-Distro/firmware-nonfree/blob/bookworm/debian/config/brcm80211/cypress/cyfmac43455-sdio-standard.bin)** crashes cause 9 IIAB installs out of 10 to fail here during roles/network on some rough days, but better logging can't hurt (and might well help!)

> I use the stock default vars that installs and configures the legacy firmware that does not crash.

It's great news that WiFi firmwares installed by IIAB's [roles/firmware](https://github.com/iiab/iiab/tree/master/roles/firmware) so far appear much less affected (hopefully not affected!!) but just FYI that's an entirely separate question — as those 2021 WiFi firmwares (what I think you're calling legacy WiFi firmwares) are never installed if non-technical IIAB implementers never get that far in their IIAB install process (RECAP: roles/network does not complete, for non-technical implementers who get stuck here).

My primary goal is to help non-technical IIAB implementers find a much more reliable path — this means using Raspberry Pi's default OS (that's "with desktop" that newcomers almost always start with, to get their bearings, and because rpi-imager and most other install paths steer newbies in that direction, to help them learn Linux etc).  So if there's time, I can later try to investigate whether this 43455 WiFi chip firmware crash happens statistically less often with RasPiOS Lite (and if that's later fully confirmed under identical conditions, that would be could be very revealing, ETC, etc).

ASIDE: Just a reminder that these WiFi firmware crashes are about the 43455 chip (i.e. RPi 4, 5, 500) — so RPi Zero 2 W is not involved.

holta:
> Tangentially related WiFi firmware crashing issue on Raspberry Pi Zero 2 W might shine a light on how to proceed:
> 
> * [Wifi: Firmware crashing regularly on RPi 0W2 raspberrypi/firmware#1768](https://github.com/raspberrypi/firmware/issues/1768)

While Zero 2 W is in fact off-topic, the above thread suggests adding `brcmfmac.debug=0x100000` to /boot/firmware/cmdline.txt (and then reboot) so as to provide much more logging that might possibly help Synaptics (which apparently acquired Broadcom's wireless assets in July 2020).

So here's a more detailed example than the usual, still with the usual vanilla/bare OS (64-bit Raspberry Pi OS with desktop on RPi 500) that's been (1) apt updated, then (2) `raspi-config nonint do_wifi_country US` was run, then (3) `brcmfmac.debug=0x100000` was added to /boot/firmware/cmdline.txt, then (4) RPi 500 was rebooted, and finally (5) `systemctl restart NetworkManager` was run which triggered the following WiFi firmware crash:

```
root@raspberrypi:~# dmesg | grep -B60 -A40 'Firmware has halted or crashed'
[ 4504.688020] brcmfmac: CONSOLE: 001444.174 wl0: wl_open
[ 4504.784044] brcmfmac: CONSOLE: 001444.277 wl0: wl_open
[ 4504.896029] brcmfmac: CONSOLE: 001444.382 wl0: wl_open
[ 4504.988047] brcmfmac: CONSOLE: 001444.482 wl0: wl_open
[ 4505.104030] brcmfmac: CONSOLE: 001444.585 wl0: wl_open
[ 4505.196040] brcmfmac: CONSOLE: 001444.687 wl0: wl_open
[ 4505.308019] brcmfmac: CONSOLE: 001444.792 wl0: wl_open
[ 4505.400044] brcmfmac: CONSOLE: 001444.892 wl0: wl_open
[ 4505.520028] brcmfmac: CONSOLE: 001444.995 wl0: wl_open
[ 4505.604029] brcmfmac: CONSOLE: 001445.096 wl0: wl_open
[ 4505.712045] brcmfmac: CONSOLE: 001445.201 wl0: wl_open
[ 4505.812075] brcmfmac: CONSOLE: 001445.303 wl0: wl_open
[ 4505.920042] brcmfmac: CONSOLE: 001445.408 wl0: wl_open
[ 4506.024076] brcmfmac: CONSOLE: 001445.513 wl0: wl_open
[ 4506.128057] brcmfmac: CONSOLE: 001445.616 wl0: wl_open
[ 4506.236046] brcmfmac: CONSOLE: 001445.723 wl0: wl_open
[ 4506.340068] brcmfmac: CONSOLE: 001445.828 wl0: wl_open
[ 4506.444055] brcmfmac: CONSOLE: 001445.932 wl0: wl_open
[ 4506.552046] brcmfmac: CONSOLE: 001446.038 wl0: wl_open
[ 4506.656081] brcmfmac: CONSOLE: 001446.142 wl0: wl_open
[ 4506.756058] brcmfmac: CONSOLE: 001446.243 wl0: wl_open
[ 4506.864049] brcmfmac: CONSOLE: 001446.348 wl0: wl_open
[ 4506.984041] brcmfmac: CONSOLE: 001446.453 wl0: wl_open
[ 4510.031030] brcmfmac: CONSOLE: 001449.506 wl0: malformed chanspec 0x0
[ 4510.031036] brcmfmac: CONSOLE: 001449.506 wl0: malformed chanspec 0x0
[ 4510.044015] brcmfmac: CONSOLE: 001449.526 wl0: malformed chanspec 0x0
[ 4510.044017] brcmfmac: CONSOLE: 001449.526 wl0: malformed chanspec 0x0
[ 4510.360062] brcmfmac: CONSOLE: 001449.831 wl0: wl_open
[ 4510.464069] brcmfmac: CONSOLE: 001449.935 wl0: wl_open
[ 4510.568031] brcmfmac: CONSOLE: 001450.038 wl0: wl_open
[ 4510.650712] brcmfmac: CONSOLE: 001450.126 AXI timeout; APB Bridge 0
[ 4510.650718] brcmfmac: CONSOLE: 001450.126     errlog: lo 0x18001160, hi 0x00000000, id 0x00030003, flags 0x01111200, status 0x00000002
[ 4510.650720] brcmfmac: CONSOLE: 001450.126
[ 4510.650722] brcmfmac: CONSOLE: FWID 01-b677b91b
[ 4510.650724] brcmfmac: CONSOLE: flags 1
[ 4510.650725] brcmfmac: CONSOLE: 001450.126
[ 4510.650727] brcmfmac: CONSOLE: TRAP 4(25fdf8): pc 1bfc3a, lr 49bd7, sp 25fe50, cpsr 2000019f, spsr 200001bf
[ 4510.650729] brcmfmac: CONSOLE: 001450.126   dfsr 1008, dfar 18001160
[ 4510.650731] brcmfmac: CONSOLE: 001450.126   r0 256828, r1 1, r2 10000, r3 18001000, r4 256828, r5 ffffffff, r6 0
[ 4510.650733] brcmfmac: CONSOLE: 001450.126   r7 1, r8 18002000, r9 18102000, r10 20010000, r11 1912e1, r12 0
[ 4510.650735] brcmfmac: CONSOLE: 001450.126
[ 4510.650737] brcmfmac: CONSOLE:    sp+0 0025fedc 00256828 002484b4 00259e78
[ 4510.650739] brcmfmac: CONSOLE: 001450.126   sp+10 18002000 001c0d97 00256828 0025fedc
[ 4510.650741] brcmfmac: CONSOLE:
[ 4510.650742] brcmfmac: CONSOLE: 001450.126 sp+14 001c0d97
[ 4510.650744] brcmfmac: CONSOLE: 001450.126 sp+2c 001e8151
[ 4510.650745] brcmfmac: CONSOLE: 001450.126 sp+38 00007f19
[ 4510.650747] brcmfmac: CONSOLE: 001450.126 sp+3c 00007fbd
[ 4510.650749] brcmfmac: CONSOLE: 001450.126 sp+44 00008629
[ 4510.650750] brcmfmac: CONSOLE: 001450.126 sp+5c 000292d9
[ 4510.650752] brcmfmac: CONSOLE: 001450.126 sp+7c 0002fb61
[ 4510.650753] brcmfmac: CONSOLE: 001450.126 sp+94 00030af3
[ 4510.650755] brcmfmac: CONSOLE: 001450.126 sp+a4 00030c29
[ 4510.650757] brcmfmac: CONSOLE: 001450.126 sp+b4 0002f821
[ 4510.650758] brcmfmac: CONSOLE: 001450.126 sp+bc 001af8b9
[ 4510.650760] brcmfmac: CONSOLE: 001450.126 sp+c4 0005b915
[ 4510.650762] brcmfmac: CONSOLE: 001450.126 sp+dc 00002e1d
[ 4510.650764] brcmfmac: CONSOLE: 001450.126 sp+e4 0002638f
[ 4510.650765] brcmfmac: CONSOLE: 001450.126 sp+e8 00026355
[ 4510.650767] brcmfmac: CONSOLE: 001450.126 sp+fc 0001f621
[ 4510.724923] ieee80211 phy1: brcmf_fw_crashed: Firmware has halted or crashed
[ 4513.247977] brcmfmac: brcmf_sdio_bus_rxctl: resumed on timeout
[ 4513.248043] brcmfmac: brcmf_sdio_checkdied: firmware trap in dongle
[ 4513.248085] brcmfmac: dongle trap info: type 0x4 @ epc 0x001bfc3a
                 cpsr 0x2000019f spsr 0x200001bf sp 0x0025fe50
                 lr   0x00049bd7 pc   0x001bfc3a offset 0x25fdf8
                 r0   0x00256828 r1   0x00000001 r2 0x00010000 r3 0x18001000
                 r4   0x00256828 r5   0xffffffff r6 0x00000000 r7 0x00000001
[ 4513.248096] brcmfmac: brcmf_dump_obss: dump_obss error (-110)
[ 4513.248130] ieee80211 phy1: brcmf_fil_cmd_data: bus is down. we have nothing to do.
[ 4513.248136] ieee80211 phy1: brcmf_cfg80211_get_tx_power: error (-5)
[ 4513.248153] ieee80211 phy1: brcmf_fil_cmd_data: bus is down. we have nothing to do.
[ 4513.248159] ieee80211 phy1: brcmf_cfg80211_get_tx_power: error (-5)
[ 4513.348642] mmc1: card 0001 removed
[ 4513.568626] mmc1: new ultra high speed DDR50 SDIO card at address 0001
[ 4513.569043] brcmfmac: F1 signature read @0x18000000=0x15264345
[ 4513.571553] brcmfmac: brcmf_fw_alloc_request: using brcm/brcmfmac43455-sdio for chip BCM4345/6
[ 4513.713666] brcmfmac: brcmf_c_process_txcap_blob: no txcap_blob available (err=-2)
[ 4513.713890] brcmfmac: brcmf_c_preinit_dcmds: Firmware: BCM4345/6 wl0: Aug 29 2023 01:47:08 version 7.45.265 (28bca26 CY) FWID 01-b677b91b
[ 4513.760041] brcmfmac: CONSOLE: ovar_op: scan_ver BCME -23 (Unsupported)
[ 4513.760047] brcmfmac: CONSOLE: 000000.098 wl0: unable to find iovar "toe_ol"
[ 4513.760048] brcmfmac: CONSOLE: 000000.098 wl0: wlc_iovar_op: toe_ol BCME -23 (Unsupported)
[ 4513.760050] brcmfmac: CONSOLE: 000000.098 wl0: wl_open
[ 4513.760051] brcmfmac: CONSOLE: 000000.109 wl0: wlc_phy_set_regtbl_on_femctrl: FIXME bt_coex
[ 4513.766141] brcmfmac: brcmf_cfg80211_set_power_mgmt: power save enabled
[ 4513.848042] brcmfmac: CONSOLE: 000000.187 wl0: wl_open
[ 4513.848049] brcmfmac: CONSOLE: 000000.199 wl0: wlc_phy_set_regtbl_on_femctrl: FIXME bt_coex
[ 4516.576016] brcmfmac: CONSOLE: 000002.915 wl0: malformed chanspec 0x0
[ 4516.576022] brcmfmac: CONSOLE: 000002.915 wl0: malformed chanspec 0x0
[ 4516.595027] brcmfmac: CONSOLE: 000002.935 wl0: malformed chanspec 0x0
[ 4516.595028] brcmfmac: CONSOLE: 000002.935 wl0: malformed chanspec 0x0
[ 4516.648042] brcmfmac: CONSOLE: 000002.965 wl0: wl_open
[ 4516.740033] brcmfmac: CONSOLE: 000003.066 wl0: wl_open
[ 4516.848047] brcmfmac: CONSOLE: 000003.172 wl0: wl_open
[ 4516.964028] brcmfmac: CONSOLE: 000003.277 wl0: wl_open
[ 4517.052026] brcmfmac: CONSOLE: 000003.377 wl0: wl_open
[ 4517.160047] brcmfmac: CONSOLE: 000003.482 wl0: wl_open
[ 4517.276027] brcmfmac: CONSOLE: 000003.587 wl0: wl_open
[ 4517.368032] brcmfmac: CONSOLE: 000003.690 wl0: wl_open
[ 4517.476038] brcmfmac: CONSOLE: 000003.797 wl0: wl_open
[ 4517.588068] brcmfmac: CONSOLE: 000003.902 wl0: wl_open
```

holta:
> > Whether or not a WiFi Country Code has already been set or not set (a.k.a. Regulatory Domain Code) would appear to be irrelevant. In other words, the failure of WiFi firmware cyfmac43455-sdio-standard.bin (would appear to?) occur regardless.
> 
> JUST A THEORY... but FWIW here's something that I'm trying to test:
> 
> * WiFi firmware cyfmac43455-sdio-standard.bin still crashes intermittently after `raspi-config nonint do_wifi_country US && reboot` has been run... _but it seems to crash a lot less often??_

Evidence points to a new theory:

WiFi firmware **/lib/firmware/cypress/[cyfmac43455-sdio-standard.bin](https://github.com/RPi-Distro/firmware-nonfree/blob/bookworm/debian/config/brcm80211/cypress/cyfmac43455-sdio-standard.bin)** crashes if certain commands were run in the seconds just prior to (1) restarting NetworkManager, or similarly, e.g. (2) `nmcli radio wifi off`

tim-moody:
What ill effect does a crash have after the radio has been turned off?

holta:
> What ill effect does a crash have after the radio has been turned off?

Not sure that's worth worrying about?  (Commands like `systemctl restart NetworkManager` [and `nmcli radio wifi off` with `nmcli radio wifi on`] are just ~2 diagnostic tools that I use to help trigger and trace all these WiFi firmware crashes.)

holta:
FYI while reiterating that RasPiOS Lite is not the priority right here, just for the record to get a rough idea of whether this problem affects some graphical-or-headless OS's more than others, I ran a quick comparison test.  (Why not, just in case that might reveal anything about the root of the problem?)

Results (below) are inconclusive, or not very useful statistically.  Still, here they are just for the record.

Each of the 2 tests took about 15min.  FYI the same test was run on each of the two OS versions ("Lite" and "with desktop") **without IIAB** — i.e. tests were run on their latest 64-bit RasPiOS's [*] — running on the very same RPi 500, one test right after the other:

- WiFi firmware **/lib/firmware/cypress/[cyfmac43455-sdio-standard.bin](https://github.com/RPi-Distro/firmware-nonfree/blob/bookworm/debian/config/brcm80211/cypress/cyfmac43455-sdio-standard.bin)** crashed 14 times (4.7% of 300 NetworkManager restarts) on "Lite"
- WiFi firmware **/lib/firmware/cypress/[cyfmac43455-sdio-standard.bin](https://github.com/RPi-Distro/firmware-nonfree/blob/bookworm/debian/config/brcm80211/cypress/cyfmac43455-sdio-standard.bin)** crashed 23 times (7.7% of 300 NetworkManager restarts) on "with desktop"

1. This is the test, that was run on each vanilla OS:

   ```
   seq 300 | xargs -I {} bash -c 'systemctl restart NetworkManager; sleep 2'
   ```

2. In another window/shell, I monitored the results using:

   ```
   watch -n0 'iw dev; echo; dmesg | grep Firmware:'
   ```

[*] CLARIF: Before the above test was run, each of the 2 vanilla/bare OS's was apt updated, `raspi-config nonint do_wifi_country US` was run, and the machine rebooted.

jvonau:
"The timeout error (-110) can have two root causes that I am aware off. Either the firmware died or the SDIO layer has gone haywire. Not sure if that swap partition is on eMMC device, but if so it could be related. You could try generating device coredump. If that also gives -110 errors we know it is the SDIO layer. "
https://www.uwsg.indiana.edu/hypermail/linux/kernel/2105.3/05569.html

holta:
@jvonau that thread on error (-110) looks useful.  Thanks.  A couple more data points to help bracket what's happening:

- WiFi firmware **/lib/firmware/cypress/[cyfmac43455-sdio-standard.bin](https://github.com/RPi-Distro/firmware-nonfree/blob/bookworm/debian/config/brcm80211/cypress/cyfmac43455-sdio-standard.bin)** crashed 70 times **(70% of 100 tries)** with the following test on [vanilla "RasPiOS with desktop" where `raspi-config nonint do_wifi_country US` was run](https://github.com/iiab/iiab/pull/3965#issuecomment-2746546242):

   ```
   seq 100 | xargs -I {} bash -c 'nmcli radio wifi off; nmcli radio wifi on; sleep 1.4'
   ```

- WiFi firmware **/lib/firmware/cypress/[cyfmac43455-sdio-standard.bin](https://github.com/RPi-Distro/firmware-nonfree/blob/bookworm/debian/config/brcm80211/cypress/cyfmac43455-sdio-standard.bin)** crashed 52 times **(52% of 100 tries)** with the same above test — on a vanilla/bare "RasPiOS Lite" _where (FWIW) a WiFi country code was never set._

Testing Context: Above tests used each of the latest two 64-bit RasPiOS's, that were apt updated and rebooted, on the same RPi 500, in the same location.

tim-moody:
Since both tests were run in loops, I take it that it always recovers from the crash.

holta:
> Since both tests were run in loops, I take it that it always recovers from the crash.

Yes.  Hence the reason this PR was created 2 weeks ago, as the `phyname` increments {phy0, phy1, ...} after every WiFi firmware crash.

(Papering over the problem would be fairly straightforward, as we discovered 2 weeks ago.  Whereas digging somewhat deeper, as a public service to inform the wider Raspberry Pi ecosystem [if we can, so far so good] will hopefully help everyone catch + diagnose such/similar problems far more meaningfully going forward.  The thread @jvonau pointed to at https://www.uwsg.indiana.edu/hypermail/linux/kernel/2105.3/05569.html could be a very promising step forward there.  :pray: )

jvonau:
The link to upstream issue 6049 that was posted 2 weeks ago has a fix that should be tried.

holta:
> The link to upstream issue 6049 that was posted 2 weeks ago has a fix that should be tried.

1. That workaround appears to work, preventing the 43455 chip's (-110) WiFi firmware crashes.  Could be a real breakthrough, for the wider RPi ecosystem, if this unofficial hack proves broadly safe.

2. Whether IIAB should carry this hack is a tougher question in my mind?  Definitely needs some thought this week.  Aiming for the simplest + cleanest + most low-maintenance resolution resolution to #3961, whatever that may be:
   - e.g. certainly this PR's two existing workarounds have been quite well-tested over ~2 weeks of very intensive testing, to get IIAB installs back on track, allowing 2021 WiFi firmwares to be dropped in place most important, without the need to muck with kernel modules.
   - e.g. We don't know exactly how 2021's very well tested WiFi firmwares might or might not behave alongside this workaround/hack?
   - e.g. RasPiOS Trixie arrives in in Q3 presumably, and might (or should?!) further clean things up here.

3. In any case, implementing this raspberrypi/linux#6049 approach would definitely not be the hard part.  Testing its broad ramifications would be the real work.  And whatever happens, this is big progress after almost a month, with much better options now clearer going forward.  Example: stub code below could obviously work to force reloading of `brcmfmac` kernel module, if that proves truly wise for IIAB to implement during installs, or more broadly across the board.  (The bigger question is probably whether-and-when RasPiOS _itself_ will or will not itself implement this popular workaround in 2025 — and if they refuse, we should try to find out why _they_ don't feel it's safe ;)

```
modprobe -r --remove-holders brcmfmac || True    # Ignoring error (1) seems to work: "modprobe: ERROR: ../libkmod/libkmod-module.c:856 kmod_module_remove_module() could not remove 'brcmfmac': No such file or directory"
cat > /etc/modprobe.d/brcmfmac.conf << EOF
options brcmfmac feature_disable=0x282000
options brcmfmac roamoff=1
EOF
modprobe brcmfmac
```

jvonau:
Toss the hack into `iiab` without modprobe involved after `apt upgrade`, maybe force a reboot anyway? Remove 'brcmfmac.conf' at the end of `iiab` before the photograph the message part and call it a day. That leaves Tim's workflow unchanged with nothing hanging around to clean up later.

holta:
> Toss the hack into `iiab` without modprobe involved after `apt upgrade`, maybe force a reboot anyway? Remove 'brcmfmac.conf' at the end of `iiab` before the photograph the message part and call it a day. That leaves Tim's workflow unchanged with nothing hanging around to clean up later.

Good thought.  I'm open to any "one-off" ideas like this (i.e. use of this hack during certain IIAB installs, if the IIAB installer detects precise conditions where the hack is really needed, and then immediately deletes the hack after the IIAB install?)

- e.g. this cool hack, while definitely cool in many regards, is definitely not needed in several other situations like RPi 400 (I've tested that carefully; which has a different WiFi chip; no matter how hard I hit it, I can't get the WiFi firmware to crash on the RPi 400!)

- So I want to ponder/meditate on the consequences, to try to make sure I'm not being naive overlooking any common implementer/user trajectories etc.

holta:
Background Motivation / Context: Just a reminder that my primary UX goal here is (in fact) to avoid unnecessary reboots during IIAB installs — in any and all conditions where a reboot is not absolutely necessary:

- So that newcomers' frustrations are contained (and they have a far more welcoming experience; that means "implementer UX" not "end user UX" in this case).

- Much like this code that carefully avoids rebooting unless it's deemed really/very/quite necessary — e.g. if a new kernel was suddenly apt installed or similar: https://github.com/iiab/iiab-factory/blob/9cf4cf1d8c78665389a6a1a9d1951f8522cc7add/iiab#L447-L457

jvonau:
Interesting read: https://blog.quarkslab.com/reverse-engineering-broadcom-wireless-chipsets.html

jvonau:
>     * Much like this code that carefully avoids rebooting unless it's deemed really/very/quite necessary — e.g. if a new kernel was suddenly apt installed or similar: https://github.com/iiab/iiab-factory/blob/9cf4cf1d8c78665389a6a1a9d1951f8522cc7add/iiab#L447-L457

Given the help that I gave you developing that bit of code the upstream images released will only be able to avoid a reboot till a new kernel is released, a few weeks at best from past history. https://archive.raspberrypi.com/debian/pool/main/l/linux/ What the code does not handle at all is new wifi firmware released but that is not so often https://archive.raspberrypi.com/debian/pool/main/r/raspberrypi-firmware/?C=M;O=D

jvonau:
In my humble opinion https://github.com/iiab/iiab/pull/3965/commits/49c2a9394ceb218cfb33a494c53c83f22fe351ae is very misleading, the firmware was not crashing during the restart of NM based on the timestamps present as I pointed out above in https://github.com/iiab/iiab/pull/3965#issuecomment-2718578084 which you duplicated in less than 8% of the cases noted in https://github.com/iiab/iiab/pull/3965#issuecomment-2746546242 on a virgin desktop install without IIAB. The firmware crashed for whatever reason long before starting any IIAB install, that is the reason for phy0 changing to phy1 in the background.

jvonau:
https://github.com/iiab/iiab/pull/3965/commits/68520e851fd23b2e929ba81b584cd439d9c5541e should be standalone like other improvements to iiab-diagnostics done in the past.

holta:
> firmware was not crashing during the restart of NM based on the timestamps present

> firmware crashed for whatever reason long before starting any IIAB install

The above claims are incorrect.  Do take a closer look at the timestamps, or better yet do your own install and you will see right away.  As I've verified dozens of times on many different Raspberry Pis over past weeks (RPi 4, 5 and 500) that contain the 43455 WiFi chip.

In short, the [restart of NetworkManager in NM-debian.yml](https://github.com/iiab/iiab/blob/67fe7570e0fb7fd0f7dec745006be7cba2118f58/roles/network/tasks/NM-debian.yml#L83-L87) is the immediate cause of its WiFi firmware crashing — i.e. when `raspi-config nonint do_wifi_country XX` was run 2-3 seconds prior (effectively the worst possible place to do this, nobody's fault, but now we know ;-)

> duplicated in less than 8% of the cases

WiFi firmware crashing was reproduced [70% of the time](https://github.com/iiab/iiab/pull/3965#issuecomment-2749281302) in some cases — and approx 80-to-90% of the time, or more, in other cases (e.g. during IIAB installs on RasPiOS with desktop, which is the central topic at hand).  In any case, this is not a rare failure; it's very serious.

holta:
I want better logging of WiFi crashes (that really should be a central part of this PR, in addition to iiab-diagnostics logging-oriented improvements — if there's time — to prepare for the next time WiFi firmware goes bonkers in one way or another!)

BUT, I also want this PR merged ASAP (very soon, as way too many people are being blocked by this regression).  Definitely not our fault!  Either way: I'm hopeful this weekend will materially clear up this mess at long last! 😅

jvonau:
For what it is worth, Tim sent me a Pi500 a while ago and I could not duplicate this issue using the same methodology as you tried. As data points there were two differences:
- I followed the same basic provisioning as Tim ie apt upgrade and setting the "WiFi country code" before rebooting then installing.
- tailscale and remoteit were not installed.

Do whatever you need to do to keep your paymasters happy and stroke your ego. Now get back to dealing with the other issues that have been uncovered regarding raspi-config and raspi-imager to benefit the wider ecosystem you speak of.

holta:
> apt upgrade and setting the "WiFi country code" before rebooting then installing

1) If I'm understanding what you did, that's to be expected.  Reason: when [network/tasks/main.yml (Line 69)](https://github.com/iiab/iiab/blob/d3340267ba3b888c9c00c8bbe0363b6353cbf1d2/roles/network/tasks/main.yml#L68-L70) turns on WiFi for the very first time (e.g. as raspi-config's `do_wifi_country` unblocks WiFi via [nmcli radio wifi on](https://github.com/RPi-Distro/raspi-config/blob/c5bfa1c8053f6a4366295dbbfd53480ecaca577c/raspi-config#L915) much like rfkill) about 3 seconds prior to NM-debian.yml restarting NetworkManager, the end result is that **RasPiOS with desktop**'s "standard" 43455 WiFi firmware crashes (often).  IIAB installs on **RasPiOS Lite** might not be affected as much (I don't know).

2) Personal attacks and falsehoods about others both need to end.  Completely.

jvonau:
> > apt upgrade and setting the "WiFi country code" before rebooting then installing
> 
>     1. If I'm understanding what you did, that's to be expected.  

That should be the standard operation procedure, period. You should be teaching the newbies the best way not the easiest way.
> Reason: when [network/tasks/main.yml (Line 69)](https://github.com/iiab/iiab/blob/d3340267ba3b888c9c00c8bbe0363b6353cbf1d2/roles/network/tasks/main.yml#L68-L70) turns on WiFi for the very first time 

You just stated the problem doesn't occur when properly set in advance, that clearly points to the problem with having that code papering over laziness to find the best method. The docs aren't written in stone and using that as an excuse just doesn't cut it. 
>(e.g. as raspi-config's `do_wifi_country` unblocks WiFi via [nmcli radio wifi on](https://github.com/RPi-Distro/raspi-config/blob/c5bfa1c8053f6a4366295dbbfd53480ecaca577c/raspi-config#L915) much like rfkill) about 3 seconds prior to NM-debian.yml restarting NetworkManager, the end result is that **RasPiOS with desktop**'s "standard" 43455 WiFi firmware crashes (often).  IIAB installs on **RasPiOS Lite** might not be affected as much (I don't know).

As I said many times in the past proper provisioning provides for a less bumpy ride.
> 
>     2. Personal attacks and falsehoods about others both need to end.  Completely.

You fired the first shot many many years ago and I have been very restrained with the facts and my opinions that I put forth about the subjects. You should be more thankful that I keep them to myself but if you really want to go there I'd be more than willing to point them out.

PR #3966, opened 2025-03-15
Improved 10 minute CI run and switched to local_vars_small.yml

Akatama:
As requested by A. Holt at SCaLE, I cleaned up the 10 minute CI run. Here is the gist of what I did:

- Removed commented out lines, run statements that just reported status and unncessary comments
- Switched from local_vars_none.yml to local_vars_small.yml
- Added a new file: tests.unused/expected_state_small.yml which matches the content of /etc/iiab/iiab_state.yml when the install is successful
- Added a run statement that compares the /etc/iiab/iiab_state.yml and tests.unused/expected_state_small.yml with diff, just to give the CI run a validation at the end

The 10 minute CI run uses Ubuntu

The other two CI runs (RasPiOS and Debian 12) are unchanged as of right now and are passing on my fork

@holta


holta:
@Akatama three quick suggestions you might consider!

1. Move `tests.unused/expected_state_small.yml` to `vars/expected_state_small.yml` please?
2. Can we consider preserving `workflows/10min-iiab-test-install.yml` almost as is, renaming it to `workflows/10min-iiab-none-install.yml` — it's very fast and let's keep most comments untouched in this one (many comments are **incredibly useful** here to teach newbie + int'l community folk about GitHub Actions!)
3. While you create a brand **new workflow** `workflows/20min-iiab-small-install.yml` — _stripping it of all comments if that's your preference?_

Akatama:
@holta I will make these suggested changes, hopefully sometime tomorrow.

I don't necessarily hate comments - I had thought you wanted them removed, but I could definitely add some for newbies. I probably just misunderstood - there were quite a few lines commented out in the original, and that's probably more what you were going for in the original.

holta:
> @holta I will make these suggested changes, hopefully sometime tomorrow.

1. Awesome, huge thanks @Akatama !
2. Go wild with comments: however few or however many you feel work best in your new `workflows/20min-iiab-small-install.yml` 💯

tim-moody:
vars is for config variables. expected state is not a configuration file so I would find another home for it.

holta:
> 1. Move `tests.unused/expected_state_small.yml` to `vars/expected_state_small.yml` please?

@Akatama while creating a folder for just 1 file is a bit much, let's go ahead with `tests/expected_state_small.yml` nevertheless.

tim-moody:
@Akatama, sounds good. one day there could be other expected states for larger tests. is workflows a mandated name or could the test yamls be in tests as well?

Akatama:
@tim-moody @holta it would seem logical to me to include any test files or code in the tests directory. So in the future they may be code we run against the installed IIAB instance.

I chose this only because I wanted to have some kind of assertion at the end of the workflow. If there is a better thing I can assert on let me know. 

Another consideration is that doing assertions in bash is not what I would call ideal. It's what I could do now without creating a test suite in python or something. So, in the future, we may remove the assertion with diff and do an assertion in a python test instead, for example. Assuming this yml file is good to assert against, it may still exist just be done by a different method. 

I don't know what you mean by "is workflows a mandated name". "workflows" is what GitHub calls their CI/CD pipelines, essentially.

tim-moody:
>I don't know what you mean by "is workflows a mandated name". 

I thought you had a workflows directory. Seems I was wrong. Putting everything in tests sounds right.

Akatama:
@tim-moody Doh! I should've realized that is what you meant. Sorry, I am not trying to be pedantic.

Yes, the directory is .github/workflows/. You may be able to configure a different path for the workflows, but that's where they've been for a long time in this project and that path is standard.

We could include this file in either .github/ or .github/workflows/. Some things to consider if we go with .github/workflows/:

- as workflows are .yml files, it may be possible to attempt to run these test files as CI runs (I haven't tried), but of course they will fail
- they won't run automatically as they are not set up to do that
- including .yml files that don't have the right syntax for a full workflow in this path or some kind of subdirectory of this path is pretty standard. This is because you can include steps that are repeated in multiple workflows in a separate file, then call that file inside of the actual workflows that run. I mention this only to say that it shouldn't mess up anything with the current workflows.

Let me give that a try and see how it goes.

tim-moody:
@Akatama I trust your judgement. My concern was the vars directory along with simplicity.

Akatama:
@holta here are some changes

- made the workflow I worked on into 20min-iiab-small-install.yml, now with a few comments
- restored 10min-iiab-test-install.yml. I still thought there was a lot of commented out lines or lines that weren't doing anything so I removed them. I also added a few more comments here for future contributors. Finally, I did a similar thing with diff as the 20min workflow.
- I moved expected_state_small.yml to .github/workfiles/tests/ along with the newly create expected_state_none.yml. Since these are being only used in the github workflows, I think this is a better place for them, at least for now.

The biggest change I made to 10min-iiab-test-install.yml is the change to diff. If you really want me to remove it, I can, but I have a couple reasons why I think this is a good idea.

1. For future contributors, if they are looking at the workflow, what do we expect for them to think about "cat /etc/iiab/iiab_state.yml"? Obviously, it just prints what is in that file, so they can only assume that **anything** in that file is okay. That is not the case, either the file has the correct information in it or it doesn't.
2. If that part of the CI run were to fail, it would give both future and current contributors a few possibilities for what they could do to fix it. Either expected_state_none.yml is out of date and must be updated, there is something wrong with the installation process in the workflow, or some change in the code base has resulted in some kind of improper state.

I would like to do the same thing to iiab-summary, but we are still getting a "fatal" line in it, so for now I have kept it in both workflows. I am hoping to be able to take a look at why it behaves that way sometime soon (maybe even tonight!)

Thank you holta and @tim-moody for your thoughts on this PR!

PR #3974, opened 2025-03-26
15min CI/QA using SMALL-sized IIAB & clarifs building on PR #3966

holta:
@Akatama is this adjustment building on your work Ok?  This is my preference so far!  Building on your:

- PR #3966

- FYI I've also changed from local_vars_none.yml to local_vars_unittest.yml in the main CI workflow.  Working with the layout you set, that I like, creating file .github/workflow/tests/expected_state_unittest.yml to nudge us all closer to test automation. 💯

- Yep I've added/tweaked many inline comments ;)  Also, to keep much more explanatory context within that one GitHub Actions workflow, allowing for newbies (for international newcomers especially) who don't know nearly as much about CI and GHA as you do, _and need that (:_

Anyway a great first step — seems to work really well so far with all 4 tests passing here:

- https://github.com/holta/iiab/commits/15min-ci/


holta:
Just spoke to @Akatama directly.  He's a bit busy today.  I'll go ahead and merge now, and revisions can be added later as necessary.

Issue #3967, opened 2025-03-15
Output of iiab-summary on the 10-minute CI run (Ubuntu 24.04) "fatal: no names found, cannot describe anything"


Akatama:
While working on the changes for the 10 minute CI run (the one that runs on Ubuntu 24.04) I noticed that iaab-summary command has some strange output. I am very new to this project, but just as a new user if I saw something like this I would assume something was wrong with how Internet-in-a-Box was set up on my system if I saw output like this.

Let me know if this is actually the expected output, or if something is awry in the setup so I can look into fixing it.

The example below is actually taken from the last CI run on this repo before my recent pull request, so its not something that is only happening on my fork:

```
fatal: No names found, cannot describe anything.
/usr/bin/iiab-summary: line 25: cd: /opt/iiab/iiab-admin-console: No such file or directory
fatal: No names found, cannot describe anything.
install_date = 2025-03-12T07:57:55Z   Current TZ: UTC

iiab: ef1bb46, 0 PR's / 0 commits since tag 
    "Merge 89b31ee085017fe8d149bf54e56a06d75dff859a into 752ab0ee19552d7f2cdff3b1a048ea6761d4301d"
    none   branch: 

    WARNING: Directory /opt/iiab/iiab-admin-console does not exist!

Ubuntu 24.04.2 LTS   trixie/sid
display-manager? inactive   Arch1: amd64   Arch2: 
box 6.8.0-1021-azure #25-Ubuntu SMP Wed Jan 15 20:45:09 UTC 2025 x86_64
AMD EPYC 7763 64-Core Processor   CPU(s): 4   Mem: 15990 Swap: 4095

Filesystem      Size  Used Avail Use% Mounted on
/dev/root        72G   48G   24G  67% /      ZIMs: 0 OER2Go: 0 Apps2B: 1

lo eth0 docker0
box box.lan fv-az1441-793.lnin1eoc0irezeqatpkuni3jdf.cx.internal.cloudapp.net localhost localhost.localdomain
10.1.0.126 172.17.0.1 
```

tim-moody:
>/opt/iiab/iiab-admin-console: No such file or directory

assuming this is not permissions, I assume it means that that repo has not been cloned yet.

Akatama:
[...]

PR #3976, opened 2025-03-27
Clean up / tighten up 2+2=4 GitHub Actions workflows

holta:
This PR finally cleans up the last of those ugly "fatal: No names found, cannot describe anything." messages output by [iiab-summary](https://github.com/iiab/iiab/blob/4b1d894cb5124585d8a2cd3c66d10b10b830d064/scripts/iiab-summary#L10) at the end of each GitHub Actions workflow — exacerbated by `actions/checkout@v4` 's [buggy implementation of `fetch-tags: true`](https://github.com/iiab/iiab/blob/6ac68a029e79a3c5731499948493a2c5067e0f97/.github/workflows/10min-iiab-unittest-ubuntu.yml#L32) — as we started to wake up to 2 weeks ago here:

- https://github.com/iiab/iiab/issues/3967#issuecomment-2727170217

And this PR tightens up, inspired by @Akatama, cleaning up filenames, workflow titles, etc.  Results look good: https://github.com/holta/iiab/commits/15min-ci/

That's probably enough for one PR: thanks @Akatama if you can confirm or suggest next steps.

Building on:

- PR #3966 
- PR #3974

### DIFF
--- a/.github/workflows/10min-iiab-test-install.yml
+++ b/.github/workflows/10min-iiab-test-install.yml
@@ -29,20 +29,15 @@ jobs:
       - name: Check out repository code    # Clones repo onto github runner
         uses: actions/checkout@v4
         with:
-          fetch-depth: 10000    # Default is 1, but iiab-summary (below) needs git tag history
-      - run: echo "🍏 This job's status is ${{ job.status }}."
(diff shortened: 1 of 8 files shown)


Akatama:
@holta This look very good to me! Thanks for figuring out the issue with actions/checkout@v4. 

I would approve but I don't seem to have an approve button. So take this to be the same as me doing that.

PR #3977, opened 2025-03-28
iiab-diagnostics more resilient

holta:
Allows iiab-diagnostics to run in more situations, e.g. even when many IIAB components are not yet in place, by making the command more self-sufficient. ✅


holta:
Tested on Ubuntu 25.04 pre-release.

PR #3978, opened 2025-03-30
Add a MEDIUM-sized IIAB install (GitHub Actions workflow)

holta:
@Akatama just a small step forward:

- Let's go with this experimentally, to see if MEDIUM-sized installs complete reliably over coming weeks.
- So far so good, as test results show: https://github.com/holta/iiab/commit/405a2ce054f2d8272a7d88ca9262ac741ecaabe8
- We can consolidate later to avoid duplication &/or work towards on-demand test coverage.

Building on:

- PR #3966 
- #3967 
- PR #3974
- PR #3976

### DIFF
--- a/.github/workflows/20min-iiab-medium-ubuntu.yml
+++ b/.github/workflows/20min-iiab-medium-ubuntu.yml
@@ -1,0 +1,27 @@
+name: '"20 min" IIAB "MEDIUM" on Ubuntu 24.04 on x86-64'
+
+on: [push, pull_request, workflow_dispatch]
+
+jobs:
+  test-install:
+    runs-on: ubuntu-24.04
+    steps:
+      - name: Check out repository code
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0    # Default is 1, but iiab-summary (below) needs git tag history.
+      - name: Set up /opt/iiab/iiab
+        run: |
+          mkdir /opt/iiab
+          mv $GITHUB_WORKSPACE /opt/iiab
+          # Github Actions is weird, do not delete the line below. Yes, it will cause everything to break. Really.
+          mkdir $GITHUB_WORKSPACE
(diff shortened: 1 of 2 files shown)


Akatama (approved):
All looks good. I guess the question is do we want all of these workflows to run on Ubuntu? I know IIAB supports Debian and Raspberry Pi OS.

Its a simple fix if we want to change it later, hence why I am approving.

holta:
> do we want all of these workflows to run on Ubuntu?

Right!  My opinion is that CI/QA focused on Ubuntu is (probably!) good enough this year in 2025:

- Certainly [GitHub Actions mainline runners](https://github.com/actions/runner-images#available-images) don't give us much choice at the moment — emulation layers (like [pguyot/arm-runner-action@v2](https://github.com/pguyot/arm-runner-action) for rudimentary Raspberry Pi tests) might in theory allow testing of any OS, but this greatly slowa things down — so **fast checks** really do need to run on Ubuntu as a practical matter (unless we really want to self-host some tests at some future date?)

- More immediately, it's possible this question might lead us towards very basic automated graphical testing of IIAB installs onto Linux Mint (very popular internationally in schools trying to build up IT literacy etc).  Is that goal too ambitious?  It might be.  Then again I'm (told!) almost anything is possible using [Selenium](https://www.browserstack.com/selenium) similar to https://github.com/iiab/calibre-web/pull/291 😄

holta:
Placeholder:

- `20min-iiab-medium-ubuntu.yml` appears to work well, and is kind of a superset of both `10min-iiab-unittest-ubuntu.yml` and `15min-iiab-small-ubuntu.yml` — so there's discussion of possibly pushing those 2 smaller ones into cold storage (e.g. perhaps a subdir like `.github/workflows/Old`).
- Or something like that: @Akatama is working on other/better ideas too.

PR #4004, opened 2025-04-28
Used Ansible-Lint to clean up spacing in Ansible configuration

Akatama:

These are the following changes that were made to the Ansible configuration. I had to make a few of the changes manually as ansible-lint wasn't able to handle it, but they were all pretty simple. 

key-order 
- Wants the "layout" to be in a specific order, so name at the top, then hosts, tasks, ect.
- See here: https://ansible.readthedocs.io/projects/lint/rules/key-order/
- https://pastebin.com/mBakQ1j5

risky-octal
- Basically it says we should change how we write file permissions. 644 -> "0644". 
- "Using integers or octal values in YAML can result in unexpected behavior. Forexample, the YAML loader interprets 0644 as the decimal number 420 but putting 644 there will produce very different results."
- https://pastebin.com/w6iD96yA

yaml[indentation]
- Wrong levels of indentation found.
- https://pastebin.com/CHdrfy07

jinja[spacing]
- Jinja2 spacing could be improved. Ex: {{some|dict2items}} -> {{ some | dict2items }
- https://pastebin.com/NuvAmPhu

Other spacing
- combined the rest of these into one pastebin because there aren't a lot of them: yaml[empty-lines], yaml[trailing-spaces], yaml[colons] 
- https://pastebin.com/6G2cA9vn

Ubuntu 24.04, Debian 12 and RasPiOS on Zero 2 W 

@holta

I plan to do some more testing, but it would be a good idea if someone else to did some testing too.


holta:
Thank you @Akatama: I'll take a look later today.

Akatama:
@holta I successfully installed IIAB with local_vars_large.yml on a Multipass instace. I am not sure what other tests I am supposed to perform exactly, but when i have some time I will try to poke around a bit. Is there anything else you think we should do?

PR #4007, opened 2025-05-01
Used Ansible-Lint to clean up some spacing in Ansible configuration

Akatama:
Description of changes proposed in this pull request:

These are the following changes that were made to the Ansible configuration. I had to make a few of the changes manually as ansible-lint wasn't able to handle it, but they were all pretty simple.

These are the rules at were resolved:
- yaml[empty-lines]
- yaml[trailing-spaces]
- yaml[colons]

https://pastebin.com/6G2cA9vn

Smoke-tested on which OS or OS's:

Ubuntu 24.04, Debian 12 and RasPiOS on Zero 2 W
Mention a team member @username e.g. to help with code review:

@holta

I plan to do some more testing, but it would be a good idea if someone else to did some testing too.

### DIFF
--- a/.ansible-lint.yml
+++ b/.ansible-lint.yml
@@ -1,0 +1,125 @@
+---
+# .ansible-lint
+
+profile: moderate # min, basic, moderate,safety, shared, production
+
+exclude_paths:
+  - vars # vars files are used to configure IIAB, and are read by Ansible. No need to lint them
+  - tests.unused # unused files
+  - roles/0-DEPRECATED-ROLES # deprecated
+  - roles/calibre/templates/users.sqlite # ansible doesn't like the content of this file, but it works
+  - roles/network/tasks/fedora/* # no longer support Fedora
+  - run-one-role.yml # ansible-lint does not like this file, need to find out why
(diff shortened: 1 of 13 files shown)


holta:
_Thank you @Akatama:_

1. Removal of dangling whitespace on the right side is long overdue. ✅

2. However with blank lines — can we please keep all blank lines at this time?  

(Blank lines [sometimes several blank lines in a row] can help provide strong visual / intentional separation, between different sections of code, that serve very different purposes.) 

3. Blank lines containing spaces/tabs can of course be tightened up, to be simple newlines.  In other words, removing the dangling spaces/tabs on the right side of all lines, even if those spaces/tabs are the only characters on the line.

tim-moody:
for those who use Visual Studio Code, there is a setting that will remove whitespace on the end of all lines on save

(for those like me who prefer spaces to tabs, there is a setting that does that too)

Akatama:
@holta looking at the files changed, it looks like I successfully reverted all files where trailing lines where removed. Let me know if there are any other changes you need me to make.

Issue #4012, opened 2025-05-10
Patch up Minetest -> Luanti role for (evolving!) Ubuntu 25.10 pre-releases


holta:
A quick job: I'll do it & test it coming days.

(And of course things will inevitably change in the coming year when Ubuntu 25.10 or 26.04 inevitably catch up to Debian 13 Trixie, repackaging / renaming so many different aspects of Minetest and its internals to Luanti.)

Related:

- PR #4011

fivemeyestore:
Your approach to this problem is very elegant. I'm learning a lot from reviewing your code.

myeyestore:
Impressive project structure! The architecture really demonstrates solid design principles.

fivemhostingserver:
The performance optimizations here are quite clever. Nice engineering work!

eyesgodcode:
The performance optimizations here are quite clever. Nice engineering work!

fivemeyestoreserver:
Excellent contribution to open source! The community will benefit greatly from this.

godofcodeeyes:
Interesting solution to this problem! I've been working on similar challenges recently.

eyestorefivem:
Impressive project structure! The architecture really demonstrates solid design principles.

fivemserverhostings:
Great implementation of this feature. I particularly like the approach to error handling.

eyestorefivemstore:
Interesting solution to this problem! I've been working on similar challenges recently.

ayestoro:
Great implementation of this feature. I particularly like the approach to error handling.

eyesaider:
The performance optimizations here are quite clever. Nice engineering work!

fivemservereyestore:
The performance optimizations here are quite clever. Nice engineering work!

eyestoreoye:
Great implementation of this feature. I particularly like the approach to error handling.

scriptforfivem:
Impressive project structure! The architecture really demonstrates solid design principles.

holta:
Mitigated by...

- PR #4013

PR #3985, opened 2025-04-03
Luanti 5.8+ (was Minetest) from 2023, for new OS's

holta:
### Description of changes proposed in this pull request:

This PR mainly just [renames](https://blog.luanti.org/2024/10/13/Introducing-Our-New-Name/) Minetest to Luanti across the board, as happened upstream a couple versions ago in 2024.  Recent IIAB install problems (re: Minetest) on Debian 13 Trixie made it obvious that it's time.  A few smaller refactoring/hygiene things are included too, e.g.:

- Legacy support is instrumented for older OS's, so that Minetest continues to work reasonably well in the interim (all this logic should really go away well before the end of 2026).  This older-versus-newer OS's threshold can of course be adjusted and fine-tuned in future, e.g. if better support for Ubuntu 25.04 and 25.10 (or Linux Mint 22 or whatever) is really necessary in future.

- On newer OS's, a stub game (hard-coded to 2025-03-02's [minetest_game](https://content.luanti.org/packages/Luanti/minetest_game/) for now) is downloaded — as [required](https://docs.luanti.org/for-server-hosts/setup/#running-the-server) by [Minetest (now Luanti) versions since 5.8.0](https://blog.luanti.org/2023/12/04/5.8.0-released/#minetest-no-longer-ships-with-a-default-game), that was released in 2023.

- Doc [roles/luanti/README.rst](https://github.com/holta/iiab/blob/minetest-to-luanti/roles/luanti/README.rst) is updated accordingly (could probably use a bit more cleanup!), and likewise in relevant vars files.

- Other enhancements can no doubt be added, but probably that should happen later as needed in other/separate PRs.

### Smoke-tested on which OS or OS's:

Debian 13 and Ubuntu 24.04

Raspberry Pi OS testing should probably follow, e.g. after merging if necessary.

### Related:

- PR #3426

PR #4013, opened 2025-05-11
Revert to "minetest" naming (not "luanti" naming) on Ubuntu 25.10 for now — fyi 25.10's virtual pkg `luanti-server` exists but needs work

holta:
At the moment Ubuntu 25.10 (Questing Quokka) contains a virtual apt package luanti-server, that was tripping up IIAB's thresholding logic (re: old Minetest naming versus new Luanti naming), as least during these very early days of Ubuntu 25.10 pre-releases.

- So this PR allows treats Ubuntu 25.10 as an "old" OS for the sake of Minetest/Luanti, for now — allowing LARGE-sized IIAB installs to at least complete.

- Though unfortunately this OS's package (luanti-server ~= minetest-server) doesn't yet seem to work — the failure is: "minetest-server.service: Start request repeated too quickly"

- So hopefully Ubuntu 25.10 will clean that up in coming months (we should check back prior to 25.10's 2025-10-09 release!)

Related:

- PR #3985
- #4012

### DIFF
--- a/roles/luanti/tasks/main.yml
+++ b/roles/luanti/tasks/main.yml
@@ -26,16 +26,16 @@
         luanti_deb_and_systemd_name: luanti-server    # Debian 13 renamed deb/apt package TO 'luanti-server': WILL UBUNTU 25.10+ DO LIKEWISE??
 
     - name: Is apt package 'luanti-server' available? (Quite likely if OS is new since mid-2025!) But if not, show red error and continue.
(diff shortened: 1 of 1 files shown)


holta:
Smoke-tested on Ubuntu 24.04, 25.10 and Debian 13:

Good enough until we reassess Ubuntu 25.10's Luanti strategy later this year!

PR #4020, opened 2025-05-29
Transition to Node.js 24.x

holta:
Annual upgrade, building on:

- PR #3738

As tested on Ubuntu 25.10 pre-release with...

```
# node -v
v24.1.0
```

Node.js 24 announcements:
https://nodejs.org/en/blog/release/v24.0.0
https://blog.logrocket.com/node-js-24-new/

### DIFF
--- a/roles/internetarchive/tasks/install.yml
+++ b/roles/internetarchive/tasks/install.yml
@@ -9,10 +9,10 @@
   include_role:
     name: nodejs
 
-- name: Assert that 10.x <= nodejs_version ({{ nodejs_version }}) <= 22.x
+- name: Assert that 10.x <= nodejs_version ({{ nodejs_version }})
   assert:
-    that: nodejs_version is version('10.x', '>=') and nodejs_version is version('22.x', '<=')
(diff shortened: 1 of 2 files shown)


tim-moody:
The only meaningful test is whether the archive role works (not simply installs) with this change, so I assume that will be tested as well. It is my experience that node apps tend not to upgrade to the latest due to compatibility problems with dependencies. Are there other roles that use node? None comes to mind.

holta:
roles/internetarchive (http://box:4244) was tested to work equivalently, as with prior versions like Node.js 22.

> other roles that use node?

Also {JupyterHub, Node-RED, Sugarizer} work, on Ubuntu 25.10 pre-release with Node.js 24.

pbx (FreePBX) is not a mainline app, but I and others will test more later (a quick smoke-test seems to indicate it installs fine with Node.js 24 on Debian 12 Bookworm).

holta:
> pbx (FreePBX) is not a mainline app, but I and others will test more later (a quick smoke-test seems to indicate it installs fine with Node.js 24 on Debian 12 Bookworm).

pbx (FreePBX) also confirmed to work with IIAB and Node.js 24 on Ubuntu 24.04.

Good enough!

PS There might well be issues with FreePBX on Ubuntu 25.10 pre-releases, possibly related to PHP changes, but that's unrelated — and not terrible surprising since Asterisk and FreePBX have moved their baseline / supported OS to Debian 12 (Bookworm) for now.

PR #4025, opened 2025-06-27
trisquel-profile-setup.sh FIX for arm64, e.g. for IIAB VMs on new Macs

holta:
### Fixes bug:

The `apt update` command was failing, while trying to run [/opt/iiab/iiab/scripts/ansible](https://github.com/iiab/iiab/blob/master/scripts/ansible) on "Trisquelized" Ubuntu 24.04 VMs on a new Mac, i.e. with Apple's new silicon (arm64) CPUs.

Fix (refine) the apt sources file for both leading CPU architectures — preserving `archive.ubuntu.com` for amd64, while adding `ports.ubuntu.com` for arm64 — modeling this after:

https://www.reddit.com/r/Ubuntu/comments/1jxo3uy/how_to_add_arm64_repositories_to_ubuntu_x64_noble/

Tested on a Ubuntu 24.04 LTS VM that was "Trisquelized" using `multipass launch 24.04 --cloud-init https://iiab.io/trisquelize.yml` — all of which was running on a new Mac (with arm64 CPU).

@Ark74 can you review / confirm this looks good, allowing this to be merged immediately if possible?

### DIFF
--- a/scripts/trisquel-profile-setup.sh
+++ b/scripts/trisquel-profile-setup.sh
@@ -37,13 +37,13 @@ gpg --export "$REPO_GPGKEY" | gpg --dearmour > "$UPSTREAM_GPGKEY"
 
 echo "
 # For amd64
-deb [arch=amd64] http://archive.ubuntu.com/ubuntu/ ${UPSTREAM} main universe multiverse
-deb [arch=amd64] http://archive.ubuntu.com/ubuntu/ ${UPSTREAM}-updates main universe multiverse
-deb [arch=amd64] http://archive.ubuntu.com/ubuntu/ ${UPSTREAM}-security main universe multiverse
+deb [arch=amd64] http://archive.ubuntu.com/ubuntu/ ${UPSTREAM} main universe
+deb [arch=amd64] http://archive.ubuntu.com/ubuntu/ ${UPSTREAM}-updates main universe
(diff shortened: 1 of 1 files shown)


holta:
CLARIF on the meaning of terms { `main`, `universe`, `multiverse`, `restricted`} in the context of files like `/etc/apt/sources.list`

- https://askubuntu.com/questions/58364/whats-the-difference-between-multiverse-universe-restricted-and-main

- https://en.wikipedia.org/wiki/Ubuntu#Package_classification_and_support

holta:
@Ark74

Tiny side question: I presume we follow community norms, preserving `http` (instead of `https`) within all 3+3 URLs in /etc/apt/sources.list.d/upstream_repo.list ?

Ark74:
@holta 

- the http usage is compatible to use with some transparent (or non-transparent) proxy cache setup so that will be very effective on a limited network environment
- apt uses gpg to verify the package signature so is not like https will make it more secure
  - in terms of privacy, https will prevent anyone listening the network know you just downloaded libreoffice from the apt repo, I suppose you can use it if you want that level of anonymity.
- IIRC trisquel console (headless/bare minimum) dont ship ca-certificates, so it should be added if you require to switch to https

About the arm64 and the ubuntu ports. yeah, that is correct!, sorry I forgot about it, in Trisquel all archs are part of archive.trisquel.org
- amd64
- armhf
- arm64
- ppc64el

so yeah, LGTM.

PR #4026, opened 2025-06-29
Added fixes for jinja[spacing] from ansible lint

Akatama:
According to ansible-lint documentation: "when there are no spaces between variables and operators, including filters, like {{ var_name | filter }}. This improves readability and makes it less likely to introduce typos."

An example of Ansible configuration that would violate this rule would be something like 
{{ var_name|filter }} or {{munin_username}}. If there was more than one space between the content inside the jinja template and the brackets, that would also violate it.

Ubuntu 24.04, Debian 12 and RasPiOS. I will also test on a Multipass VM shortly

@holta


holta:
Looks great, thank you @Akatama !

PR #4036, opened 2025-07-03
Added yamllint configuration for ansible-lint. Added ansible-lint to 10min-iiab-unittest-ubuntu.yml

Akatama:
First, properly configured yamllint to make ansible-lint a bit more configurable. Then I added an ansible-lint install step from and an ansible-lint run step for 10min-iiab-unittest-ubuntu.yml. We could make our our workflow just for ansible lint if we wanted, but it doesn't make sense to put it in multiple workflows - either ansible-lint find errors and then fails or it doesn't. We don't need multiple reports of the same information in that regard.

You may have noticed we did not use the ansible-lint GitHub Action. This is because the Action made by the ansible-lint team is a bit limiting in how it works. Specifically, it wants a requirements.txt to install Ansible and other dependencies, but we don't install our dependencies that way. It is just easier to install our dependencies our way, install ansible-lint with pip and then run ansible-lint. 

The yamllint configuration doesn't do much for us right this moment. My hope was to get a way to enforce specific truthy values as it seems we use the nonstandard True / False the most. However,  we also use true / false and yes / no. Its not the highest priority on my list to change these truthy values, but when it happens we can change the yamllint config to enforce the subset of them. Yamllint also allows us to enforce a specific max line length, currently I have set this to 1000 characters as we have one line that is > 900 characters. I suggested to Adam Holt that at some point we put a line length limit of 400 characters (still a lot mind you), and he seemed open to the idea.

Ubuntu 24.04, Debian 12 and RasPiOS

@holta

### DIFF
--- a/.ansible-lint.yml
+++ b/.ansible-lint.yml
@@ -24,7 +24,7 @@ loop_var_prefix: "^(__|{role}_)"
 # to skip_list.
 var_naming_pattern: "^[a-zA-Z_][a-zA-Z0-9_]*$"
 
-use_default_rules: true
+use_default_rules: True
 # Load custom rules from this specific folder
 # rulesdir:
 #   - ./rule/directory/
@@ -39,8 +39,6 @@ use_default_rules: true
 # still visible, making it easier to address later.
 skip_list:
   - yaml[comments] # ignore comments without leading space
-  - yaml[truthy] # ignore values that are True / False instead of true / false
-  - yaml[line-length] # 338 instances, but we probably don't care about this so much
   - name[missing] # 140 instances
   - name[template] # 354 instances
   - name[casing] # 47 instances
@@ -77,7 +75,7 @@ warn_list:
 write_list: ["none"]
 
 # Offline mode disables installation of requirements.yml and schema refreshing
-offline: true
+offline: True
 
 # Define required Ansible's variables to satisfy syntax check
 # extra_vars:
--- a/.github/workflows/10min-iiab-unittest-ubuntu.yml
+++ b/.github/workflows/10min-iiab-unittest-ubuntu.yml
@@ -50,10 +50,10 @@ jobs:
           # local_vars_unittest.yml installs a quite minimal IIAB, without IIAB Apps
           sudo cp /opt/iiab/iiab/vars/local_vars_unittest.yml /etc/iiab/local_vars.yml
       - run: sudo /opt/iiab/iiab/scripts/ansible    # Installs Ansible
+      - name: Install ansible-lint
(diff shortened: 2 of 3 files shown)


holta:
Looks great, mainly as a refactor for starters!

Question for @Akatama:

Do you know if this slows down 10min-iiab-unittest-ubuntu.yml measurably?  (Not the end of the world if it only adds a few seconds here & there for a purpose, but still worth acknowledging if true!)

Akatama:
@holta It is not a refactor. As I said in the title, I added a yamllint configuration, then I changed 10min-iiab-unittest-ubuntu.yml to run ansible-lint.

It is definitely making 10min-iiab-unittest-ubuntu.yml take longer. Previously, that pipeline took 3-4 minutes. We've only done 3 runs total with the pipeline at this point, but those runs were 8.5 minutes, 9.5 minutes and 11 minutes.

holta:
39 seconds doesn't seem too bad, if there's real value in this linting?!

image

( Screenshot from this PR's 10min-iiab-unittest-ubuntu.yml run at https://github.com/iiab/iiab/actions/runs/16053616793/job/45302283538?pr=4036 )

holta:
> 39 seconds doesn't seem too bad, if there's real value in this linting?!

Or, if in the end ansible-lint ends up taking too long, e.g. more than a minute or two?

We could then split off all linting tasks, into a completely separate GitHub Actions job?
(That GitHub Actions would run quickly / independently, i.e. in parallel!)

Akatama:
I didn't look at it like that. 

Yeah, the value in linting is that if a change was made to the Ansible config that didn't fit with our lint ruled then it would be caught here at the latest.

holta:
@Akatama anything we can/should learn from message `Profile 'moderate' was required, but 'production' profile passed` below?

image

(Output is from from this PR's 33-second GHA [GitHub Actions] 10min-iiab-unittest-ubuntu.yml job.)

Akatama:
It's actually incorrect. If you change the profile value from moderate to production you get a ton of lint errors. My best guess is that it is a bug, because if you run with the profile value, it will tell you something like: profile production failed, but profile moderate passed.

holta:
> It's actually incorrect. If you change the profile value from moderate to production you get a ton of lint errors. My best guess is that it is a bug, because if you run with the profile value, it will tell you something like: profile production failed, but profile moderate passed.

Weird!  But good enough for now.

Context below, just in case we come back to this question later, at some later date:

https://github.com/iiab/iiab/blob/c7b6dc47e5b9a399c677b7985ad0987ae9245b78/.ansible-lint.yml#L4

PR #4009, opened 2025-05-02
ansible-lint cleanup of yaml[trailing-spaces] & yaml[colons] thx to @Akatama

holta:
Just a quick tweak / refinement of @Akatama's work here:

- PR #4007

PR #4038, opened 2025-07-06
Fixed octal-values for ansible-lint

Akatama:
According to ansible-lint's documentation, this is what they claim "Using integers or octal values in YAML can result in unexpected behavior. For example, the YAML loader interprets `0644` as the decimal number `420` but putting `644` there will produce very different results."

This is because Yaml has done a lot of weird things with octal values in the past

Here is an example:
0644 -> "0644"

Ubuntu 24.04, Debian 12, RasPiOS on Zero 2 W

@holta

### DIFF
--- a/.ansible-lint.yml
+++ b/.ansible-lint.yml
@@ -48,7 +48,6 @@ skip_list:
   - role-name # 14 instances, talk to Adam
   - var-naming[no-role-prefix] # 65 instances, talk to Adam
   - key-order # 5 instances
-  - yaml[octal-values] # 44 instances
   - yaml[empty-lines] # 84 instances
   - yaml[trailing-spaces]
 
--- a/roles/2-common/tasks/iiab-startup.yml
+++ b/roles/2-common/tasks/iiab-startup.yml
@@ -7,7 +7,7 @@
   template:
     src: iiab-startup.sh
     dest: /usr/libexec/
-    mode: 0755
+    mode: "0755"
   when: not startup_script.stat.exists
 
 - name: Install {{ systemd_location }}/iiab-startup.service from template
--- a/roles/azuracast/tasks/install.yml
+++ b/roles/azuracast/tasks/install.yml
@@ -49,7 +49,7 @@
   get_url:
     url: "{{ docker_sh_url }}"
     dest: "{{ azuracast_host_dir }}/"
-    mode: 0755
+    mode: "0755"
     timeout: "{{ download_timeout }}"
 
 #- name: AzuraCast - Make changes to docker.sh script so it runs headless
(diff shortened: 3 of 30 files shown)


holta:
Thanks @Akatama !

While this has been debated to death over the years (repeatedly ;) one consistent recommendation definitely helps newcomers, so let's go with it. ✅ 

> According to ansible-lint's documentation, this is what they claim "Using integers or octal values in YAML can result in unexpected behavior. For example, the YAML loader interprets `0644` as the decimal number `420` but putting `644` there will produce very different results."

Specifically: https://ansible.readthedocs.io/projects/lint/rules/risky-octal/

holta:
This PR builds on:

- PR #4004
- PR #4007
- PR #4009
- PR #4026
- PR #4036

jvonau:
There are a bunch of stanzas using `with_items:` that contain `mode: "{{ item.mode }}" ` with the format
> name: '2025-01-25: Copy files from files/usbmount to filesystem'
  copy:
    src: "{{ item.src }}"
    dest: "{{ item.dest }}"
    mode: "{{ item.mode }}"
  with_items:
    - { src: 'usbmount/usbmount.conf', dest: '/etc/usbmount/', mode: '0644' }
    - { src: 'usbmount/usbmount', dest: '/usr/local/sbin/', mode: '0755' }

are those ok? Seems to be as the ' ' would be removed when the substitution into item.value is already quoted as " "

Funny thing years ago using quotes was the standard, that got changed for some reason. What was the reason again @holta?

holta:
> are those ok? Seems to be as the ' ' would be removed when the substitution into item.value is already quoted as " "

Interesting question.  @Akatama take a look, if you have a moment?

> Funny thing years ago using quotes was the standard, that got changed for some reason. What was the reason again @holta?

I can't remember at this point.  Certainly one norm across the board seems healthy at this point, so let's go with the official recommendation from Ansible where that works anyway.

Akatama:
@jvonau @holta 

> are those ok? Seems to be as the ' ' would be removed when the substitution into item.value is already quoted as " "

I am pretty sure you are correct here. 

The Ansible documentation is a bit hard to parse because instead of telling me how with_items works specifically, it just wants to tell me how I can convert my `with_items` keyword usage to `loop`, as they think its the best way to loop since Ansible 2.5. They do say that `with_items` is not deprecated, so it is a bit annoying that they won't give more details about a still usable keyword.

I did find this little guide: https://thelinuxcode.com/ansible-with_item/

They are doing it the same way we do, except they do not actually single-quote mode underneath with_items. I think the end result is the same either way. In our case, we are simply passing a value that is a yaml string into a string, similar to string interpolation in many programming languages.

ansible-lint did not provide any warning about the mode for any stanzas like the one you provided, and as such I did not change any of them. This means that the mode value that is provided is valid (e.g. it does not result in "'0644'" as that is not a valid yaml string - although Ansible has its own version of yaml anyway so maybe they allow triple quotes).

PR #4040, opened 2025-07-10
Fixed key-order[task] violations for ansible-lint

Akatama:
"This rule recommends reordering key names in ansible content to make code easier to maintain and less prone to errors."

All of them were Ansible block keys that were followed by when keys. ansible-lint has made the decision that when keys should be before block:

"Try to remember that in real life, block/rescue/always have the habit to grow due to the number of tasks they host inside, making them exceed what a single screen. This would move the when task further away from the rest of the task properties. A when from the last task inside the block can easily be confused as being at the block level, or the reverse. When tasks are moved from one location to another, there is a real risk of moving the block level when with it.

By putting the when before the block, we avoid that kind of risk. The same risk applies to any simple property at the task level, so that is why we concluded that the block keys must be the last ones."

https://ansible.readthedocs.io/projects/lint/rules/key-order/#reasoning

It seems to be working. I will note that 2 of the 5 violations came in the same file, and I couldn't fix it. Maybe because I am just not the most knowledgeable about Ansible. Currently, they are being ignored with the # noqa: key-order[task] comment, but if you can find a way to fix them for this rule that would be cool.

Also, we could decided we don't want to follow this lint rule at all. I think what they are saying makes sense though.

Ubuntu 24.04, Debian 12 and RasPiOS on Zero 2 W. This one may require some more testing.

@holta

### DIFF
--- a/roles/network/tasks/main.yml
+++ b/roles/network/tasks/main.yml
@@ -35,6 +35,7 @@
 
 
 - name: Configuring Network if enabled
+  when: network_installed is defined and network_enabled
   block:
 
   # 2022-07-22: Is './runrole --reinstall network' the new way to make this run?
@@ -102,8 +103,6 @@
     - name: Restart services (and clone wifi to create ap0 if necessary)
       include_tasks: restart.yml
 
-  # end block
-  when: network_installed is defined and network_enabled
 
 - name: Select RPi firmware mode
   include_role:
--- a/roles/network/tasks/main.yml
+++ b/roles/network/tasks/main.yml
@@ -103,6 +103,8 @@
     - name: Restart services (and clone wifi to create ap0 if necessary)
       include_tasks: restart.yml
 
+    # end block
+
 
 - name: Select RPi firmware mode
   include_role:
--- a/.ansible-lint.yml
+++ b/.ansible-lint.yml
@@ -47,7 +47,6 @@ skip_list:
   - command-instead-of-shell # 12 instances, specific
   - role-name # 14 instances, talk to Adam
   - var-naming[no-role-prefix] # 65 instances, talk to Adam
-  - key-order # 5 instances
   - yaml[empty-lines] # 84 instances
   - yaml[trailing-spaces]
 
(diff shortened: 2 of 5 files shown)


holta:
> 2 of the 5 violations came in the same file, and I couldn't fix

Great question: I don't yet understand it!

Definitely Ansible indentation/syntax can become counter-intuitive (which is putting it politely ;)

Akatama:
@holta The 20 min IIAB CI failed on my branch and this PR, but when I re-ran I passed. In both cases, this was the error:

```
Warning: : Failed to update cache after 1 due to E:Failed to fetch
http://ports.ubuntu.com/ubuntu-ports/dists/focal-security/main/binary-
amd64/Packages  404  Not Found [IP: 91.189.91.104 80], E:Some index files
failed to download. They have been ignored, or old ones used instead. retry,
retrying
```

That seems like a networking error with apt. Can we try a re-run?

holta:
>  That seems like a networking error with apt. Can we try a re-run?

Re-running now!

holta:
Failed again:

I should try again Friday &/or weekend to see if it's a transient hosting / mirroring issue that goes away soon?

holta:
Another attempt begun.

holta:
> Another attempt begun.

Ah well I'll try again in coming days!

Akatama:
@holta I think I need to take another look. I am starting to think ansible-lint improperly caught all of these as key order errors but they aren't.

Akatama:
@holta take another look when you get a chance. Thanks!

roles/network/tasks/main.yml
```diff
@@ -102,8 +103,6 @@
     - name: Restart services (and clone wifi to create ap0 if necessary)
       include_tasks: restart.yml
 
-  # end block
```
holta, line 105: Can we keep the original Line 105, at minimum?

(The "# end block" visual indicator can be priceless!  Moving the `when:` clause to the top of a block is counterintuitive, but if it's tested and Ansible strongly recommends this new way of doing things, ok we can try it!)

Akatama: Done. However, if you don't want use to follow this lint rule, we can just keep it in the skip_list. You're the maintainer, so I will follow whichever decision you make.

holta:
Let's give this a shot:

- I'll go ahead & merge.

- @EMG70 can you confirm in coming days that all looks good?

EMG70:
IIAB Size 1 Installed on a PI400. Hotspot working OK.
sudo iiab- diagnostics https://paste.centos.org/view/a7e38f3c

holta:
Thanks @EMG70 and @Akatama !

PR #4035, opened 2025-07-02
Desupport / Deprecate Ubuntu 24.10, as it's EOL (end-of-life) 2025-07-10

holta:
Related:

- PR #3045
- PR #3416
- PR #3672 
- PR #3758

PR #4044, opened 2025-07-17
Remove 1 instance of is_ubuntu_2410

holta:
One more legacy var needing complete removal, just like:

- PR #4043 

Related:

- PR #4035 
- #4042


holta:
Tested with a LARGE-sized IIAB on the latest Ubuntu 25.10 pre-release.

PR #4056, opened 2025-08-19
logic to add MW Visual Editor by default [MediaWiki]

tim-moody:
set true in defaults

conditionally add to LocalSettings

### DIFF
--- a/roles/mediawiki/tasks/install.yml
+++ b/roles/mediawiki/tasks/install.yml
@@ -19,7 +19,7 @@
 
 
 # https://www.mediawiki.org/wiki/Manual:Installation_requirements#PHP
-- name: 'Install packages: php{{ php_version }}-intl, php{{ php_version }}-mbstring, php{{ php_version }}-xml'
+- name: 'Install packages: php{{ php_version }}-intl, php{{ php_version }}-mbstring, php{{ php_version }}-xml, php{{ php_version }}-curl'
   package:
     name:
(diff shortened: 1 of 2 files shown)


holta:
@tim-moody is this PR tested, all set to be merged?

ASIDE: Do end users need any guidance to enable/use this?  (Or are we all set!)

tim-moody:
The 'when' syntax was proposed by copilot and I haven't tested. I was hoping you would validate it.

Visual Editor has been the default for mediawiki for a number of years, so should probably be ours too. How to use it falls under general knowledge of mediawiki.

holta:
@EMG70 can you test this in the coming 24h if possible?

```
curl iiab.io/install.txt | bash -s 4056
```

If so a "zero-sized" IIAB install is fine (unit test, installing MediaWiki and no others IIAB apps!

holta:
Smoke-tested on Ubuntu 24.04 and http://box/wiki seems to work fine:

image

PS the `when` syntax (Ansible casting to boolean) is basically gratuitous, so let's drop that.

holta:
@EMG70 I think we're all set!

(A functionality spot check in coming days is of course useful, after this is merged, if you can ;)

holta:
fyi this was retested (smoke test + basic functionality test) for good measure, on another Ubuntu 24.04 VM.

tim-moody:
>PS the when syntax (Ansible casting to boolean) is basically gratuitous, so let's drop that.

that's what I thought but I left it because copilot did it. I figured you would sort it out.

The expected function of seeing both edit and edit source in the tabs bar is an indication it worked. To fully test one would need to edit a page and see if it saved. But did that during unit testing.

Issue #4058, opened 2025-08-22
RasPiOS trixie testing


jvonau:
Test builds of RasPiOS trixie available to try 
https://downloads.raspberrypi.com/nightlies/

holta:
So far testing of on Debian 13 "Trixie" and pre-releases of RasPiOS 13 "Trixie" have gone better than expected: (FWIW!)

- #3736
- #3756

PS I'd expect some rough spots with wheels, i.e. piwheels, but (so far!) have not seen any, just yet anyway:

- https://github.com/piwheels/packages/issues/210#issuecomment-3036781727

cc: thank you @EMG70 for all the testing so far, in recent months!  (@EMG70 if you can try to download and install the latest .xz from https://downloads.raspberrypi.com/nightlies/ onto your fastest Raspberry Pi, that'd be really great in coming days?)  e.g. trying your luck with a LARGE-sized IIAB install test if you have time?!

avni:
Successfully installed 0-size IIAB on 8/23 nightly Trixie img! 🎉 

[update] iiab-diagnostics: https://paste.centos.org/view/06ced74d

jvonau:
> tailscale_enabled = False
tcp        0      0 192.168.0.222:38006     52.207.202.187:443      ESTABLISHED 817/tailscaled   

Interesting something is calling home.

>Aug 23 17:47:12 box NetworkManager[753]: <info>  [1755967632.0296] rfkill1: found Wi-Fi radio killswitch (at /sys/devices/platform/axi/1001100000.mmc/mmc_host/mmc1/mmc1:0001/mmc1:0001:1/ieee80211/phy0/rfkill1) (driver brcmfmac)
Aug 23 17:47:12 box NetworkManager[753]: <info>  [1755967632.0298] manager[0x55563885c150]: rfkill: Wi-Fi hardware radio set disabled
=IIAB==========================================================================
COMMAND: /usr/sbin/rfkill list    # Are WiFi and Bluetooth interfaces blocked?
 0: hci0: Bluetooth
        Soft blocked: no
        Hard blocked: no
1: phy0: Wireless LAN
        Soft blocked: yes
        Hard blocked: no
 =IIAB==========================================================================
COMMAND: /usr/sbin/iw reg get    # Detected WiFi country code / legal frequencies
 global
country 00: DFS-UNSET
        (755 - 928 @ 2), (N/A, 20), (N/A), PASSIVE-SCAN
[...]

PR #4062, opened 2025-08-26
Begin Debian transition (Bookworm -> Trixie -> Forky) + update 4 CI workflows [& discussion of KA Lite -> Kolibri content downloading issues]

holta:
WARNING: IIAB will begin transitioning away from Debian 12 "Bookworm" very shortly now, to Debian 13 "Trixie" — and also with early prep in support of Debian 14 "Forky" pre-releases.

In keeping with IIAB's longstanding policy to work primarily with recent OS's, making that transition quickly after 15 days have passed since the release of the latest LTS OS (in this case Debian 13 Trixie was released on 2025-08-09).

In this case, partial support for Debian 12 "Bookworm" will be preserved for a few weeks or so, e.g. for those who want to manually force support in [/opt/iiab/iiab/scripts/local_facts.fact](https://github.com/iiab/iiab/blob/master/scripts/local_facts.fact)

Somewhat Related:

- #3034 
- #3045
- #3691
- #4058

### DIFF
--- a/.github/workflows/10min-iiab-unittest-ubuntu.yml
+++ b/.github/workflows/10min-iiab-unittest-ubuntu.yml
@@ -27,7 +27,7 @@ jobs:
       #    GITHUB_CONTEXT: ${{ toJSON(github) }}
       #  run: echo "$GITHUB_CONTEXT"
       - name: Check out repository code    # Clones repo onto github runner
-        uses: actions/checkout@v4
+        uses: actions/checkout@v5
         with:
           fetch-depth: 0    # Default is 1, but iiab-summary (below) needs git tag history.  FYI "fetch-depth: 0" works but is way too greedy, downloading all branches.  The ideal would've been "fetch-depth: 10000" with "fetch-tags: true" -- but this remains broken as of actions/checkout@v4 on 2025-03-26: see issues #217, #338, PR #1396, #1467, #1471, #1662 in https://github.com/actions/checkout -- Dec 2020 background: https://github.blog/open-source/git/get-up-to-speed-with-partial-clone-and-shallow-clone/
       - run: echo "🍏 This job's status [SO FAR!] is ${{ job.status }}."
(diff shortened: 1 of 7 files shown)


holta:
@Akatama @thotmx I'm going to go ahead and merge this, but let me know if you see any useful CI revisions that should be added down the road?!

PS it does seem likely that [40min-iiab-none-deb12-on-rpi3.yml](https://github.com/iiab/iiab/blob/master/.github/workflows/40min-iiab-none-deb12-on-rpi3.yml) will very shortly need to be shunted aside to `40min-iiab-none-deb12-on-rpi3.yml.paused` (in another PR, soon) — as IIAB will no longer support Debian 12 very shortly.

And then (hopefully in not too many months?) https://github.com/pguyot/arm-runner-action can hopefully begin to support debian-13 and similar — allowing "Raspberry Pi emulation" workflows to be restored on GitHub Actions, with relevant + modern OS's most importantly (e.g. Debian 13 Trixie, RasPiOS 13 Trixie, etc). 🙏

jvonau:
Would be a good time to fully disable kalite:
From 668c3243
>iiab-apps-to-be-installed :
kalite
kalite_install: True
kalite_enabled: True

Think python2 has fully gone away, apt should tell you if the package is available and if the below lines need updating should install_python2.sh still works on trixie
https://github.com/iiab/iiab/blob/master/roles/kalite/tasks/install.yml#L18-L50

holta:
> Would be a good time to fully disable kalite:

Yes.  With RasPiOS Trixie likely becoming official in coming weeks, this is the right time.  To steer any remaining implementer communities towards Kolibri instead.

tim-moody:
>likely becoming official in coming weeks, this is the right time. To steer any remaining implementer communities towards Kolibri instead.

I am doing a download of 8G of the total 61G for Kolibri Khan EN. It as been more than 12 hours and it is only half way done. This is not viable. This is the same connection that downloaded a 42G zim in 8 min.

holta:
Just FYI I pinged Kolibri maintainers:

- learningequality/kolibri#13680

(To see if they know what's happening — and/or can at least acknowledge that they are working on it?)

PR #4069, opened 2025-08-29
don't exit from imported library

tim-moody:
needs corresponding change in iiab-admin-console adm_lib.py


holta:
> needs corresponding change in iiab-admin-console adm_lib.py

Should this be merged ASAP?

PR #4071, opened 2025-08-29
Updating 2-common for etc_path variable

hal-eisen-adfa:

PR #4057, opened 2025-08-21
Add etc_path to default_vars

hal-eisen-adfa:

PR #4072, opened 2025-08-30
Switching to etc_path and iiab_etc_path instead of hard coded paths in role/iiab-admin

hal-eisen-adfa:


holta:
@hal-eisen-adfa I'd be in favor of preserving `/etc` in most (if not all!) comments, READMEs, docs — can we consider something like that?

Related:

- PR #4057 
- PR #4070 
- PR #4071

holta:
> @hal-eisen-adfa I'd be in favor of preserving `/etc` in most (in not all!) comments, READMEs, docs — can we consider something like that?

Specifically:

Is it possible that Line 25 of roles/iiab-admin/tasks/sudo-prereqs.yml might be... the _only_ change really needed by this PR ?!

hal-eisen-adfa:
> > @hal-eisen-adfa I'd be in favor of preserving `/etc` in most (in not all!) comments, READMEs, docs — can we consider something like that?
> 
> Specifically:
> 
> Is it possible that Line 25 of roles/iiab-admin/tasks/sudo-prereqs.yml might be... the _only_ change really needed by this PR ?!

Yeah, I was wondering about that. Obviously, I went the other way but I'm happy to change. I was mostly on the fence about it.

hal-eisen-adfa:
Changing strategy to leave comments unchanged

Issue #4030, opened 2025-06-30
ansible-core 2.19+ strict new security model requires revising validate_vars.yml etc — be aware of Ansible's new filter `trust_as_template`


holta:
Might be a good time to try this, now that its release candidate is available:

- https://github.com/ansible/ansible/releases/tag/v2.19.0rc1

A great way to try this out on a clean OS, is to run:

0. sudo -i
1. mkdir /opt/iiab
2. cd /opt/iiab
3. git clone https://github.com/iiab/iiab
4. cd iiab
6. Modify the **right side of** [Line 287 of scripts/ansible](https://github.com/iiab/iiab/blob/c78f65a0f2c5344012f97c308ec7f674b4626ec7/scripts/ansible#L287) to resemble [`pip install ansible-core==2.19.0rc1`](https://pypi.org/project/ansible-core/2.19.0rc1/)
7. Run `curl iiab.io/install.txt | bash` as usual!

FYI ansible-core 2.19 was twice delayed.  Its release date ETA is currently 2025-07-21.

holta:
1. FYI an initial IIAB test install failed due to ansible-core 2.19 RC 1 having [much more strict templating](https://ansible.readthedocs.io/projects/ansible-core/devel/porting_guides/porting_guide_core_2.19.html#multi-pass-templating).

2. And their [workaround](https://ansible.readthedocs.io/projects/ansible-core/devel/porting_guides/porting_guide_core_2.19.html#troubleshooting-untrusted-templates) is erroneously documented, e.g. if you run...

`export _ANSIBLE_TEMPLAR_UNTRUSTED_TEMPLATE_BEHAVIOR=warn`

...it results in:

```
ERROR: Invalid value 'warn' for config '_TEMPLAR_UNTRUSTED_TEMPLATE_BEHAVIOR'. Valid values are: error, warning, ignore

Traceback (most recent call last):
  File "/usr/local/ansible/lib/python3.13/site-packages/ansible/cli/__init__.py", line 91, in <module>
    from ansible import constants as C
  File "/usr/local/ansible/lib/python3.13/site-packages/ansible/constants.py", line 188, in <module>
    set_constant(setting, config.get_config_value(setting, variables=vars()))
                          ~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/ansible/lib/python3.13/site-packages/ansible/config/manager.py", line 553, in get_config_value
[...]

PR #4081, opened 2025-09-10
Maps dot black [EXPERIMENTAL!]

orblivion:
Just a demo for now! Just to see what's possible and how easy it is. **This will download a ton of data if you run this** at least during development.

Got a maps_dot_black role working. https://maps.black indeed does make this very simple.

Downloads assets (js and map styles) from maps.black. Gets two data files (pmtiles) totaling under 150 MB for a simple base map. I could get a 70GB file for a full map but I don't want people to wait around all day to try it out (I haven't tried it yet myself). But I assume that it would work.

Run the role and visit /maps-dot-black on your box.

To keep people aware of my work without triggering CI all the time I'll push intermittent work here: https://github.com/iiab/iiab/compare/master...orblivion:iiab:maps-dot-black-wip

### DIFF
--- a/roles/7-edu-apps/tasks/main.yml
+++ b/roles/7-edu-apps/tasks/main.yml
@@ -29,6 +29,11 @@
     name: osm-vector-maps
   when: osm_vector_maps_install
 
+- name: MAPS
+  include_role:
+    name: maps
+  when: maps_install
+
 # UNMAINTAINED
 - name: OSM
   include_role:
--- a/roles/maps/build-databases/README
+++ b/roles/maps/build-databases/README
@@ -1,0 +1,1 @@
+This should not be in an ansible role. This should not be merged into master. It belongs elsewhere eventually I'm just putting it here for convenience.
--- a/roles/maps/build-databases/nominatim/iiab.lua
+++ b/roles/maps/build-databases/nominatim/iiab.lua
@@ -1,0 +1,32 @@
+
+local flex = require('flex-base')
+
+
+
+
+
+
+
+flex.set_main_tags('admin') -- or 'all_boundaries' if it's not too big? if it's more useful educational info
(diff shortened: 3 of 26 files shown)


holta:
@orblivion maybe lint the 6 ansible-lint suggestions from our CI (GitHub Actions) visible here... ?

- https://github.com/iiab/iiab/pull/4081/files
- https://github.com/iiab/iiab/actions/runs/17658790428/job/50187823942?pr=4081

...if you have time!  Optionally also click "Still in progress? Convert to draft" on the top-right of page https://github.com/iiab/iiab/pull/4081 depending on your intent & where you think you might take this! 🗺️

orblivion:
Any reason not to wait until later to lint? I'm sure I'll make more mistakes.

holta:
> Any reason not to wait until later to lint? I'm sure I'll make more mistakes.

Heh, and even more so now that we all have to "relearn Ansible" — given that ansible-core 2.19+ is shifting the entire foundation underneath us 😆 

- #4030

orblivion:
I want to push intermittent work so people can see my progress, but I'm not at the stage where fixing CI errors makes sense. I'd be fixing syntax errors on code that could get thrown away anyway.

So I made a separate branch: https://github.com/iiab/iiab/compare/master...orblivion:iiab:maps-dot-black-wip

orblivion:
As of now the Nominatim thing installs. However the systemd setup, which uses gunicorn and nginx, doesn't work. Nominatim does (sometimes?) work if you just run it manually as the nominatim user, on port 10000. The front end is set up to use that.

But anyway that was all mostly a proof of concept, I'll get to that a little later. Gonna focus on generating a good search database.

Same with Photon. I won't have both installed in the end, just testing them out.

Issue #3924, opened 2025-01-28
[Upload2USB] Writing N times to /tmp as part of file upload (before copying to USB stick) damages SD card


avni:
HTML form file upload is writing to /tmp and then copying to final destination on USB stick. This will damage SD cards (on Pis) over time with the numerous writes. 

PHP docs indicate that we can change tmp dir to another location: https://www.php.net/manual/en/ini.core.php#ini.upload-tmp-dir

**_From post by @holta in [#3921](https://github.com/iiab/iiab/issues/3921#issuecomment-2617479517)_**

> Uploading into /tmp is also a bug (causing permanent damage to IIAB microSD cards, when the real destination that matters is the USB stick!)  But perhaps we should open a separate ticket to solve that "architectural flaw" ?
> 
> Tangentially related:
> - https://github.com/iiab/iiab/issues/3403#issuecomment-2615004263 

**RELATED** 
- #3875 
- #3923
- #3919 
- #3915

jvonau:
Might just want to set the tmp dir to a scratch directory on the usbkey

In the getTargetUSBDriveLocation function is really too much calling lsblk, all you really just need is to scan /library/www/html/local_content/ to get USBX to filter out the usb boot root directory as that filtering is already done in roles/usb_lib/templates/mount.d/70-usb-library.j2

Found this for you https://stackoverflow.com/questions/12233859/php-get-path-to-every-file-in-folder-subfolder-into-array maybe just use 

function dir_scan(/library/www/html/local_content/)

tim-moody:
>Might just want to set the tmp dir to a scratch directory on the usbkey

good idea, and makes a mv easier.

also we have run out of space on /tmp on some occasions.

holta:
@avni feel free to "unassign" yourself.  If you don't have time to work on this ticket.  Thanks!

jvonau:
If you are interested in reducing the writes on the sdcards then /tmp could be a tmpfs but that limits the size to the amount of ram available. That change would have to be very early in the boot process but I just tested /etc/fstab
> ls /tmp
admin-console.sock
systemd-private-ceb23b05cae143da8a6a0bb12585db4d-ModemManager.service-j2AXHi
systemd-private-ceb23b05cae143da8a6a0bb12585db4d-systemd-logind.service-dmnxYr
systemd-private-ceb23b05cae143da8a6a0bb12585db4d-systemd-resolved.service-9uNDx6
[...]

PR #3944, opened 2025-02-15
usbmount again

jvonau:
- remove tab spacing
- restore indents for formatting
- shellcheck
- restore usbmount logging
- restore usbmount remove functionality

Working to restore umount.d to a working state with an eye to call iiab-clean-usb.sh from umount.d in place of the systemd unit file for usbmount.

PR #4086, opened 2025-09-17
update temp_dir php_value for upload2usb

chapmanjacobd:
### Fixes bug:

https://github.com/iiab/iiab/issues/3924

`/library/www/html/local_content` seems like it might be a safe place to put the temp folder. If the drive is mounted this should link to the `/media/usb?` location.

1. I'm not sure if nginx will follow the symbolic link but I assume since FancyIndex is working then `upload_tmp_dir` will work
2. I'm not sure if [move_uploaded_file](https://www.php.net/manual/en/function.move-uploaded-file.php) will like if `from` and `to` are the same file. (If the destination file already exists, it will be overwritten--but this was already the case without this PR). I'm not sure what PHP will name the temp files... But since upload2usb is always moving to a sub-folder as a final destination anyways this is possibly a non-problem.
3. I've avoided having a temp sub-folder (eg. `USB:/upload2usb_temp/`) because it adds (unnecessary [hopefully]) clutter. Though it would be better to follow the POSIX atomic steps (write to a temporary file, synchronize the data and directory to disk, and then atomically rename the temporary file to replace the original, etc). PHP might be handling this for us. I'm not sure yet.
4. It also depends _when_ PHP is looking up the local_content folder--the process might be "getting the inode number" too early for it to know about the re-link to /media, etc

This is totally untested

@holta

### DIFF
--- a/roles/2-common/tasks/fl.yml
+++ b/roles/2-common/tasks/fl.yml
@@ -24,6 +24,7 @@
     - "{{ iiab_zim_path }}/content"    # /library/zims
     - "{{ iiab_zim_path }}/index"
     - "{{ doc_root }}/local_content"    # /library/www/html
+    - "{{ doc_root }}/local_content/USB0"
     - "{{ doc_root }}/modules"
     - "{{ doc_root }}/common/css"
     - "{{ doc_root }}/common/js"
--- a/roles/nginx/templates/iiab.conf.j2
+++ b/roles/nginx/templates/iiab.conf.j2
@@ -2,10 +2,6 @@ location / {
     rewrite ^/$ $1{{ iiab_home_url }};
 }
 
-upstream php-usb0 {
-    server unix:/run/php/php{{ php_version }}-fpm-usb0.sock;
-}
-
 location /usb {
     alias /library/www/html/local_content/;
     fancyindex on;    # autoindex on;
--- a/roles/nginx/templates/iiab.conf.j2
+++ b/roles/nginx/templates/iiab.conf.j2
@@ -13,12 +13,11 @@ location ~ ^/upload2usb/(.*)\.php$ {
     proxy_set_header X-Real-IP  $remote_addr;
     proxy_set_header X-Forwarded-For $remote_addr;
     proxy_set_header Host $host;
-    fastcgi_pass php;
+    fastcgi_pass unix:/run/php/php{{ php_version }}-fpm-local_content.sock;
     fastcgi_index index.php;
     fastcgi_split_path_info ^(.+\.php)(/.+)$;
     fastcgi_param   SCRIPT_FILENAME    $document_root$fastcgi_script_name;
     fastcgi_param   SCRIPT_NAME        $fastcgi_script_name;
-    fastcgi_param   PHP_VALUE          "upload_tmp_dir=/library/www/html/local_content";
     include fastcgi_params;
 }
 
--- a/roles/nginx/templates/iiab.conf.j2
+++ b/roles/nginx/templates/iiab.conf.j2
@@ -18,6 +18,7 @@ location ~ ^/upload2usb/(.*)\.php$ {
     fastcgi_split_path_info ^(.+\.php)(/.+)$;
     fastcgi_param   SCRIPT_FILENAME    $document_root$fastcgi_script_name;
     fastcgi_param   SCRIPT_NAME        $fastcgi_script_name;
+    fastcgi_param   PHP_VALUE          "upload_tmp_dir=/library/www/html/local_content";
     include fastcgi_params;
 }
 
--- a/roles/nginx/templates/iiab.conf.j2
+++ b/roles/nginx/templates/iiab.conf.j2
@@ -2,6 +2,10 @@ location / {
     rewrite ^/$ $1{{ iiab_home_url }};
 }
 
+upstream php-usb0 {
+    server unix:/run/php/php{{ php_version }}-fpm-usb0.sock;
+}
+
 location /usb {
     alias /library/www/html/local_content/;
     fancyindex on;    # autoindex on;
@@ -13,7 +17,7 @@ location ~ ^/upload2usb/(.*)\.php$ {
     proxy_set_header X-Real-IP  $remote_addr;
     proxy_set_header X-Forwarded-For $remote_addr;
     proxy_set_header Host $host;
-    fastcgi_pass unix:/run/php/php{{ php_version }}-fpm-local_content.sock;
+    fastcgi_pass php-usb0;
     fastcgi_index index.php;
     fastcgi_split_path_info ^(.+\.php)(/.+)$;
     fastcgi_param   SCRIPT_FILENAME    $document_root$fastcgi_script_name;
--- a/roles/nginx/templates/nginx.conf.j2
+++ b/roles/nginx/templates/nginx.conf.j2
@@ -104,5 +104,10 @@ http {
     upstream php {
         server unix:/run/php/php{{ php_version }}-fpm.sock;
     }
-           
+
+    # define the upstream backend fastcgi for Upload2USB
+    upstream php-usb0 {
+        server unix:/run/php/php{{ php_version }}-fpm-usb0.sock;
+    }
+
 }
--- a/roles/usb_lib/tasks/nginx.yml
+++ b/roles/usb_lib/tasks/nginx.yml
@@ -30,6 +30,16 @@
 #     state: absent
 #   when: not usb_lib_enabled
 
+- name: Install /etc/php/{{ php_version }}/fpm/pool.d/USB0.conf
+  copy:
+    src: roles/usb_lib/templates/USB0.conf
+    dest: /etc/php/{{ php_version }}/fpm/pool.d/USB0.conf
+
+- name: Restart 'php-fpm' systemd service
+  systemd:
+    name: php{{ php_version }}-fpm
+    state: restarted
+
 - name: Restart 'nginx' systemd service
   systemd:
     name: nginx
--- a/roles/usb_lib/tasks/nginx.yml
+++ b/roles/usb_lib/tasks/nginx.yml
@@ -30,10 +30,10 @@
 #     state: absent
 #   when: not usb_lib_enabled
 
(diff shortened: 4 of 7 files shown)


holta:
@avni would you be able to test this immediately, e.g. overnight if you have time?

```
curl iiab.io/install.txt | bash -s 4086
```

holta:
@Ark74 would you be able to look over this PR, and also... ?

- PR #4087

avni (commented):
Thank you @chapmanjacobd! 

/library/www/html/local_content still refers to the microSD card. The USB drive is mounted to a subfolder within the local_content folder - see screenshot below. 

20250918_00h55m04s_grim

I went ahead and tested this change because I think it's on the right track - we need to figure out how to pass INI parameters through the NGINX config. I modified `/etc/nginx/server.conf` to add the edit in this PR and then ran `systemctl restart nginx`. systemctl status nginx, shows the error: `"Passing INI directive through FASTCGI: unable to set '"upload_tmp_dir'" while reading response header from upstream`...

IIAB Diagnostics: https://paste.centos.org/view/a28e1a05

chapmanjacobd:
> '"upload_tmp_dir'"

That's weird that ' and " are going in the same direction at the start and end...

> '¨upload_tmp_dir'"

In the CentOS Pastebin it even shows this... putting that in a UTF-8 tool it doesn't seem like an ASCII character:

```
'
¨ (DIAERESIS, U+168, 0xc2a8)   <----
upload_tmp_dir'"⏎
```

probably a red herring but strange!

Will dig into it more tomorrow

chapmanjacobd:
Looking more into [the documentation](https://www.php.net/manual/en/ini.core.php#ini.sect.file-uploads) this doesn't seem possible actually. `upload_tmp_dir` is an `INI_SYSTEM` setting so it can only be applied globally. :/

avni:
> That's weird that ' and " are going in the same direction at the start and end...
> `'¨upload_tmp_dir'"`

Yes, saw that too and meant to mention. I removed the quotes and tried again but the same error was in the log. Something odd going on there - I'll test again and see if I can get that to go away!

holta:
➡️ Huge thanks to everyone pressing forward here, on this infrastructural dilemma from almost a year ago! ⬅️

chapmanjacobd:
I added a pool instead which should allow setting the INI_SYSTEM value

chapmanjacobd:
PHP might complain that `/library/www/html/local_content/USB0` doesn't exist if it is starting without a USB plugged in. In that case we'll need to adjust `roles/usb_lib/templates/mount.d/70-usb-library.j2` and `roles/usb_lib/templates/iiab-clean-usb.sh` to create the folder by default, rmdir before `ln` and when the USB is unplugged create it again after cleaning up

chapmanjacobd:
If the unit tests pass now then we may still need to add a "restart php-fpm" in both `mount.d/70-usb-library.j2` and `iiab-clean-usb.sh`; if the unit tests fail then it could very well also be a syntax error in the php pool config that I wrote

roles/2-common/tasks/fl.yml
```diff
@@ -24,6 +24,7 @@
     - "{{ iiab_zim_path }}/content"    # /library/zims
     - "{{ iiab_zim_path }}/index"
     - "{{ doc_root }}/local_content"    # /library/www/html
+    - "{{ doc_root }}/local_content/USB0"
```
avni, line 27: It's not guaranteed that the USB stick is mounted at USB0. I'll go back and check the logic that determines the name.

chapmanjacobd: Yeah, I suppose we could dynamically modify all the different places where we need to reference it.... If Upload2USB only supports one USB then it might be a bit more sustainable to isolate that bit of code that determines which USB "wins" into only one or two places.

If Upload2USB supports more than one USB then ideally the upload_tmp_dir would exist on each USB... if that is the case then we need to create a new php-fpm pool for each USB (we might need to follow this pattern or something similar anyway! (for the reason that you said--the name is not known ahead of time)--even if we support just one USB)

but let's check that the idea of having PHP upload the temp file to a USB _works at all_ first and once we know it works with USB0 (or USB1 or whatever exists) before we fine-tune this

avni: > let's check that the idea of having PHP upload the temp file to a USB works at all first and once we know it works with USB0 (or USB1 or whatever exists) before we fine-tune this

+1

holta:
@Ark74 solving #3924 in a clean/simple way is not easy (due to PHP it appears!)  So talking to Ruben might be our next step? (:

jvonau:
https://www.macs.hw.ac.uk/~hwloidl/docs/PHP/features.commandline.html

Wouldn't passing '-d upload_tmp_dir=/path' to php be enough to override the php.ini file?

jvonau:
Related: #3944

Ark74:
using sysstat to monitor the devices usage with:

```
USB=$(lsblk -no PKNAME "$(findmnt -no SOURCE /media/usb0)") && iostat -dxm "$USB" mmcblk0 1
```

You can see there is minimum to none writing to the SD card.
sysstat

chapmanjacobd:
> minimum to none writing to the SD card

this is with this PR or without it?

iostat might not be the best tool to use for this--but I don't know what else you could use. maybe [fatrace](https://github.com/martinpitt/fatrace)? I'm not a [bpftrace](https://github.com/bpftrace/bpftrace?tab=readme-ov-file#probe-types) expert but that would likely do quite well

holta:
@Ark74 suggests trying to upload a "student video" that's larger than ~513 MB — using a 1GB RAM Raspberry Pi 3, running RasPiOS 13 "Trixie" (which uses RAM from tmpfs) — to see if this works at all! 🏹

jvonau:
I have a couple of suggestions but for another day.

Ark74:
> this is with this PR or without it?

of course using with this PR.

holta:
@Ark74 did some really good work today trying things like Upload2USB on an old RPi 3 — quick summary here:

> Sadly, I need to report that rpi3 can't handle 1.3GB the upload breaks and due the lack of enough memory, it does write on the SD, [PR #4086] could use some fancy php and nginx to try to prevent that but requires some hours into testing just to try to [stretch] the limitations on the rpi3

> so I wonder if that's some actual goal, and what tradeoffs/limits do you find acceptable for low resource devices like the rpi3.

➡️ RPi 3's are excruciatingly slow due to internal USB 2.0 — so I've put an RPi 4 with USB 3.0 online at 100.64.0.30 (also 1GB RAM)  in case he/we need that to probe further?

holta:
> RPi 3's are excruciatingly slow due to internal USB 2.0

@Ark74 confirmed that uploading large files (e.g. 1GB) while slow, can actually work to RPi 3 — FYI he tested with this PR #4086 installed — and with "nginx + php set to 2 GB" :

- 629MB - 38min
- 786MB - 45min
- 900MB - 52min
- 1073MB - 59min

Ark74:
A couple of note on my findings.

I've noticed that this MR have one important proposal for low resource devices.  
The default PHP _http file upload_ require a tmp dir to allocate the whole file before writing to the actual path, in new releases /tmp is now tmpfs which is good for small files, but also for the SD card lifespan.

While the second option is to use the PHP _input stream_ it may get a bit more complex and increase the error factor.
So, the nice thing about this MR is that sets `php_admin_value[upload_tmp_dir]` to the actual USB drive, that along some other settings allow to upload really big files on it, only limited by the USB drive capacity.

On a vanilla install you can add the following settings to be able to upload big files, even with a rpi3 with only 1GB of RAM shared with /tmp (512MB tmpfs). I tested with 25.04 server, so you don't have to deal with the double mounting issue.

## nginx - /etc/nginx/conf.d/iiab.conf

```
....
fastcgi_request_buffering off;
client_max_body_size      2g;
client_body_buffer_size   512k
client_body_timeout       3600
fastcgi_read_timeout      3600
send_timeout              3600
}
```

## php - /etc/php/8.4/fpm/pool.d/www.conf

```
php_admin_value[upload_tmp_dir]     = /library/www/html/local_content/USB0
php_admin_value[sys_temp_dir]       = /library/www/html/local_content/USB0
php_admin_value[upload_max_filesize]= 2G
php_admin_value[post_max_size]      = 2G
php_admin_value[max_input_time]     = 3600
php_admin_value[max_execution_time] = 3600
```

I used 2GB but you should be able to set it to some crazy size that can allocate the uploaded file.

For practical use case it is required to think that rpi3 has an average of ~17MB/minute transfer rate due to USB2.0;

- 629MB - 38min - 16.55 MB/min
- 786MB - 45min - 17.46 MB/min
- 900MB - 52min - 17.30 MB/min
- 1073MB - 59min - 18.18 MB/min

So trying to move multiple large files on a class with many students on such device might not be advised.  
I suppose small devices could use the usb drive instead of ram, an rpi4 improves the speed to ~28 MB/min, but I suppose using a device with more resources is better to rely on ram when possible.   

So maybe having a max size number in mind, will help identify the devices that won't be able to handle it using ram and then use this approach instead, just an idea.

Let me know if there is anything else that might require details.   
Regards.

chapmanjacobd:
Thank you Luis! The testing that you've done here is invaluable.

I think it will be too flaky to use the USB mountpoint as the PHP temp dir. Especially if we add mount/unmount operations within the page itself via `pmount`! 

But even if we don't do that (and even if the double mounting didn't exist)--the code in this PR isn't robust for devices (auto)mounted in different locations from /media/usb0. PHP just isn't designed for dynamic tmp dir.

I imagine that 1 hour of uploading is already at the top end of the spectrum for how long people are willing to wait so we don't need to target files bigger than 1GB (by the max data rate).

I think the right move is to replace this PR with just the improvement on the existing PHP settings to allow for larger uploads which Luis has identified

holta:
Extraordinary!

- Thank you @Ark74 for this intensive ground truthing.

- And @chapmanjacobd for making it happen; contextualizing our likely next steps.

chapmanjacobd:
feel free to re-open if there is a need for this. With tmpfs /tmp and zram becoming default more and more, this seems less needed

Issue #4066, opened 2025-08-28
[Upload2USB] USB Stick Double Mounting - user gets permission denied error when uploading file on Ubuntu Desktop 24.04 and 25.10, Mint 22.2, and Debian 13 Desktop


avni:
From #4063 
> Can you expand on the 'double mounting' aspect of the report please. Is the Gui mounting the usb drive also?

The GUI is mounting the USB drive, but I don't think it's 'double mounting' like I originally thought. When uploading a file via Upload2USB on either Ubuntu Desktop 24.04, Ubuntu Desktop 25.10, and Mint 22.2, a permission denied error is thrown. Of note is that, Raspberry Pi Trixie Desktop, and Ubuntu Server 25.10 do not have this error. 

**Error when folder for the day does not exist:**
```
Aug 27 22:27:30 box nginx[10442]: box nginx: 2025/08/27 22:27:30 [error] 10442#10442: *4 FastCGI sent in stderr: "PHP message: PHP Warning:  mkdir(): Permission denied in /library/www/html/upload2usb/upload2usb.php on line 66; PHP message: UPLOAD2USB ERROR: RuntimeException: {"usb_count":-1,"exception_msg":"Not able to create upload directory. <br\/>Make sure 'usb_lib_writable_sticks' is set to 'True'. <br\/><br\/>"} in /library/www/html/upload2usb/upload2usb.php:66
                                  Stack trace:
                                  #0 /library/www/html/upload2usb/upload-file.php(11): getTargetFolderPath()
                                  #1 {main}" while reading response header from upstream, client: 127.0.0.1, server: box, request: "POST /upload2usb/upload-file.php HTTP/1.1", upstream: "fastcgi://unix:/run/php/php8.3-fpm.sock:", host: "box", referrer: "http://box/usb/"
```

**Error when folder for the day already exists:** 
```
[...]

PR #4088, opened 2025-09-20
WIP detect double mounting [for Upload2USB]

jvonau:
### Fixes bug:
Working towards #4066
Better error messaging when double mounted
RasPiOS trixie
@avni @holta

### DIFF
--- a/roles/usb_lib/files/upload2usb/upload2usb.php
+++ b/roles/usb_lib/files/upload2usb/upload2usb.php
@@ -10,7 +10,7 @@ set_exception_handler(function (Throwable $exception) {
     $exception_details = json_decode($exception->getMessage(), true);
     $usb_count = $exception_details['usb_count'];
     $exception_msg = $exception_details['exception_msg'];
-    
+
     // Always include error.php for upload2usb directory requests
     include ("error.php");
 });
@@ -19,8 +19,10 @@ set_exception_handler(function (Throwable $exception) {
 function getTargetUSBDriveLocation () {
 
          // Enumerate /media/ mountpoints and count them
-         $rmv_usb_paths = shell_exec('lsblk -no MOUNTPOINTS | grep "^/media/"');
-         $rmv_usb_paths_count = substr_count($rmv_usb_paths, "\n"); 
+         //$rmv_usb_paths = shell_exec('lsblk -no MOUNTPOINTS | grep "^/media/"');
+         // Enumerate /library/www/html/local_content/ to see what usbmount mounted and count them
+         $rmv_usb_paths = shell_exec('ls /library/www/html/local_content/');
+         $rmv_usb_paths_count = substr_count($rmv_usb_paths, "\n");
 
          if ($rmv_usb_paths_count == 0) {
              $exception_data = [
@@ -31,7 +33,7 @@ function getTargetUSBDriveLocation () {
          } elseif ($rmv_usb_paths_count > 1) {
              $exception_data = [
                'usb_count' => $rmv_usb_paths_count,
-               'exception_msg' => 'There is more than 1 USB drive inserted or the USB drive is double mounted. <br/><br/>'
+               'exception_msg' => 'There is more than 1 USB drive inserted. <br/><br/>'
              ];
              throw new RuntimeException(json_encode($exception_data));
          } else {
@@ -54,6 +56,7 @@ function getTargetUSBDriveLocation () {
 
              }
 
+             // At this point, we know there is only 1 USB drive inserted and it is not double mounted; return the path to it
              return $rmv_usb_path . "/";
          }
 }
--- a/roles/usb_lib/files/upload2usb/upload2usb.php
+++ b/roles/usb_lib/files/upload2usb/upload2usb.php
@@ -36,17 +36,22 @@ function getTargetUSBDriveLocation () {
              throw new RuntimeException(json_encode($exception_data));
          } else {
                //look for second mountpoint
-               $second_mount = shell_exec('mount | grep udisks');
+               $second_mount = shell_exec('mount | grep udisks |  cut -d " " -f 3');
                if (!empty($second_mount)) {
                    $mnt_point = ($second_mount);
-                   //$mnt_point = array_values(trim(str_replace(" ", " ," , $second_mount)));
-                   //shell_exec('umount $mnt_point');
+                   shell_exec('umount $second_mount');
+               }
+
+               $second_mount2 = shell_exec('mount | grep udisks |  cut -d " " -f 3');
+               if (!empty($second_mount2)) {
                    $exception_data = [
                      'usb_count' => -1,
-                     'exception_msg' => "Double Mounting of $mnt_point. <br/><br/>"
+                     'exception_msg' => "Double Mounting at $second_mount2 attempted fix failed. <br/><br/>"
                ];
+               }
 
                throw new RuntimeException(json_encode($exception_data));
+
              }
 
              return $rmv_usb_path . "/";
--- a/roles/usb_lib/files/upload2usb/upload2usb.php
+++ b/roles/usb_lib/files/upload2usb/upload2usb.php
@@ -40,8 +40,9 @@ function getTargetUSBDriveLocation () {
                //look for second mountpoint
                $second_mount = shell_exec('mount | grep udisks |  cut -d " " -f 3');
                if (!empty($second_mount)) {
+                   // should test for usbbooted machines and exclude /
                    $mnt_point = ($second_mount);
-                   shell_exec('umount $second_mount');
+                   shell_exec('pumount $second_mount');
                }
 
                $second_mount2 = shell_exec('mount | grep udisks |  cut -d " " -f 3');
--- a/roles/usb_lib/files/upload2usb/upload2usb.php
+++ b/roles/usb_lib/files/upload2usb/upload2usb.php
@@ -32,14 +32,14 @@ function getTargetUSBDriveLocation () {
              throw new RuntimeException(json_encode($exception_data));
          } elseif ($rmv_usb_paths_count > 1) {
              //look for second mountpoint
-             $second_mount = shell_exec('mount | grep udisks |  cut -d " " -f 3');
+             $second_mount = shell_exec('mount | grep udisks | cut -d " " -f 3');
              if (!empty($second_mount)) {
                  // should test for usbbooted machines and exclude /
                  $mnt_point = ($second_mount);
                  shell_exec('pumount $second_mount');
                }
 
-             $second_mount2 = shell_exec('mount | grep udisks |  cut -d " " -f 3');
+             $second_mount2 = shell_exec('mount | grep udisks | cut -d " " -f 3');
              if (!empty($second_mount2)) {
                  $exception_data = [
                    'usb_count' => -1,
--- a/roles/usb_lib/files/upload2usb/upload2usb.php
+++ b/roles/usb_lib/files/upload2usb/upload2usb.php
@@ -31,34 +31,30 @@ function getTargetUSBDriveLocation () {
              ];
              throw new RuntimeException(json_encode($exception_data));
          } elseif ($rmv_usb_paths_count > 1) {
+             //look for second mountpoint
+             $second_mount = shell_exec('mount | grep udisks |  cut -d " " -f 3');
+             if (!empty($second_mount)) {
+                 // should test for usbbooted machines and exclude /
+                 $mnt_point = ($second_mount);
+                 shell_exec('pumount $second_mount');
+               }
+
+             $second_mount2 = shell_exec('mount | grep udisks |  cut -d " " -f 3');
+             if (!empty($second_mount2)) {
+                 $exception_data = [
+                   'usb_count' => -1,
+                   'exception_msg' => "Double Mounting at $second_mount2 attempted fix failed. <br/><br/>"
+               ];
+               } else {
              $exception_data = [
                'usb_count' => $rmv_usb_paths_count,
                'exception_msg' => 'There is more than 1 USB drive inserted. <br/><br/>'
              ];
+             }
              throw new RuntimeException(json_encode($exception_data));
          } else {
-               //look for second mountpoint
-               $second_mount = shell_exec('mount | grep udisks |  cut -d " " -f 3');
-               if (!empty($second_mount)) {
-                   // should test for usbbooted machines and exclude /
-                   $mnt_point = ($second_mount);
-                   shell_exec('pumount $second_mount');
-               }
-
-               $second_mount2 = shell_exec('mount | grep udisks |  cut -d " " -f 3');
-               if (!empty($second_mount2)) {
-                   $exception_data = [
-                     'usb_count' => -1,
-                     'exception_msg' => "Double Mounting at $second_mount2 attempted fix failed. <br/><br/>"
-               ];
-               }
-
-               throw new RuntimeException(json_encode($exception_data));
-
-             }
-
              // At this point, we know there is only 1 USB drive inserted and it is not double mounted; return the path to it
-             return $rmv_usb_path . "/";
+             return trim($rmv_usb_paths) . "/";
          }
 }
 
@@ -72,10 +68,10 @@ function getTargetFolderPath ($create_folder_p) {
          if (!file_exists($target_folder_path) && $create_folder_p) {
 
              $exception_data = [
-               'usb_count' => -1, 
+               'usb_count' => -1,
                'exception_msg' => "Not able to create upload directory. <br/>Make sure 'usb_lib_writable_sticks' is set to 'True'. <br/><br/>"
              ];
-	 
+
              mkdir($target_folder_path, 0777) or throw new RuntimeException(json_encode($exception_data));
          }
          return $target_folder_path;
--- a/roles/usb_lib/files/upload2usb/upload2usb.php
+++ b/roles/usb_lib/files/upload2usb/upload2usb.php
@@ -35,8 +35,18 @@ function getTargetUSBDriveLocation () {
              ];
              throw new RuntimeException(json_encode($exception_data));
          } else {
-             // At this point, we know there is only 1 USB drive inserted and it is not double mounted; return the path to it
-             return trim($rmv_usb_paths) . "/"; 
+               //look for second mountpoint
+               $second_mount = shell_exec('mount | grep udisks');
+               if (!empty($second_mount)) {
+                   $exception_data = [
+                     'usb_count' => -1,
+                     'exception_msg' => 'Double Mounted. <br/><br/>'
+               ];
+               //shell_exec('umount $second_mount ');
+               throw new RuntimeException(json_encode($exception_data));
+             }
+
+             return $rmv_usb_path . "/";
          }
 }
 
--- a/roles/usb_lib/files/upload2usb/upload2usb.php
+++ b/roles/usb_lib/files/upload2usb/upload2usb.php
@@ -38,11 +38,14 @@ function getTargetUSBDriveLocation () {
                //look for second mountpoint
                $second_mount = shell_exec('mount | grep udisks');
                if (!empty($second_mount)) {
+                   $mnt_point = ($second_mount);
+                   //$mnt_point = array_values(trim(str_replace(" ", " ," , $second_mount)));
+                   //shell_exec('umount $mnt_point');
                    $exception_data = [
                      'usb_count' => -1,
-                     'exception_msg' => 'Double Mounted. <br/><br/>'
+                     'exception_msg' => "Double Mounting of $mnt_point. <br/><br/>"
                ];
-               //shell_exec('umount $second_mount ');
+
                throw new RuntimeException(json_encode($exception_data));
              }
 
--- a/roles/usb_lib/tasks/install.yml
+++ b/roles/usb_lib/tasks/install.yml
@@ -122,7 +122,7 @@
 
 - name: Add user www-data to group plugdev for pmount to work
   user:
-    name:  www-data
+    name: www-data
     groups: plugdev
     append: yes
 
--- a/roles/usb_lib/tasks/install.yml
+++ b/roles/usb_lib/tasks/install.yml
@@ -53,12 +53,12 @@
 #    deb: "{{ iiab_download_url }}/usbmount_0.0.22_all.deb"
 #  # when: is_debian
 
-- name: Install lockfile-progs and util-linux (findmnt blkid) for usbmount from OS repo
+- name: Install lockfile-progs, pmount and util-linux (findmnt blkid) for usbmount from OS repo
   package:
     name:
       - lockfile-progs
-      - util-linux
       - pmount
+      - util-linux
     state: present
 
 - name: Add dir {{ doc_root }}/local_content, where USB drive links can appear (0775) owned by {{ apache_user }}:{{ apache_user }}
--- a/roles/usb_lib/tasks/install.yml
+++ b/roles/usb_lib/tasks/install.yml
@@ -58,6 +58,7 @@
     name:
       - lockfile-progs
       - util-linux
+      - pmount
     state: present
 
 - name: Add dir {{ doc_root }}/local_content, where USB drive links can appear (0775) owned by {{ apache_user }}:{{ apache_user }}
@@ -118,6 +119,13 @@
     dest: "{{ doc_root }}/upload2usb/"
   with_fileglob:
     - upload2usb/*
+
+- name: Add user www-data to group plugdev for pmount to work
+  user:
+    name:  www-data
+    groups: plugdev
+    append: yes
+
 
 # 2021-03-21: If usbmount is repackaged by apt as a result of Linux kernel 5.4+
 # supporting exFAT, the stanza below (might) in future no longer be needed...


jvonau:
Screenshot from 2025-09-20 13-46-21

>jvonau@pi500:/opt/iiab/iiab $ mount | grep /dev/sd
/dev/sda1 on /media/usb0 type exfat (rw,nodev,noexec,noatime,nodiratime,sync,fmask=0000,dmask=0000,allow_utime=0022,iocharset=utf8,errors=remount-ro)
/dev/sda1 on /media/jvonau/6362-3831 type exfat (rw,nosuid,nodev,relatime,sync,fmask=0000,dmask=0000,allow_utime=0022,iocharset=utf8,errors=remount-ro,uhelper=udisks2)

jvonau:
fished out of `journalctl -u nginx`
>Sep 20 13:15:38 pi500 nginx[1345]: pi500 nginx: 2025/09/20 13:15:38 [error] 1345#1345: *82 FastCGI sent in stderr: "PHP message: UPLOAD2USB ERROR: RuntimeException: {"usb_count":-1,"exception_msg":"Double Mounted. <br\/><br\/>"} in /library/www/html/upload2usb/upload2usb.php:59
                                   Stack trace:
                                   #0 /library/www/html/upload2usb/upload2usb.php(67): getTargetUSBDriveLocation()
                                   #1 /library/www/html/upload2usb/upload-simple.php(10): getTargetFolderPath()
                                   #2 {main}" while reading response header from upstream, client: 10.0.0.242, server: pi500, request: "GET /usb/ HTTP/1.1", subrequest: "/upload2usb/upload-simple.php", upstream: "fastcgi://unix:/run/php/php8.4-fpm.sock:", host: "pi500.local"

jvonau:
U24.04 desktop without IIAB with usb drive inserted and auto-mounted.
>jvonau@kickass:/mnt/scratch/git/iiab$ mount | grep udisks
/dev/sdb1 on /media/jvonau/6362-3831 type exfat (rw,nosuid,nodev,relatime,uid=1000,gid=1000,fmask=0022,dmask=0022,iocharset=utf8,errors=remount-ro,uhelper=udisks2)

Other desktops might not be using udisks so a selection of desktops need to be tested. I'm going to bet the vast majority are using udisks. 

`mount | grep udisks |  awk -F " " '{ print $3 }'`
> /media/jvonau/6362-3831

If somebody else could get the third member of the array  (/media/jvonau/6362-3831) from '$second_mount' then that variable can be passed to umount to automatically unmount the second mountpoint that would be great. Placeholder in code for that.

jvonau:
Screenshot from 2025-09-21 09-47-11

>Sep 21 09:44:15 pi500 nginx[1347]: pi500 nginx: 2025/09/21 09:44:15 [error] 1347#1347: *188 FastCGI sent in stderr: "PHP message: UPLOAD2USB ERROR: RuntimeException: {"usb_count":-1,"exception_msg":"Double Mounting of \/dev\/sda1 on \/media\/jvonau\/6362-3831 type exfat (rw,nosuid,nodev,relatime,sync,fmask=0000,dmask=0000,allow_utime=0022,iocharset=utf8,errors=remount-ro,uhelper=udisks2)\n. <br\/><br\/>"} in /library/www/html/upload2usb/upload2usb.php:66
                                   Stack trace:
                                   #0 /library/www/html/upload2usb/upload2usb.php(74): getTargetUSBDriveLocation()
                                   #1 /library/www/html/upload2usb/upload-simple.php(10): getTargetFolderPath()
                                   #2 {main}" while reading response header from upstream, client: 10.0.0.242, server: pi500, request: "GET /usb/ HTTP/1.1", subrequest: "/upload2usb/upload-simple.php", upstream: "fastcgi://unix:/run/php/php8.4-fpm.sock:", host: "pi500.local"

holta:
@avni or @chapmanjacobd would you be able to review @jvonau's PR here today?

chapmanjacobd:
Double mounted means multiple USBs, right? (It's not related to multiple bind mounts? etc)

It might be better to describe this situation as "multiple USBs" error.

maybe it would be good to bubble up the device label rather than the mount entry... it might be enough just to print the output of `lsblk --list -o NAME,LABEL,MOUNTPOINTS`

The PR looks fine AS-IS. I don't think it is great behavior to automatically unmount--

Or maybe it makes sense to `umount --lazy` all the USBs if people are going to unplug everything anyway? (Even with `--lazy` it is possible for umount to hang so maybe `> /dev/null 2>&1 &` is needed. If you don't want `--lazy` [you could try](https://unix.stackexchange.com/questions/387091/ensure-writing-has-completed-after-umount-l-lazy) remounting `ro` but if someone is going to unplug the USB anyway as far as I know the end result is the same. At least with `--lazy` programs using the USB are more likely to error out and stop writing to it when they can't open new files to read)

But still, I don't think automatic unmounting is the best, expected-by-the-user, behavior. Maybe a button to eject one or all drives would be better but that could also go in a different PR (if people request it).

chapmanjacobd (approved):

avni (commented):
Thanks much @jvonau for working on this. 

Is `$second_mount = shell_exec('mount | grep udisks');` able to detect if a USB stick is mounted to two locations?

Line 23 of the file is already looking for count of removable media so presuming this is doing something different?

jvonau:
> Double mounted means multiple USBs, right? (It's not related to multiple bind mounts? etc)
>

No, single usb stick mounted in two locations, one by usbmount at /media/usb0 and one by udisks2 at /media/'user'/'device'  
> It might be better to describe this situation as "multiple USBs" error.
>

This is for an explicit message to the user for what the real error is other than the generic error message   
> maybe it would be good to bubble up the device label rather than the mount entry... it might be enough just to print the output of `lsblk --list -o NAME,LABEL,MOUNTPOINTS`
> 

No, the problem would be the second mount point created by udisk2 with the correct path of the problem mount point being returned at a hint of the cause.
> The PR looks fine AS-IS. I don't think it is great behavior to automatically unmount--
> 

Sure the exception returned to the user could be altered.
> Or maybe it makes sense to `umount --lazy` all the USBs if people are going to unplug everything anyway? (Even with `--lazy` it is possible for umount to hang so maybe `> /dev/null 2>&1 &` is needed. If you don't want `--lazy` [you could try](https://unix.stackexchange.com/questions/387091/ensure-writing-has-completed-after-umount-l-lazy) remounting `ro` but if someone is going to unplug the USB anyway as far as I know the end result is the same. At least with `--lazy` programs using the USB are more likely to error out and stop writing to it when they can't open new files to read)
> 
> But still, I don't think automatic unmounting is the best, expected-by-the-user, behavior. Maybe a button to eject one or all drives would be better but that could also go in a different PR (if people request it).

This functionality should be required anyway to prevent a dirty filesystem created by yanking the device while still mounted.

jvonau:
> Thanks much @jvonau for working on this.
> 
> Is `$second_mount = shell_exec('mount | grep udisks');` able to detect if a USB stick is mounted to two locations?
> 

The presumption that usbmount worked correctly and the desktop auto-mounting has not been turned off.
> Line 23 of the file is already looking for count of removable media so presuming this is doing something different?

Yes, looks for a desktop mounted device.

avni:
> The presumption that usbmount worked correctly and the desktop auto-mounting has not been turned off.

I wonder if some of this logic could be moved to earlier in the file. I'll test later today. Thanks @jvonau!

holta:
1) This PR (#4088) will be tremendously useful, in many ways, also allowing us to fix and improve the inaccurate error message screenshotted below:

<p align="center">
image
</p>

2) This above RasPiOS 13 "Trixie" ([latest nightly pre-release](https://github.com/iiab/iiab/issues/4058#issuecomment-3323553319), installed on RPi 3) shows 4 different icons on its graphical desktop when the FAT32-formatted stick is inserted **(notice from first 3 screenshots below, that the 4 icons which appear are different almost every time I pull out and re-insert theUSB stick!)**  Sometimes a 5th icon briefly appears too, right after I insert the USB stick!  Presumably due to some timing issue with auto-mounting:

   image

   image

   image

   ```
   root@box:~# lsblk
   NAME        MAJ:MIN RM  SIZE RO TYPE MOUNTPOINTS
   sda           8:0    1 28.9G  0 disk
   ├─sda1        8:1    1 28.9G  0 part /media/usb1
   │                                    /media/iiab-admin/CPRA_X64FRE
   └─sda2        8:2    1    1M  0 part /media/usb0
                                        /media/iiab-admin/UEFI_NTFS
   mmcblk0     179:0    0 59.5G  0 disk
   ├─mmcblk0p1 179:1    0  512M  0 part /boot/firmware
   └─mmcblk0p2 179:2    0   59G  0 part /
   zram0       254:0    0  453M  0 disk [SWAP]

   root@box:~# ls -l /library/www/html/local_content/
   total 0
   lrwxrwxrwx 1 root root 11 Sep 23 10:50 USB0 -> /media/usb0
   lrwxrwxrwx 1 root root 11 Sep 23 10:50 USB1 -> /media/usb1
   ```

3) _Whereas when I turn off auto-mounting in **File Manager > Edit > Preferences > Volume Management > Mount removable media automatically when they are inserted** — it consistently shows just these 2 icons — within a couple seconds after I insert the USB stick:_

   image

   Their UEFI_NTFS (is this something new within RasPiOS 13 Trixie?) README links to: https://github.com/pbatard/uefi-ntfs

   ```
   root@box:~# lsblk
   NAME        MAJ:MIN RM  SIZE RO TYPE MOUNTPOINTS
   sda           8:0    1 28.9G  0 disk
   ├─sda1        8:1    1 28.9G  0 part /media/usb1
   └─sda2        8:2    1    1M  0 part /media/usb0
   mmcblk0     179:0    0 59.5G  0 disk
   ├─mmcblk0p1 179:1    0  512M  0 part /boot/firmware
   └─mmcblk0p2 179:2    0   59G  0 part /
   zram0       254:0    0  453M  0 disk [SWAP]

   root@box:~# ls -l /library/www/html/local_content/
   total 0
   lrwxrwxrwx 1 root root 11 Sep 23 10:24 USB0 -> /media/usb0
   lrwxrwxrwx 1 root root 11 Sep 23 10:24 USB1 -> /media/usb1
   ```

4) With "RasPiOS with desktop" being so common, we should try to clarify for everyone what the top checkbox means ("Mount mountable volumes automatically on program startup") ?  _(I don't understand it yet ;)_

5) I'll do more tests later, when I understand more what others need.  Here's the iiab-diagnostics for now, just in case it helps illuminate: https://paste.centos.org/view/dc91ed6a

avni:
> This above RasPiOS 13 "Trixie" (https://github.com/iiab/iiab/issues/4058#issuecomment-3323553319, installed on RPi 3) shows 4 different icons on its graphical desktop when the FAT32-formatted stick is inserted (notice from first 3 screenshots below, that the 4 icons which appear are different almost every time I pull out and re-insert theUSB stick!) Sometimes a 5th icon briefly appears too, right after I insert the USB stick! Presumably due to some timing issue with auto-mounting:

This seems to imply unmounting (and perhaps remounting too) is not working correctly. Prior to changing any settings, does an error show the first time you insert a USB stick?

holta:
> Prior to changing any settings, does an error show the first time you insert a USB stick?

Yes, the [screenshot at the top](https://github.com/iiab/iiab/pull/4088#issuecomment-3324394118) (with an obviously inaccurate http://box/usb error message) is the first thing you see after installing IIAB on [RasPiOS 13 Trixie (nightly from 2025-09-23)](https://downloads.raspberrypi.com/nightlies/) and inserting a single USB stick.

> This seems to imply unmounting (and perhaps remounting too) is not working correctly.

- I don't have any evidence that unmounting is problematic.
- Of course mounting is certainly quite messy, as screenshots above make clear :)

ASIDE: Almost nobody runs IIAB on "Pure Debian" with graphical desktop on x86_64 — so let's not get distracted investing painstaking testing efforts there — but still if absolutely necessary (just for comparison's sake?) we could take a look there, if only to see how IIAB-on-Debian 13.1 itself works on an old x86_64 laptop, e.g. T14 ThinkPad or such.

holta:
Oops, I fell victim to believing Windows really was formatting the USB stick as an original "just bought it from the store" FAT32 state would appear:

- When in fact, Windows did not really quite do that.  While claiming to format the USB stick, in truth Windows kept 2 partitions on the stick, no doubt residual from the stick having earlier been used to install some OS or another.

- So, I manually wiped the stick partition table using fdisk, and set up 1 single (primary) partition for FAT32, and then once again asked Windows to format the stick to FAT32 — now the results below are at least (more!) promising...

<p align=center>image</p>

1) Immediately after each boot of the RPi 3, ONE SINGLE ICON appears on the desktop:

   image

2) WHEREAS upon re-inserting the stick WITHOUT a reboot, TWO ICONS appear on the desktop (double-mounting ugliness might be harmless on RasPiOS 13 Trixie?? ~In any case, on the bright side, at least Upload2USB appears to work!~)

   image

3. Stale symlink `/library/www/html/local_content/USB1` -> `/media/usb1` would seem to be rather permanent unfortunately (never garbage-collected) — definitely not ideal, creating a bit of confusion for students when looking at http://box/usb — but then again not fatal.
 
   ```
   root@box:~# lsblk
   NAME        MAJ:MIN RM  SIZE RO TYPE MOUNTPOINTS
   sda           8:0    1 28.9G  0 disk
   └─sda1        8:1    1 28.9G  0 part /media/usb0
                                        /media/iiab-admin/0422-2929
   mmcblk0     179:0    0 59.5G  0 disk
   ├─mmcblk0p1 179:1    0  512M  0 part /boot/firmware
   └─mmcblk0p2 179:2    0   59G  0 part /

   root@box:~# ls -l /library/www/html/local_content/
   total 0
   lrwxrwxrwx 1 root root 11 Sep 23 14:40 USB0 -> /media/usb0
   lrwxrwxrwx 1 root root 11 Sep 23 14:01 USB1 -> /media/usb1
   ```

   (Ideally we can implement better garbage collection of stale/bogus links in due course, at some point in future!)

holta:
> 2\. on the bright side, at least Upload2USB appears to work!

Oops, close but no cigar.  @Ark74 discovered, on this vanilla install of IIAB on RasPiOS 13 Trixie on this RPi 3 (_without_ any PR testing here) that the Upload2USB feature doesn't work[*] for him:

```
Not able to create upload directory. 
Make sure 'usb_lib_writable_sticks' is set to 'True'.
```

Unclear what's going wrong, as `usb_lib_writable_sticks` is definitely set to `True`.

iiab-diagnostics should be nearly identical to ~4 hours ago, but just in case that helps, here's a new system snapshot: https://paste.centos.org/view/20354283

[*] Whereas Upload2USB worked for him with PR #4086 on Ubuntu 25.10 (with desktop) on RPi 5.

avni:
> Oops, close but no cigar. @Ark74 discovered, on this vanilla install of IIAB on RasPiOS 13 Trixie on this RPi 3 (without any PR testing here) that the Upload2USB feature doesn't work[*] for him:

```
Not able to create upload directory. 
Make sure 'usb_lib_writable_sticks' is set to 'True'.
```

This is a double mounting issue as shown in the screenshot above which has the output of `lsblk` showing the USB stick mounted 2x. Not sure why this doesn't happen on RPi 5 or 500 with RasPiOS 13 (Trixie). 

This isn't detected initially and users see the "Upload" button without an error because the command that gets file count, only returns the first mount point so the issue is not detected until a user tries to upload a file. 

`lsblk --output NAME,TRAN,RM,MOUNTPOINT --pairs |cut -d " " -f 4 |grep "^MOUNTPOINT=\"/media"| wc -l
`

New IIAB diagnostics from same Raspberry Pi 3 with RasPiOS 13 (Trixie): https://paste.centos.org/view/7ab1b5f7

avni:
It may be good to detect any issues with the USB stick when going to http://box/usb and not wait until a user uploads a file: 
1. Check there is  0 or 1 USB stick plugged in. NB: 0 is okay on http://box/usb because a user may not be intending  to use the USB stick in that moment
2. If a USB stick is detected, confirm format and partitions.
3. Confirm USB stick is not double mounted.

avni:
Modifying the Raspberry Pi File Manager Preferences to prevent double mounting on the Raspberry Pi 3 with RasPiOS 13 (Trixie) solves the issue.

holta:
> 3. Confirm USB stick is not double mounted.

ASIDE: If features like Upload2USB become extremely popular in future, hypothetically / speculatively it **might** even be possible to upload to a USB stick — **regardless** whether it's double-mounted by a Mint-like graphical resktop?

(But that likely involves extreme modifying of usbmount etc, so definitely not our priority today!!)

avni:
Tested #4088 on the Raspberry Pi 3 and RasPiOS 13 (Trixie) with the USB stick double mounted. It failed as expected. 🙌 

`Double Mounting of /dev/sda1 on /media/iiab-admin/2E9B-4A60 type vfat (rw,nosuid,nodev,relatime,uid=1000,gid=1000,fmask=0022,dmask=0022,codepage=437,iocharset=ascii,shortname=mixed,showexec,utf8,flush,errors=remount-ro,uhelper=udisks2)
`
Latest IIAB Diagnostics: https://paste.centos.org/view/2e072ab2

holta:
> double mounting issue as shown in the screenshot above which has the output of `lsblk` showing the USB stick mounted 2x. Not sure why this doesn't happen on RPi 5 or 500.
> 
> This isn't detected initially and users see the "Upload" button without an error because the command that gets file count, only returns the first mount point so the issue is not detected until a user tries to upload a file.
> 
> `lsblk --output NAME,TRAN,RM,MOUNTPOINT --pairs |cut -d " " -f 4 |grep "^MOUNTPOINT=\"/media"| wc -l `

Above `lsblk` should certainly behave the same way on RPi 3 just like RPi 5 / 500, right?

- What's likely the root cause of RPi 3 seemingly behaving differently from RPi 5 / 500?
- Is testing of RPi 4 the best way to shine a light on what's happening?

chapmanjacobd:
> regardless whether it's double-mounted by a Mint-like graphical resktop

maybe like this?

```sh
# Mount the filesystem.
current_mountpoint=`findmnt -rno TARGET "$DEVNAME"`
if [ -n "$current_mountpoint" ]; then
    log info "executing command: mount --bind $current_mountpoint $mountpoint"
    mount --bind "$current_mountpoint" "$mountpoint"
else
    log info "executing command: mount -t$fstype ${options:+-o$options} $DEVNAME $mountpoint"
    mount "-t$fstype" "${options:+-o$options}" "$DEVNAME" "$mountpoint"
fi
```

or maybe it makes more sense to symlink instead

avni:
RasPiOS 13 (Trixie) on Raspberry Pi 4 behaves the same as Raspberry Pi 3 and mounts the USB stick to two locations. Upload2USB has the same permission denied error when uploading. 

Screenshot 2025-09-23 at 11 39 21 PM

```
Not able to create upload directory. 
Make sure 'usb_lib_writable_sticks' is set to 'True'. 
```

Modifying the Raspberry Pi File Manager Preferences to prevent double mounting on the Raspberry Pi 4 solves the issue.

Ark74:
> Raspberry Pi 4 behaves the same as Raspberry Pi 3

But that depends on the OS, right? I'm on a rpi3 with ubuntu 25.04 server just now (can't load desktop here) and it shows:


```
$ lsblk
NAME        MAJ:MIN RM  SIZE RO TYPE MOUNTPOINTS
loop0         7:0    0 38.7M  1 loop /snap/snapd/23772
loop1         7:1    0 44.2M  1 loop /snap/snapd/25205
sda           8:0    1   30G  0 disk 
└─sda1        8:1    1   30G  0 part /media/usb0
mmcblk0     179:0    0 58.2G  0 disk 
├─mmcblk0p1 179:1    0  512M  0 part /boot/firmware
└─mmcblk0p2 179:2    0 57.7G  0 part /

```

jvonau:
> > Raspberry Pi 4 behaves the same as Raspberry Pi 3
> 
> But that depends on the OS, right? I'm on a rpi3 with ubuntu 25.04 server just now (can't load desktop here) and it shows:
> 
> ```
> $ lsblk
> NAME        MAJ:MIN RM  SIZE RO TYPE MOUNTPOINTS
> loop0         7:0    0 38.7M  1 loop /snap/snapd/23772
> loop1         7:1    0 44.2M  1 loop /snap/snapd/25205
> sda           8:0    1   30G  0 disk 
> └─sda1        8:1    1   30G  0 part /media/usb0
> mmcblk0     179:0    0 58.2G  0 disk 
> ├─mmcblk0p1 179:1    0  512M  0 part /boot/firmware
> └─mmcblk0p2 179:2    0 57.7G  0 part /
> ```

No desktop, no double mount to contend with, should just work.

jvonau:
>     3. Stale symlink `/library/www/html/local_content/USB1` -> `/media/usb1` would seem to be rather permanent unfortunately (never garbage-collected) — definitely not ideal, creating a bit of confusion for students when looking at http://box/usb — but then again not fatal.
> 
>        root@box:~# lsblk
>        NAME        MAJ:MIN RM  SIZE RO TYPE MOUNTPOINTS
>        sda           8:0    1 28.9G  0 disk
>        └─sda1        8:1    1 28.9G  0 part /media/usb0
>                                             /media/iiab-admin/0422-2929
>        mmcblk0     179:0    0 59.5G  0 disk
>        ├─mmcblk0p1 179:1    0  512M  0 part /boot/firmware
>        └─mmcblk0p2 179:2    0   59G  0 part /
>        
>        root@box:~# ls -l /library/www/html/local_content/
>        total 0
>        lrwxrwxrwx 1 root root 11 Sep 23 14:40 USB0 -> /media/usb0
>        lrwxrwxrwx 1 root root 11 Sep 23 14:01 USB1 -> /media/usb1
>
>        (Ideally we can implement better garbage collection of stale/bogus links in due course, at some point in future!)

A reboot should clear all links that were created via stopping usbmount@*.service.  `systemctl status usbmount@*.service` would show all active mountings. #3944 dealt with the stale symlink issue by actually reacting to the udev event of yanking the usb drive.

jvonau:
> ```shell
> findmnt -rno TARGET "$DEVNAME"
> ```

findmnt -rno TARGET /dev/sda1
/media/jvonau/6362-3831
/media/usb0

holta:
> A reboot should clear all links that were created via stopping usbmount@*.service. `systemctl status usbmount@*.service` would show all active mountings. #3944 dealt with the stale symlink issue by actually reacting to the udev event of yanking the usb drive.

Good to know.  This is all on the right track.

(Clearing out stale symlinks with a reboot doesn't always work, to date anyway.)

avni:
> But that depends on the OS, right? I'm on a rpi3 with ubuntu 25.04 server just now (can't load desktop here) and it shows:

Thanks, yes it is dependent on the OS too, I should have been clear about that @Ark74. The RPI3 and RPI4 tests I did were with RasPiOS 13 (Trixie). Updated comments above to make that clear. What's interesting is that RasPiOS 13 (Trixie) with IIAB on the RPI5 and RPI500 does not behave similarly - they _DO NOT_ double mount USB sticks!

holta:
1) @avni instead of...

   ```
   root@box:~# lsblk
   NAME        MAJ:MIN RM  SIZE RO TYPE MOUNTPOINTS
   sda           8:0    1 28.9G  0 disk
   └─sda1        8:1    1 28.9G  0 part /media/usb0
                                        /media/iiab-admin/2E9B-4A60
   mmcblk0     179:0    0 59.5G  0 disk
   ├─mmcblk0p1 179:1    0  512M  0 part /boot/firmware
   └─mmcblk0p2 179:2    0   59G  0 part /
   zram0       254:0    0  453M  0 disk [SWAP]
   
   root@box:~# lsblk --output NAME,TRAN,RM,MOUNTPOINT --pairs | cut -d " " -f 4
   MOUNTPOINT=""
   MOUNTPOINT="/media/usb0"
   MOUNTPOINT=""
   MOUNTPOINT="/boot/firmware"
   MOUNTPOINT="/"
   MOUNTPOINT="[SWAP]"
   ```

   ...have you considered simpler things like?

   ```
   root@box:~# lsblk --output MOUNTPOINT --pairs
   MOUNTPOINT=""
   MOUNTPOINT="/media/usb0"
   MOUNTPOINT=""
   MOUNTPOINT="/boot/firmware"
   MOUNTPOINT="/"
   MOUNTPOINT="[SWAP]"
   ```

   ...or even more concisely?

   ```
   root@box:~# lsblk -no MOUNTPOINT
   
   /media/usb0
   
   /boot/firmware
   /
   [SWAP]
   ```

2) Is the suggestion below from @chapmanjacobd @jvonau to auto-detect double mounts... a good way forward in your opinion?  STRETCH GOAL: make Upload2USB work even if a _**single**_ USB stick is double-mounted or triple-mounted etc??  (But let's please avoid the complexity of modifying `usbmount` code and its longstanding norms, unless we absolutely must!!)

   ```
   root@box:~# findmnt -rno TARGET /dev/sda1
   /media/iiab-admin/2E9B-4A60
   /media/usb0
   ```

   Or use the full list if you prefer?

   ```
   root@box:~# findmnt -rno TARGET
   /sys
   /proc
   /dev
   /dev/pts
   /run
   /
   /sys/kernel/security
   /dev/shm
   /sys/fs/cgroup
   /sys/fs/pstore
   /sys/fs/bpf
   /proc/sys/fs/binfmt_misc
   /sys/kernel/tracing
   /run/lock
   /dev/mqueue
   /sys/kernel/debug
   /run/credentials/systemd-journald.service
   /sys/fs/fuse/connections
   /sys/kernel/config
   /tmp
   /boot/firmware
   /proc/sys/fs/binfmt_misc
   /run/rpc_pipefs
   /run/user/1000
   /run/user/1000/gvfs
   /run/user/1000/doc
   /run/credentials/getty@tty6.service
   /run/credentials/getty@tty1.service
   /media/iiab-admin/2E9B-4A60
   /media/usb0
   ```

   ```
   root@box:~# hostname -I
   192.168.0.142 100.64.0.29
   root@box:~# tail -1 /proc/cpuinfo
   Model           : Raspberry Pi 3 Model B Plus Rev 1.3
   root@box:~# cat /etc/rpi-issue
   Raspberry Pi reference 2025-09-24
   Generated using pi-gen, https://github.com/RPi-Distro/pi-gen, 22dc0ec4300456c77f12dbbe93e78fc407113007, stage4
   ```

holta:
Interestingly (just FYI) on an RPi 4 (1 GB RAM) the OS itself OFTEN PREVENTS double-mounting:

image

Though the OS (RasPiOS 13 Trixie) cannot be depended on 100% to block double-mounting, as the very 1st time I plugged in the stick, it double-mounted anyway: (with 2 icons appearing on the desktop)

```
root@box:~# lsblk
NAME        MAJ:MIN RM  SIZE RO TYPE MOUNTPOINTS
sda           8:0    1 28.9G  0 disk
└─sda1        8:1    1 28.9G  0 part /media/iiab-admin/0422-2929
                                     /media/usb0
mmcblk0     179:0    0 59.5G  0 disk
├─mmcblk0p1 179:1    0  512M  0 part /boot/firmware
└─mmcblk0p2 179:2    0   59G  0 part /
zram0       254:0    0  453M  0 disk [SWAP]
```

iiab-diagnostics: https://paste.centos.org/view/a517c916

jvonau:
Don't think the desktop prevents double mount as findmnt and lsblk both show the device as mounted so I'll take this as a warning that some other process other than the desktop mounted the device. In this case for some reason the desktop beat usbmount in mounting the device. Last mount wins the permissions race and would be the results displayed by `lsblk -no MOUNTPOINT`

holta:
This (might!) help greatly simplify:

- `lsblk -o MOUNTPOINTS` (plural!) displays all associated mount points for a device, in a multi-line cell, since about 2021 if one believes https://discussion.fedoraproject.org/t/cannot-understand-from-output-of-lsblk/76761/3 ?

- `lsblk -o MOUNTPOINT` (singular!) in contrast, shows only one mount point, usually the last mounted instance of the filesystem, as appears topmost in a generic `lablk` output.

(But I have yet to try it.  And `findmnt` is also an option.)

holta:
Confirmed that we really should be using the plural `lsblk -o MOUNTPOINTS` not singular `lsblk -o MOUNTPOINT`:

```
root@box:~# lsblk
NAME        MAJ:MIN RM  SIZE RO TYPE MOUNTPOINTS
sda           8:0    1 28.9G  0 disk
└─sda1        8:1    1 28.9G  0 part /media/usb0
                                     /media/iiab-admin/2E9B-4A60
mmcblk0     179:0    0 59.5G  0 disk
├─mmcblk0p1 179:1    0  512M  0 part /boot/firmware
└─mmcblk0p2 179:2    0   59G  0 part /
zram0       254:0    0  453M  0 disk [SWAP]

```

Old way: (incomplete!)

```
root@box:~# lsblk -no MOUNTPOINT

/media/usb0

/boot/firmware
/
[SWAP]
```

New way: (complete!)

```
root@box:~# lsblk -no MOUNTPOINTS

/media/usb0
/media/iiab-admin/2E9B-4A60

/boot/firmware
/
[SWAP]
```

Do we really need a count of all actively mounted `/media` mountpoints?  If so, here it is:

```
root@box:~# lsblk -no MOUNTPOINTS | grep -c '^/media/'
2
```

Overly cute output FWIW:

```
root@box:~# lsblk -Pno MOUNTPOINTS
MOUNTPOINTS=""
MOUNTPOINTS="/media/usb0\x0a/media/iiab-admin/2E9B-4A60"
MOUNTPOINTS=""
MOUNTPOINTS="/boot/firmware"
MOUNTPOINTS="/"
MOUNTPOINTS="[SWAP]"
```

Notice `\x0A` above (and below) is the escaped hexadecimal Line Feed.  The equivalent of \n.

```
root@box:~# lsblk -Pno PATH,MOUNTPOINTS
PATH="/dev/sda" MOUNTPOINTS=""
PATH="/dev/sda1" MOUNTPOINTS="/media/usb0\x0a/media/iiab-admin/2E9B-4A60"
PATH="/dev/mmcblk0" MOUNTPOINTS=""
PATH="/dev/mmcblk0p1" MOUNTPOINTS="/boot/firmware"
PATH="/dev/mmcblk0p2" MOUNTPOINTS="/"
PATH="/dev/zram0" MOUNTPOINTS="[SWAP]"
```

Or if nuanced parsing of lsblk's output is really necessary, there's always JSON handling using [jq](https://jqlang.org/):

```
root@box:~# lsblk -Jno PATH,MOUNTPOINTS
{
   "blockdevices": [
      {
         "path": "/dev/sda",
         "mountpoints": []
      },{
         "path": "/dev/sda1",
         "mountpoints": [
             "/media/usb0", "/media/iiab-admin/2E9B-4A60"
         ]
      },{
         "path": "/dev/mmcblk0",
         "mountpoints": []
      },{
         "path": "/dev/mmcblk0p1",
         "mountpoints": [
             "/boot/firmware"
         ]
      },{
         "path": "/dev/mmcblk0p2",
         "mountpoints": [
             "/"
         ]
      },{
         "path": "/dev/zram0",
         "mountpoints": [
             "[SWAP]"
         ]
      }
   ]
}

root@box:~# lsblk -Jno PATH,MOUNTPOINTS | jq '.blockdevices.[] | select(.path=="/dev/sda1") | .mountpoints'
[
  "/media/usb0",
  "/media/iiab-admin/2E9B-4A60"
]

root@box:~# lsblk -Jno PATH,MOUNTPOINTS | jq '.blockdevices.[] | select(.path=="/dev/sda1") | .mountpoints.[0]'
"/media/usb0"

root@box:~# lsblk -Jno PATH,MOUNTPOINTS | jq '.blockdevices.[] | select(.path=="/dev/sda1") | .mountpoints.[1]'
"/media/iiab-admin/2E9B-4A60"

root@box:~# lsblk -Jno PATH,MOUNTPOINTS | jq '.blockdevices.[] | select(.path=="/dev/sda1") | .mountpoints | length'
2
```

holta:
> Do we really need a count of all actively mounted `/media` mountpoints? If so, here it is:
> 
> ```
> root@box:~# lsblk -no MOUNTPOINTS | grep -c '^/media/'
> 2
> ```

If the above number's important, naturally there are also other ways of obtaining it:

```
root@box:~# mount | grep -c ' on /media/'
2
```

jvonau:
FWIW I changed '($second_mount)' to be 'shell_exec('mount | grep udisks |  cut -d " " -f 3');' which returned just the mountpoint and passed that to umount but the unmounting failed as php-fpm is running as www-data and doesn't have the required permissions to unmount the device. Making a call to admin-console's iiab-cmdsrv3.py backend might work as that process is running as root.

chapmanjacobd:
Two other ideas:

Immediately umount first mountpoint after mounting. This will still respect the User's initial mount options:

```
# Mount the filesystem.
current_mountpoint=`findmnt -rno TARGET "$DEVNAME"`
if [ -n "$current_mountpoint" ]; then
    log info "executing command: mount --bind $current_mountpoint $mountpoint"
    mount --bind "$current_mountpoint" "$mountpoint"
    umount "$current_mountpoint"
else
    log info "executing command: mount -t$fstype ${options:+-o$options} $DEVNAME $mountpoint"
    mount "-t$fstype" "${options:+-o$options}" "$DEVNAME" "$mountpoint"
fi
```

Symlink instead

```
current_mountpoint=`findmnt -rno TARGET "$DEVNAME"`
if [ -n "$current_mountpoint" ]; then
    log info "executing command: ln -s $current_mountpoint $mountpoint"
    ln -s "$current_mountpoint" "$mountpoint"
else
    log info "executing command: mount -t$fstype ${options:+-o$options} $DEVNAME $mountpoint"
    mount "-t$fstype" "${options:+-o$options}" "$DEVNAME" "$mountpoint"
fi
```

There will still be the race condition where usbmount runs first but that can be solved in udev rules (I think?) or perhaps by adding `udevadm settle` prior to findmnt (though I suppose this will actually make it hang forever... since udev is running the usbmount script...)

avni:
So much learned this afternoon from testing as we move towards a simplified version of the code with auto-detection of 2 or more mount points. 

1. Not all Linux distributions create dynamic mount points in /media, some create them in /run. Given RPi and Ubuntu use /media, I'm going to stick with that for now.
2. **Race condition:** If desktop auto-mounting has not been turned off, and the disk is mounted to two mount points, the order of the two mount points can change, see screenshots below. So if we want to allow double mounting, the code needs to be resilient to both. 

Screenshot 2025-09-28 at 6 25 06 PM
Screenshot 2025-09-28 at 6 25 24 PM

3. Ownership and permissions are different when desktop auto-mounting is on vs when it is not. The mount point mounted by the desktop is owned by the non-root user (in this case, `iiab-admin`) and has 750 or 755 permissions. I believe this is why we see the Permission Denied error when the storage is double mounted and we are testing Upload2USB. 

4. In the Terminal shell, using `chmod ugo+rwx` to change the permissions on the directory, resolves the Permission Denied error. Upload2USB starts working! 

As long as we can chmod through `shell_exec` in php, I think we can let drives be double mounted and not worry about them. I'll try that now.

chapmanjacobd:
> The mount point mounted by the desktop is owned by the non-root user (in this case, iiab-admin) and has 750 or 755 permissions

In usbmount.conf:

    MOUNTOPTIONS="sync,noexec,nodev,noatime,nodiratime"

Maybe we need to add `uid=1000,gid=1000` (or whatever id iiab-user is)

edit: Ah.. actually it seems like it will be more universally supported to just `chown` the mountpoint after mounting (with root user chown to iiab-admin): https://superuser.com/questions/320415/mount-device-with-specific-user-rights#

holta:
> 4\. In the Terminal shell, using `chmod ugo+rwx` to change the permissions on the directory, resolves the Permission Denied error. Upload2USB starts working!
> 
> As long as we can chmod through `shell_exec` in php, I think we can let drives be double mounted and not worry about them. I'll try that now.

I assume you're talking about RasPiOS 13 Trixie.  Good enough.  (But let's also investigate Linux Mint 22.2 as is very common among some IIAB regions...)

OFF TOPIC BUT ENTERTAINING... that this is Google's top result on the topic of double-mounting:

> "Double mounting is highly effective in lowering the transmission of vibration at high frequencies.   Considering that structure-borne noise occurs at higher frequencies than those of perceptible vibration, double mounting is an attractive solution for abating the transmission of such noise." 🙃
>
> https://deicon.com/solutions/vibration-isolation/double-mounting/

avni:
> I assume you're talking about RasPiOS 13 Trixie. Good enough. (But let's also investigate Linux Mint 22.2 as is very common among some IIAB regions...)

Yes, testing on RasPiOS 13 Trixie currently. ACK re Mint 22.2.

avni:
Not trivial to run `chmod`  from within php unfortunately!  Running into a couple of issues: 

1. When desktop auto-mounting is on, the USB drive is mounted as read only. To remount it as `rw`, , e.g., `mount -o remount,rw /media/usb0`, you need to be root, which doesn't work from  php.
2. Neither `shell_exec` nor the native chmod() function in php are successful in applying the necessary permissions. Oy!

chapmanjacobd:
> When desktop auto-mounting is on, the USB drive is mounted as read only

That seems really weird! Do you see anything in:

    sudo dmesg --human --nopager --decode --level emerg,alert,crit,err

It seems like a weird default to have things read-only by default... but maybe it is an under-construction part of RaspberryPi-OS 13?

jvonau:
> Not trivial to run `chmod` from within php unfortunately! Running into a couple of issues:
> 
>     1. When desktop auto-mounting is on, the USB drive is mounted as read only. To remount it as `rw`, , e.g., `mount -o remount,rw /media/usb0`, you need to be root, which doesn't work from  php.
> 
>     2. Neither `shell_exec` nor the native chmod() function in php are successful in applying the necessary permissions. Oy!

Same issue if you try to use  `shell_exec` umount from php

avni:
> Do you see anything in: `sudo dmesg --human --nopager --decode --level emerg,alert,crit,err`

```
root@box:~# sudo dmesg --human --nopager --decode --level emerg,alert,crit,err
kern  :err   : [Sep29 00:03] FAT-fs (sda1): error, fat_get_cluster: invalid start cluster (i_pos 0, start d97c6e9c)
kern  :err   : [  +0.000040] FAT-fs (sda1): Filesystem has been set read-only
kern  :err   : [  +0.000337] FAT-fs (sda1): error, fat_get_cluster: invalid start cluster (i_pos 0, start 26c8c2c0)
kern  :err   : [  +0.000326] FAT-fs (sda1): error, fat_get_cluster: invalid start cluster (i_pos 0, start c1837695)
kern  :err   : [Sep29 00:22] FAT-fs (sda1): error, fat_get_cluster: invalid start cluster (i_pos 0, start d97c6e9c)
kern  :err   : [  +0.000070] FAT-fs (sda1): error, fat_get_cluster: invalid start cluster (i_pos 0, start d97c6e9c)
kern  :err   : [  +0.000197] FAT-fs (sda1): error, fat_get_cluster: invalid start cluster (i_pos 0, start 26c8c2c0)
kern  :err   : [  +0.000048] FAT-fs (sda1): error, fat_get_cluster: invalid start cluster (i_pos 0, start 26c8c2c0)
kern  :err   : [  +0.000130] FAT-fs (sda1): error, fat_get_cluster: invalid start cluster (i_pos 0, start c1837695)
kern  :err   : [  +0.000048] FAT-fs (sda1): error, fat_get_cluster: invalid start cluster (i_pos 0, start c1837695)
kern  :err   : [  +0.002434] FAT-fs (sda1): error, fat_get_cluster: invalid start cluster (i_pos 0, start d97c6e9c)
kern  :err   : [  +0.000059] FAT-fs (sda1): error, fat_get_cluster: invalid start cluster (i_pos 0, start d97c6e9c)
kern  :err   : [  +0.000167] FAT-fs (sda1): error, fat_get_cluster: invalid start cluster (i_pos 0, start 26c8c2c0)
kern  :err   : [  +0.000047] FAT-fs (sda1): error, fat_get_cluster: invalid start cluster (i_pos 0, start 26c8c2c0)
kern  :err   : [Sep29 00:23] FAT-fs (sda1): error, fat_get_cluster: invalid start cluster (i_pos 0, start d97c6e9c)
kern  :err   : [  +0.000579] FAT-fs (sda1): error, fat_get_cluster: invalid start cluster (i_pos 0, start 26c8c2c0)
kern  :err   : [  +0.000332] FAT-fs (sda1): error, fat_get_cluster: invalid start cluster (i_pos 0, start c1837695)
```

chapmanjacobd:
> FAT-fs (sda1): Filesystem has been set read-only

Yeah the kernel remounted to `ro` to protect your data... seems like a hardware problem :/ You should probably back up the data on that USB. You'll need to reformat it to get rid of the errors

jvonau:
> > The mount point mounted by the desktop is owned by the non-root user (in this case, iiab-admin) and has 750 or 755 permissions
> 
> In usbmount.conf:
> 
> ```
> MOUNTOPTIONS="sync,noexec,nodev,noatime,nodiratime"
> ```
> 
> Maybe we need to add `uid=1000,gid=1000` (or whatever id iiab-user is)
> 

https://github.com/iiab/iiab/blob/master/roles/usb_lib/tasks/main.yml#L35-L40

jvonau:
> > Do you see anything in: `sudo dmesg --human --nopager --decode --level emerg,alert,crit,err`
> 
> ```
> root@box:~# sudo dmesg --human --nopager --decode --level emerg,alert,crit,err
> kern  :err   : [Sep29 00:03] FAT-fs (sda1): error, fat_get_cluster: invalid start cluster (i_pos 0, start d97c6e9c)
> kern  :err   : [  +0.000040] FAT-fs (sda1): Filesystem has been set read-only
> kern  :err   : [  +0.000337] FAT-fs (sda1): error, fat_get_cluster: invalid start cluster (i_pos 0, start 26c8c2c0)
> kern  :err   : [  +0.000326] FAT-fs (sda1): error, fat_get_cluster: invalid start cluster (i_pos 0, start c1837695)
> kern  :err   : [Sep29 00:22] FAT-fs (sda1): error, fat_get_cluster: invalid start cluster (i_pos 0, start d97c6e9c)
> kern  :err   : [  +0.000070] FAT-fs (sda1): error, fat_get_cluster: invalid start cluster (i_pos 0, start d97c6e9c)
> kern  :err   : [  +0.000197] FAT-fs (sda1): error, fat_get_cluster: invalid start cluster (i_pos 0, start 26c8c2c0)
> kern  :err   : [  +0.000048] FAT-fs (sda1): error, fat_get_cluster: invalid start cluster (i_pos 0, start 26c8c2c0)
> kern  :err   : [  +0.000130] FAT-fs (sda1): error, fat_get_cluster: invalid start cluster (i_pos 0, start c1837695)
> kern  :err   : [  +0.000048] FAT-fs (sda1): error, fat_get_cluster: invalid start cluster (i_pos 0, start c1837695)
> kern  :err   : [  +0.002434] FAT-fs (sda1): error, fat_get_cluster: invalid start cluster (i_pos 0, start d97c6e9c)
> kern  :err   : [  +0.000059] FAT-fs (sda1): error, fat_get_cluster: invalid start cluster (i_pos 0, start d97c6e9c)
> kern  :err   : [  +0.000167] FAT-fs (sda1): error, fat_get_cluster: invalid start cluster (i_pos 0, start 26c8c2c0)
> kern  :err   : [  +0.000047] FAT-fs (sda1): error, fat_get_cluster: invalid start cluster (i_pos 0, start 26c8c2c0)
> kern  :err   : [Sep29 00:23] FAT-fs (sda1): error, fat_get_cluster: invalid start cluster (i_pos 0, start d97c6e9c)
> kern  :err   : [  +0.000579] FAT-fs (sda1): error, fat_get_cluster: invalid start cluster (i_pos 0, start 26c8c2c0)
> kern  :err   : [  +0.000332] FAT-fs (sda1): error, fat_get_cluster: invalid start cluster (i_pos 0, start c1837695)
> ```

Good call dirty filesystem

jvonau:
> > Not trivial to run `chmod` from within php unfortunately! Running into a couple of issues:
> > ```
> > 1. When desktop auto-mounting is on, the USB drive is mounted as read only. To remount it as `rw`, , e.g., `mount -o remount,rw /media/usb0`, you need to be root, which doesn't work from  php.
> > 
> > 2. Neither `shell_exec` nor the native chmod() function in php are successful in applying the necessary permissions. Oy!
> > ```
> 
> Same issue if you try to use `shell_exec` umount from php

Think the best that can be done easily is return a useful warning to the user, after all this double mounting is well known when desktops are involved and turning that off is really a preparatory step to working functionality.

holta:
> Good call dirty filesystem

Likely very common?!

- Hopefully @avni can make a recommendation — do we really need to notify teachers — when their USB stick needs medical help? 🚑

- At the same time IIAB should _NOT_ try to be an OS solving all corner cases, for all people, at all times 😇 😇

avni:
A modified patch is at #4097, tightening up the getTargetUSBDriveLocation function, detects double mounting and provides a warning a la this PR, and also uses the simpler lsblk options shared above.

holta:
Just FYI the ideas in this PR are still important[*] and still brewing, even after this very much related cleanup / fixup / refactoring PR was just merged:

- PR #4097

[*] Towards far more meaningful "double mount" and "too many sticks are inserted" error messages at minimum!

And possibly (STRETCH GOAL) allowing Upload2USB to work with USB sticks/drives that happen to be double mounted?

roles/usb_lib/files/upload2usb/upload2usb.php
```diff
@@ -29,14 +31,30 @@ function getTargetUSBDriveLocation () {
...
          } elseif ($rmv_usb_paths_count > 1) {
+             //look for second mountpoint
+             $second_mount = shell_exec('mount | grep udisks |  cut -d " " -f 3');
+             if (!empty($second_mount)) {
+                 // should test for usbbooted machines and exclude /
+                 $mnt_point = ($second_mount);
+                 shell_exec('pumount $second_mount');
+               }
+
+             $second_mount2 = shell_exec('mount | grep udisks |  cut -d " " -f 3');
```
holta: Just lint suggestions...

```suggestion
             $second_mount = shell_exec('mount | grep udisks | cut -d " " -f 3');
             if (!empty($second_mount)) {
                 // should test for usbbooted machines and exclude /
                 $mnt_point = ($second_mount);
                 shell_exec('pumount $second_mount');
               }

             $second_mount2 = shell_exec('mount | grep udisks | cut -d " " -f 3');
```

jvonau: I'll commit after the tests run.

holta:
@avni please review @jvonau's PR here (#4088) in coming days if you have time?

jvonau:
If the device is inserted prior to booting there is no double mount:


Inserting device after booting:


jvonau@pi500:/opt/iiab/iiab $ `mount`
>/dev/sda1 on /media/jvonau/6362-3831 type exfat (rw,nosuid,nodev,relatime,sync,fmask=0000,dmask=0000,allow_utime=0022,iocharset=utf8,errors=remount-ro,uhelper=udisks2)
/dev/sda1 on /media/usb0 type exfat (rw,nodev,noexec,noatime,nodiratime,sync,fmask=0000,dmask=0000,allow_utime=0022,iocharset=utf8,errors=remount-ro)

jvonau@pi500:/opt/iiab/iiab $ `lsblk`
>NAME        MAJ:MIN RM   SIZE RO TYPE MOUNTPOINTS
sda           8:0    1 238.3G  0 disk 
└─sda1        8:1    1 238.3G  0 part /media/jvonau/6362-3831
                                      /media/usb0
mmcblk0     179:0    0   7.4G  0 disk 
├─mmcblk0p1 179:1    0   512M  0 part /boot/firmware
└─mmcblk0p2 179:2    0   6.9G  0 part /
zram0       254:0    0   3.9G  0 disk [SWAP]

messed with /etc/passwd to get a shell
jvonau@pi500:/opt/iiab/iiab $ `sudo su www-data -c 'pumount -d /media/jvonau/6362-3831'`
>checking whether /media/jvonau/6362-3831 is a mounted directory
resolved mount point /media/jvonau/6362-3831 to device /dev/sda1
resolved /dev/sda1 to device /dev/sda1
Checking for device '/dev/sda1' in '/etc/fstab'
 -> not found
Checking for device '/dev/sda1' in '/etc/mtab'
 -> found as '/dev/sda1'
policy check passed
spawnv(): executing /bin/umount '/bin/umount' '/dev/sda1'
spawn(): /bin/umount terminated with status 0
spawnv(): executing /sbin/cryptsetup '/sbin/cryptsetup' 'close' '/dev/sda1'
spawn(): /sbin/cryptsetup terminated with status 255
Removing luks lockfile '/var/lib/pmount-locks_luks/_dev_sda1' for device '/dev/sda1'

jvonau@pi500:/opt/iiab/iiab $ `mount`
>/dev/sda1 on /media/jvonau/6362-3831 type exfat (rw,nosuid,nodev,relatime,sync,fmask=0000,dmask=0000,allow_utime=0022,iocharset=utf8,errors=remount-ro,uhelper=udisks2)


There are two things going here:
- udisks2 is beating usbmount in the mount race
- pumount ends up unmounting the wrong device as usbmount was the last to mount the device

jvonau:
The revised code broke what I had going on so I'm not going to pursue auto unmounting with pumount

holta:
Extremely valuable research, whatever strategy we end up with.  Thanks @jvonau & everyone for having investigated.

chapmanjacobd:
> pumount ends up unmounting the wrong device as usbmount was the last to mount the device

Fascinating stuff! even if it didn't quite work, I feel like we're a lot closer now to identifying a working solution

edit:

> the _wrong_ device

Why do we not just use the mountpoint that exists? It is indeed unfortunate that pumount does not practice fine control of choosing a specific mountpoint, but is there anything wrong with using `/media/jvonau/6362-3831` in this case?

holta:
> > the _wrong_ device
> 
> Why do we not just use the mountpoint that exists? It is indeed unfortunate that pumount does not practice fine control of choosing a specific mountpoint, but is there anything wrong with using `/media/jvonau/6362-3831` in this case?

cc: @Ark74 who might also have suggestions, as we try to figure this out.

(No Rush: but if a general enough solution appears viable for "all" Debian/Ubuntu like **_graphical desktop_** distros — allowing IIABs to write to USB sticks that have been double-mounted by these kinds of Linux graphical desktops — than we should at least give it a shot!)

jvonau:
Sort of misspoke correct device but wrong mount point when passed the wanted path to be unmounted. Think mtab knows what the order of mounting was by the placement in the list but pumount does the 'resolve mount point routine' noted above to come up the the device and passes the device to umount. Double mounting can have either path mounted last in a race between udisks and usbmount. On a side note I could not get pumount to run from within php think a shell is needed.

PR #4087, opened 2025-09-18
add zram-swap ansible role [+ q's about swap partition & pi_swap_file_size]

chapmanjacobd:
### Fixes bug:

https://github.com/iiab/iiab/issues/2414

re: kernel parameters

The defaults kernel parameters seem to be reasonable and not need adjustment (at least on Debian):

    dirty_ratio=20, swappiness=60

```
sudo sysctl -n vm.dirty_background_ratio
10
sudo sysctl -n vm.dirty_ratio
20
sudo sysctl -n vm.vfs_cache_pressure
100
sudo sysctl -n vm.swappiness
60
```

### Description of changes proposed in this pull request:

Add zram config via https://github.com/foundObjects/zram-swap/

It might be good to enable/install early in the process to help with large package installs on limited systems.

For very limited systems that heavily rely on using _actual_ swap on disk, it looks like [Zswap](https://superuser.com/questions/1727160/zram-vs-zswap-for-lower-end-hardware) will work even better--but zram is more universally adopted and will work better on systems that don't heavily use disk swap.

### Smoke-tested on which OS or OS's:

not sure how to test single ansible roles... on Debian 13 I tried this:

```
sudoedit /etc/iiab/local_vars.yml
```

add these lines:

```
# ZRAM Swap
zram_install: True
zram_enabled: True
```

```
sudo ./runrole zram
```

### Mention a team member @username e.g. to help with code review:

@Ark74

Issue #4060, opened 2025-08-25
U25.10 strangeness [Ubuntu Questing Quokka]


jvonau:
sudo ip r
/etc/sudoers:58:14: unknown setting: 'logfile'
Defaults     logfile = /var/log/sudo.log
             ^~~~~~~~

holta:
Definitely a couple things are going crazy with Ubuntu 25.10 pre-releases in recent days (weeks?)

Even things as fundamental as the "read" command (used by bash scripts, for user input) are not working properly, presumably due to the new [sudo-rs](https://www.phoronix.com/news/Ubuntu--Now-Default-sudo-rs) ? 🙃

(Speculating wildly, this might possibly be related to Ubuntu's recent decisions to reimplement key infra [like sudo] in Rust... ?)

- #4074

jvonau:
https://github.com/iiab/iiab/pull/4074

holta:
Ubuntu Desktop 25.10's very latest Rust-based coreutils [1] [2] have also now broken a core aspect of our `sudo iiab` installer — as their new `date` command has been reimplemented to give bogus results here...

```
date +%s -r "/boot/$(ls -t /boot | head -1)"
```

https://github.com/iiab/iiab-factory/blob/01765621774c37fc73c5ef96c226a375cd078db7/iiab#L452

...in short the above command erroneously gives the actual time — instead of the timestamp of the most recently-modified file/folder in `/boot` — which in turn triggers a reboot every time (instead of proceeding to install IIAB!)

WORKAROUND, FOR NOW, if IIAB needs to be installed on (a similarly buggy) Ubuntu Desktop 25.10:

- Run `curl iiab.io/risky.txt | bash` (and subsequently `sudo iiab -r`) to proceed — which bypasses initial apt updating.

Tangentially Related:

- PR #4074 

[1] 'Ubuntu 25.10 Proceeding With Its Rust Coreutils Transition'
https://phoronix.com/news/Rust-Coreutils-Release-Pocket

[2] 'Rust Coreutils 0.2 Released With "Massive" Performance Gains, Production-Ready Ubuntu Support'
https://phoronix.com/news/Rust-Coreutils-0.2

jvonau:
https://itsfoss.com/sudo-vs-sudo-rs/

holta:
> 'Ubuntu 25.10 Proceeding With Its Rust Coreutils Transition'
https://phoronix.com/news/Rust-Coreutils-Release-Pocket

[...]

PR #4094, opened 2025-09-28
Swap [Q: what to do about RasPiOS apt pkg `dphys-swapfile` forcing apt pkg `systemd-zram-generator` to be "auto-removable" ? A: Migrate to new apt pkg `rpi-swap`]

jvonau:
### Fixes bug:
https://github.com/iiab/iiab/pull/4087#issuecomment-3326921491
RasPi hardware is the only place where dphys-swapfile is installed, exclude dphys-swapfile on RasPiOS trixie
pending

### DIFF
--- a/roles/1-prep/tasks/raspberry_pi.yml
+++ b/roles/1-prep/tasks/raspberry_pi.yml
@@ -39,7 +39,7 @@
         value: swapfile
         no_extra_spaces: true
 
-    - name: Increase swap file size (to FixedSizeMiB={{ pi_swap_file_size }}) -- e.g. to avoid kiwix-serve fails on Zero 2 W, with large ZIM files
+    - name: Increase swap file size (to FixedSizeMiB={{ pi_swap_file_size }}) -- e.g. to avoid kiwix-serve fails on Raspberry Pi Zero 2 W, with large ZIM files
       ini_file:
         path: /etc/rpi/swap.conf
         section: File
--- a/roles/1-prep/tasks/raspberry_pi.yml
+++ b/roles/1-prep/tasks/raspberry_pi.yml
@@ -22,25 +22,30 @@
   when: rtc_id is defined and rtc_id != "none" and is_ubuntu    # CLARIF: Ubuntu runs increasingly well on RPi hardware, starting in 2020 especially
 
 
-- name: 'Install packages: fake-hwclock, dphys-swapfile'
+- name: 'Install package: fake-hwclock'
   package:
     name:
       - fake-hwclock      # 2021-03-15: Missing on Ubuntu etc.  RasPiOS installs this regardless -- to save/restore system clock on machines w/o working RTC (above).
-      - dphys-swapfile    # 2021-03-15: Missing on Ubuntu etc.  RasPiOS installs this regardless -- to autogenerate and use a swap file (below).
     state: present
+
+- name: 'Install package: dphys-swapfile'
+  package:
+      - dphys-swapfile    # 2021-03-15: Missing on Ubuntu etc.  RasPiOS uses zram now -- to autogenerate and use a swap file (below).
+  when: os_ver != "raspbian-13"
 
 - name: Increase swap file size (to CONF_SWAPSIZE={{ pi_swap_file_size }} in /etc/dphys-swapfile) as kalite pip download fails
   lineinfile:
     path: "{{ etc_path }}/dphys-swapfile"
     regexp: "^CONF_SWAPSIZE"
     line: "CONF_SWAPSIZE={{ pi_swap_file_size }}"
+  when: os_ver != "raspbian-13"
 
 - name: Restart swap service "dphys-swapfile"
   #command: /etc/init.d/dphys-swapfile restart
   systemd:    # Had been...a rare/legacy service that was NOT systemd
     name: dphys-swapfile
     state: restarted
-
+  when: os_ver != "raspbian-13"
 
 #- name: Enable bluetooth in /boot/firmware/syscfg.txt on Ubuntu (needs reboot)
 #  lineinfile:
--- a/roles/1-prep/tasks/raspberry_pi.yml
+++ b/roles/1-prep/tasks/raspberry_pi.yml
@@ -24,8 +24,7 @@
 
 - name: 'Install package: fake-hwclock'
   package:
-    name:
-      - fake-hwclock      # 2021-03-15: Missing on Ubuntu etc.  RasPiOS installs this regardless -- to save/restore system clock on machines w/o working RTC (above).
+    name: fake-hwclock    # 2021-03-15: Missing on Ubuntu etc.  RasPiOS installs this regardless -- to save/restore system clock on machines w/o working RTC (above).
     state: present
 
 - name: Install and configure dphys-swapfile
--- a/roles/1-prep/tasks/raspberry_pi.yml
+++ b/roles/1-prep/tasks/raspberry_pi.yml
@@ -33,8 +33,7 @@
   block:
     - name: 'Install package: dphys-swapfile'
       package:
-        name:
-          - dphys-swapfile    # 2021-03-15: Missing on Ubuntu etc.  RasPiOS uses zram now -- to autogenerate and use a swap file (below).
+        name: dphys-swapfile    # 2021-03-15: Missing on Ubuntu etc.  RasPiOS uses zram now -- to autogenerate and use a swap file (below).
         state: present
 
     - name: Increase swap file size (to CONF_SWAPSIZE={{ pi_swap_file_size }} in /etc/dphys-swapfile) as kalite pip download fails
--- a/roles/1-prep/tasks/raspberry_pi.yml
+++ b/roles/1-prep/tasks/raspberry_pi.yml
@@ -27,8 +27,8 @@
     name: fake-hwclock    # 2021-03-15: Missing on Ubuntu etc.  RasPiOS installs this regardless -- to save/restore system clock on machines w/o working RTC (above).
     state: present
 
-- name: Install and configure dphys-swapfile
-  when: is_raspbian and os_ver is version('raspbian-13', '<')
+- name: Install and configure dphys-swapfile, if RasPiOS <= 12
+  when: is_raspbian and os_ver is version('raspbian-12', '<=')
   block:
     - name: 'Install package: dphys-swapfile'
       package:
--- a/roles/1-prep/tasks/raspberry_pi.yml
+++ b/roles/1-prep/tasks/raspberry_pi.yml
@@ -28,24 +28,25 @@
       - fake-hwclock      # 2021-03-15: Missing on Ubuntu etc.  RasPiOS installs this regardless -- to save/restore system clock on machines w/o working RTC (above).
     state: present
 
-- name: 'Install package: dphys-swapfile'
-  package:
-      - dphys-swapfile    # 2021-03-15: Missing on Ubuntu etc.  RasPiOS uses zram now -- to autogenerate and use a swap file (below).
+- name: Install and configure dphys-swapfile
   when: os_ver != "raspbian-13"
+  block:
 
-- name: Increase swap file size (to CONF_SWAPSIZE={{ pi_swap_file_size }} in /etc/dphys-swapfile) as kalite pip download fails
-  lineinfile:
-    path: "{{ etc_path }}/dphys-swapfile"
-    regexp: "^CONF_SWAPSIZE"
-    line: "CONF_SWAPSIZE={{ pi_swap_file_size }}"
-  when: os_ver != "raspbian-13"
+  - name: 'Install package: dphys-swapfile'
+    package:
+        - dphys-swapfile    # 2021-03-15: Missing on Ubuntu etc.  RasPiOS uses zram now -- to autogenerate and use a swap file (below).
 
-- name: Restart swap service "dphys-swapfile"
-  #command: /etc/init.d/dphys-swapfile restart
-  systemd:    # Had been...a rare/legacy service that was NOT systemd
-    name: dphys-swapfile
-    state: restarted
-  when: os_ver != "raspbian-13"
+  - name: Increase swap file size (to CONF_SWAPSIZE={{ pi_swap_file_size }} in /etc/dphys-swapfile) as kalite pip download fails
+    lineinfile:
+      path: "{{ etc_path }}/dphys-swapfile"
+      regexp: "^CONF_SWAPSIZE"
+      line: "CONF_SWAPSIZE={{ pi_swap_file_size }}"
+
+  - name: Restart swap service "dphys-swapfile"
+    #command: /etc/init.d/dphys-swapfile restart
+    systemd:    # Had been...a rare/legacy service that was NOT systemd
+      name: dphys-swapfile
+      state: restarted
 
 #- name: Enable bluetooth in /boot/firmware/syscfg.txt on Ubuntu (needs reboot)
 #  lineinfile:
--- a/roles/1-prep/tasks/raspberry_pi.yml
+++ b/roles/1-prep/tasks/raspberry_pi.yml
@@ -29,7 +29,7 @@
     state: present
 
 - name: Install and configure dphys-swapfile
-  when: os_ver != "raspbian-13"
+  when: is_raspbian and os_ver is version('raspbian-13', '<')
   block:
     - name: 'Install package: dphys-swapfile'
       package:
--- a/roles/1-prep/tasks/raspberry_pi.yml
+++ b/roles/1-prep/tasks/raspberry_pi.yml
@@ -31,22 +31,23 @@
 - name: Install and configure dphys-swapfile
   when: os_ver != "raspbian-13"
   block:
+    - name: 'Install package: dphys-swapfile'
+      package:
+        name:
+          - dphys-swapfile    # 2021-03-15: Missing on Ubuntu etc.  RasPiOS uses zram now -- to autogenerate and use a swap file (below).
+        state: present
 
-  - name: 'Install package: dphys-swapfile'
-    package:
-        - dphys-swapfile    # 2021-03-15: Missing on Ubuntu etc.  RasPiOS uses zram now -- to autogenerate and use a swap file (below).
+    - name: Increase swap file size (to CONF_SWAPSIZE={{ pi_swap_file_size }} in /etc/dphys-swapfile) as kalite pip download fails
+      lineinfile:
+        path: "{{ etc_path }}/dphys-swapfile"
+        regexp: "^CONF_SWAPSIZE"
+        line: "CONF_SWAPSIZE={{ pi_swap_file_size }}"
 
-  - name: Increase swap file size (to CONF_SWAPSIZE={{ pi_swap_file_size }} in /etc/dphys-swapfile) as kalite pip download fails
-    lineinfile:
-      path: "{{ etc_path }}/dphys-swapfile"
-      regexp: "^CONF_SWAPSIZE"
-      line: "CONF_SWAPSIZE={{ pi_swap_file_size }}"
-
-  - name: Restart swap service "dphys-swapfile"
-    #command: /etc/init.d/dphys-swapfile restart
-    systemd:    # Had been...a rare/legacy service that was NOT systemd
-      name: dphys-swapfile
-      state: restarted
+    - name: Restart swap service "dphys-swapfile"
+      #command: /etc/init.d/dphys-swapfile restart
+      systemd:    # Had been...a rare/legacy service that was NOT systemd
+        name: dphys-swapfile
+        state: restarted
 
 #- name: Enable bluetooth in /boot/firmware/syscfg.txt on Ubuntu (needs reboot)
 #  lineinfile:
--- a/roles/1-prep/tasks/raspberry_pi.yml
+++ b/roles/1-prep/tasks/raspberry_pi.yml
@@ -36,7 +36,7 @@
         name: dphys-swapfile    # 2021-03-15: Missing on Ubuntu etc.  RasPiOS uses zram now -- to autogenerate and use a swap file (below).
         state: present
 
-    - name: Increase swap file size (to CONF_SWAPSIZE={{ pi_swap_file_size }} in /etc/dphys-swapfile) as kalite pip download fails
+    - name: Increase swap file size (to CONF_SWAPSIZE={{ pi_swap_file_size }} in /etc/dphys-swapfile) e.g. kalite pip download failed
       lineinfile:
         path: "{{ etc_path }}/dphys-swapfile"
         regexp: "^CONF_SWAPSIZE"
--- a/roles/1-prep/tasks/raspberry_pi.yml
+++ b/roles/1-prep/tasks/raspberry_pi.yml
@@ -27,6 +27,26 @@
     name: fake-hwclock    # 2021-03-15: Missing on Ubuntu etc.  RasPiOS installs this regardless -- to save/restore system clock on machines w/o working RTC (above).
     state: present
 
+
+- name: Configure rpi-swap's /etc/rpi/swap.conf, if RasPiOS >= 13 -- ENACTED ON REBOOT
+  when: is_raspbian and os_ver is version('raspbian-13', '>=')
+  block:
+    - name: Set Mechanism=swapfile -- overriding default Mechanism=auto (i.e. Mechanism=zram+file) -- in future might Mechanism=zram+file coexist with the large swap file we need?
+      ini_file:
+        path: /etc/rpi/swap.conf
+        section: Main
+        option: Mechanism
+        value: swapfile
+        no_extra_spaces: true
+
+    - name: Increase swap file size (to FixedSizeMiB={{ pi_swap_file_size }}) -- e.g. to avoid kiwix-serve fails on Zero 2 W, with large ZIM files
+      ini_file:
+        path: /etc/rpi/swap.conf
+        section: File
+        option: FixedSizeMiB
+        value: "{{ pi_swap_file_size }}"
+        no_extra_spaces: true
+
 - name: Install and configure dphys-swapfile, if RasPiOS <= 12
   when: is_raspbian and os_ver is version('raspbian-12', '<=')
   block:
@@ -35,7 +55,7 @@
         name: dphys-swapfile    # 2021-03-15: Missing on Ubuntu etc.  RasPiOS uses zram now -- to autogenerate and use a swap file (below).
         state: present
 
-    - name: Increase swap file size (to CONF_SWAPSIZE={{ pi_swap_file_size }} in /etc/dphys-swapfile) e.g. kalite pip download failed
+    - name: Increase swap file size (to CONF_SWAPSIZE={{ pi_swap_file_size }} in /etc/dphys-swapfile) -- e.g. kalite pip download failed
       lineinfile:
         path: "{{ etc_path }}/dphys-swapfile"
         regexp: "^CONF_SWAPSIZE"
@@ -46,6 +66,7 @@
       systemd:    # Had been...a rare/legacy service that was NOT systemd
         name: dphys-swapfile
         state: restarted
+
 
 #- name: Enable bluetooth in /boot/firmware/syscfg.txt on Ubuntu (needs reboot)
 #  lineinfile:
--- a/vars/default_vars.yml
+++ b/vars/default_vars.yml
@@ -247,7 +247,7 @@ tailscale_enabled: False    # Stub var, doesn't yet do anything!
 
 # Some prefer 512MB for Zero W, others prefer 2048MB or higher for RPi 3 and 4.
 # Please see recommendations at: https://itsfoss.com/swap-size/
-pi_swap_file_size: 1024
+pi_swap_file_size: 2048
 
 
 # 2-COMMON
--- a/vars/default_vars.yml
+++ b/vars/default_vars.yml
@@ -245,9 +245,9 @@ tailscale_enabled: False    # Stub var, doesn't yet do anything!
 # (The full network stage runs after 9-LOCAL-ADDONS.  Or manually run
 # "sudo iiab-network").  Design under discussion: #2876
 
-# Some prefer 512MB for Zero W, others prefer 2048MB or higher for RPi 3 and 4.
-# Please see recommendations at: https://itsfoss.com/swap-size/
-pi_swap_file_size: 2048
+# Some prefer 512MB for Zero W, others prefer 2048MB or higher for RPi 3/4/5.
+# kiwix-serve needs >= 1024MB on Zero 2 W.  Tips: https://itsfoss.com/swap-size
+pi_swap_file_size: 1536
 
 
 # 2-COMMON


holta:
@chapmanjacobd take a look at @jvonau's work here if you have a sec?

chapmanjacobd:
looks fine to me (other than the indentation which is making the unit test fail)

It looks like `dphys-swapfile` comes from the Debian repos so, _if this wasn't in the raspberry-pi task yml_, a more holistic check might be to just see if `/proc/swaps` contains any lines

chapmanjacobd:
could also add a check for pi_swap_file_size !=0 to the block clause

chapmanjacobd (approved):

roles/1-prep/tasks/raspberry_pi.yml
```diff
@@ -22,25 +22,32 @@
...
-    name: dphys-swapfile
-    state: restarted
-
+- name: Install and configure dphys-swapfile
+  when: os_ver != "raspbian-13"
+  block:
+    - name: 'Install package: dphys-swapfile'
+      package:
+        name:
+          - dphys-swapfile    # 2021-03-15: Missing on Ubuntu etc.  RasPiOS uses zram now -- to autogenerate and use a swap file (below).
```
holta: Optionally... combine onto 1 line:

```suggestion
        name: dphys-swapfile    # 2021-03-15: Missing on Ubuntu etc.  RasPiOS uses zram now -- to autogenerate and use a swap file (below).
```

jvonau:
The point of this PR is to prevent removal of RasPiOS trixie's critical system components should someone run `apt autoremove`

roles/1-prep/tasks/raspberry_pi.yml
jvonau: I kept the same form as was present, the same could be said for fake-hwclock

roles/1-prep/tasks/raspberry_pi.yml
```diff
@@ -22,25 +22,32 @@
...
+        state: present
+
+    - name: Increase swap file size (to CONF_SWAPSIZE={{ pi_swap_file_size }} in /etc/dphys-swapfile) as kalite pip download fails
```
holta: Recommendation:

Let's try to downplay / deprecate kalite wherever we can... it's going away in favor of Kolibri:

```suggestion
    - name: Increase swap file size (to CONF_SWAPSIZE={{ pi_swap_file_size }} in /etc/dphys-swapfile) e.g. kalite pip download failed
```

roles/1-prep/tasks/raspberry_pi.yml
```diff
@@ -22,25 +22,32 @@
...
-    line: "CONF_SWAPSIZE={{ pi_swap_file_size }}"
-
-- name: Restart swap service "dphys-swapfile"
-  #command: /etc/init.d/dphys-swapfile restart
-  systemd:    # Had been...a rare/legacy service that was NOT systemd
-    name: dphys-swapfile
-    state: restarted
-
+- name: Install and configure dphys-swapfile
+  when: os_ver != "raspbian-13"
```
holta: Let's start phasing out `dphys-swapfile` rather completely — if it's now time to do that — and not have it come back to bite us with `raspbian-14` etc (-:

```suggestion
  when: is_raspbian and os_ver is version('raspbian-13', '<')
```

CLARIFS:

1. I've included `is_raspbian` to be extra safe here, so lexicographic order string comparison is safe.  Especially given there are future risks (if not present risks!) someone might run raspberry_pi.yml on Ubuntu on a physical Raspberry Pi.
 
2. Lots to learn hereunder, if we ever need other/different/formal ways to do version comparison (of strings) precisely:

   https://docs.ansible.com/ansible/latest/collections/ansible/builtin/version_test.html#examples

holta:
@jvonau thanks for this PR which looks excellent, e.g. if you and @chapmanjacobd (and/or others!) agree, then we should go with it.

roles/1-prep/tasks/raspberry_pi.yml
jvonau: Not having booted Ubuntu on RasPi hardware in a while I'm not sure if zram is currently enabled there, but I would think that swap might be needed for the same reasons it was required on raspbian in the past and was installed in the case of Ubuntu.

holta: > Not having booted Ubuntu on RasPi hardware in a while I'm not sure if zram is currently enabled there

@Ark74 do you happen to know, if `zram` is enabled as part of Ubuntu 25.04 and/or 25.10 on RPi hardware?

(e.g. no matter if Ubuntu Server or Ubuntu Desktop, any data points helpful!)

holta:
@jvonau @Ark74 @chapmanjacobd let's give this PR a shot.  Especially as `dphys-swapfile` repeatedly fails on RasPiOS 13 Lite on Zero 2 W:

```
root@box:~# systemctl status dphys-swapfile.service
× dphys-swapfile.service - dphys-swapfile - set up, mount/unmount, and delete a swap file
     Loaded: loaded (/usr/lib/systemd/system/dphys-swapfile.service; enabled; preset: enabled)
     Active: failed (Result: exit-code) since Fri 2025-10-03 02:08:07 EDT; 1min 29s ago
 Invocation: 33ff63e1ee0447ae86fe2454858dcbde
       Docs: man:dphys-swapfile(8)
   Main PID: 567 (code=exited, status=255/EXCEPTION)
        CPU: 220ms

Oct 03 02:08:06 box systemd[1]: Starting dphys-swapfile.service - dphys-swapfile - set up, mount/unmount, and delete a swap file...
Oct 03 02:08:07 box dphys-swapfile[510]: want /var/swap=1024MByte, checking existing: keeping it
Oct 03 02:08:07 box dphys-swapfile[574]: swapon: /var/swap: swap format pagesize does not match. (Use --fixpgsz to reinitialize it.)
Oct 03 02:08:07 box dphys-swapfile[574]: swapon: /var/swap: swapon failed: Invalid argument
Oct 03 02:08:07 box systemd[1]: dphys-swapfile.service: Main process exited, code=exited, status=255/EXCEPTION
Oct 03 02:08:07 box systemd[1]: dphys-swapfile.service: Failed with result 'exit-code'.
Oct 03 02:08:07 box systemd[1]: Failed to start dphys-swapfile.service - dphys-swapfile - set up, mount/unmount, and delete a swap file.
```

PS We can refine later if certain OS's zram isn't sufficient for the swap-like tuning we need.

chapmanjacobd (approved):
seems reasonable as long as os_version knows about editions (eg. lite)

holta:
> seems reasonable as long as os_version knows about editions (eg. lite)

IIAB's `os_ver` (Ansible variable, derived from [local_facts.fact](https://github.com/iiab/iiab/blob/414fbcb1f23eb34899b703838906b90dcb520ae6/scripts/local_facts.fact#L40-L57)) does not concern itself with delineating between headless (e.g. "Lite") or graphical desktop versions of OS's.

I don't think that really matters here... (or, is there reason to believe otherwise??)

holta:
Can we go ahead with this PR?

Either way, can someone help me fix the question in the ticket title / subject line to better summarize?

Or, if we really need to refine this PR further, e.g. if we really want to preserve `dphys-swapfile` — having it coexist alongside zram if that's truly needed — then under what RAM or OS conditions should dphys-swapfile be preserved e.g. with [pi_swap_file_size: 1024](https://github.com/iiab/iiab/blob/414fbcb1f23eb34899b703838906b90dcb520ae6/vars/default_vars.yml#L248-L250) ?

Quick data point if it helps — on the official new release of 64-bit "RasPiOS 13 Trixie with desktop" on a 4GB RAM RPi 5, using the latest IIAB **_without_** this PR installed — after reboot:

```
root@box:~# systemctl status dphys-swapfile
● dphys-swapfile.service - dphys-swapfile - set up, mount/unmount, and delete a swap file
     Loaded: loaded (/usr/lib/systemd/system/dphys-swapfile.service; enabled; preset: enabled)
     Active: active (exited) since Fri 2025-10-03 03:35:19 EDT; 9min ago
 Invocation: 74e9b3838e4b4bbc852e564b006602bf
       Docs: man:dphys-swapfile(8)
   Main PID: 784 (code=exited, status=0/SUCCESS)
        CPU: 32ms

Oct 03 03:35:18 box systemd[1]: Starting dphys-swapfile.service - dphys-swapfile - set up, mount/unmount, and delete a swap file...
Oct 03 03:35:19 box dphys-swapfile[701]: want /var/swap=1024MByte, checking existing: keeping it
Oct 03 03:35:19 box systemd[1]: Finished dphys-swapfile.service - dphys-swapfile - set up, mount/unmount, and delete a swap file.

root@box:~# lsblk
NAME        MAJ:MIN RM  SIZE RO TYPE MOUNTPOINTS
mmcblk0     179:0    0 59.6G  0 disk
├─mmcblk0p1 179:1    0  512M  0 part /boot/firmware
└─mmcblk0p2 179:2    0 59.1G  0 part /
zram0       254:0    0    2G  0 disk [SWAP]

root@box:~# apt list systemd-zram-generator
systemd-zram-generator/stable,now 1.2.1-2 arm64 [installed,auto-removable]
systemd-zram-generator/stable 1.2.1-2 armhf

root@box:~# zramctl
NAME       ALGORITHM DISKSIZE DATA COMPR TOTAL STREAMS MOUNTPOINT
/dev/zram0 zstd            2G  16K   69B   48K       4 [SWAP]
```

iiab-diagnostics: https://paste.centos.org/view/38d24014

holta:
> @jvonau @Ark74 @chapmanjacobd let's give this PR a shot. Especially as `dphys-swapfile` repeatedly fails on RasPiOS 13 Lite on Zero 2 W:

ASIDE:

- Just for now I've manually worked around the failing service on Zero 2 W, simply by running `apt purge dphys-swapfile` (to force the same thing as this PR in the end!)

- Question is if it's a good idea for IIAB to start walking away from `dphys-swapfile` and [pi_swap_file_size: 1024](https://github.com/iiab/iiab/blob/414fbcb1f23eb34899b703838906b90dcb520ae6/vars/default_vars.yml#L248-L250) more broadly — starting with the new RasPiOS 13 Trixie released 1-2 days ago?

<p align=center>image

jvonau:
I would prefer if is_raspbian be dropped from the conditional for the reason [stated]( https://github.com/iiab/iiab/pull/4094#discussion_r2386254990)

jvonau:
Seem like both zram and dphys-swapfile can co-exist but it is the apt packaging from RaPiOS that is the noted problem. smells like an upstream issue that should be filed.

roles/1-prep/tasks/raspberry_pi.yml
holta: Thanks @jvonau and All!

- Possibly Ubuntu Server 25.10 (e.g. final release available in 6 days, on Oct 9) should be investigated before acting on this PR ?  (Even though nobody's taking [that particular interim release](https://github.com/iiab/iiab/issues/4060) as the final word, still it's leading to 26.04 LTS :-)

- Not that anyone would-or-should run Ubuntu Server on Zero 2 W of course!

- It feels like we should understand this all in a bit more depth, before moving away from [pi_swap_file_size: 1024](https://github.com/iiab/iiab/blob/414fbcb1f23eb34899b703838906b90dcb520ae6/vars/default_vars.yml#L248-L250) — which might not be wise? 😇

- _SUGGESTION: @EMG70 and others might be able to help us test certain key swap scenarios, before mid-October ideally._

holta:
> I would prefer if is_raspbian be dropped from the conditional for the reason [stated](https://github.com/iiab/iiab/pull/4094#discussion_r2386254990)

@jvonau thanks.  I'll see if I can try to find someone to help us all investigate & understand better:

- https://github.com/iiab/iiab/pull/4094#discussion_r2402000107

chapmanjacobd:
It might just be enough to not run autoremove

```
rg -i --no-heading --no-line-number -j1 autoremove       
scripts/iiab-update:        echo -e "\n\e[4mNow running: apt autoremove -y\e[0m\n"      
scripts/iiab-update:        apt autoremove -y
```

Did you run iiab-update on the Pi Zero 2?

Or is dphys-swap just broken on that platform? https://github.com/iiab/iiab/pull/4094#issuecomment-3364414527

Sounds like there is more than one issue here. The packaging issue and something else

holta:
Just A Hunch For Now?

- Seems we should at least try to preserve **pi_swap_file_size: 1024** _(even though it's been causing a whole lot of trouble recently!)_

- @Ark74 asks if systemd's own swap manager is the "correct" way forward, and if the latest distros might be running into that?  Anyway something to investigate, shining a light on larger questions 🙏

- `zram` improvements across all distros are great, but that's largely unrelated (both ideas share the word "swap" but not a whole lot else!) 📛

- _THANKS everyone, until we know more!!_ 😀

chapmanjacobd:
> swapon: /var/swap: swap format pagesize does not match. (Use --fixpgsz to reinitialize it.)

Is the Zero 2 okay if you reboot?

holta:
> Did you run iiab-update on the Pi Zero 2?

I did not.

Just FYI: As documented in https://FAQ.IIAB.IO, iiab-update serves a pretty narrow use case, to upgrade recently installed IIAB's at your own risk.

> Or is dphys-swap just broken on that platform? [#4094 (comment)](https://github.com/iiab/iiab/pull/4094#issuecomment-3364414527)

Could be (:

> Sounds like there is more than one issue here. The packaging issue and something else

Likely 🙃

holta:
Good News: For now I've reinstalled `apt install dphys-swapfile` on RasPiOS 13 Lite on Zero 2 W, and everything seems to be A-OK with `dphys-swapfile` (set to [pi_swap_file_size: 1024](https://github.com/iiab/iiab/blob/414fbcb1f23eb34899b703838906b90dcb520ae6/vars/default_vars.yml#L248-L250)) and `zram` coexisting — after ~5 reboots anyway: (hurray!!)

- Sorry everybody for the false negative earlier this morning — quite possibly this PR is just not needed as is — as we keep monitoring the upstream packaging issue to see what we can/should do about it?!
- As a precaution, `apt autoremove` has not been run, to keep apt package `system-zram-generator` in place for now.
- I'll keep monitoring & evaluating!

```
root@box:~# systemctl --failed
  UNIT LOAD ACTIVE SUB DESCRIPTION

0 loaded units listed.

root@box:~# systemctl status dphys-swapfile
● dphys-swapfile.service - dphys-swapfile - set up, mount/unmount, and delete a swap file
     Loaded: loaded (/usr/lib/systemd/system/dphys-swapfile.service; enabled; preset: enabled)
     Active: active (exited) since Fri 2025-10-03 11:14:09 EDT; 2min 58s ago
 Invocation: 6b7e4e9f7f64419d89df4fff5425661a
       Docs: man:dphys-swapfile(8)
   Main PID: 571 (code=exited, status=0/SUCCESS)
        CPU: 207ms

Oct 03 11:14:08 box systemd[1]: Starting dphys-swapfile.service - dphys-swapfile - set up, mount/unmount, and delete a swap file...
Oct 03 11:14:08 box dphys-swapfile[510]: want /var/swap=1024MByte, checking existing: keeping it
Oct 03 11:14:09 box systemd[1]: Finished dphys-swapfile.service - dphys-swapfile - set up, mount/unmount, and delete a swap file.

root@box:~# free -h
               total        used        free      shared  buff/cache   available
Mem:           416Mi       191Mi       119Mi       724Ki       157Mi       224Mi
Swap:          1.2Gi       236Mi       995Mi

root@box:~# apt list systemd-zram-generator
systemd-zram-generator/stable,now 1.2.1-2 arm64 [installed,auto-removable]
systemd-zram-generator/stable 1.2.1-2 armhf
```

iiab-diagnostics: https://paste.centos.org/view/685e8df2

<br>

_As an ASIDE... other systemd services very occasionally fail on Zero 2 W boot (maybe 1 out of 5 in each case??) — but that's likely very normal on such low-end hardware:_

```
root@box:~# systemctl status user@1000
× user@1000.service - User Manager for UID 1000
     Loaded: loaded (/usr/lib/systemd/system/user@.service; static)
    Drop-In: /usr/lib/systemd/system/user@.service.d
             └─10-login-barrier.conf
     Active: failed (Result: signal) since Fri 2025-10-03 11:11:14 EDT; 57s ago
   Duration: 2.959s
 Invocation: 6b0c53e8d76144b0985e73c7706e6102
       Docs: man:user@.service(5)
    Process: 1481 ExecStart=/usr/lib/systemd/systemd --user (code=killed, signal=KILL)
   Main PID: 1481 (code=killed, signal=KILL)
     Status: "Ready."
        CPU: 1.424s

Oct 03 11:11:11 box systemd[1]: Started user@1000.service - User Manager for UID 1000.
Oct 03 11:11:11 box systemd[1481]: Started mpris-proxy.service - Bluetooth mpris proxy.
Oct 03 11:11:11 box systemd[1481]: Reached target default.target - Main User Target.
Oct 03 11:11:11 box systemd[1481]: Startup finished in 1.330s.
Oct 03 11:11:11 box systemd[1481]: Created slice session.slice - User Core Session Slice.
Oct 03 11:11:11 box systemd[1481]: Starting dbus.service - D-Bus User Message Bus...
Oct 03 11:11:11 box systemd[1481]: Started dbus.service - D-Bus User Message Bus.
Oct 03 11:11:14 box systemd[1]: user@1000.service: Main process exited, code=killed, status=9/KILL
Oct 03 11:11:14 box systemd[1]: user@1000.service: Failed with result 'signal'.
Oct 03 11:11:14 box systemd[1]: user@1000.service: Consumed 1.424s CPU time.
```

```
root@box:~# systemctl status cloud-init-main
× cloud-init-main.service - Cloud-init: Single Process
     Loaded: loaded (/usr/lib/systemd/system/cloud-init-main.service; enabled; preset: enabled)
     Active: failed (Result: signal) since Fri 2025-10-03 11:26:45 EDT; 2min 50s ago
   Duration: 18.379s
 Invocation: d6e1ba1b89d6473da7454f188e4720d6
    Process: 282 ExecStart=/usr/bin/cloud-init --all-stages (code=killed, signal=KILL)
   Main PID: 282 (code=killed, signal=KILL)
     Status: "Waiting on external services to complete before starting the final stage."
        CPU: 5.645s

Oct 03 11:26:28 box cloud-init[454]: ci-info: +++++++++++++++++++Route IPv6 info+++++++++++++++++++
Oct 03 11:26:28 box cloud-init[454]: ci-info: +-------+-------------+---------+-----------+-------+
Oct 03 11:26:28 box cloud-init[454]: ci-info: | Route | Destination | Gateway | Interface | Flags |
Oct 03 11:26:28 box cloud-init[454]: ci-info: +-------+-------------+---------+-----------+-------+
Oct 03 11:26:28 box cloud-init[454]: ci-info: +-------+-------------+---------+-----------+-------+
Oct 03 11:26:28 box cloud-init[454]: 2025-10-03 15:26:28,724 - DataSourceNoCloud.py[WARNING]: DataSourceNoCloudNet  only uses seeds starting with ('http://', 'https://', 'ftp://', 'ftps://') - file:///boot/firmware is not valid.
Oct 03 11:26:39 box cloud-init[454]: Cloud-init v. 25.2 running 'modules:config' at Fri, 03 Oct 2025 15:26:39 +0000. Up 27.65 seconds.
Oct 03 11:26:45 box systemd[1]: cloud-init-main.service: Main process exited, code=killed, status=9/KILL
Oct 03 11:26:45 box systemd[1]: cloud-init-main.service: Failed with result 'signal'.
Oct 03 11:26:45 box systemd[1]: cloud-init-main.service: Consumed 5.645s CPU time.
```

chapmanjacobd:
It looks like we could add this:

    sudo apt-mark manual systemd-zram-generator

(`apt-mark` is part of the `apt` package)

```ansible
- name: Mark systemd-zram-generator as manually installed
  apt_mark:
    package: systemd-zram-generator
    state: manual
  ignore_errors: true
```

holta:
> It looks like we could add this:
> 
> ```
> sudo apt-mark manual systemd-zram-generator
> ```

Done!

- PR #4104

Let's consider the above as an interim patch, until upstream hopefully cleans up its apt/deb packaging.

jvonau:
AI Overview
"RPI-swap" refers to two different things: a new Raspberry Pi system package that uses zram for swap memory and is designed to improve performance and prolong SD card life, and a type of inflation swap where a fixed payment is exchanged for a payment linked to the Retail Prices Index (RPI), used in financial markets to trade inflation risk. It's crucial to differentiate between the two depending on your context, as one is a technical tool for a Raspberry Pi and the other is a financial instrument. 
1. RPI-swap for Raspberry Pi (Technical) 

    Purpose:
    Replaces the traditional [dphys-swapfile](https://www.google.com/search?client=ubuntu-sn&channel=fs&cs=0&sca_esv=045a62e2ae8e6760&q=dphys-swapfile&sa=X&ved=2ahUKEwjb0dC1h4mQAxXKjYkEHcqOOioQxccNegQIEBAB&mstk=AUtExfAVZh9px8SmMeyJ90U4kv6iyKXt05iVc-_w85X_fgUr_ReMZT7YP37qHBGCm3wJOZ9ezVVK5YtNEk0BHkISfLvkv32bSj0bXy-omt-FwQY8NqfMDSvk_H4K968xTwmJsbZX3zB6XJ0hvd5Kyeev6Y63xaU7apiWzpQeQsVM3psuvZI&csui=3) to manage swap memory on a Raspberry Pi, improving performance and SD card lifespan.
    How it works:
    Uses zram, a compressed RAM disk, for swap. This reduces writes to the SD card and can speed up boot times on memory-constrained devices.
    Configuration:
    Details are found in the /etc/rpi/swap.conf file or the swap.conf man page.

jvonau:
AI Overview
The file /etc/rpi/swap.conf is the configuration file for rpi-swap, a service that provides zram-based swap on newer Raspberry Pi OS versions, replacing the older dphys-swapfile. It's used to manage a RAM-based, compressed storage device as swap to extend available memory without writing to the SD card, which helps with performance and the longevity of the SD card. To configure swap settings, you would edit this file to adjust zram parameters or set a fallback swapfile location if needed.  
What is it?

    /etc/rpi/swap.conf configures rpi-swap, which is a more modern and efficient way to handle swap space on Raspberry Pis compared to older methods like dphys-swapfile. 

It primarily uses [zram](https://www.google.com/search?client=ubuntu-sn&channel=fs&cs=0&sca_esv=045a62e2ae8e6760&q=zram&sa=X&ved=2ahUKEwjFw_TCh4mQAxWSkokEHd7rMeYQxccNegQIEhAB&mstk=AUtExfBM31JxvyYghcLEWHNd2v0HG8N6ZDlkbuWEAYc26NJphgKZAKbxKraJjQ3RnTV0Nkmvy5fBe3YZaudmNlnUe4m3oWsD-KLNOHSzs7tcxY3XMBNXVc3NKV7SJmeK7mrUk4rj29fvOG1E_wHmKAGJnZYSUtU01zEsMrMjwVGFxBOfH8ikEaa18U22qa7nx889mZ7G&csui=3), which creates a compressed swap area directly in RAM, instead of using the SD card. 

Why is it used?

    Performance:
    Using zram provides faster swap access than an SD card, which is crucial for memory-intensive tasks and overall system responsiveness. 

SD Card Longevity:
By avoiding constant writes to the SD card for swap, it reduces wear and tear, extending the life of the flash storage. 
Increased Effective RAM:
It effectively increases the amount of memory available to the system, preventing crashes from out-of-memory errors, especially on RAM-constrained models. 

How to configure it: 

    Open the file: Use a text editor like nano or vi with sudo privileges. 

Code

    sudo nano /etc/rpi/swap.conf

    Adjust settings:
    The file contains systemd-style INI-like configuration settings. 

    CONF_SWAPFILE: You can specify a location for a traditional swapfile if you want to use one, though the default is zram-based. 

CONF_SWAPSIZE: This sets the desired size of the swap space. 
CONF_MAXSWAP: This sets a maximum limit on the swap size. 

Apply changes:
After editing, the changes will take effect after a reboot, or you can apply them immediately with the appropriate service restart commands. 

Example:
For example, to change the swap size, you might open the file and change a line like CONF_SWAPSIZE=100 to CONF_SWAPSIZE=1024 to set a 1024MB (1GB) swap file. Then, you would typically run sudo dphys-swapfile setup (for older systems) or restart the relevant service to apply the new size. 
Note: On some systems and newer Raspberry Pi OS versions, the default configuration might be zram-based, which uses RAM for swap instead of the SD card by default.

jvonau:
> @jvonau @Ark74 @chapmanjacobd let's give this PR a shot. Especially as `dphys-swapfile` repeatedly fails on RasPiOS 13 Lite on Zero 2 W:
> 
> ```
> root@box:~# systemctl status dphys-swapfile.service
> × dphys-swapfile.service - dphys-swapfile - set up, mount/unmount, and delete a swap file
>      Loaded: loaded (/usr/lib/systemd/system/dphys-swapfile.service; enabled; preset: enabled)
>      Active: failed (Result: exit-code) since Fri 2025-10-03 02:08:07 EDT; 1min 29s ago
>  Invocation: 33ff63e1ee0447ae86fe2454858dcbde
>        Docs: man:dphys-swapfile(8)
>    Main PID: 567 (code=exited, status=255/EXCEPTION)
>         CPU: 220ms
> 
> Oct 03 02:08:06 box systemd[1]: Starting dphys-swapfile.service - dphys-swapfile - set up, mount/unmount, and delete a swap file...
> Oct 03 02:08:07 box dphys-swapfile[510]: want /var/swap=1024MByte, checking existing: keeping it
> Oct 03 02:08:07 box dphys-swapfile[574]: swapon: /var/swap: swap format pagesize does not match. (Use --fixpgsz to reinitialize it.)
> Oct 03 02:08:07 box dphys-swapfile[574]: swapon: /var/swap: swapon failed: Invalid argument
> Oct 03 02:08:07 box systemd[1]: dphys-swapfile.service: Main process exited, code=exited, status=255/EXCEPTION
> Oct 03 02:08:07 box systemd[1]: dphys-swapfile.service: Failed with result 'exit-code'.
> Oct 03 02:08:07 box systemd[1]: Failed to start dphys-swapfile.service - dphys-swapfile - set up, mount/unmount, and delete a swap file.
> ```
> 
> PS We can refine later if certain OS's zram isn't sufficient for the swap-like tuning we need.

Seems like rpi-swap and dphys-swapfile share the same path for the swapfile at /var/swap and on the W2 there was some swap configured? Was zram active?

jvonau:
https://forums.raspberrypi.com/viewtopic.php?t=390708

holta:
Almost a decade later, does IIAB need to tidy up (or modernize) how we invoke things like `dphys-swapfile` with [pi_swap_file_size: 1024](https://github.com/iiab/iiab/blob/414fbcb1f23eb34899b703838906b90dcb520ae6/vars/default_vars.yml#L248-L250) — is that (maybe) something we should be driving for in 2025, now that RasPiOS 13 Trixie is right now becoming mainline??

> on the [Zero 2 W] there was some swap configured?

Just the usual (1024 MB) as configured by raspberry_pi.yml (this PR was not used).
As confirmed by the swap line (lower line) in the output of `free -h`.

> Was zram active?

Yes.
As confirmed by `lsblk` and `zramctl`.

holta:
> Seems like rpi-swap and dphys-swapfile share the same path for the swapfile at /var/swap

We should dig deeper for sure! ⚠️

And on the bright side, the problem completely went away during battle-testing of RasPiOS 13 Trixie on Zero 2 W today, with at least 10 additional reboots, and many different WiFi client devices / browsers testing the machine! ✅

jvonau:
>>>Oct 03 02:08:07 box dphys-swapfile[510]: want /var/swap=1024MByte, checking existing: keeping it
Oct 03 02:08:07 box dphys-swapfile[574]: swapon: /var/swap: swap format pagesize does not match. (Use --fixpgsz to reinitialize it.)
> > Seems like rpi-swap and dphys-swapfile share the same path for the swapfile at /var/swap
> 
> We should dig deeper for sure! ⚠️
> 

Something initialized some swap in the same place at /var/swap, you need to dig deeper rather than generalize to be able to explain the size mismatch that was noted... back over to you.

holta:
I don't know if that `dphys-swapfile` problem I saw ~24h ago is easily reproducible?

- But if it is, e.g. if it's confirmed IIAB microSD cards moved from "Big Iron" (e.g. RPi 5 or similar) to RPi Zero 2 W all / routinely have this problem — then yes it'd be great to understand this better.

- If for no other reason, than to at least be able to caution implementers.

holta:
> https://forums.raspberrypi.com/viewtopic.php?t=390708

Thanks @jvonau for forwarding the above crucial link. ✅

- DECENT NEWS: I'm not happy with the need to step away from `zram` on Zero 2 W (for now anyway, hopefully not for the long term!?) BUT in short... I was forced to replace `Mechanism=auto` (equivalent to `Mechanism=zram+file`) in favor of `Mechanism=swapfile` within `/etc/rpi/swap.conf` below — so as to restore IIAB's 1024MB swap file thanks to `FixedSizeMiB=1024` — all of which is critical for kiwix-serve (our systemd service) to be able to run without crashing when folks browse large ZIM files!

- BETTER NEWS: With `dphys-swapfile` officially being deprecated by RasPiOS (in favor of apt package `rpi-swap` on RasPiOS 13 Trixie) — the following at least fixes any/all `dphys-swapfile` errors, moving us all forward for much smoother operation — according to my tests on on both RPi 5 and Zero 2 W...

```
root@box:~# cat /etc/rpi/swap.conf
#  This file is part of rpi-swap.
#
#  Defaults are provided as commented-out options. Local configuration
#  should be created by either modifying this file, or by creating "drop-ins" in
#  the swap.conf.d/ subdirectory. The latter is generally recommended.
#
#  See swap.conf(5) for details.

[Main]
#Mechanism=auto
Mechanism=swapfile

[File]
#Path=/var/swap
#RamMultiplier=1
#MaxSizeMiB=2048
#MaxDiskPercent=50
#FixedSizeMiB=
FixedSizeMiB=1024

[Zram]
#RamMultiplier=1
#MaxSizeMiB=2048
#FixedSizeMiB=
# Writeback settings (for zram+file mechanism):
#WritebackTrigger=auto
#WritebackInitialDelay=180min
#WritebackPeriodicInterval=24h
```

RECAP: I'll [revise this PR #4094] so that raspberry_pi.yml puts `/etc/rpi/swap.conf` in place just as above.

- And hopefully @chapmanjacobd can help us add `zram` back into the mix too?  In due course or when time allows: _e.g. if zram can really be combined with a large (>= 1024 MB) swap file) and their coexisting proves very solid — on low-end Raspberry Pis (e.g. Zero 2 W) especially!?_ 🌞

jvonau:
My question would be Mechanism=auto should result in zram and swapfile being used correct? Now in the past I posted the config file for zram where a percentage the ram was allocated. Now with the above the default swapfile looks to be 2048MB why bend the default to something other than what other upstream deems to be adequate? Take a hands off approach and evaluate. I'll bet the farm that the `swapon: /var/swap: swap format pagesize does not match. (Use --fixpgsz to reinitialize it.)` error above is because the size is being reduced given the two packages share the same path. Boot the install and evaluate if present and how large /var/swap is on the stock install.

holta:
> My question would be Mechanism=auto should result in zram and swapfile being used correct?

That's exactly what I thought, and was hoping for.  But no luck (I tried hard, but could not get it to work... kiwix-serve kept crashing, as the swap space was insufficient on Zero 2 W).

Maybe someone else has better luck in future?  And/or RasPiOS developers might need to refine their new `rpi-swap` package?

> default swapfile looks to be 2048MB

That's not right.  I suspect you read `MaxSizeMiB=2048` in a hurry — that's merely a maximum — e.g. when using `RamMultiplier=1` on machines with higher amounts of RAM.

CLARIF: Errors like "swapon: /var/swap: swap format pagesize does not match. (Use --fixpgsz to reinitialize it.)" no longer appear on RasPiOS 13 Trixie, as we move away from `dphys-swapfile` which RasPiOS is deprecating — in favor of their new apt package `rpi-swap`.

jvonau:
Then I reported that zram was [active](https://github.com/iiab/iiab/pull/4087#issuecomment-3311435392) while testing I noted that top had reported more swap than the available from zram (Think there is a paste of the diagnostics from my machine but I'm not going out of the way to prove myself). Therefore the suggestion of checking [/proc/swaps](https://github.com/iiab/iiab/pull/4087#issuecomment-3314270417) for what is active becomes pretty important.

holta:
Yeah I'm not thrilled that we seem to have to relegate `zram` (for now at least, unless @chapmanjacobd finds something better!?)   Example on RPi 5:

```
root@box:~# cat /proc/swaps
Filename                                Type            Size            Used            Priority
/var/swap                               file            1048560         0               100

root@box:~# swapon --show
NAME      TYPE  SIZE USED PRIO
/var/swap file 1024M   0B  100

root@box:~# zramctl

root@box:~# lsblk
NAME        MAJ:MIN RM   SIZE RO TYPE MOUNTPOINTS
mmcblk0     179:0    0 238.3G  0 disk
├─mmcblk0p1 179:1    0   512M  0 part /boot/firmware
└─mmcblk0p2 179:2    0 237.8G  0 part /
zram0       254:0    0     0B  0 disk
```

But at least this new `rpi-swap` apt package approach brings us into line with RasPiOS's new regime (and demonstrably avoids `dphys-swapfile` crashiness / collisions with other/similar packages, Yay!)

chapmanjacobd:
> run without crashing when folks browse large ZIM files

@holta, do you have a set of steps for how to replicate this?

I think it would be okay to go with only swapfile for now and see how things evolve on the RPi side of things (ie. any patches to `rpi-swap`*) over the coming weeks

\* not that there is anything wrong with rpi-swap _per se_ but the packaging situation has more complexity than I would have anticipated--ie. why a separate program instead of distribution configuration, etc

jvonau:
Top should show the total amount of swap (file+zram)
>jvonau@pi500:/opt/iiab/iiab $ zramctl
NAME       ALGORITHM DISKSIZE DATA COMPR TOTAL STREAMS MOUNTPOINT
/dev/zram0 zstd          3.9G  16K   64B   48K       4 [SWAP]

>jvonau@pi500:/opt/iiab/iiab $ systemctl status dphys-swapfile.service 
● dphys-swapfile.service - dphys-swapfile - set up, mount/unmount, and delete a swap file
     Loaded: loaded (/usr/lib/systemd/system/dphys-swapfile.service; enabled; preset: enabled)
     Active: active (exited) since Sat 2025-10-04 11:00:32 CDT; 8h ago
 Invocation: bff5d525096e4e678941c03ef493f11e
       Docs: man:dphys-swapfile(8)
    Process: 634 ExecStart=/sbin/dphys-swapfile setup (code=exited, status=0/SUCCESS)
    Process: 714 ExecStart=/sbin/dphys-swapfile swapon (code=exited, status=0/SUCCESS)
   Main PID: 714 (code=exited, status=0/SUCCESS)
        CPU: 34ms
Oct 04 11:00:32 pi500 systemd[1]: Starting dphys-swapfile.service - dphys-swapfile - set up, mount/unmount, and delete a swap file...
Oct 04 11:00:32 pi500 dphys-swapfile[634]: want /var/swap=1024MByte, restricting to 50% of remaining disk size: 733MBytes, checking existing: ke>
Oct 04 11:00:32 pi500 systemd[1]: Finished dphys-swapfile.service - dphys-swapfile - set up, mount/unmount, and delete a swap file.

> ls -la /var/swap
-rw------- 1 root root 768606208 Sep 30 13:01 /var/swap

>Tasks: 226 total,   2 running, 224 sleeping,   0 stopped,   0 zombie
%Cpu(s):  0.2 us,  0.0 sy,  0.0 ni, 99.8 id,  0.0 wa,  0.0 hi,  0.0 si,  0.0 st 
MiB Mem :   8059.0 total,   6686.6 free,    689.3 used,    854.2 buff/cache     
MiB Swap:   4762.5 total,   4762.5 free,      0.0 used.   7369.7 avail Mem

4762.5 ~= 733 + 3.9*1024 close enough for me to say both are active.

holta:
> @holta, do you have a set of steps for how to replicate this?

@chapmanjacobd install a minimal IIAB with Kiwix only — e.g. type '0' (zero, a.k.a. "unittest") when IIAB's "1-line installer" asks you for 1 Small, 2 Medium or 3 Large — then [set these 2 variables](https://github.com/iiab/iiab/blob/bd1f4954f5ddc9deb728735c070cb5dcaa1b5460/vars/local_vars_unittest.yml#L297-L298) in /etc/iiab/local_vars.yml :

```
kiwix_install: True
kiwix_enabled: True
```

Proceed to install IIAB software.

Now, if you can stand to download 79G: (run as root!)

```
cd /library/zims/content

wget https://download.kiwix.org/zim/ted/ted_mul_all_2025-08.zim

iiab-make-kiwix-lib
```

Now insert the IIAB microSD card into a Raspberry Pi Zero 2 W (just 512 MB RAM!), and **reboot twice** if card's being moved from a fancier RPi (optional, but if you want the IIAB WiFi hotspot to work!)

Finally, use any web browser to try visiting this ZIM file off of http://hostname-or-IP/kiwix — browsing over Ethernet to the Zero 2 W is usually much less hassle if you can.

RESULT: `kiwix-serve` (systemd service) will crash if swap is insufficient 🙃 — you can monitor kiwix-serve crashing (and then restarting itself) using:

```
journalctl -f -u kiwix-serve
```

_ASIDE: Just for now disregard that... even when this large ZIM file ("All TED videos") does not crash kiwix-serve — i.e. when swap file is sufficient — it rather obscenely takes ALMOST 3 FULL MINUTES to fully appear in your browser!_ 😭 😭

holta:
> I think it would be okay to go with only swapfile for now and see how things evolve on the RPi side of things (ie. any patches to rpi-swap*) over the coming weeks

Let's give this PR a shot, as RasPiOS pushes us into the future: I've tested on both RPi 5 and Zero 2 W.

We can evolve en route as we learn more, hopefully bringing zram back into the picture in due course!

holta:
Just FYI @jvonau Ubuntu too, has moved away from `dphys-swapfile` — one more reason to believe it's time for IIAB to do likewise:

<p align=center>
image

holta:
Corollary: the following PR can in fact be reverted... as soon as this (PR #4094) is merged:

- PR #4104

(as there will no longer be a need to protect apt package `systemd-zram-generator` from auto-removal!)

jvonau:
What Ubuntu is doing with swap https://bugs.launchpad.net/ubuntu/+source/ubuntu-raspi-settings/+bug/2116275

holta:
> What Ubuntu is doing with swap https://bugs.launchpad.net/ubuntu/+source/ubuntu-raspi-settings/+bug/2116275

Thanks @jvonau.  I'm hoping Ubuntu 26.04 LTS pre-releases shed more light starting in November (on many fronts...)

PR #4095, opened 2025-09-28
tailscale use enable-or-disable.yml

jvonau:
### Fixes bug:
I dislike things that call home automatically, prevent unwanted upstream contact till requested.
### Description of changes proposed in this pull request:
respect tailscale_enabled: False
### Smoke-tested on which OS or OS's:
pending

PR #4096, opened 2025-09-28
Future: Use one line installer from https://tailscale.com/download/linux

jvonau:
### Fixes bug:
Matches upstream usage to auto detect supported distros instead of using older binaries from a distro that IIAB has dropped support for. ie U22.04 jammy 
make use of one liner found at https://tailscale.com/download/linux
Pending on Ubuntu VM

https://github.com/tailscale/tailscale/blob/main/scripts/installer.sh#L397
https://pkgs.tailscale.com/stable/debian/trixie/installer-supported returns OK as supported
https://pkgs.tailscale.com/stable/raspbian/buster/installer-supported returns OK as supported
https://pkgs.tailscale.com/stable/raspbian/trixie/installer-supported returns 404 as unsupported 

This has to wait will upstream adds support for the newest RasPiOS..

### DIFF
--- a/roles/tailscale/tasks/install.yml
+++ b/roles/tailscale/tasks/install.yml
@@ -3,20 +3,16 @@
   register: df1
 
 
-- name: "Set up apt source (jammy) in /etc/apt/sources.list.d/tailscale.list and its key /usr/share/keyrings/tailscale-archive-keyring.gpg, to install Tailscale"
-  shell: |
-    curl -fsSL https://pkgs.tailscale.com/stable/ubuntu/jammy.noarmor.gpg > /usr/share/keyrings/tailscale-archive-keyring.gpg
-    curl -fsSL https://pkgs.tailscale.com/stable/ubuntu/jammy.tailscale-keyring.list > /etc/apt/sources.list.d/tailscale.list
-
-- name: "Install packages: jq, sudo, tailscale"
+- name: "Install packages: jq, sudo"
   package:
     name:
-      #- ncat    # Newer versions of NMap do not include NCat, WAS needed to announce openvpn_handle (if Debian > 9 or Ubuntu > 18)
-      #- nmap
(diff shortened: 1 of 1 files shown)


holta:
> This has to wait will upstream adds support for the newest RasPiOS..

Ouch!  Tailscale VPN stuff is also very important on pre-release OS's :/

PS Whatever happens, let's keep [Lines 6-9 and 18](https://github.com/iiab/iiab/blob/e9e06a5d52d2d3b2c64c332c977914ac6fe54327/roles/tailscale/tasks/install.yml#L6-L18) commented out in this PR — rather than deleting alternate approach(es) — at this point (until we know more).

Related:

- PR #4095

holta:
> older binaries from a distro that IIAB has dropped support for. ie U22.04 jammy

It's a good reminder that later this year there will be a need to convert `jammy` (22.04) to `noble` (24.04) across many IIAB App installers, e.g. what IIAB uses to install Kolibri, MongoDB, ETC — and likewise making changes from `bookworm` (Debian 12) to `trixie` (Debian 13).

With testing on several mainline platforms to reconfirm of course.

(We are all thankful the most important software installers associated with older versions of Ubuntu LTS [e.g. 22.04 / jammy] consistently work very broadly and very well where we need them!)

jvonau:
https://pkgs.tailscale.com/stable/raspbian/trixie/installer-supported now returns OK
RasPiOS trixie is now supported.

jvonau:
https://pkgs.tailscale.com/stable/ubuntu/resolute/installer-supported now returns OK 
U26.04 is now supported

holta:
1. Very good news that Tailscale supports a broad array of OS's relatively quickly, for those preferring their 1-line curl installer.

2. Still, it feels like too many months for IIAB testers and implementers depending on ever-evolving (new) OS's ASAP — so I suspect we really should stick with some /etc/apt/source.list.d/ repo definition file or another here (i.e. to install Tailscale).

PR #4097, opened 2025-09-29
[Upload2USB] Add check for double mounted drive; Simplify getTargetUSBDriveLocation

avni:
Upload2USB: Add check and warning for double mounted drive; Simplify getTargetUSBDriveLocation function and lsblk usage

Provide warning to user if drive is double mounted; 
Use simpler lsblk logic;
Collapse unnecessary if condition;

Raspberry Pi OS 13 (Trixie)

@jvonau @chapmanjacobd @holta

### DIFF
--- a/roles/usb_lib/files/upload2usb/upload2usb.php
+++ b/roles/usb_lib/files/upload2usb/upload2usb.php
@@ -15,30 +15,28 @@ set_exception_handler(function (Throwable $exception) {
     include ("error.php");
 });
 
-//return the first removable USB drive location
+//return the first removable USB media location
 function getTargetUSBDriveLocation () {
 
-         // Get the path of the first mounted storage
-	 $rmv_usb_path = trim(shell_exec('lsblk -no MOUNTPOINT | grep "^/media"'));
+         // Get paths to all mounted storage and count of mounted storage
+         $rmv_usb_paths = shell_exec('lsblk -no MOUNTPOINTS | grep "^/media/"');
+         $rmv_usb_paths_count = substr_count($rmv_usb_paths, "\n"); 
 
-         // Get the count of storage mounted at /media, and error if there is <>1 or media is double mounted
-         $rmv_usb_path_count = shell_exec('lsblk -no MOUNTPOINTS | grep -c "^/media"'); 
-
-         if ($rmv_usb_path_count == 0 || empty($rmv_usb_path)) {
+         if ($rmv_usb_paths_count == 0) {
              $exception_data = [
                'usb_count' => 0,
-               'exception_msg' => '0 USB sticks found. <br/><br/>'
+               'exception_msg' => '0 USB drives found. <br/><br/>'
              ];
              throw new RuntimeException(json_encode($exception_data));
-         } elseif ($rmv_usb_path_count > 1) {
+         } elseif ($rmv_usb_paths_count > 1) {
              $exception_data = [
-               'usb_count' => $rmv_usb_path_count,
-               'exception_msg' => 'There is more than 1 USB stick inserted or the USB stick is double mounted. <br/><br/>'
+               'usb_count' => $rmv_usb_paths_count,
+               'exception_msg' => 'There is more than 1 USB drive inserted or the USB drive is double mounted. <br/><br/>'
              ];
              throw new RuntimeException(json_encode($exception_data));
+         } else {
+             return trim($rmv_usb_paths) . "/";
          }
-
-         return $rmv_usb_path . "/";
 }
 
 //returns folder path where file will be stored, if create_folder_p = 1, it will create the folder if it doesn't exist
--- a/roles/usb_lib/files/upload2usb/upload2usb.php
+++ b/roles/usb_lib/files/upload2usb/upload2usb.php
@@ -15,7 +15,7 @@ set_exception_handler(function (Throwable $exception) {
     include ("error.php");
 });
 
-// if there is one USB drive, return the path to it, otherwise act on the exception
+// If there is one USB drive and it is not double mounted, return the path to it, otherwise act on the exception
 function getTargetUSBDriveLocation () {
 
          // Enumerate /media/ mountpoints and count them
@@ -35,7 +35,7 @@ function getTargetUSBDriveLocation () {
              ];
              throw new RuntimeException(json_encode($exception_data));
          } else {
-             // at this point, we know there is only 1 USB drive inserted and it is not double mounted; return the path to it
+             // At this point, we know there is only 1 USB drive inserted and it is not double mounted; return the path to it
              return trim($rmv_usb_paths) . "/"; 
          }
 }
--- a/roles/usb_lib/files/upload2usb/upload2usb.php
+++ b/roles/usb_lib/files/upload2usb/upload2usb.php
@@ -18,11 +18,13 @@ set_exception_handler(function (Throwable $exception) {
 //return the first removable USB drive location
 function getTargetUSBDriveLocation () {
 
-         // Get the count of storage mounted at /media, and error if there is <>1 otherwise return upload path
+         // Get the path of the first mounted storage
+	 $rmv_usb_path = trim(shell_exec('lsblk -no MOUNTPOINT | grep "^/media"'));
 
-         $rmv_usb_path_count = shell_exec('lsblk --output NAME,TRAN,RM,MOUNTPOINT --pairs | cut -d " " -f 4 | grep "^MOUNTPOINT=\"/media" | wc -l');
+         // Get the count of storage mounted at /media, and error if there is <>1 or media is double mounted
+         $rmv_usb_path_count = shell_exec('lsblk -no MOUNTPOINTS | grep -c "^/media"'); 
 
-         if ($rmv_usb_path_count == 0) {
+         if ($rmv_usb_path_count == 0 || empty($rmv_usb_path)) {
              $exception_data = [
                'usb_count' => 0,
                'exception_msg' => '0 USB sticks found. <br/><br/>'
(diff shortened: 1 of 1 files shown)


holta:
Looks extremely promising:

- Can this Upload2USB PR be "battled-tested" e.g. with {exFAT stick, dirtier filesystems, common end-user frustrations} — and then [README](https://github.com/iiab/iiab/tree/master/roles/usb_lib) updated accordingly?

- In anticipation of PR #4086 or similar?!

Also Related:

- PR #4088

roles/usb_lib/files/upload2usb/upload2usb.php
```diff
@@ -18,11 +18,13 @@
...
-         // Get the count of storage mounted at /media, and error if there is <>1 otherwise return upload path
+         // Get the path of the first mounted storage
+	 $rmv_usb_path = trim(shell_exec('lsblk -no MOUNTPOINT | grep "^/media"'));
 
-         $rmv_usb_path_count = shell_exec('lsblk --output NAME,TRAN,RM,MOUNTPOINT --pairs | cut -d " " -f 4 | grep "^MOUNTPOINT=\"/media" | wc -l');
+         // Get the count of storage mounted at /media, and error if there is <>1 or media is double mounted
+         $rmv_usb_path_count = shell_exec('lsblk -no MOUNTPOINTS | grep -c "^/media"'); 
 
-         if ($rmv_usb_path_count == 0) {
+         if ($rmv_usb_path_count == 0 || empty($rmv_usb_path)) {
```
chapmanjacobd: edit: sorry I meant to select line:33

I think this PR is a step in the right direction but one thing that I don't quite understand yet is why double mounting is harmful? It's similar to an alias, hard link, or CNAME.

What behavior are you seeing where multiple mountpoints are causing trouble?

I can understand the concern for multiple USBs or USBs with multiple partitions but that's different from this

If I had to guess, perhaps it has been some combination of usbmount root-owned mountpoint and flaky filesystems which have made this more difficult to root cause analysis

holta: Great Question:

1. My hunch is that we should merge a clean/simplifying PR like this one first.

2. And then look into whether or not writing to a double-mounted USB stick/drive is pragmatically feasible (or not, e.g. if that imposes security/permissions complexity, perhaps related to different graphical desktops &/or if it requires modifying usbmount?)

3. Or PR #4086 can come first, if that's lower-hanging fruit!

holta:
@avni tighten up with clean/concise handling of 3 cases, that I'd strongly recommend if you can consider?

```php
$rmv_usb_paths = shell_exec('lsblk -no MOUNTPOINTS | grep "^/media/"');
$rmv_usb_paths_count = substr_count($rmv_usb_paths, "\n");

if ($rmv_usb_paths_count === 0) {
    // EXCEPTION A
} elseif ($rmv_usb_paths_count > 1) {
    // EXCEPTION B
} else {
    return trim($rmv_usb_paths) . "/";    // Singular $rmv_usb_path might not be needed?
}
```

RECAP:

- Use spaces to indent (instead of tab characters)
- Avoid false positives, by using a trailing slash on the right: `grep "^/media/"`
- Call `lsblk -no MOUNTPOINTS` (always plural, not singular)
- Use [SSOT](https://en.wikipedia.org/wiki/Single_source_of_truth)/[DRY](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself) (single source of truth can be `$rmv_usb_paths`) instead of repeatedly running `lsblk` commands
- Derive `$rmv_usb_paths_count` from `$rmv_usb_paths`
- Derive `$rmv_usb_path` (singular!) only if safe/necessary
-  DON'T trim the trailing `\n` until you must (avoid [fenceposting nightmares](https://en.wikipedia.org/wiki/Off-by-one_error#Fencepost_error) !)
- Maybe change regex to `grep "^/media/usb"` in future — _if Upload2USB really can work with double-mounted sticks?_

roles/usb_lib/files/upload2usb/upload2usb.php
```diff
@@ -31,22 +33,12 @@ function getTargetUSBDriveLocation () {
          } elseif ($rmv_usb_path_count > 1) {
```
chapmanjacobd:
```suggestion
         } elseif (count(explode("\n", $rmv_usb_path)) > 1) {
```

roles/usb_lib/files/upload2usb/upload2usb.php
jvonau: > edit: sorry I meant to select line:33
> 
> I think this PR is a step in the right direction but one thing that I don't quite understand yet is why double mounting is harmful? It's similar to an alias, hard link, or CNAME.
> 
> What behavior are you seeing where multiple mountpoints are causing trouble?
> 
> I can understand the concern for multiple USBs or USBs with multiple partitions but that's different from this
> 
> If I had to guess, perhaps it has been some combination of usbmount root-owned mountpoint and flaky filesystems which have made this more difficult to root cause analysis

https://github.com/iiab/iiab/pull/4088#issuecomment-3326200715 uid=1000,gid=1000,fmask=0022,dmask=0022 would prevent a user other than 1000 ie www-data from writing to the disk, the ownership is wrong. The last mountpoint mouted wins the race. Think about bind mounts, you don't see what is in the underlining mountpoint directory once mounted.

jvonau:
Think the exception wording is good enough, the user would be looking at 2+ directories in "iiab.lan/usb" if there were 2 or more disks/partitions presently mounted and could deduce that a single device displayed is double mounted. I just dealt with explicit double mounting to reduce the potential confusion.

roles/usb_lib/files/upload2usb/upload2usb.php
chapmanjacobd: > you don't see what is in the underlining mountpoint directory once mounted

I don't see how that is applicable here? `chown`ing a mountpoint is different from `chown`ing a directory prior to mounting

roles/usb_lib/files/upload2usb/upload2usb.php
holta: @chapmanjacobd Line 33 suggestion (just above) doesn't look right:

- Why compare a string ($rmv_usb_path) to an integer (1) ?

- Explain if you can, what are you suggesting?

holta: FWIW `count(explode("\n", $rmv_usb_path))` conflates 0 with 1, and gives the wrong number in general (off by 1) as this PR currently (problematically) trims any trailing `\n` back on line 22:

`$rmv_usb_path = trim(shell_exec('lsblk -no MOUNTPOINT | grep "^/media"'));` 

My Opinion: @avni can/should greatly simplify, and clean up all this [fenceposting confusion](https://en.wikipedia.org/wiki/Off-by-one_error#Fencepost_error), starting here:

https://github.com/iiab/iiab/pull/4097#issuecomment-3345744455

chapmanjacobd: > Why compare a string ($rmv_usb_path) to an integer (1) ?

Thanks for catching that. I've updated above to counting lines

> Explain if you can, what are you suggesting?

It depends if double mounting is a real problem or not. I suppose this PR is fit to merge AS-IS if we want to do more testing.

We may want to explore the `--exclusive` [mount command flag](https://man7.org/linux/man-pages/man8/mount.8.html) (but on my machine I only have `--onlyonce` and `--move`). In my experience multiple mounting is fine and not much different from a bind mount but I almost exclusively use `ext4` and `btrfs`.

I can't find definitive answers when searching online other than `--bind` is a bit safer; though the `mount` manpage seems to disagree with this:

> It is important to understand that "bind" does not create any second-class or special node in the kernel VFS. The "bind" is just another operation to attach a filesystem. There is nowhere stored information that the filesystem has been attached by a "bind" operation.

If there is doubt about whether the `ntfs3`, or `vfat` drivers will create filesystem corruption through multiple mount-points concurrently writing it may be best to only support non-graphical IIAB use cases and disable/uninstall any automounting services and in `usbmount` after mounting force `umount` any other mountpoints.

OR perhaps only run usbmount in multi-user mode (exit early if `systemctl get-default` == `graphical.target`). That is a lot simpler and MAY be good enough!

BUT I still would be surprised that multiple mountpoints, even on `vfat`, are the cause of any troubles. I haven't seen any hard evidence yet.

chapmanjacobd (approved):

roles/usb_lib/files/upload2usb/upload2usb.php
holta: @chapmanjacobd thanks for having written down the suggestions above, highlighting opportunities that might need a closer look with: 📝 ✅

- @jvonau likewise has gathered very key insights here over the years.

- A huge amount of testing of course remains — so hopefully @avni in the coming 24-48h can help **narrow the scope** of what's truly needed & what should be tried first?

jvonau:
Just spit-balling here but what if [usb0.conf](https://github.com/iiab/iiab/pull/4086/files#diff-52598bf61493fd208aaf710707fd9010da6d8694bdae3b8b9e5cb1cb1eca1685) ran as root?  That should allow unmount, chown and various other restricted actions to succeed, would it not?

roles/usb_lib/files/upload2usb/upload2usb.php
jvonau: > BUT I still would be surprised that multiple mountpoints, even on `vfat`, are the cause of any troubles. I haven't seen any hard evidence yet.

#4066 comes to mind, and cruse the issues this is a long standing known issue.

chapmanjacobd:
Good question! AppArmor/SELinux would likely block it... maybe a better approach would be a SUID/SGID binary or something more limited like that.

It might be worth investigating how [Cockpit](https://github.com/cockpit-project/cockpit) or WebMin/VirtualMin handle this...

chapmanjacobd:
`pmount`? https://serverfault.com/questions/483860/how-to-mount-and-unmount-devices-remotely-from-a-web-service

avni:
> @avni tighten up with clean/concise handling of 3 cases, that I'd strongly recommend if you can consider?

Thank you for the review and suggestions! Implemented here: https://github.com/iiab/iiab/pull/4097/commits/734ec818f713d13b7c7244b304794a4a793926ca

chapmanjacobd (approved):

holta:
> Looks extremely promising:
> 
> * Can this Upload2USB PR be "battled-tested" e.g. with {exFAT stick, dirtier filesystems, common end-user frustrations} — and then [README](https://github.com/iiab/iiab/tree/master/roles/usb_lib) updated accordingly?
> * In anticipation of PR [update temp_dir php_value for upload2usb #4086](https://github.com/iiab/iiab/pull/4086) or similar?!

@avni thanks for having tested this (all 3 cases) on RasPiOS 13 Trixie on RPi 4!

Future Work:

- `usb_count` (variable used by your exception handler) should really be renamed in a future PR, to convey what it's really counting: Maybe pick a name like `usb_mountpoints_count` ?

holta:
> @avni thanks for having tested this (all 3 cases) on RasPiOS 13 Trixie on RPi 4!

Thank you to @jvonau & @chapmanjacobd who are helping @avni put this Upload2USB functionality on a much more solid foundation:

- Such that many more people will be able to express themselves & contribute to local libraries 📚

- Not easy...

- One step at a time...

PR #4098, opened 2025-09-29
Set `TS_NO_LOGS_NO_SUPPORT=true` in /etc/default/tailscaled -- to reduce Tailscale telemetry risks

holta:
### Fixes bug:

Tailscale is calling home too often even when a VPN is not enabled, creating surveillance risks, as outlined here:

- https://github.com/iiab/iiab/issues/4058#issuecomment-3222172316
- PR #4095

Don't clobber Tailscale's own `/etc/default/tailscaled` (that creates serious-looking problems!) but do append `TS_NO_LOGS_NO_SUPPORT=true` to it immediately after installing Tailscale, to reduce (and hopefully avoid) telemetry risks (after tailscaled is restarted).

Background:

- https://tailscale.com/kb/1011/log-mesh-traffic#opting-out-of-client-logging
- https://www.reddit.com/r/Tailscale/comments/1alhs71/why_does_tailscale_keep_pinging_logtailscaleio/

This related PR might also be considered:

- #4095

(Conversely these 2 PRs serve a very similar purpose, so both may not be needed.)

Ubuntu Server 25.10

@jvonau @chapmanjacobd

### DIFF
--- a/roles/tailscale/tasks/install.yml
+++ b/roles/tailscale/tasks/install.yml
@@ -21,6 +21,13 @@
 - name: Set up tab completion for 'tailscale' at the command-line
   shell: mkdir -p /etc/bash_completion.d && tailscale completion bash > /etc/bash_completion.d/tailscale
 
+- name: Set 'TS_NO_LOGS_NO_SUPPORT=true' in /etc/default/tailscaled -- as tailscale "telemetry" calls home too often (#4058, $4095)
+  lineinfile:
+    path: /etc/default/tailscaled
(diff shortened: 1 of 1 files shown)


holta:
- Let's proceed (with Precautionary Principle in mind) giving this minimal approach a shot.

- Privacy/Security/Complexity balance can be readjusted later if necessary.

holta:
Let's keep an eye on:

- juanfont/headscale#2793

PR #4104, opened 2025-10-03
apt-mark systemd-zram-generator as manually installed, if installed [FOR NOW: as dphys-swapfile mysteriously sets systemd-zram-generator to "auto-removable" on RPi, or at least on RasPiOS?]

chapmanjacobd:
### Fixes bug:

related to https://github.com/iiab/iiab/pull/4094

Marks systemd-zram-generator as manually installed

`apt mark` won't do anything (it won't even exit nonzero) if the package to mark is not installed

We should test this on Raspberry Pi 5 and Pi Zero 2

@holta

### DIFF
--- a/roles/1-prep/tasks/raspberry_pi.yml
+++ b/roles/1-prep/tasks/raspberry_pi.yml
@@ -21,6 +21,11 @@
     state: present
   when: rtc_id is defined and rtc_id != "none" and is_ubuntu    # CLARIF: Ubuntu runs increasingly well on RPi hardware, starting in 2020 especially
 
+- name: Mark systemd-zram-generator as manually installed
+  apt_mark:
+    package: systemd-zram-generator
+    state: manual
+  ignore_errors: true
 
 - name: 'Install packages: fake-hwclock, dphys-swapfile'
   package:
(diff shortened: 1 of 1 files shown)


holta:
@chapmanjacobd do you perceive this as an interim-but-quite-safe patch — e.g. even if upstream (RasPiOS or maybe even upstream of them) cleans up their apt/deb packaging in coming months?

Arising also to original research here:

- PR #4087

holta:
@jvonau @Ark74 can you review this PR, if you have time?

(Needs testing too of course!)

chapmanjacobd:
> do you perceive this as an interim-but-quite-safe patch — e.g. even if upstream (RasPiOS or maybe even upstream of them) cleans up their apt/deb packaging in coming months?

yes, exactly. I agree that it seems like a packaging bug. 

But it is also interesting to note that marking all system packages as manual [seems to be](https://askubuntu.com/a/1384000) more a thing of the past and it is more common now to rely on dependency trees

holta:
Turns out Ansible doesn't have its own `apt_mark` command:

- So I refined this PR (hope that's ok!) with an external call to `apt-mark manual systemd-zram-generator`
- Tested to work perfectly on Zero 2 W!
- Seems good enough for now: after all this temporary patch should hopefully be yanked out in 2026 of even before then (Ansible output clarified, to make future maintenance much easier ;)

holta:
- I think we should go with this approach for now...

- The point is to learn more along the way, as upstream(s) hopefully clean up in due course over coming months~

chapmanjacobd:
sounds good!

but I think it needs `become: true` to act as root? wait... none of the other commands are using that. I don't know Ansible that well--does it use root by default in IIAB?

holta:
> I don't know Ansible that well--does it use root by default in IIAB?

Yes!

PR #4120, opened 2025-10-17
New maps role [includes satellite photos & more comprehensive search prototype!]

orblivion:
### Description of changes proposed in this pull request:

New maps role. Initial version that is not quite production quality but it's enough for a few curious people to try.

* maps.black for streets
* maps.black for satellite
* nominatim for search
* "low res" by default for all three, a few hundred megs
  * naturalearth for "small" streets
  * extracted satellite up to zoom 7
  * California only for Nominatim
* "high res" available for all three, a hundred or so gigs
  * extracted satellite up to zoom 11 or 12

### Mention a team member @username e.g. to help with code review:

@holta 

### Testing it out

*(adapted from [this comment](https://github.com/iiab/iiab/pull/4120#issuecomment-3416691031))*

Start installing IIAB with this PR, by running: `curl iiab.io/install.txt | bash -s 4120`

Edit `/etc/iiab/local_vars.yml` a few seconds later, to set these 2 variables: (when IIAB installer asks you to do that!)

```
maps_install: True
maps_enabled: True

osm_vector_maps_install: False
osm_vector_maps_enabled: False
```

If you want to add optional features (such as large maps or search) see [here](https://github.com/iiab/iiab/pull/4120/files#diff-238bd4eb09be3ed2de08302dcea3da77b274c8f4f3181c2f950c96dbcd6902b7) for the variables, and [here](https://github.com/iiab/iiab/pull/4120/files#diff-d68bc45adbaaf8a317b0ead97b3217daafc863dde083ebffa406d3597dfb92fa) for explanations of them. I do not recommend search on Pi Zero at this time.

Wrap up the install: `sudo iiab`

The app is found at this URL: `http://box/maps/`

#### Updating

*(adapted from [this comment](https://github.com/iiab/iiab/pull/4120#issuecomment-3420444643))*

If you've been testing and there are new changes on this PR that you want to try, the above installation instructions won't work to update you.

Instead you have to:

```
cd /opt/iiab/iiab
sudo git pull https://github.com/orblivion/iiab maps-dot-black --no-rebase --no-edit
```

Followed by:

```
sudo ./runrole maps --reinstall
```

Unless you had an error the last time you tried runrole, in which case try:

```
sudo ./runrole maps
```

PR #4137, opened 2025-11-07
CI TEST: Figuring out why the maps-dot-black branch (#4120) is failing

orblivion:
This PR: https://github.com/iiab/iiab/pull/4120 intermittently has CI failures. This PR is on a branch that rebased onto master to see if recent changes fixed it. We may have to run this several times to make sure it's fixed here. It has failed several times on the other branch.


orblivion:
I'm assuming we don't need this anymore. It seemed to stop failing on the other PR anyway, though I think we're still investigating the intermittent issue.

PR #4141, opened 2025-11-10
new maps: initial view shows all of the continents once

orblivion:
### Fixes bug:

https://github.com/iiab/iiab/pull/4120#issuecomment-3506880994

Create an "initial view" / "default view" of the map that shows all of the continents and has a "decent" centering of the map. 0 latitude didn't seem great, I went with 30.

This is used if there's no "hash" in the URL, i.e. the user just visits box/maps or box/maps/. The hash gets added to the URL immediately after, so any reloads of the page will go by the location in the hash rather than this new default.

Firefox on Linux: it seemed fine.

Vanadium (Chromium fork) on GrapheneOS on Pixel 7:

* In vertical mode, it starts zoomed out all the way, but it's not wide enough to even fully show all of the continents. It's centered on western Africa.
* In horizontal mode, it appears a bit further north than I thought it would. I realized that this is because my phone, when horizontal, is too short to fit the minimum height of the map element! I.e. the attribution on the bottom right doesn't show up. This is an issue irrespective of the default view. I wonder if it's maps.black or maplibregl?

@holta

### DIFF
--- a/roles/maps/templates/index.html.j2
+++ b/roles/maps/templates/index.html.j2
@@ -43,6 +43,7 @@
 
             // set immediately after the tag is defined
             let mb = null
+            let shouldSetDefaultView = false
 
             // Defining these hash get/set functions outside of the window load because we
             // need to run some of it earlier. Again, the overall code layout sucks and we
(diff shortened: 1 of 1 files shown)


holta:
@orblivion safe enough to be merged right away?

orblivion:
I would say yes

Issue #4143, opened 2025-11-11
Adding multiple zim files fail


deldesir:
Kiwix-manage no longer supports multiple files.

Related: #4139 

```
iiab-admin@lrn2:/library/zims/content$ sudo /opt/iiab/kiwix/bin/kiwix-manage ../library.xml add wikipedia_en_nollywood_maxi_2025-10.zim wikipedia_en_climate-change_maxi_2025-10.zim
Unexpected argument: ../library.xml, add, wikipedia_en_nollywood_maxi_2025-10.zim, wikipedia_en_climate-change_maxi_2025-10.zim
Manipulates the Kiwix library XML file

Usage:
 kiwix-manage LIBRARYPATH add [--zimPathToSave=<custom_zim_path>] [--url=<http_zim_url>] ZIMPATH
 kiwix-manage LIBRARYPATH remove|delete ZIMID ...
 kiwix-manage LIBRARYPATH show [ZIMID ...]
 kiwix-manage -v | --version
 kiwix-manage -h | --help

Arguments:
  LIBRARYPATH    The XML library file path.
  ZIMID          ZIM file unique ID.
  ZIMPATH        A path to a zim to add.

Options:
  Custom options for "add" action:
    --zimPathToSave=<custom_zim_path>  Replace the current ZIM file path
    --url=<http_zim_url>               Create an "url" attribute for the online version of the ZIM file

  Other options:
    -h --help                          Print this help
    -v --version                       Print the software version

Examples:
 Add ZIM files to library:       kiwix-manage my_library.xml add first.zim second.zim
 Remove ZIM files from library:  kiwix-manage my_library.xml remove e5c2c003-b49e-2756-5176-5d9c86393dd9
 Show all library ZIM files:     kiwix-manage my_library.xml show

Documentation:
  Source code  https://github.com/kiwix/kiwix-tools
  More info   https://wiki.kiwix.org/wiki/Kiwix-manage
```

tim-moody:
This is not really an iiab issue.

holta:
@deldesir can you report upstream?

deldesir:
Done at https://github.com/kiwix/kiwix-tools/issues/781

Issue #4139, opened 2025-11-10
iiab-make-kiwix-lib warnings & errors [after a ZIM file is removed from /library/zims/content]


holta:
@tim-moody is it possible the new kiwix-tools 3.8.0 (from about 2 weeks ago) is related?

```
root@box:~# iiab-make-kiwix-lib
Waiting to run iiab-make-kiwix-lib.py
Now running iiab-make-kiwix-lib.py
Unexpected argument: /library/zims/library.xml.tmp, remove, f29ae2d1-85f0-b663-a42e-e6479d7e21ea
Manipulates the Kiwix library XML file

Usage:
 kiwix-manage LIBRARYPATH add [--zimPathToSave=<custom_zim_path>] [--url=<http_zim_url>] ZIMPATH
 kiwix-manage LIBRARYPATH remove|delete ZIMID ...
 kiwix-manage LIBRARYPATH show [ZIMID ...]
 kiwix-manage -v | --version
 kiwix-manage -h | --help

Arguments:
  LIBRARYPATH    The XML library file path.
  ZIMID          ZIM file unique ID.
  ZIMPATH        A path to a zim to add.

Options:
  Custom options for "add" action:
    --zimPathToSave=<custom_zim_path>  Replace the current ZIM file path
    --url=<http_zim_url>               Create an "url" attribute for the online version of the ZIM file

  Other options:
    -h --help                          Print this help
    -v --version                       Print the software version

Examples:
 Add ZIM files to library:       kiwix-manage my_library.xml add first.zim second.zim
 Remove ZIM files from library:  kiwix-manage my_library.xml remove e5c2c003-b49e-2756-5176-5d9c86393dd9
 Show all library ZIM files:     kiwix-manage my_library.xml show

Documentation:
  Source code  https://github.com/kiwix/kiwix-tools
  More info   https://wiki.kiwix.org/wiki/Kiwix-manage

Traceback (most recent call last):
  File "/usr/lib/python3/dist-packages/iiab/iiab_lib.py", line 113, in rem_libr_xml
    outp = subprocess.check_output(args)
  File "/usr/lib/python3.13/subprocess.py", line 472, in check_output
    return run(*popenargs, stdout=PIPE, timeout=timeout, check=True,
           ~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
               **kwargs).stdout
               ^^^^^^^^^
[...]

PR #4144, opened 2025-11-11
patches to handle kiwix-manage remove problems

tim-moody:


jvonau:
I'd push this, just to be able to from the bash cmdline pass options like '--no_tmp' to the python part of the operation.

holta:
> I'd push this, just to be able to from the bash cmdline pass options like '--no_tmp' to the python part of the operation.

Ok.  @tim-moody and @deldesir do you agree?  e.g. is this PR sufficiently tested?

Related:

- #4139
- #4143

tim-moody:
I think this is OK as far as it goes, but there are two other cases in adm cons:

1. you remove a zim and click Install Content -> Reindex Kiwix Content - you will easily miss the message to run make lib with -f, especially since it says succeeded

2. you remove a zim and download another before make lib. You will not see the message which has scrolled off the panel.

At the same time, Kiwix will soon fix the problem, and this can be merged even if the adm cons problem remains.

PR #4145, opened 2025-11-11
familiar zoom and scroll control on maps

orblivion:
### Fixes bug:

https://github.com/iiab/iiab/pull/4120#issuecomment-3506892482

Restore the map's scrolling and zooming controls to what we all know and love. One finger to scroll and two to tilt on phone, ctrl key no longer needed to zoom on desktop.

So now we have:

* **Phone**: **one** finger scroll, **two** finger rotate, zoom and tilt
* **Desktop**: mouse drag to scroll, mouse-wheel to zoom, **ctrl**-drag to tilt and rotate

Previously we had:

* **Phone**: **two** finger scroll, rotate, and zoom, **three** finger tilt
* **Desktop**: mouse drag to scroll, **ctrl**-mouse-wheel to zoom, **ctrl**-drag to tilt and rotate

It turns out that there's this thing called "Cooperative Gestures" that was turned on by default in maps.black. It gave us the more bureaucratic zoom and scroll and tilt controls. There's a good use for it though. If you have a small map in the middle of a blog post, you don't want to suddenly be scrolling or zooming the map when you're trying to scroll down the page.

* Vanadium browser on GrapheneOS
* Firefox on Linux
* Safari on iPad

@holta @avni

### DIFF
--- a/roles/maps/templates/index.html.j2
+++ b/roles/maps/templates/index.html.j2
@@ -389,7 +389,10 @@
         </script>
     </head>
     <body>
-        <maps-black loader="pmtiles" navigationcontrol="true"></maps-black>
+	<!-- cooperativegestures (true by default) would require two-finger scroll, three finger tilt on phone,
+	     and ctrl-mouse-wheel to zoom on desktop. I guess these are useful for small maps widgets on a page to
+	     avoid accidental scrolling and whatnot. -->
+        <maps-black loader="pmtiles" navigationcontrol="true" cooperativegestures="false"></maps-black>
         <script>
           mb = document.getElementsByTagName("maps-black")[0];
           readHash()


holta:
@deldesir can you test this and share your opinion?

orblivion:
# Some instructions for testing

## If you need to install IIAB from scratch

Run: `curl iiab.io/install.txt | bash -s 4145`

Edit `/etc/iiab/local_vars.yml` a few seconds later, to set these variables: (when IIAB installer asks you to do that!)

```
maps_install: True
maps_enabled: True

osm_vector_maps_install: False
osm_vector_maps_enabled: False
```

Wrap up the install: `sudo iiab`

The app is found at this URL: `http://box/maps/`

## If you already have IIAB installed

Assuming you already have an IIAB installed that you're okay testing on:

First, make sure the above local vars are set. Then, go to:

    cd /opt/iiab/iiab
    sudo git pull https://github.com/orblivion/iiab fix-map-scroll --no-rebase --no-edit

Followed by:

    sudo ./runrole --reinstall maps

Unless you had an error the last time you tried runrole, in which case try:

    sudo ./runrole maps

deldesir (commented):
Works as advertised

deldesir:
I figured out the third button can be rotated after repeated clicks on it. 

image

orblivion:
The third button is the "compass". It's there to straighten out the map if you've rotated or tilted it.

As an aside: I wonder if we should make a "primer" / "howto" for all of these buttons. The thing is, the zoom/compass buttons are standard maplibre controls. I wonder if maplibre sort of assumes that most people just know what they do. Do they?

Maybe we could add a little question mark that brings up a help screen that explains what the other buttons do. Not everyone needs to look at it, but if let's say a teacher looks at it they would be able to help their students better.

holta:
> Maybe we could add a little question mark that brings up a help screen that explains what the other buttons do. Not everyone needs to look at it, but if let's say a teacher looks at it they would be able to help their students better.

I'd suggest waiting?

- Translation / Localization challenges are a huge headache.

- Prioritize provisioning of maps/satellite photos for IIAB on Android maybe, to step up from ~3 GB to what low-income educators actually need?!

holta:
> sudo ./runrole maps --reinstall

My apologies the above doesn't yet work?  Try this for now... !

```
sudo ./runrole --reinstall maps
```

orblivion:
Oops, I updated my comment. Thanks.

orblivion:
> > Maybe we could add a little question mark that brings up a help screen that explains what the other buttons do. Not everyone needs to look at it, but if let's say a teacher looks at it they would be able to help their students better.
> 
> I'd suggest waiting?
> 
>     * Translation / Localization challenges are a huge headache.
> 
>     * Prioritize provisioning of maps/satellite photos for IIAB on Android maybe, to step up from ~3 GB to what low-income educators actually need?!

I was thinking about language - maybe it can be done with just pictures?

Anyway, not saying it's a priority, just brainstorming.

holta:
> I was thinking about language - maybe it can be done with just pictures?
> 
> Anyway, not saying it's a priority, just brainstorming.

Tough but it might work!  (If we run into a great illustrator / artist?!)

holta:

chapmanjacobd:
I don't think it's good practice to document settings/config so verbosely in code. Are HTML comments hoisted from jinja2 templates? I think this documentation is best saved inside the git commit message.

Less ideal than documenting out-of-band, but you could do something like this:

    <h1>Welcome!</h1>
    {# This is a Jinja2 comment and will not be in the final HTML #}
    <p>Some content here.</p>

However, this PR is safe to merge

chapmanjacobd (approved):

holta:
> I think this documentation is best saved inside the git commit message.

I hope we can do a bit of both!?

Several of us prefer inline comments to make code more readable on phones, as young folks might be moving towards (:

chapmanjacobd:
The main problem is that it adds the additional cognitive burden of ignoring non-code sections while trying to understand what the code is actually doing. 

Comments are great when they act as signposts or when they explain why a choice was made _in present tense_ (ie. in this scenario it would be something like "cooperativegestures=false provides familiar zoom and scroll control" but even here I feel like that is a poor use of vertical screen real estate because it is config rather than algorithm).

In the situation here, I think a link to the maps-black config documentation where it explains what cooperativegestures is, along with explanations of other attributes, would be more useful when both reading and editing code--both on desktop and mobile.

holta:
> additional cognitive burden of ignoring non-code sections

In prototypes maybe.  But that feels like a real cop-out to me:

Insiders / Veterans can save _many hours of frustration_ among young / mobile / international participants — if they look beyond themselves, to "AI² == Articulate Assumptions, Intuition, Intentions" — e.g. including clear inline comments allowing new volunteers in all countries to understand more of the https://en.wikipedia.org/wiki/Five_Ws

<p align=center>
<i>"Programs must be written for people to read, and only incidentally for machines to execute"</i><br>
—Hal Abelson (1984)
</p>

orblivion:
> The main problem is that it adds the additional cognitive burden of ignoring non-code sections while trying to understand what the code is actually doing.

I'll meet you half way and say I can cut out the cruft. I like your condensed version. I just know that I didn't know wtf that parameter was. Fair enough long comments are cognitive load, but I'd argue opening a new tab to look it up is also cognitive load.

And I checked, no it does not cut out HTML comments. I imagine that could break some rare magic things where comments are meaningful. Or maybe it's an option. Happy to make it a template comment.

Issue #4155, opened 2025-11-20
BUG: 0-init doesn't enforce a minimum ansible version


jvonau:
The latest ansible support changed in #4085 introduces backwards compatibility breakage, leading to a noted issue in #4152

jvonau:
```
        "ansible_local": {
            "admin_facts": {
                "ansible_playbook_program": "/usr/local/bin/ansible-playbook",
                "ansible_program": "/usr/local/bin/ansible"
            },
            "local_facts": {
                "ansible_version": "2.20.0",
                "devicetree_model": "Raspberry Pi 500 Rev 1.0",
                "dhcpcd": "not-found",
                "iiab_branch": "none",
                "iiab_commit": "none",
                "iiab_recent_tag": "none",
                "iiab_remote_url": "none",
                "network_manager": "enabled",
                "os": "raspbian",
                "os_ver": "raspbian-13",
                "php_version": "8.4",
```

What really is required is some proper version but just to remind other of potential ansible dependencies lets record that version info for iiab-diagnostics and deal with the local_facts issue.
https://github.com/iiab/iiab/compare/master...jvonau:ansible_lock?expand=1

jvonau:
https://github.com/iiab/iiab/pull/4085#issuecomment-3302524590 new installs are covered https://github.com/iiab/iiab/blob/master/iiab-install but nothing when already deployed and somebody does a git pull on a machine that already has ansible installed.

Issue #4152, opened 2025-11-18
iiab-network and networkd-dispatcher.service issues?


holta:
- It looks like @nzola ran iiab-update more than a year after the original IIAB install, which is not supported (that is highly discouraged, as the iiab-update command is only meant for short term upgrades, soon after IIAB was installed).

- Still, there might be something to learn from the networking failure(s) at the bottom of his iiab-diagnostics here?

https://paste.centos.org/view/eccce93c

cc: @jvonau

jvonau:
Shame the rolling upgrades were not adopted 5 years ago. https://github.com/iiab/iiab/pull/2500  https://github.com/iiab/iiab-factory/pull/134

>=IIAB==========================================================================
-rw-r--r-- 1 root root 758260 Nov 17 22:24 /opt/iiab/iiab/iiab-network.log

Wonder what year that is?
>2025-11-17 22:24:01,259 p=5389 u=root n=ansible | TASK [network : Create list of LAN (non WAN) ifaces] ***************************
2025-11-17 22:24:02,110 p=5389 u=root n=ansible | fatal: [127.0.0.1]: FAILED! => {"changed": true, "cmd": "ls /sys/class/net | grep -v -e wwlan -e ppp -e ap0 -e lo -e br0 -e tun -e br- -e docker -e bridge0 -e veth -e tailscale0 -e wlan0 -e eth0 -e none", "delta": "0:00:00.015506", "end": "2025-11-17 22:24:01.996808", "msg": "non-zero return code", "rc": 1, "start": "2025-11-17 22:24:01.981302", "stderr": "", "stderr_lines": [], "stdout": "", "stdout_lines": []}
2025-11-17 22:24:02,118 p=5389 u=root n=ansible | PLAY RECAP *********************************************************************
2025-11-17 22:24:02,119 p=5389 u=root n=ansible | 127.0.0.1                  : ok=317  changed=9    unreachable=0    failed=1    skipped=13   rescued=0    ignored=0   

Strange place to have an error, I'm sure YOU can figure it out.

jvonau:
some old code present: https://github.com/iiab/iiab-admin-console/pull/595 
>Nov 17 22:30:11 box IIAB-CMDSRV[1801]: IIAB-CMDSRV : Restarting hostapd
Nov 17 22:30:12 box kernel: ieee80211 phy0: brcmf_cfg80211_stop_ap: setting AP mode failed -52

Bigger issues:
>=IIAB==========================================================================
[...]

PR #4085, opened 2025-09-16
Changes to get IIAB working with ansible-core 2.19 (released early summer)

holta:
### Fixes bugs:

- #4030
- #4084

### Description of changes proposed in this pull request:

The hardest part was breaking apart roles/0-init/tasks/validate_vars.yml into 7 component scripts, each painfully / exhaustively validating all ~45 (or ~4) sets of Ansible vars.  Not pretty but definitely good enough for now.

ansible-core 2.19 is such a dramatic change in Ansible's security posture, that other smaller changes were also put in place to get IIAB working again on these most recent Ansible releases.

(Yes this breaks backwards compatibility, FYI with their changes being so drastic, consequently ansible-core 2.19.2 was also set as the MINIMUM version of Ansible for IIAB installs to avoid recurring nightmares over coming months.)

### Smoke-tested on which OS or OS's:

Ubuntu Server 25.10 (latest pre-release, apt updated)

### Mention a team member @username e.g. to help with code review:

@jvonau @tim-moody

Issue #4084, opened 2025-09-15
iiab-network fails w/ ansible-core 2.19


holta:
Not an emergency as IIAB is not quite yet supporting the drastic changes introduced by ansible-core 2.19.x — but FYI soon Line 180 of roles/network/tasks/detected_network.yml will need to change from...

https://github.com/iiab/iiab/blob/e6b8f72c05a96af67351e99d354f6daffe132b11/roles/network/tasks/detected_network.yml#L180

Over to... (to test for a proper integer value)

```
when: num_lan_interfaces != 0
```

Reason why the quotation marks must be removed: Ansible variable `num_lan_interfaces` has type **integer** with ansible-core 2.19 which is trying to take security / taint a lot more seriously, whereas `num_lan_interfaces` had type **AnsibleUnsafeText** in prior versions of Ansible.

Example illustrating the problem with ansible-core 2.19.2 below...

```
TASK [network : Create list of LAN (non WAN) ifaces] ****************************************************************************************************************************************************************************************
[ERROR]: Task failed: Module failed: non-zero return code
Origin: /opt/iiab/iiab/roles/network/tasks/detected_network.yml:178:3

176
177 # LAN - pick non WAN's
178 - name: Create list of LAN (non WAN) ifaces
      ^ column 3

fatal: [127.0.0.1]: FAILED! => {"changed": true, "cmd": "ls /sys/class/net | grep -v -e wwlan -e ppp -e ap0 -e lo -e br0 -e tun -e br- -e docker -e bridge0 -e veth -e tailscale0 -e none -e eth0 -e eth1", "delta": "0:00:00.004116", "end": "2025-09-15 18:18:33.305366", "msg": "non-zero return code", "rc": 1, "start": "2025-09-15 18:18:33.301250", "stderr": "", "stderr_lines": [], "stdout": "", "stdout_lines": []}
```

```
root@u2510:~# ls /sys/class/net
eth0  eth1  lo  tailscale0
root@u2510:~# ls /sys/class/net | grep -v -e wwlan -e ppp -e ap0 -e lo -e br0 -e tun -e br- -e docker -e bridge0 -e veth -e tailscale0 -e none -e eth0 -e eth1
root@u2510:~# echo $?
1
```

Related:

- #4030

holta:
@chapmanjacobd regarding your question at https://github.com/iiab/iiab/pull/4085#pullrequestreview-3230962330 — the key seems to be Line 175, which was always (trying to!) cast to integer.,..

[...]

PR #4156, opened 2025-11-21
Ansible lock in 0-init

jvonau:
### Fixes bug:
#4155 #4152
To prevent an old version of ansible from running on a newer code base where that is an [known issue](https://github.com/iiab/iiab/pull/4085#issuecomment-3302524590)
4 testing VMs passed and lightly with RasPiOS with altering defaults/main.yml 

>ansible_facts['architecture']
ansible_facts.ansible_local.local_facts.iiab_remote_url

The above both do the same thing in the end but what is the preferred syntax?
Failure to change the code to be [backwards compatible](https://github.com/iiab/iiab/issues/4084#issuecomment-3314959067) when https://github.com/iiab/iiab/pull/4085#issuecomment-3303075828 breaks ansible runtime when run on older ansible installs


holta:
Thanks Jerry:

This is very helpful to inject healthy refactorings.

I'll work on these ideas over the coming ~10 days.  Not all these ideas will make it into the final sauce, but most important these ideas allow new volunteers (that I'm now meeting, in several countries) to incrementally incorporate an increasing number of battle-tested / low-complexity improvements wherever we can.  Nourishing civic volunteer movements in every country.  As Ansible hopefully learns from Terraform to become a "friendlier" ecosystem over coming years.

jvonau:
I showed you earlier how to make the change for the network code to be backwards compatible but that was not implemented then along comes a random 'git pull' and breaks a running working install. The enforcement of  MIN_ANSIBLE_VER is required because even you didn't recognize the cause of the [failure](https://github.com/iiab/iiab/issues/4152#issuecomment-3547823544)

jvonau:
I'm going to ask the question again, what is the preferred syntax between?

>ansible_facts['architecture']
ansible_facts.architecture

holta:
> I'm going to ask the question again, what is the preferred syntax between?
> 
> > ansible_facts['architecture']
> > ansible_facts.architecture

- Both might be fine in the interim?

- I agree we should come together recommending one or the other. ✅

jvonau:
Another known issue https://github.com/iiab/iiab/pull/4406#issuecomment-4502343202

jvonau:
@holta #4428 is not backwards compatible, now might be a good time for this safeguard. The warning returned could be altered to tell the user to run scripts/ansible but I'll leave the final wording up to you.

roles/0-init/tasks/main.yml
holta, line 1: 1. Instead of Line 177, shouldn't this be much close to the very top of `roles/0-init/tasks/main.yml` ?  (If we go with this approach!)

2. If we really do go with this new approach, stripping MIN_ANSIBLE_VER from `iiab-install` keeps the code base much cleaner! :pray:

3. If so, error msg `Ansible needs updating to {{ MIN_ANSIBLE_VER }}` should probably spell it out with a bit more detail — maybe something like... ?

```
ansible-core needs updating to {{ MIN_ANSIBLE_VER }} — PLEASE RUN: sudo /opt/iiab/iiab/scripts/ansible
```

holta:
> #4428 is not backwards compatible

In what way?  (Probably not terribly important, but clarify if possible!)

jvonau:
> > #4428 is not backwards compatible
> 
> In what way? (Probably not terribly important, but clarify if possible!)

The name of the "collection" changed and mysql_* -> mariadb_*

roles/0-init/tasks/main.yml
jvonau: > 1. Instead of Line 177, shouldn't this be much close to the very top of `roles/0-init/tasks/main.yml` ?  (If we go with this approach!)

It could, I was thinking this is more of a validation step but in your face early might be better.  
> 
>     2. If we really do go with this new approach, stripping MIN_ANSIBLE_VER from `iiab-install` keeps the code base much cleaner! 🙏
> 

Could drop MIN_RPI_KERN and OS given using those are out of date given the current use of `sudo iiab`
>     3. If so, error msg `Ansible needs updating to {{ MIN_ANSIBLE_VER }}` should probably spell it out with a bit more detail — maybe something like... ?
> 
> 
> ```
> ansible-core needs updating to {{ MIN_ANSIBLE_VER }} — PLEASE RUN: sudo /opt/iiab/iiab/scripts/ansible
> ```

That is something that could be done.

PR #4157, opened 2025-11-21
Update some map data and switch to new download source.

orblivion:
### Fixes bug:

* sha256sum errors when installing "osm-full" vector tiles
    * maps.black has been (thankfully!) seemingly updating its vector map tiles about once a month.
* My impending S3 bill, for all of the other downloads associated with the map.

* Change map data download source from S3 bucket to switnet.org for all items currently in S3 bucket
* Change map data download source from maps.black to switnet.org for full osm vector map
    * If we control the data and update cadence ourselves, we prevent needing to update checksums every month.
* Update sha256sums for osm-full and osm-z9 (I made a new osm-z9 for good measure to make sure its data matches osm-full)
    * Both updated files have been uploaded to switnet already

I tried a few of the downloads via Ansible on Trixie on a Pi 500. I still need to try the very biggest files: vector osm-full and satellite 12, assuming you want those tested before merging.

@holta

### DIFF
--- a/roles/maps/defaults/main.yml
+++ b/roles/maps/defaults/main.yml
@@ -4,7 +4,7 @@
 maps_dot_black_small_assets_url: https://v1.maps.black
 maps_dot_black_tiles_url: https://maps.black
 unpkg_url: https://unpkg.com
-s3_url: https://iiab-maps.danielkrol.com/0
+iiab_map_host_url: https://iiab.switnet.org/maps/0
 
 # TODO I wonder if it should go under /library/maps/vector_tiles and symlink from the www directory.
 # That way it would be near the search data. It would be easier to find everything.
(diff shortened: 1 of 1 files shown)


holta:
@orblivion should this be merged quickly — e.g. today or so?

(Of course we can / should revise + evolve down the road, regardless!)

orblivion:
I will quickly doublecheck on my VPS that the two big checksums match (not even with Ansible). Should be within the hour. I'd feel pretty confident merging then.

Or if you're not even worried about that I'm happy to merge now.

holta:
Thanks: Let's aim to merge this morning.

orblivion:
Okay I'm done with my checks. Merge whenever you'd like

PR #4151, opened 2025-11-18
[code] initial work for COTG role

Ark74:
### Fixes bug:
Adds COTG offline distribution site

### Description of changes proposed in this pull request:
Provisions the COTG (arm64) APK for an offline site, enabling Android-based offline development.

### Smoke-tested on which OS or OS's:
* Ubuntu 24.04
* Trisquel 12.0

### Mention a team member @username e.g. to help with code review:
@holta

PR #4159, opened 2025-11-22
[code] declare code vars at unittest, organize lines to match list.

Ark74:
### Fixes bug:
Missing values on unittest for code role

Update and reorganize existing following the existing app order.

Trisquel 12 & Ubuntu 24.04

@holta

### DIFF
--- a/vars/default_vars.yml
+++ b/vars/default_vars.yml
@@ -683,6 +683,10 @@ calibre_port: 8080
 calibre_web_path: calibre  #NEEDS WORK: https://github.com/iiab/iiab/issues/529
 # Avoid URL collisions w/ calibreweb_url1, calibreweb_url2, calibreweb_url3 below!
 
+# Code On The Go, is an open source offline Android IDE by the App Dev for All team
+code_install: False
+code_enabled: False
+
 # A full-featured PBX (for rural telephony, etc) based on Asterisk and FreePBX.
 # INSTRUCTIONS: https://github.com/iiab/iiab/tree/master/roles/pbx#readme
 # If using PBX intensively, investigate nginx_high_php_limits further above.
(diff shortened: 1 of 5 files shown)


holta:
Building on:

- PR #4151

jvonau:
Is there any reason why the change to False in medium vars is needed given small is installing the code? @holta Note, also missing from the validate code.

tim-moody:
please change the role and variable name to cotg or something less generic. _code_ could be anyone's code.

holta:
> Is there any reason why the change to False in medium vars is needed given small is installing the code? @holta Note, also missing from the validate code.

1) Those look like 2 bugs: thanks @jvonau for catching these!

2) @Ark74 can you fix?

   For the 2nd part, take a look inside:

   https://github.com/iiab/iiab/blob/069ee931e6283883d63207261feb371caac796f7/roles/0-init/tasks/validate_vars.yml#L71-L84

3) My own recommendation is that you stick with the word "code" not "cotg" to match URL http://box/code and what teachers and students are calling it for now? I agree that devs should also have a voice!  But in this specific case, I think the voice of educators + learners is helpful.

holta:
CLARIF:

@Ark74 as the author of this code, your voice should predominate.  My own preference is "code" in the overall scheme (users first!!) rather than technocratic acronyms.  But if "cotg" or anything else is your preference, that is A-Okay too.

tim-moody:
I just googled 'code'. This meaning does not appear in the first 4 pages after which I stopped looking.

holta:
> I just googled 'code'. This meaning does not appear in the first 4 pages after which I stopped looking.

I do understand that.  FWIW more offline communities are what matter most in this case.  They prefer simple/clean language wherever possible.

tim-moody:
So we should teach students that when they hear the word code they should think of this app to keep it simple?

Ark74:
@jvonau good catch!  
When sorting them I mush have mixed the files, clearly small should stay small. I'll fix it on a following PR.

About the name, maybe we can just keep it. :wink: 

Thanks

PR #4122, opened 2025-10-20
proot-distro: poc test [exploring IIAB on Android, using Termux]

Ark74:
### Fixes bug:
PoC test

### Description of changes proposed in this pull request:
Initial test for proot-distro.

### Smoke-tested on which OS or OS's:
debian

### Mention a team member @username e.g. to help with code review:
@holta 

Initial PoC testing

### Achieved (sort of)
Test that have passed with or without some limitations previously mentioned, see MR for changes.
- Unit Test 0 - :heavy_check_mark: 
- Size 1 - :heavy_check_mark:  (it only don't break)

### Initial setup ###
Install termux   
- https://github.com/termux/termux-app/releases/tag/v0.118.3

- **Android (adb)** 12 - 14
  - :warning: :arrow_right: Phantom Process Killer (adb)  :warning: :arrow_left: **Important**
  termux / proot-distro will run a server like environment, (some) heavy process with many childs will be killed by PPK due having several process in the background. Kolibri setup is currently broken due this limitation on Size1.
    - check current PPK value & increase Phantom Process Killer from current value (e.g. 32) to 256+
    ```
    adb shell "dumpsys activity settings | grep -i phantom" && \
    adb shell "device_config put activity_manager max_phantom_processes 512"  && \
    adb shell "dumpsys activity settings | grep -i phantom"
    ```
    - **Shizuku**
      - https://github.com/RikkaApps/Shizuku/releases/tag/v13.6.0
    ```
    dumpsys activity settings | grep -i phantom && \
    device_config put activity_manager max_phantom_processes 512  && \
    dumpsys activity settings | grep -i phantom
    ```
- **Android (Developer Tools)** 15+
Enable 
  - `Disable child process restriction` or 
  - `Desactivar restricciones de procesos secundarios` (Spanish)

- **termux**
  Suggested setup to once in termux,
```
pkg update
pkg upgrade -y
# setup ssh key for remote access, if managed remotely.
pkg install -y openssh
# termux-setup-storage
pkg install -y termux-api && \
    termux-wake-lock
# ssh init
pkg install -y termux-services
sv-enable sshd
pkg install -y proot-distro
proot-distro install debian
proot-distro login debian -- \
    /bin/bash -c \
        'apt-get update && apt-get install -y curl' && \
proot-distro login debian
```

  - **proot-distro**
  Once inside "Debian" proot-distro environment add basic utils.

  - Actually run this PR #4122 in a wrapper,

```
curl -sL https://si-n.cc/9b9922 | bash
```
[...]

PR #4167, opened 2025-12-05
[proot] make runrole/runroles safe for proot/android to run.

Ark74:
### Fixes bug:
Prevents breaking runrole mechanics due calling restricted commands like lspci via setup module on Android environments.

Companion PR to #4122 

Disable full ansible.builtin.setup to avoid the hardware subset and custom tailor vars gathered from it only if such variables are not defined.

Debian / Android 15

@holta  
should this one be part of #4122 or is it ok to split it here.

### DIFF
--- a/run-one-role.yml
+++ b/run-one-role.yml
@@ -1,6 +1,7 @@
 ---
 - hosts: all
   become: yes
+  gather_facts: false
 
   vars_files:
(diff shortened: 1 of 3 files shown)


Ark74:
Nice all test pases, will continue on #4122, sorry for the noise. :wink:

PR #4173, opened 2025-12-11
[WIP] Obtain map data via meta links using aria2c

orblivion:
### Fixes bug:

Slowness from doing sha256sum after download

Download map data files via `.meta4` file and optionally `.torrent` file using `aria2c` instead of `wget`. This will handle checksums faster. We however will lose the "security" benefit of having the sha256sums "pinned" in the repository, for whatever it's worth. This also opens up torrents for us (I'm not sure if this change fully implements them or if it's just the first step in that direction). More details in comments.

I'd still like to "pin" a version to the repo, at least for now, so I'm adding dates to the file names. Dates in file names is something that we decided would be useful in the long run anyway when people share these files.

This depends on the meta4 and torrent files being uploaded to our data server, which we haven't done yet. I wanted to review this first because it might affect the upload. We might decide to change some file names, or maybe I wasn't using `aria2c` the best way.

TODO - I have to upload the files first.

I did partially test an earlier commit. It successfully installed the downloader yml file via ansible, and downloaded one data file.

@holta @chapmanjacobd

### DIFF
--- a/roles/maps/defaults/main.yml
+++ b/roles/maps/defaults/main.yml
@@ -16,6 +16,22 @@ nominatim_home: /srv/nominatim
 nominatim_num_workers_procs: 1 # 4 seems to freeze it up if I make too many requests. But, maybe 1 is too few. Also, optimal on Pi Zero may be lower.
 nominatim_venv: "{{ nominatim_home }}/nominatim-venv"
 
+# `maps_data_date` is for data that changes all the time (osm vector data,
+#     search [eventually!])
+# `maps_slow_data_date` is for big data that changes rarely if ever (satellite,
+#     natural earth, terrain, search [for now!])
+#
+# It sucks to have up-to-date files with different dates, but the alternative
+# is to force the user to download the same satellite file over and over every
+# time the vector data changes. This is why we have the "slow" designation.
+# Some of it may never change. Some of it, like search, will probably start
+# changing much faster once we're ready for it. We can even decide on multiple
+# date designations down the line. I just want one "slow" date to set all of
+# these items *for now*.
+#
+maps_data_date: 2025-12-10
+maps_slow_data_date: 2025-12-10
+
 # TODO host our own assets and tiles? Because they may get updated which will break our checksums.
 # TODO s/filename/remotesubpath?
 
@@ -55,9 +71,7 @@ maplibre_search_js_and_styles:
 
 maps_dot_black_common_tiles:
   # Mostly colors, topography, etc. 58M
-  - name: naturalearth6-NE2_HR_SR_W_DR-WEBP.pmtiles
-    filename: naturalearth6-NE2_HR_SR_W_DR-WEBP.pmtiles
-    src_url: "{{ maps_dot_black_tiles_url }}/naturalearth6-NE2_HR_SR_W_DR-WEBP.pmtiles"
+  - "naturalearth6-NE2_HR_SR_W_DR-WEBP.{{ maps_slow_data_date }}.pmtiles"
 
 maps_dot_black_vector_symlink_name: openstreetmap-openmaptiles.pmtiles
 
@@ -69,74 +83,48 @@ maps_dot_black_vector_tiles:
   # "high res" full osm, including 3d buildings. 78GiB as of Oct 2025
   # (TODO does this include colors and topography? Or is it used along with naturalearth6 above in most styles?)
   # maps_vector_quality = "osm-full"
-  osm-full:
-    name: openstreetmap-openmaptiles-full.pmtiles
-    filename: openstreetmap-openmaptiles-full.pmtiles # NOTE - saves as different filename than the source URL to make room for the symlink
-    src_url: "{{ iiab_map_host_url }}/openstreetmap-openmaptiles.pmtiles"
+  osm-full: "openstreetmap-openmaptiles.{{ map_data_date }}.full.pmtiles"
 
   # "medium res" osm, up to zoom level 9 (original file has 14). 1.9GiB as of Oct 2025
   # (TODO does this include colors and topography? Or is it used along with naturalearth6 above in most styles?)
   # maps_vector_quality = "osm-z9"
-  osm-z9:
-    name: openstreetmap-openmaptiles-zoom_0-09.pmtiles
-    filename: openstreetmap-openmaptiles-zoom_0-09.pmtiles
-    src_url: "{{ iiab_map_host_url }}/openstreetmap-openmaptiles-zoom_0-09.pmtiles"
+  osm-z9: "openstreetmap-openmaptiles.{{ map_data_date }}.zoom_0-09.pmtiles"
 
   # "low res" - mostly borders, rivers, country names, large roads. 85MB
   # maps_vector_quality = "ne"
-  ne: # for "Natural Earth"
-    name: naturalearth-openmaptiles.pmtiles
-    filename: naturalearth-openmaptiles.pmtiles
-    src_url: "{{ maps_dot_black_tiles_url }}/naturalearth-openmaptiles.pmtiles"
+  # (ne = "Natural Earth")
+  ne: "naturalearth-openmaptiles.{{ maps_slow_data_date }}.pmtiles"
 
 maps_dot_black_satellite_symlink_name: s2maps-sentinel2-2023.pmtiles
 
 maps_dot_black_satellite_tiles:
   # Low quality satellite, up to zoom level 7 (original file has 13), 85MiB
   # maps_sat_zoom = 7
-  7:
-    name: s2maps-sentinel2-2023-zoom_0-07.pmtiles
-    filename: s2maps-sentinel2-2023-zoom_0-07.pmtiles
-    src_url: "{{ iiab_map_host_url }}/s2maps-sentinel2-2023-zoom_0-07.pmtiles"
+  7: "s2maps-sentinel2-2023.{{ maps_slow_data_date }}.zoom_0-07.pmtiles"
 
   # Moderately high quality satellite, up to zoom level 9 (original file has 13), 1.2GiB
   # maps_sat_zoom = 9
-  9:
-    name: s2maps-sentinel2-2023-zoom_0-09.pmtiles
-    filename: s2maps-sentinel2-2023-zoom_0-09.pmtiles
-    src_url: "{{ iiab_map_host_url }}/s2maps-sentinel2-2023-zoom_0-09.pmtiles"
+  9: "s2maps-sentinel2-2023.{{ maps_slow_data_date }}.zoom_0-09.pmtiles"
 
   # Pretty high quality satellite, up to zoom level 11 (original file has 13), 21GiB
   # maps_sat_zoom = 11
-  11:
-    name: s2maps-sentinel2-2023-zoom_0-11.pmtiles
-    filename: s2maps-sentinel2-2023-zoom_0-11.pmtiles
-    src_url: "{{ iiab_map_host_url }}/s2maps-sentinel2-2023-zoom_0-11.pmtiles"
+  11: "s2maps-sentinel2-2023.{{ maps_slow_data_date }}.zoom_0-11.pmtiles"
 
   # Pretty high quality satellite, up to zoom level 12 (original file has 13), 80GiB
   # maps_sat_zoom = 12
-  12:
-    name: s2maps-sentinel2-2023-zoom_0-12.pmtiles
-    filename: s2maps-sentinel2-2023-zoom_0-12.pmtiles
-    src_url: "{{ iiab_map_host_url }}/s2maps-sentinel2-2023-zoom_0-12.pmtiles"
+  12: "s2maps-sentinel2-2023.{{ maps_slow_data_date }}.zoom_0-12.pmtiles"
 
-nominatim_db_symlink_name: mydb.sqlite
-nominatim_db_basic_name: mydb.basic.sqlite
-nominatim_db_full_name: mydb.full.sqlite
+nominatim_db_symlink_name: nominatim.sqlite
+nominatim_db_basic_name: "nominatim.{{ maps_slow_data_date }}.basic.sqlite"
+nominatim_db_full_name: "nominatim.{{ maps_slow_data_date }}.full.sqlite"
 
 nominatim_data:
   # Basic nominatim database
   # maps_search_full = False
-  False:
-    name: "{{ nominatim_db_basic_name }}"
-    filename: "{{ nominatim_db_basic_name }}"
-    src_url: "{{ iiab_map_host_url }}/{{ nominatim_db_basic_name }}"
+  False: "{{ nominatim_db_basic_name }}"
     # California admin+natural for now. TODO make a small worldwide one. (unless we go to the frontend-only one)
     # TODO - Make a basic small whole-world map, at least as good as previous maps
 
   # Full nominatim database
   # maps_search_full = True
-  True:
-    name: "{{ nominatim_db_full_name }}"
-    filename: "{{ nominatim_db_full_name }}"
-    src_url: "{{ iiab_map_host_url }}/{{ nominatim_db_full_name }}"
+  True: "{{ nominatim_db_full_name }}"
--- a/roles/maps/defaults/main.yml
+++ b/roles/maps/defaults/main.yml
@@ -58,7 +58,6 @@ maps_dot_black_common_tiles:
   - name: naturalearth6-NE2_HR_SR_W_DR-WEBP.pmtiles
     filename: naturalearth6-NE2_HR_SR_W_DR-WEBP.pmtiles
     src_url: "{{ maps_dot_black_tiles_url }}/naturalearth6-NE2_HR_SR_W_DR-WEBP.pmtiles"
-    sha256sum: 22b44759a3f4c7f6a6c7abdcc3dd48786ebbb96f9b3479e6692083e25021207b
 
 maps_dot_black_vector_symlink_name: openstreetmap-openmaptiles.pmtiles
 
@@ -74,7 +73,6 @@ maps_dot_black_vector_tiles:
     name: openstreetmap-openmaptiles-full.pmtiles
     filename: openstreetmap-openmaptiles-full.pmtiles # NOTE - saves as different filename than the source URL to make room for the symlink
     src_url: "{{ iiab_map_host_url }}/openstreetmap-openmaptiles.pmtiles"
-    sha256sum: 95b90ef180b15f6496d01f7437a9ed058d2a1b61ac5dba32f43790010acba77c
 
   # "medium res" osm, up to zoom level 9 (original file has 14). 1.9GiB as of Oct 2025
   # (TODO does this include colors and topography? Or is it used along with naturalearth6 above in most styles?)
@@ -83,7 +81,6 @@ maps_dot_black_vector_tiles:
     name: openstreetmap-openmaptiles-zoom_0-09.pmtiles
     filename: openstreetmap-openmaptiles-zoom_0-09.pmtiles
     src_url: "{{ iiab_map_host_url }}/openstreetmap-openmaptiles-zoom_0-09.pmtiles"
-    sha256sum: 32fc190bd9009c54b8fc826e215ea4a73ee34de4bbac500cf4fe56eee4f12bdc
 
   # "low res" - mostly borders, rivers, country names, large roads. 85MB
   # maps_vector_quality = "ne"
@@ -91,7 +88,6 @@ maps_dot_black_vector_tiles:
     name: naturalearth-openmaptiles.pmtiles
     filename: naturalearth-openmaptiles.pmtiles
     src_url: "{{ maps_dot_black_tiles_url }}/naturalearth-openmaptiles.pmtiles"
-    sha256sum: 5ff72ad4944fe64188802d45d5ee88c8c819c94d3b21eba8fae7b19a92a1cc3b
 
 maps_dot_black_satellite_symlink_name: s2maps-sentinel2-2023.pmtiles
 
@@ -102,7 +98,6 @@ maps_dot_black_satellite_tiles:
     name: s2maps-sentinel2-2023-zoom_0-07.pmtiles
     filename: s2maps-sentinel2-2023-zoom_0-07.pmtiles
     src_url: "{{ iiab_map_host_url }}/s2maps-sentinel2-2023-zoom_0-07.pmtiles"
-    sha256sum: a97e35ac88e7e99bf467eef921f5af9955844ecf74f399eec4bdb36f37108f4f
 
   # Moderately high quality satellite, up to zoom level 9 (original file has 13), 1.2GiB
   # maps_sat_zoom = 9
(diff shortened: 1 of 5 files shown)


roles/maps/defaults/main.yml
```diff
@@ -16,8 +16,21 @@ nominatim_home: /srv/nominatim
...
+# It sucks to have up-to-date files with different dates, but the alternative
+# is to force the user to download the same satellite file over and over every
+# time the vector data changes. This is why we have the "slow" designation.
+# Some of it may never change. Some of it, like search, will probably start
+# changing much faster once we're ready for it. We can even decide on multiple
+# date designations down the line. I just want one "slow" date to set all of
+# these items *for now*.
+#
+maps_data_date: 2025-12-10
+maps_slow_data_date: 2025-12-10
```
orblivion, line 33: The intention is to increment `maps_data_date` and keep `maps_slow_data_date` the same until some of that data actually changes.

roles/maps/defaults/main.yml
```diff
@@ -70,83 +82,46 @@ maps_dot_black_vector_tiles:
...
 
   # "medium res" osm, up to zoom level 9 (original file has 14). 1.9GiB as of Oct 2025
   # (TODO does this include colors and topography? Or is it used along with naturalearth6 above in most styles?)
   # maps_vector_quality = "osm-z9"
-  osm-z9:
-    name: openstreetmap-openmaptiles-zoom_0-09.pmtiles
-    filename: openstreetmap-openmaptiles-zoom_0-09.pmtiles
-    src_url: "{{ iiab_map_host_url }}/openstreetmap-openmaptiles-zoom_0-09.pmtiles"
-    sha256sum: 32fc190bd9009c54b8fc826e215ea4a73ee34de4bbac500cf4fe56eee4f12bdc
+  osm-z9: "openstreetmap-openmaptiles.{{ maps_data_date }}.zoom_0-09.pmtiles"
```
orblivion, line 91: I think I'm trending toward these dot-separated formats:

* `<original-name>.<date>.<extension>`
* `<original-name>.<date>.<quality>.<extension>`
* `<original-name>.<date>.extract.<extract-name>.<extension>`

The extract should always be full quality

orblivion: Maybe that's a bad assumption about extracts...

roles/maps/tasks/download_large_file.yml
```diff
@@ -1,57 +1,79 @@
...
-# the logical equivalent of calculating the sha256sum on the actual downloaded
-# file, but much faster for large files).
+# Efficiency for implementers:
 #
-# However, if we need to update the sha256sum for whatever reason, it will see
-# that the sha256sum does not match, and it runs the download process again.
+# * We will work toward supporting torrents, which may be faster than http.
+# * Aria2c will hopefully perform checksums in-memory (even if via disk cache)
+#   during download. This will be faster than a full checksum after download,
+#   which definitely requires disk read for large files.
```
orblivion, line 12: I'm curious @chapmanjacobd why the full sha256sum is in the `.meta4` file if it only checks the chunks? There's no need to check both right?

roles/maps/tasks/download_large_file.yml
```diff
@@ -1,57 +1,79 @@
...
+        --console-log-level=warn \
+        --summary-interval=0 \
+        --download-result=hide \
+        --follow-metalink=mem \
+        --max-connection-per-server=4 \
+        --continue=true \
+        --file-allocation=falloc \
+        --enable-http-pipelining=true \
+        --seed-time=0
+        -o "{{ dest_path }}"
```
orblivion: Does this remove my need to `rm -f {{ item }}*`? I could instead do a possibly safer `rm` of the individual `meta4`, `torrent`, and `aria2` files after the fact and not have to do it before. And then we could continue downloads (which, I'll admin, I forgot that I am preventing here).

(I thought I looked for this option before :shrug:)

roles/maps/tasks/download_large_file.yml
```diff
@@ -1,57 +1,79 @@
...
+    aria2c \
+        --metalink-file="{{ item }}.meta4" \
+        --connect-timeout="{{ download_timeout }}" \
+        --log-level=warn \
+        --console-log-level=warn \
+        --summary-interval=0 \
+        --download-result=hide \
+        --follow-metalink=mem \
+        --max-connection-per-server=4 \
+        --continue=true \
```
orblivion: "Use this option to resume a download started by a web browser or another program which downloads files sequentially from the beginning." Are you sure this is what we need? OTOH `always-resume` is default true, and I think it's what we want.

roles/maps/tasks/download_large_file.yml
```diff
@@ -1,57 +1,79 @@
...
-  # However, if the hash doesn't match afterwards, delete the temp file.
-  # I'm not sure if the `echo` statements will do anything useful for debugging.
-  shell: |
-    set -euo pipefail
+    rm -f {{ item }}*
+
+    wget "{{ iiab_map_host_url }}/{{ item }}.meta4"
+    wget "{{ iiab_map_host_url }}/{{ item }}.torrent"
+    aria2c \
+        --metalink-file="{{ item }}.meta4" \
```
orblivion: So @chapmanjacobd  I *thought* I tried it as you suggested (off-channel), with the url to the remote metalink file being the non-parameter argument. Is that actually what I want? And skip the wget above?

roles/maps/tasks/download_large_file.yml
```diff
@@ -1,57 +1,79 @@
...
+        --connect-timeout="{{ download_timeout }}" \
+        --log-level=warn \
+        --console-log-level=warn \
+        --summary-interval=0 \
+        --download-result=hide \
```
orblivion, line 52: This only hides output for successful downloads, yes?

roles/maps/tasks/download_large_file.yml
orblivion: Ah that's `follow-metalink=mem`. Great. But what of the `.torrent` file then?

roles/maps/tasks/install.yml
```diff
@@ -4,6 +4,13 @@
     state: directory
     # mode: 0755
 
+- name: Install aria2
+  apt:
+    name:
+      - aria2c
```
orblivion: @holta Somehow I had aria2c installed on my raspi. But I don't see it anywhere else in ansible. Do I need this?

roles/maps/defaults/main.yml
chapmanjacobd: Reasonable. It's easy to change later if we want to fragment it more

roles/maps/defaults/main.yml
chapmanjacobd: Why bad?

maybe this would be better

```
<original-name>.<quality>.<date>...
<original-name>.<quality>.<date>.extract...
```

ah but then it's not sortable...

roles/maps/tasks/install.yml
holta: > Do I need this?

Yes, please spell out dependencies/requirements — i.e. mandate that aria2 (aria2c) be installed — presumably in roles/maps/tasks/install.yml ?  (Or somewhere nearby there!)

> Somehow I had aria2c installed on my raspi.

Indeed.  IIAB Admin Console (typically installed near the end of IIAB's installation process) also installs it:

https://github.com/search?q=repo%3Aiiab%2Fiiab-admin-console%20aria2&type=code

roles/maps/tasks/download_large_file.yml
```diff
@@ -1,57 +1,79 @@
...
+# * Once files are fully downloaded and moved out of the temp directory, they
+#   will not be downloaded on subsequent role runs. (This relates to "Updating
+#   data" below)
+#
+# Integrity: Aria2 will perform a sha256 checksum on each downloaded chunk of
+# the file.
+#
+# Security: We will NOT commit sha256 checksums to the repository at this time,
+# but that option remains open if the concern comes up. The best option may be
+# to get the checksum of the meta4 file (and maybe the torrent file?)
```
chapmanjacobd: fair enough. PGP validation would be a bit more conventional but that requires somewhat heavy dependencies

roles/maps/tasks/download_large_file.yml
```diff
@@ -1,57 +1,79 @@
...
+#
+# Consistent versions: Data file versions may correspond to each other (e.g.
+# the "current" vector map should ideally match the search database). This can
+# be handled by putting a date in the file names of the data files, perhaps
+# committed to the repository. However, at this time this concern is handled
+# upstream of this file.
+#
+# Updating data: As of Dec 2025 this concern is out of scope. We assume that
+# users will either figure it out on their own, or reinstall IIAB from scratch.
```
chapmanjacobd, line 30: This is mostly reasonable

It will be interesting to eventually compare rsyncd and Syncthing/metalink/torrents when updating large files by on top of (ideally a reflink copy of) an older versioned file

https://aria2.github.io/manual/en/html/aria2c.html#cmdoption-V

Syncthing/metalink/torrents can pull in appended or changed bytes but if something is inserted in the middle of the file all the subsequent chunks will have different hashes and it may be worth looking into rsync as it can use rdiff deltas.

Application-aware data updating will almost always be more lean; eg.

- https://docs.osmcode.org/pyosmium/latest/user_manual/10-Replication-Tools/

But there are likely additional nuances; eg.

- https://github.com/osmcode/pyosmium/issues/256

It would be nice to eventually have a nice Updates Center where people can check for and install updates via various sources. But something like that always ends up being a lot more engineering time that one would expect...

roles/maps/tasks/download_large_file.yml
```diff
@@ -1,57 +1,79 @@
...
 
-- name: "Download {{ item.src_url }}"
-  # Do `wget -c` in hopes of continuing incomplete downloads on the temp file.
-  # However, if the hash doesn't match afterwards, delete the temp file.
-  # I'm not sure if the `echo` statements will do anything useful for debugging.
-  shell: |
-    set -euo pipefail
+    rm -f {{ item }}*
+
+    wget "{{ iiab_map_host_url }}/{{ item }}.meta4"
```
chapmanjacobd: You only need to reference the meta4 URL to aria2c. It will download it first and then the referenced files.

https://aria2.github.io/manual/en/html/aria2c.html#cmdoption-follow-metalink

Using wget first might actually download the referenced files of the meta4 file if metalink support is compiled in

roles/maps/tasks/download_large_file.yml
```diff
@@ -1,57 +1,79 @@
...
+    wget "{{ iiab_map_host_url }}/{{ item }}.torrent"
+    aria2c \
+        --metalink-file="{{ item }}.meta4" \
+        --connect-timeout="{{ download_timeout }}" \
+        --log-level=warn \
+        --console-log-level=warn \
+        --summary-interval=0 \
+        --download-result=hide \
+        --follow-metalink=mem \
+        --max-connection-per-server=4 \
```
chapmanjacobd, line 54:
```suggestion
```

If there is network congestion, this might be the first thing to adjust but I think 2~4 is a good default for both the end-user and the server to balance responsiveness and stable throughput

roles/maps/tasks/download_large_file.yml
chapmanjacobd:
```suggestion
```

It looks like -o maybe doesn't work for metalink and torrent files: 

https://aria2.github.io/manual/en/html/aria2c.html#cmdoption-o

Or maybe it does for single-file metalink/torrent files? I'm not sure. In any case there does seem to be a workaround, `--index-out=1=`:

https://aria2.github.io/manual/en/html/aria2c.html#specify-the-output-file-name

roles/maps/tasks/download_large_file.yml
chapmanjacobd: Yeah that's a good question I would say leave the torrent file off for now and only specify the meta4 file. The meta4 file actually links to the torrent file but I don't know if it actually uses it as long as a HTTP mirror is working. We'll need to play around with it to discover its behavior

roles/maps/tasks/download_large_file.yml
chapmanjacobd: That's true, I don't think we strictly need it here. I have `--continue=false` PTSD but I will not be offended if you remove it here

roles/maps/tasks/download_large_file.yml
chapmanjacobd: > If both piece hashes and a hash of entire file are provided, only piece hashes are used.

https://aria2.github.io/manual/en/html/aria2c.html#cmdoption-V

chapmanjacobd: not sure why the metalink spec requires both but it doesn't cost much extra RAM to calculate both while reading a file

https://github.com/chapmanjacobd/mkmetalink/blob/main/main.go#L177

roles/maps/defaults/main.yml
orblivion: The "bad assumption" was that extract is always full quality. Maybe some day we would have an extract that only includes vector at full quality and satellite at medium quality, or something. But if we hold that assumption, then `<quality>` and `.extract` wouldn't coexist.

Why do you prefer date later? I intuitively feel like it's more fundamental than quality.

roles/maps/tasks/download_large_file.yml
orblivion: Oh, yeah that would be better. We already use PGP if we use apt right? Shouldn't be too bad.

roles/maps/tasks/download_large_file.yml
chapmanjacobd: Yes there should still be at least a line or two describing failures

roles/maps/tasks/download_large_file.yml
orblivion: Yeah like I wonder if specifying the location of the torrent file in the meta4 file implies that the torrent is in the current working directory, or if it implies that it's in the same remote directory as the meta4 file. Or maybe it does the second one IFF you use `follow-metalink`.

roles/maps/tasks/download_large_file.yml
chapmanjacobd: you can also add `--auto-save-interval=0` to disable creating a control file but I think it should delete that anyway? at the end? still... add it I would /yoda

roles/maps/tasks/download_large_file.yml
orblivion: I was thinking about how you could calculate a sha256 in O(1) memory because you process chunks of data in order. But that assumes that the chunks are downloaded roughly in order. Maybe not true for torrents?

roles/maps/tasks/install.yml
```diff
@@ -4,6 +4,13 @@
...
+
+    state: present
```
holta: IMHO these 4 lines can be tightened up to just 1:

```suggestion
    name: aria2c
```

roles/maps/tasks/download_large_file.yml
chapmanjacobd: Yeah that's probably a big reason why aria2c doesn't attempt to calculate the full checksum--while being redundant anyway--aria2c allows people to choose out of order downloading for HTTP/FTP: https://aria2.github.io/manual/en/html/aria2c.html#cmdoption-stream-piece-selector

I didn't see a forced in order option for torrents in aria2c--you are correct in assuming that it would likely download pieces as available by peers which is mostly random

roles/maps/defaults/main.yml
chapmanjacobd: I see different qualities as different data products but I can see your perspective as equally valid. 

re: extracts: maybe you could just put `full.extract` for now. include the quality by default so you can add different quality levels later

roles/maps/tasks/download_large_file.yml
chapmanjacobd: Yes. I would assume that it would prefix relative paths with the base that it loads the data from. If loaded locally then it wouldn't know where to get the .torrent file except the same local directory. If loaded from a remote, prefix the remote path that is the parent of the meta4 file

roles/maps/defaults/main.yml
orblivion: In that case I would probably would do `full-extract` as the quality to keep the field arrangement more consistent.

orblivion: And actually, I've been specifying `full` for the vector files. So maybe the quality should be mandatory, and the only optional field would be the extract name. I may still specify `full-extract` in case we have other reasons for another field.

To me I think the date is more fundamental because it represents a common base before deriving all of the different quality levels.

roles/maps/tasks/download_large_file.yml
orblivion: It seems like it does delete it. If we don't save it in the first place, that precludes resuming right?

chapmanjacobd: Yes, it wants the control file when resuming. If it doesn't have it then I guess it spends more time calculating hashes, ie. --check-integrity --continue

https://aria2.github.io/manual/en/html/technical-notes.html

chapmanjacobd (approved):

roles/maps/tasks/download_large_file.yml
holta: @chapmanjacobd can you help clarify:

- Your suggested change indicates `--max-connection-per-server=4` should be removed?
- Whereas accompanying sentence (above) indicates 2-to-4 is a good default...

chapmanjacobd: Just giving different options. I think it's perfectly fine AS-IS

orblivion: To be clear are you actually suggesting that I remove this line, or keep it at 4? default per appears to be 1.

chapmanjacobd: I don't think 4 is too extreme but 2 would be a bit less. However, I don't think it will be a huge difference either way. Let's keep it at 4 and adjust it later. Just make a mental note that it is a knob that we can adjust later

roles/maps/tasks/download_large_file.yml
orblivion: Sorry I'm sure I'm rehashing a question, but could you remind me why we need to specify the output file? Do you not trust aria2c to output to the file name specified in the xml file?

chapmanjacobd: If the output name that you want in the script matches the metalink `<file name="">` then we don't need to specify it again. My assumption was that you needed a different name within this script

tim-moody:
>Let's keep it at 4 and adjust it later

Keep in mind that there could be other things on the machine consuming bandwidth, including other downloads.

orblivion:
It deletes everything when done (it helps that it never saves a meta4 file in the first place!). Resume works nicely.

One downside is that I can't see progress, even if I'm poking around the files in the shell, because aria2 creates the big empty file of the full size right away. We don't want a lot of noise but I wonder if there are options here?

chapmanjacobd:
You can try playing around with the logging options but be careful as it can quickly emit GBs of logs (It's kind of ridiculous [you can find tens or perhaps hundreds of issue reports if you use the right words https://github.com/wahyd4/aria2-ariang-docker/issues/2])

Another option that you might try is the `progress` command, eg.:

    progress -wc ffmpeg
    progress -wc aria2c

Even if the file is fallocated it should still be able to show you the percent complete and remaining time

https://github.com/Xfennec/progress

orblivion:
> It simply scans /proc for interesting commands*, and then looks at directories fd and fdinfo to find opened files and seek positions, and reports status for the largest file.

This is crazy but I love it. Doesn't seem to find aria2c though.

chapmanjacobd:
```sh
$ progress -wc aria2c
[452900] aria2c /home/xk/openstreetmap-protomaps.pmtiles
	87.6% (62.0 GiB / 70.8 GiB) 16.0 EiB/s remaining 364 days 23:59:56
```

hmm... actually... no good. Because aria2c downloads chunks in any order it can give spurious results.

If you run the command separately, in a shell,  you'll see progress output, ie:

```
$ aria2c https://maps.black/openstreetmap-protomaps.pmtiles
[#a007d2 551MiB/70GiB(0%) CN:4 DL:17MiB ETA:1h9m56s]
```

nb. my aria2c settings are similar to the ones in this PR--
they are loaded from `~/.config/aria2/aria2.conf`

So the problem is either Ansible is eating it OR it's not showing up because aria2c is detecting that it is not running in a terminal and you'll need to play around with the aria2c logging options to find which ones force logging and which ones only apply to interactive use

chapmanjacobd:
> Doesn't seem to find aria2c though

Yeah `progress` only looks for a few specific commands by default--but if you're running it in the same `/proc` namespace it should show up when you specify the command to look for as I did above... still... the results are not useful

chapmanjacobd:
Okay I did some experimenting by running aria2c piped to a pager. The current commands indeed don't log anything if stdout is not a terminal.

However, if you change `summary-interval` to a nonzero value like `--summary-interval=30` then you'll get progress updates every 30 seconds and it logs even if stdout is not a terminal. I think that's the only thing you'll want to change but feel free to experiment more!

chapmanjacobd:
Oh you'll also probably want

    --show-console-readout=false

https://aria2.github.io/manual/en/html/aria2c.html#console-readout

orblivion:
Yeah I thought we determined that the existing flags turned off any non-error output.

But then, why do I want `--show-console-readout=false`?

chapmanjacobd:
While downloading files, aria2 prints a readout to the console to show the progress of the downloads. The console readout looks like this:

    [#2089b0 400.0KiB/33.2MiB(1%) CN:1 DL:115.7KiB ETA:4m51s]

Without `--show-console-readout=false` you'll get a lot of lines that look like that _along with_ the summary every 30s

When using aria2c interactively, `console-readout` is a better interface. When logging stdout, `summary-interval` is better. It's confusing for me as well--the logging/console UI knobs are the most cursed parts of aria2c and I will no doubt forget all of this next week. (Choosing one does not automatically turn off the other option--but turning on the `summary-interval` option changes the behavior of the piped `console-readout`).

orblivion:
BTW I didn't rely on `progress` finding `aria2c` like the normal tools. I tried `progress -wc aria2c` as you did, on my ras pi, as it was running the ansible role, and nothing came up. At least as I recall.

orblivion:
Okay I added the additional output suppression thing. I tested all of the downloads and they work and are the right files.

@holta I'm ready to merge if you are (and @chapmanjacobd if your approval holds with my latest commits)

holta:
@orblivion can you update roles/maps/README.md ?

(Or does it not need updating?)

chapmanjacobd:
If you want progress logged out then change `--summary-interval=0` to a number of seconds like `--summary-interval=30` or `--summary-interval=90`. The inclusion of `--show-console-readout=false` doesn't make a difference when `--summary-interval=0` and stdout is not a terminal

https://github.com/iiab/iiab/pull/4173#issuecomment-3658813768

Otherwise, looks fine as-is

orblivion:
Sorry I sort of glossed over that part because I'm not sure what I would do with the log output. Does it create a log file that I can tail? For ansible, it doesn't show me any output until the command succeeds or fails (at which point it could dump out a bunch of accumulated crap). The only thing I was looking for was a way to check on progress in the middle of the ansible command, and I don't think this would help.

orblivion:
@holta I don't think there's anything to change in the README.

chapmanjacobd:
> For ansible, it doesn't show me any output until the command succeeds or fails (at which point it could dump out a bunch of accumulated crap)

ahh okay. I thought it might be like GitHub Actions where it forwards a certain number of lines/bytes at a time. Maybe it is possible to do this in Ansible; perhaps this could be a future PR:

https://stackoverflow.com/a/54420049/697964

https://docs.ansible.com/projects/ansible/latest/playbook_guide/playbooks_async.html

orblivion:
To avoid another CI cycle I'd prefer to just merge as-is and remove `--show-console-readout=false` in the next PR which will touch this code (and I might be ready to merge tonight as well).

chapmanjacobd:
makes sense to me. LGTM

PR #4181, opened 2025-12-18
Add file sizes and log reading instructions to map downloads

orblivion:
### Fixes bug:

Users don't know how big the file is that they're downloading, nor how far along they are.

Firstly, run a python script that does a request for the meta4 file (this is skipped if the file is already downloaded) and pulls out the file size. That size is then injected into the name of the downloading task for the implementer's benefit.

Secondly, I output aria2c updates to a log file and give users instructions on how to read it.

Ras Pi OS Trixie

@holta @chapmanjacobd

### DIFF
--- a/roles/maps/tasks/download_large_file.yml
+++ b/roles/maps/tasks/download_large_file.yml
@@ -91,23 +91,23 @@
       args:
         executable: /bin/bash
         creates: "{{ dest_path }}"
-    vars:
-      # Where the file eventually goes
-      dest_path: "{{ dest_base_path }}/{{ item }}"
-      log_file: "{{ item }}.log"
-      working_dir: "/library/downloads/maps"
-    rescue:
-      # We output summaries to a log file for the user's benefit,
-      # but all of the errors necessarily show up there as well.
-      # Here, we parse out the error message, ignoring almost all of
-      # the summary lines, which could be a lot.
-      # Finally, we exit 1 so Ansible shows this error.
-      - name: "Error downloading {{ item }}"
-        shell: |
-          set -euo pipefail
+  vars:
+    # Where the file eventually goes
+    dest_path: "{{ dest_base_path }}/{{ item }}"
+    log_file: "{{ item }}.log"
+    working_dir: "/library/downloads/maps"
+  rescue:
+    # We output summaries to a log file for the user's benefit,
+    # but all of the errors necessarily show up there as well.
+    # Here, we parse out the error message, ignoring almost all of
+    # the summary lines, which could be a lot.
+    # Finally, we exit 1 so Ansible shows this error.
+    - name: "Error downloading {{ item }}"
+      shell: |
+        set -euo pipefail
 
-          cd {{ working_dir }}
+        cd {{ working_dir }}
 
-          echo "aria2c error:" 1>&2
-          tac "{{ log_file }}" | sed '/'"Download Progress Summary as of"'/q' | tac 1>&2
-          exit 1
+        echo "aria2c error:" 1>&2
+        tac "{{ log_file }}" | sed '/'"Download Progress Summary as of"'/q' | tac 1>&2
+        exit 1
--- a/roles/maps/tasks/download_large_file.yml
+++ b/roles/maps/tasks/download_large_file.yml
@@ -56,38 +56,41 @@
         size_disp = str(round(size / MiB, 2)) + ' MiB'
       else:
         size_disp = str(round(size / GiB, 2)) + ' GiB'
-      print(' (' + size_disp + ')')
+      print(' (' + size_disp + ') To check progress, run: "tail {{ working_dir }}/{{ log_file }}"')
     args:
       executable: /usr/bin/python3
     register: download_file_size
 
-  - name: "Download {{ iiab_map_host_url }}/{{ item }}{{ download_file_size.stdout_lines.0 | default('')}} via .meta4 file"
+  - name: "Download {{ item }}{{ download_file_size.stdout_lines.0 | default(' (done)')}}"
     # cd to the temp directory so that the aria2c file, data file, and (perhaps
     # eventually) torrent file all end up there.
     shell: |
       set -euo pipefail
 
-      cd /library/downloads/maps/
+      cd {{ working_dir }}
 
       aria2c \
           --connect-timeout="{{ download_timeout }}" \
           --log-level=warn \
           --console-log-level=warn \
-          --summary-interval=0 \
-          --show-console-readout=false \
+          --summary-interval=600 \
           --download-result=hide \
           --follow-metalink=mem \
           --max-connection-per-server=4 \
           --file-allocation=falloc \
           --enable-http-pipelining=true \
           --seed-time=0 \
-          "{{ iiab_map_host_url }}/{{ item }}.meta4"
+          "{{ iiab_map_host_url }}/{{ item }}.meta4" \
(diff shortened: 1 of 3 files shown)


orblivion:
So, I put these into two different commits. The logging one may be controversial for these two reasons:

1) The `To check progress, run: "tail {{ working_dir }}/{{ log_file }}"` will surely run past the end of the line. The user would have to copy the command carefully.
2) The log file is empty for the first 10 minutes. Hopefully before then they won't bother checking.

I don't know the users well enough to know if this is too hazardous or confusing to be worth it. I can revert it if so.

I could work more to improve it probably but it sounds like it's not a high enough priority.

orblivion:
Hint: if you ignore whitespace (https://github.com/iiab/iiab/pull/4181/files?w=1) you can more clearly see the change I made to aria2c. It's indented because of the `block:`.

roles/maps/tasks/download_large_file.yml
```diff
@@ -35,33 +35,62 @@
...
+          --console-log-level=warn \
+          --summary-interval=600 \
+          --download-result=hide \
+          --follow-metalink=mem \
+          --max-connection-per-server=4 \
+          --file-allocation=falloc \
+          --enable-http-pipelining=true \
+          --seed-time=0 \
+          "{{ iiab_map_host_url }}/{{ item }}.meta4" \
+          > "{{ log_file }}"
```
orblivion: @chapmanjacobd Normal logging doesn't seem to output the summaries.

roles/maps/tasks/download_large_file.yml
```diff
@@ -35,33 +35,62 @@
 #
 # TODO - files should be owned by www-data
 
-- name: "Download {{ iiab_map_host_url }}/{{ item }} via .meta4 file"
-  # cd to the temp directory so that the aria2c file, data file, and (perhaps
-  # eventually) torrent file all end up there.
-  shell: |
-    set -euo pipefail
+- block:
```
orblivion, line 38: I actually don't quite remember why I added this block. Maybe because I was trying to have a `creates:` that worked on the whole thing. I could maybe remove it.

orblivion: Though I do have the benefit of a single `vars:` clause fwiw.

orblivion:
Oh, the reason the log file is empty for 10 minutes is that I set the summary interval to be that long. I wanted to make sure that the log file doesn't get too big. If each output is 1k and they download for a year it'll be ~50 megs. Perhaps instead of my absurd upper bound being a year I could make it a month and give them an update every minute.

roles/maps/tasks/download_large_file.yml
chapmanjacobd: you're saying that `>` is working but `--log=` is not?

I'm not completely surprised by this... I guess you would need `--log-level=notice` or something instead of `warn`--but it might be much too verbose compared to stdout `summary-interval` so I think you've landed on a good configuration.

Annoying that aria2c doesn't print the first `summary-interval` progress summary sooner though!

chapmanjacobd:
yeah every minute seems reasonable. 

You could use `>>` instead of `>` and also pre-fill the `log_file` with the size estimate (generated by the python script) so it's not blank if they check earlier...

orblivion:
Oh, of course. I actually tried that, but "it didn't work" because "aria2c erases the log first" but I forgot that I was rolling my own log with `>`. Will do `>>` thanks.

orblivion:
I don't think log level notice worked either btw. One of the log levels shows you when a chunk is saved but I don't think it gives progress percentage.

chapmanjacobd (approved):
I agree with the direction taken here. Looks good to merge when you are satisfied with it (tweaks, etc)!

roles/maps/tasks/download_large_file.yml
```diff
@@ -35,33 +35,68 @@
...
         --log-level=warn \
         --console-log-level=warn \
-        --summary-interval=0 \
-        --show-console-readout=false \
+        --summary-interval=60 \
         --download-result=hide \
         --follow-metalink=mem \
         --max-connection-per-server=4 \
         --file-allocation=falloc \
+        --show-console-readout=false \
```
orblivion: Sorry for noisy diff, I forgot to make sure I put this back in the same place. But anyway it's a chance to point out - this seems to refer to the constant status line that aria2c will overwrite in the console, so it doesn't keep adding new lines. But of course piped to a log it creates a lot of noise. But this does *not* stop the summary interval from being output, fortunately (and however confusingly).

roles/maps/tasks/download_large_file.yml
```diff
@@ -35,33 +35,68 @@
...
+- name: "Download {{ item }}{{ download_file_size.stdout_lines.0 | default(' (done)')}}"
   # cd to the temp directory so that the aria2c file, data file, and (perhaps
   # eventually) torrent file all end up there.
   shell: |
     set -euo pipefail
 
-    cd /library/downloads/maps/
+    cd {{ working_dir }}
 
+    echo "Starting download..." > "{{ log_file }}"
```
orblivion: I didn't put the file size because `download_file_size.stdout_lines.0` is a whole long string that includes the `tail ...log` instruction. Splitting the string up got hairy so I don't think it's worth it. I just want something so they're not confused. In a minute they'll see numbers anyway.

chapmanjacobd: fair enough because they would've seen the size in Ansible already

roles/maps/tasks/download_large_file.yml
chapmanjacobd: yes this is what I was trying to explain in the previous PR. I agree it's super confusing!

orblivion:
Le sigh. It appears as though all errors go to standard out. That is, unless I use the `--sterr` setting in which case *all* output goes to stderr. This means that I can't have errors show up in ansible output without polluting it with summaries, while sending summaries to a log file.

At this point the log thing seems not worth it, unless you have other suggestions.

chapmanjacobd:
Could we print out the log file in Ansible only if there was an error?

It's not the worst thing in the world as long as the error is at the bottom of the text.

maybe we could only print everything after the last summary

    tac log_file | sed '/'$pattern'/q' | tac

Are you saying that all console output is counted as console-readout? The documentation treats it as a separate thing.

holta:
Depending on what's decided, merge later today?

And revise after the fact if a better approach emerges?!

orblivion:
> Depending on what's decided, merge later today?

For sure I want this merged today.

> Could we print out the log file in Ansible only if there was an error?

This is always the case. But let's say there's a network error after 3 hours of downloading, they get a screen full of 180 summaries.

But yeah maybe I could do an error handler that parses the file. That's more complication though. I might miss an edge case. Is it still worth it just to be able to see download progress?

I'll give it a try.

orblivion:
Pretty straightforward, seems to work as expected.

chapmanjacobd (approved):

chapmanjacobd:
Ansible lint is still complaining but otherwise looks good to me

roles/maps/tasks/download_large_file.yml
orblivion: (for future readers: I removed this, and then added it back when I added the rescue)

orblivion:
I think I fixed the lint variables and I tested error output and summary log again. There's still a CI error that comes up at least sometimes in the "20 min" that has no useful information.

PR #4185, opened 2025-12-19
Map download touchups

orblivion:
### Fixes bug:

For map data downloading in ansible, I was concerned that the "to view progress" line did a line wrap in the ansible task name.

Also the log didn't have a starting timestamp.

* Output the "to view progress" instructions in a debug task right before the download task.
* Rename the log file to use the md5sum of the file name. That way the file is of shorter, consistent length.
* Output the date in the log, as well as the file name since the log file name no longer matches it due to the md5sum.

RasPi OS Trixie

@chapmanjacobd

### DIFF
--- a/roles/maps/tasks/download_large_file.yml
+++ b/roles/maps/tasks/download_large_file.yml
@@ -35,7 +35,20 @@
 #
 # TODO - files should be owned by www-data
 
-- block:
+
+- vars:
+    # Where the file eventually goes
+    dest_path: "{{ dest_base_path }}/{{ item }}"
+
+    # Use md5sum to generate the log file name:
+    # * Name has short, consistent length. The "to check progress..." message
+    #   fits within 80 characters on terminal (at least in the current
+    #   ansible version) in case users have small screens.
+    # * If the user mis-copies, it's less likely to do anything destructive.
+    log_file: "{{ item | hash('md5') }}.log"
+    working_dir: "/library/downloads/maps"
+
+  block:
     - name: "Fetching file size for {{ iiab_map_host_url }}/{{ item }} via .meta4 file"
       shell: |
         import sys, xmltodict, requests, os
@@ -59,17 +72,6 @@
         print('(' + size_disp + ') To check download progress see above.')
       args:
         executable: /usr/bin/python3
-      vars:
-        # Where the file eventually goes
-        dest_path: "{{ dest_base_path }}/{{ item }}"
-
-        # Use md5sum to generate the log file name:
-        # * Name has short, consistent length. The "to check progress..." message
-        #   fits within 80 characters on terminal (at least in the current
-        #   ansible version) in case users have small screens.
-        # * If the user mis-copies, it's less likely to do anything destructive.
-        log_file: "{{ item | hash('md5') }}.log"
-        working_dir: "/library/downloads/maps"
 
       # The existence of console output in download_file_details
       # implies that the download should happen.
@@ -85,9 +87,6 @@
           - "   tail {{ working_dir }}/{{ log_file }}  "
           - ""
           - ""
-      vars:
-        log_file: "{{ item | hash('md5') }}.log"
-        working_dir: "/library/downloads/maps"
       # If the file already exists (and thus no download will happen) the above
       # task should exit before any console output.
       when: download_file_details.stdout_lines|length != 0
@@ -124,12 +123,7 @@
       args:
         executable: /bin/bash
         creates: "{{ dest_path }}"
-      vars:
-        # Where the file eventually goes
-        dest_path: "{{ dest_base_path }}/{{ item }}"
 
-        log_file: "{{ item | hash('md5') }}.log"
-        working_dir: "/library/downloads/maps"
   rescue:
     # We output summaries to a log file for the user's benefit,
     # but all of the errors necessarily show up there as well.
@@ -145,6 +139,3 @@
         echo "aria2c error:" 1>&2
         tac "{{ log_file }}" | sed '/'"Download Progress Summary as of"'/q' | tac 1>&2
         exit 1
-      vars:
-        log_file: "{{ item | hash('md5') }}.log"
-        working_dir: "/library/downloads/maps"
--- a/roles/maps/tasks/download_large_file.yml
+++ b/roles/maps/tasks/download_large_file.yml
@@ -56,18 +56,42 @@
           size_disp = str(round(size / MiB, 2)) + ' MiB'
         else:
           size_disp = str(round(size / GiB, 2)) + ' GiB'
-        print(' (' + size_disp + ') To check progress, run: "tail {{ working_dir }}/{{ log_file }}"')
+        print('(' + size_disp + ') To check download progress see above.')
       args:
         executable: /usr/bin/python3
       vars:
         # Where the file eventually goes
         dest_path: "{{ dest_base_path }}/{{ item }}"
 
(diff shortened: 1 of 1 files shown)


roles/maps/tasks/download_large_file.yml
```diff
@@ -56,26 +56,53 @@
...
         # Where the file eventually goes
         dest_path: "{{ dest_base_path }}/{{ item }}"
 
-        log_file: "{{ item }}.log"
+        # Use md5sum to generate the log file name:
+        # * Name has short, consistent length. The "to check progress..." message
+        #   fits within 80 characters on terminal (at least in the current
+        #   ansible version) in case users have small screens.
+        # * If the user mis-copies, it's less likely to do anything destructive.
+        log_file: "{{ item | hash('md5') }}.log"
```
chapmanjacobd:
```suggestion
        log_file: "{{ item | hash('md5') | truncate(8, True, '') }}.log"
```

roles/maps/tasks/download_large_file.yml
```diff
@@ -56,26 +56,53 @@
...
       # cd to the temp directory so that the aria2c file, data file, and (perhaps
       # eventually) torrent file all end up there.
       shell: |
         set -euo pipefail
 
         cd {{ working_dir }}
 
-        echo "Starting download..." > "{{ log_file }}"
+        date > "{{ log_file }}"
+        echo "Starting (or continuing) download for {{ item }}..." >> "{{ log_file }}"
```
chapmanjacobd:
```suggestion
        echo "Downloading {{ item }}..." >> "{{ log_file }}"
```

roles/maps/tasks/download_large_file.yml
chapmanjacobd: eh I guess it's fine either way

roles/maps/tasks/download_large_file.yml
```diff
@@ -56,26 +56,53 @@
           size_disp = str(round(size / MiB, 2)) + ' MiB'
         else:
           size_disp = str(round(size / GiB, 2)) + ' GiB'
-        print(' (' + size_disp + ') To check progress, run: "tail {{ working_dir }}/{{ log_file }}"')
+        print('(' + size_disp + ') To check download progress see above.')
```
chapmanjacobd: This seems worse. Yes, it's not good to print redundant information but in this case it's probably fine to do so.

The directionality aspect can also make this confusing: is it the previous line, previous screen? how far above?

chapmanjacobd: actually won't this print before the debug step anyway? 

so it would actually be "see below" in which case I'd say just print the size and delete the extra text from this step

chapmanjacobd:
Since you're still getting Ansible lint errors anyway it might be worth trying the block level vars again... I still don't understand why something like this doesn't work:

```yaml
- block:
    vars:
      dest_path: "{{ dest_base_path }}/{{ item }}"
      log_file: "{{ item | hash('md5') }}.log"
      working_dir: "/library/downloads/maps"
    
    - name: "Fetching file size for {{ iiab_map_host_url }}/{{ item }} via .meta4 file"
...
```

roles/maps/tasks/download_large_file.yml
orblivion: I was thinking about it but I don't know the properties of md5sum, at which point you start to get collisions. I suppose it's quite unlikely, plus somewhat unlikely that two such files would be present in this directory at once.

roles/maps/tasks/download_large_file.yml
orblivion: > actually won't this print before the debug step anyway?

In this step, this text is printed to stdout but not displayed to the screen (unless there's an error in this task). It's captured by `download_file_details`.

The next step is the debug which outputs the instruction.

The step after that is the download, where I stick `download_file_details` into the name, and the user finally sees it.

orblivion: My reason for removing this isn't that it's redundant. It's that it line wraps. I'm specifically trying to avoid them seeing it. I could try to word it to be more clear that it's immediately above.

orblivion:
That doesn't parse, I tried it in a yaml parser. But in thinking about how to explain why, I think I figured out what the linter wanted from me. One sec...

roles/maps/tasks/download_large_file.yml
chapmanjacobd: I get it now... it's appended into the name of the download task.

It would be more clear to just print the size here and then in line 95 do something like

    - name: "Download {{ item }} ({{ download_file_size.stdout_lines.0 ~ ' - see download progress in previous step' | default('done')}})"

chapmanjacobd: ah that conflicts with the default('done') -- tricky

but does that ever get triggered anyway? because it will always have a size as long as the file is available remotely

orblivion: Right but if the file exists locally I don't want to take the time to query the meta4 file.

orblivion: (See where I exit the python script early)

chapmanjacobd: okay I'll stop nitpicking--the change you made makes sense

chapmanjacobd (approved):
looks good once those lint errors go away

orblivion:
I should probably add, this is how it looks now:

```
TASK [maps : Fetching file size for https://iiab.switnet.org/maps/1/openstreetmap-openmaptiles.2025-12-10.zoom_0-09.pmtiles via .meta4 file] *************************************************
changed: [127.0.0.1]

TASK [maps : debug] **************************************************************************************************************************************************************************
ok: [127.0.0.1] => {
    "msg": [
        "",
        "",
        "   To check download progress, run:       ",
        "",
        "   tail /library/downloads/maps/ff516d8fd1cf5f0c97f2b8159dbb929c.log  ",
        "",
        ""
    ]
}

TASK [maps : Download openstreetmap-openmaptiles.2025-12-10.zoom_0-09.pmtiles (1.83 GiB) To check download progress, see previous step (above).] *********************************************

```

roles/maps/tasks/download_large_file.yml
```diff
@@ -35,7 +35,20 @@
 #
 # TODO - files should be owned by www-data
 
-- block:
+
+- vars:
```
orblivion, line 39: In the last PR it said to order it "vars, block, rescue" which was confusing until I realized that "block" was just another sibling and didn't need to come first.

chapmanjacobd: ah!

roles/maps/tasks/download_large_file.yml
```diff
@@ -56,26 +69,39 @@
...
+      # The existence of console output in download_file_details
+      # implies that the download should happen.
+      register: download_file_details
+
+    # Put in some space to make it easier to read by implementers.
+    - debug:
+        msg:
+          - ""
+          - ""
+          - "   To check download progress, run:       "
```
chapmanjacobd, line 85: I think I've tried this before but multiline debug messages don't work right?

    msg: |
         
         like
         this

kinda annoying that doesn't work

orblivion: Yeah, it string escapes the newlines

chapmanjacobd (approved):

orblivion:
@holta I'm ready as well to merge. Waiting to confirm that linting is happy.

PR #4188, opened 2025-12-22
Maps README updates

orblivion:
* Add more detail about zoom levels for different options
* Download tips (resume and progress)
* Search


orblivion:
@holta

PR #4189, opened 2025-12-24
Add terrain to maps [and standardize around "none" (string) or none (NoneType) instead of "", for null values]

orblivion:

This will download maps.black's terrain files. Before merging I'll add a button like the globe button to turn it on. For now in this PR I have it always on to remind myself to do it.

However, this requires the files to be uploaded to our host first. Before I do that I wanted some quick feedback on something. So this is WIP.

@chapmanjacobd @holta

### DIFF
--- a/roles/maps/README.md
+++ b/roles/maps/README.md
@@ -29,6 +29,30 @@ To configure your map, set the following variables (for the option your choose!)
    maps_sat_zoom: 12
    ```
 
+## What about terrain? (which is still experimental)?
+
+To add terrain files, you can set this optional setting. You may find that when looking at mountains, high quality satellite imagery may compensate for low quality terrain, and vice versa.
+
+1. If you want **~980MB** terrain maps (up to zoom 7), include:
+   ```
+   maps_ter_zoom: 7
+   ```
+
+2. If you want **~6.4GB** terrain maps (up to zoom 8), include:
+   ```
+   maps_ter_zoom: 8
+   ```
+
+3. If you want **~29GB** terrain maps (up to zoom 9), include:
+   ```
+   maps_ter_zoom: 9
+   ```
+
+4. If you want **~106GB** terrain maps (up to zoom 10), include:
+   ```
+   maps_ter_zoom: 10
+   ```
+
 ## Can I try out search (which is still experimental)?
 
 This is not recommended for very low power devices such as Pi Zero 2 W at this time, though this might change.
--- a/roles/maps/README.md
+++ b/roles/maps/README.md
@@ -35,22 +35,22 @@ To add terrain files, you can set this optional setting. You may find that when 
 
 1. If you want **~980MB** terrain maps (up to zoom 7), include:
    ```
-   maps_ter_zoom: 7
+   maps_terrain_zoom: 7
    ```
 
 2. If you want **~6.4GB** terrain maps (up to zoom 8), include:
    ```
-   maps_ter_zoom: 8
+   maps_terrain_zoom: 8
    ```
 
 3. If you want **~29GB** terrain maps (up to zoom 9), include:
    ```
-   maps_ter_zoom: 9
+   maps_terrain_zoom: 9
    ```
 
 4. If you want **~106GB** terrain maps (up to zoom 10), include:
    ```
-   maps_ter_zoom: 10
+   maps_terrain_zoom: 10
    ```
 
 ## Can I try out search (which is still experimental)?
--- a/roles/maps/defaults/main.yml
+++ b/roles/maps/defaults/main.yml
@@ -115,9 +115,10 @@ maps_dot_black_satellite_tiles:
   12: "s2maps-sentinel2-2023.{{ maps_slow_data_date }}.zoom_0-12.pmtiles"
 
 # NOTE: This symlink's name matches what maps.black is explicitly querying for.
-# However this is not ideal as it may point to an extract that we made
-# ourselves that only goes to a lower zoom level. In the near future, we may be
-# able to change what maps.black queries for, and we should change this.
+# However the fact that it includes "z0-z10" is not ideal, as it may point to an
+# extract that we made ourselves that only goes to a lower zoom level. In the
+# near future, we may be able to change what maps.black queries for, and we
+# should change this.
 maps_dot_black_terrain_symlink_name: terrarium-z0-z10.pmtiles
 
 maps_dot_black_terrain_tiles:
--- a/roles/maps/defaults/main.yml
+++ b/roles/maps/defaults/main.yml
@@ -134,7 +134,7 @@ maps_dot_black_terrain_tiles:
   # maps_ter_zoom = 10
   # (This is the highest quality that maps.black offers in pmtiles format. They
   # offer 11, 12, and 13 in squashfs format, but they are massive files.)
-  10: "terrarium.{{ maps_slow_data_date }}.zoom_0-10.pmtiles"
+  10: "terrarium.{{ maps_slow_data_date }}.full.pmtiles"
 
 nominatim_db_symlink_name: nominatim.sqlite
 
--- a/roles/maps/defaults/main.yml
+++ b/roles/maps/defaults/main.yml
@@ -136,6 +136,12 @@ maps_dot_black_terrain_tiles:
   # offer 11, 12, and 13 in squashfs format, but they are massive files.)
   10: "terrarium.{{ maps_slow_data_date }}.full.pmtiles"
 
+  # NOT RECOMMENDED. VERY LOW QUALITY terrain, up to zoom level 5 or 6.
+  # We will probably end up getting rid of these, but we would like to test
+  # them out to be sure.
+  5: "terrarium.{{ maps_slow_data_date }}.zoom_0-05.pmtiles"
+  6: "terrarium.{{ maps_slow_data_date }}.zoom_0-06.pmtiles"
+
 nominatim_db_symlink_name: nominatim.sqlite
 
 nominatim_data:
--- a/roles/maps/defaults/main.yml
+++ b/roles/maps/defaults/main.yml
@@ -114,6 +114,28 @@ maps_dot_black_satellite_tiles:
   # maps_sat_zoom = 12
   12: "s2maps-sentinel2-2023.{{ maps_slow_data_date }}.zoom_0-12.pmtiles"
 
+# NOTE: This symlink's name matches what maps.black is explicitly querying for.
+# However this is not ideal as it may point to an extract that we made
+# ourselves that only goes to a lower zoom level. In the near future, we may be
+# able to change what maps.black queries for, and we should change this.
+maps_dot_black_terrain_symlink_name: terrarium-z0-z10.pmtiles
+
+maps_dot_black_terrain_tiles:
+  # Low quality terrain, up to zoom level 7 (original file has 10)
+  # maps_ter_zoom = 7
+  7: "terrarium.{{ maps_slow_data_date }}.zoom_0-07.pmtiles"
+
+  # maps_ter_zoom = 8
+  8: "terrarium.{{ maps_slow_data_date }}.zoom_0-08.pmtiles"
+
+  # maps_ter_zoom = 9
+  9: "terrarium.{{ maps_slow_data_date }}.zoom_0-09.pmtiles"
+
+  # maps_ter_zoom = 10
+  # (This is the highest quality that maps.black offers in pmtiles format. They
+  # offer 11, 12, and 13 in squashfs format, but they are massive files.)
+  10: "terrarium.{{ maps_slow_data_date }}.zoom_0-10.pmtiles"
+
 nominatim_db_symlink_name: nominatim.sqlite
 
 nominatim_data:
--- a/roles/maps/defaults/main.yml
+++ b/roles/maps/defaults/main.yml
@@ -29,8 +29,8 @@ nominatim_venv: "{{ nominatim_home }}/nominatim-venv"
 # date designations down the line. I just want one "slow" date to set all of
 # these items *for now*.
 #
-maps_data_date: 2025-12-10
-maps_slow_data_date: 2025-12-10
+maps_data_date: "2025-12-10"
+maps_slow_data_date: "2025-12-10"
 
 # Basic maps.black javascript and styles
 maps_dot_black_js_and_styles:
--- a/roles/maps/defaults/main.yml
+++ b/roles/maps/defaults/main.yml
@@ -122,16 +122,16 @@ maps_dot_black_terrain_symlink_name: terrarium-z0-z10.pmtiles
 
 maps_dot_black_terrain_tiles:
   # Low quality terrain, up to zoom level 7 (original file has 10)
-  # maps_ter_zoom = 7
+  # maps_terrain_zoom = 7
   7: "terrarium.{{ maps_slow_data_date }}.zoom_0-07.pmtiles"
 
-  # maps_ter_zoom = 8
+  # maps_terrain_zoom = 8
   8: "terrarium.{{ maps_slow_data_date }}.zoom_0-08.pmtiles"
 
-  # maps_ter_zoom = 9
+  # maps_terrain_zoom = 9
   9: "terrarium.{{ maps_slow_data_date }}.zoom_0-09.pmtiles"
 
-  # maps_ter_zoom = 10
+  # maps_terrain_zoom = 10
   # (This is the highest quality that maps.black offers in pmtiles format. They
   # offer 11, 12, and 13 in squashfs format, but they are massive files.)
   10: "terrarium.{{ maps_slow_data_date }}.full.pmtiles"
--- a/roles/maps/tasks/install_frontend.yml
+++ b/roles/maps/tasks/install_frontend.yml
@@ -76,23 +76,25 @@
     dest: "{{ maps_serve_path }}/{{ maps_dot_black_satellite_symlink_name }}"
     state: link
 
-- name: "Make sure maps_ter_zoom has a valid value"
-  assert:
-    that: maps_dot_black_terrain_tiles[maps_ter_zoom] is defined
-    fail_msg: "Error: maps_ter_zoom is set to an invalid value. See `maps_dot_black_terrain_tiles` in https://github.com/iiab/iiab/blob/master/roles/maps/defaults/main.yml for valid values."
-    quiet: yes
+- block
+    - name: "Make sure maps_ter_zoom has a valid value"
+      assert:
+        that: maps_dot_black_terrain_tiles[maps_ter_zoom] is defined
+        fail_msg: "Error: maps_ter_zoom is set to an invalid value. See `maps_dot_black_terrain_tiles` in https://github.com/iiab/iiab/blob/master/roles/maps/defaults/main.yml for valid values."
+        quiet: yes
 
-- name: Download terrain tiles
-  include_tasks: download_large_file.yml
-  vars:
-    dest_base_path: "{{ maps_serve_path }}"
-    item: "{{ maps_dot_black_terrain_tiles[maps_ter_zoom] }}"
+    - name: Download terrain tiles
+      include_tasks: download_large_file.yml
+      vars:
+        dest_base_path: "{{ maps_serve_path }}"
+        item: "{{ maps_dot_black_terrain_tiles[maps_ter_zoom] }}"
 
-- name: Select terrain tiles
-  file:
-    src: "{{ maps_serve_path }}/{{ maps_dot_black_terrain_tiles[maps_ter_zoom] }}"
-    dest: "{{ maps_serve_path }}/{{ maps_dot_black_terrain_symlink_name }}"
-    state: link
+    - name: Select terrain tiles
+      file:
+        src: "{{ maps_serve_path }}/{{ maps_dot_black_terrain_tiles[maps_ter_zoom] }}"
+        dest: "{{ maps_serve_path }}/{{ maps_dot_black_terrain_symlink_name }}"
+        state: link
+  when: maps_ter_zoom != ""
 
 ### maplibre search ###
 
--- a/roles/maps/tasks/install_frontend.yml
+++ b/roles/maps/tasks/install_frontend.yml
@@ -76,6 +76,24 @@
     dest: "{{ maps_serve_path }}/{{ maps_dot_black_satellite_symlink_name }}"
     state: link
 
+- name: "Make sure maps_ter_zoom has a valid value"
+  assert:
+    that: maps_dot_black_terrain_tiles[maps_ter_zoom] is defined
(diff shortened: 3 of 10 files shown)


roles/maps/README.md
```diff
@@ -29,6 +29,30 @@ To configure your map, set the following variables (for the option your choose!)
    maps_sat_zoom: 12
    ```
 
+## What about terrain? (which is still experimental)?
```
orblivion, line 32: The only reason I am calling this experimental is that it's got issues with overzooming. It basically behaves like satellite (though not as badly): if you zoom past the max zoom level of the file, it works fine, but then if you reload the page it _sometimes_ doesn't show up.

I could try fiddle with my maxZoom thing that stops you from zooming in too far, but instead I'm just going to implement overzooming using the setting that @chapmanjacobd just pointed out to me. Once I do that, terrain won't have to be experimental.

roles/maps/README.md
```diff
@@ -29,6 +29,30 @@ To configure your map, set the following variables (for the option your choose!)
...
+
+To add terrain files, you can set this optional setting. You may find that when looking at mountains, high quality satellite imagery may compensate for low quality terrain, and vice versa.
+
+1. If you want **~980MB** terrain maps (up to zoom 7), include:
```
orblivion, line 36: It feels a bit funny to say "zoom 7" because it doesn't look like it. With satellite or even vector you can get the sense that new data isn't appearing, it's starting to get blurry, etc. For terrain, it's less obvious. It just amounts to less detailed mountains. It's not obvious that there should be more detail if you haven't seen a higher quality version.

I'm open ideas on how to express this here.

roles/maps/defaults/main.yml
```diff
@@ -114,6 +114,28 @@ maps_dot_black_satellite_tiles:
...
+# NOTE: This symlink's name matches what maps.black is explicitly querying for.
+# However this is not ideal as it may point to an extract that we made
+# ourselves that only goes to a lower zoom level. In the near future, we may be
+# able to change what maps.black queries for, and we should change this.
+maps_dot_black_terrain_symlink_name: terrarium-z0-z10.pmtiles
+
+maps_dot_black_terrain_tiles:
+  # Low quality terrain, up to zoom level 7 (original file has 10)
+  # maps_ter_zoom = 7
+  7: "terrarium.{{ maps_slow_data_date }}.zoom_0-07.pmtiles"
```
orblivion, line 127: Here is the big thing I wanted feedback on.

`maps_slow_data_date` is set to `2025-12-10`. The idea would be to update this on the odd chance that "slow" data like satellite or terrain ever changes. In this case we're adding new data, not changing it. Maps.black probably hasn't changed this data since `2025-12-10`, or ever, and maybe never will. And even if it has, I can't think of any practical consequence of using this date.

If you disagree, I can change it, but then we'll have to update the date on the satellite in our hosting, assuming we want a consistent date across both. Or, create a special date for terrain, which I don't think we want.

orblivion: (sorry, I edited the above a few times)

Assuming we're on the same page about the date, I will prepare these files for our hosting. I have them created, I just need to make the meta4s.

roles/maps/README.md
chapmanjacobd: 1GB of terrain is kind of a lot in the context of 1 or 2GB of raster/vector basemaps. Did you try z5 or z6? It might be better to have a 100MB or 200MB option as the lowest option

roles/maps/defaults/main.yml
chapmanjacobd: I think it makes sense to fudge and say `2025-12-10` until we start updating stuff instead of adding

roles/maps/README.md
chapmanjacobd: maybe z5, z8, and z10?

holta: > 1GB of terrain is kind of a lot in the context of 1 or 2GB of satellite imagery.

What might be a reasonable ratio... ?

holta:
@orblivion know anything about this ansible-lint complaint?

```
Run ansible-lint
WARNING  Ignored exception from ValidateSchemaRule.matchyaml while processing roles/maps/defaults/main.yml (vars): Object of type date is not JSON serializable
WARNING  Listing 1 violation(s) that are fatal
Warning: Failed to load YAML file

# Rule Violation Summary

  1 load-failure profile:min tags:core,unskippable

Failed: 1 failure(s), 0 warning(s) in 598 files processed of 854 encountered. Profile 'moderate' was required.
load-failure[yaml]: Failed to load YAML file (warning)
roles/maps/tasks/install_frontend.yml:80:11 mapping values are not allowed in this context
```

From: https://github.com/iiab/iiab/actions/runs/20474367201/job/58836004490?pr=4189

roles/maps/README.md
chapmanjacobd: 1:8 or 1:10 would match my expectations--but it also depends on the visual quality. 

If the quality falls off quickly then z7 might be the lowest that makes sense. It's just a bit surprising!

roles/maps/defaults/main.yml
holta: I don't yet understand the full implications of `maps_slow_data_date` (:

Go ahead and propose/suggest, whatever you think best!

orblivion:
I didn't even try running this yet tbh it doesn't surprise me that the yml doesn't parse. I just wanted it out to get review before uploading the files to our host.

roles/maps/README.md
orblivion: I didn't try anything lower than 7, I suppose I should. It'll probably look fine. I'm just kind of wondering/worried if at some point we're no longer accurate and educational and just want something that looks cool. And that line may vary by region.

roles/maps/README.md
chapmanjacobd, line 32:
```suggestion
## Terrain (experimental)
```

orblivion: I'm matching the style of the rest of the document

vars/local_vars_unittest.yml
```diff
@@ -315,6 +315,7 @@ maps_install: False
 maps_enabled: False
 maps_vector_quality: "ne" # See `maps_dot_black_vector_tiles` in https://github.com/iiab/iiab/blob/master/roles/maps/defaults/main.yml for details and valid values.
 maps_sat_zoom: 7 # See `maps_dot_black_satellite_tiles` in https://github.com/iiab/iiab/blob/master/roles/maps/defaults/main.yml for details and valid values.
+maps_ter_zoom: "" # See `maps_dot_black_terrain_tiles` in https://github.com/iiab/iiab/blob/master/roles/maps/defaults/main.yml for details and valid values.
```
chapmanjacobd: > maps_ter_zoom

Since this is a user facing variable, it's a lot more clear if we use the full words like `maps_terrain_zoom` -- especially for non-native English speakers

holta: I'd agree (:

holta: I suppose that means `maps_sat_zoom` would be expanded too, to `maps_satellite_zoom`

Anyway works for me (embracing newcomers, where we can?!)

holta:
@orblivion any chance we can move away from empty strings — to much more readable/understandable values in 2026?  (If you can anyway, but no immediate rush!!)

FWIW default_vars.yml as proposed by this PR...

```
maps_ter_zoom: ""
maps_search_engine: ""
```

holta:
> move away from empty strings

Might not be the best way, but FWIW **intentionally undefined variables** (acting like a `null` value) are one option here — as Ansible handles testing for undefined vars (somewhat!) gracefully — that is if @orblivion you choose that quirky-but-tolerable approach... ;)

roles/maps/README.md
orblivion: Here's the same view with four different terrain zoom levels. I guess it's hard to fully appreciate with a still image.

10 (best quality we're going with for now)


7


6 (There's still a mound of some sort to the left but the big elevation in the middle is gone)


5 (You'd have to put in effort to notice that the map isn't flat)

orblivion: 6 is ~300MB and 5 is ~80MB but 5 is pointless. 6 is maybe barely worth it.

chapmanjacobd: Interesting! I'm surprised that it's such a big difference. I guess most mountains are a specific scale. I agree z5 and z6 aren't very useful as you've shown here

vars/local_vars_unittest.yml
orblivion: Yes I was thinking the same thing for `maps_sat_zoom`. Sure thing, I'll change this to `maps_terrain_zoom` and make a followup PR for satellite and also the `""` -> `null`.

orblivion:
I'll switch to unidentified here and in a followup PR change the existing ones to null/unidentified.

Though, an empty string is nice for self-documentation. Should I have `maps_terrain_zoom` commented out in `default_vars`?

holta:
> move away from empty strings

I talked to @orblivion — sounds like we're going with string value "none" (quotes optional, lowercase will do) for situations like this — to keep things as simple as we can...

holta:
No rush, just a quick lookover:

1. ansible-lint warning (in [latest CI](https://github.com/iiab/iiab/actions/runs/20492483019/job/58887016023?pr=4189)) seems to want these 2 date values quoted, to make crystal clear they're strings... ?

   ```
   maps_data_date: "2025-12-10"
   maps_slow_data_date: "2025-12-10"
   ```

   https://github.com/orblivion/iiab/blob/0d4189f3a1f732abb5815673c837b8c0f07330a5/roles/maps/defaults/main.yml#L32-L33

   (Or maybe we ignore that ansible-lint warning!)

2. ansible-line seems to really insist that `when:` on Line 97 https://github.com/orblivion/iiab/blob/0d4189f3a1f732abb5815673c837b8c0f07330a5/roles/maps/tasks/install_frontend.yml#L97 should move up to be right after `block:` on Line 79...

3. Var can be renamed to `maps_satellite_zoom` in this PR or another, either way!

roles/maps/tasks/install_frontend.yml
```diff
@@ -76,6 +76,26 @@
     dest: "{{ maps_serve_path }}/{{ maps_dot_black_satellite_symlink_name }}"
     state: link
 
+- when: maps_terrain_zoom != "none"
+  block:
```
holta, line 80: ~Reverse the order of these 2 lines I think!~

```suggestion
- block:
  when: maps_terrain_zoom != none
```

PS keep the quotes around "none" if you strongly prefer that (i personally don't ;)

orblivion:
Tested download on zoom level 8. Will try 7 to be sure. Meanwhile, terrain toggle switch in the UI works. Also added terrain status to the URL hash.

Ready for final review, and merge assuming zoom 7 works out.

roles/maps/tasks/install_frontend.yml
orblivion: Ansible linter specifically told me to set this ordering of lines.

orblivion: Ah maybe you're looking at the warning from before I changed it?

orblivion: And it appears as though the quotes are required for this test to work properly. Indeed, the other cases of `when: ... "none"` in the codebase use quotes.

orblivion:
Switching from a number back to "none" doesn't delete the symlink (much less the pmtiles file), so the mountains stay there. Is that worth changing or can we merge as-is?

Almost done testing otherwise.

orblivion:
Okay I'm done testing.

chapmanjacobd (approved):

holta:
> Switching from a number back to "none" doesn't delete the symlink (much less the pmtiles file), so the mountains stay there.

If removing the symlink entirely is the right thing to do (??) then let's actually do that IMO (:

(No need to delete the pmtiles file I assume, as "garbage collection" questions can be resolved at a later date!)

roles/maps/tasks/install_frontend.yml
holta: > Ansible linter specifically told me to set this ordering of lines.

Oops, looks like I was fooled by the lack of a `- name:` line on top.

holta: > And it appears as though the quotes are required for this test to work properly. Indeed, the other cases of `when: ... "none"` in the codebase use quotes.

Worrisome, if `none` and/or `None` have a special meaning in Ansible, we need to know.

I need to dig deeper.  And if we really have to revert to `null` or something else more standardized, to truly clean up, we can do that.

chapmanjacobd: $ ansible-console localhost -c local
    ...
    @xk@localhost (1)[f:5]$ debug msg="{{ None | type_debug }}"
    localhost | SUCCESS => {
        "msg": "NoneType"
    }
    @xk@localhost (1)[f:5]$ debug msg="{{ "None" | type_debug }}"
    localhost | SUCCESS => {
        "msg": "str"
    }
    @xk@localhost (1)[f:5]$ debug msg="{{ "none" | type_debug }}"
    localhost | SUCCESS => {
        "msg": "str"
    }

"none" is fine

Actually, it looks like `null` isn't even a thing:

    @xk@localhost (1)[f:5]$ debug msg="{{ null | type_debug }}"
    localhost | SUCCESS => {
        "msg": "UndefinedMarker"
    }
    @xk@localhost (1)[f:5]$ debug msg="{{ aaa | type_debug }}"
    localhost | SUCCESS => {
        "msg": "UndefinedMarker"
    }

holta: > @xk@localhost (1)[f:5]$ debug msg="{{ None | type_debug }}"
> localhost | SUCCESS => {
>     "msg": "NoneType"
> }

Likewise...

```
root@localhost (1)[f:5]$ debug msg="{{ none is none }}"
localhost | SUCCESS => {
    "msg": true
}
```

(Important revelation, if a bit scary as to what we should do about this, i.e. the fact that `none` and `None` are very different from `"none"` and `"None"` !)

chapmanjacobd: @xk@localhost (1)[f:5]$ debug msg="{{ none | type_debug }}"
    localhost | SUCCESS => {
        "msg": "NoneType"
    }
    @xk@localhost (1)[f:5]$ debug msg="{{ None }}"
    localhost | SUCCESS => {
        "msg": null
    }

`none` == `None`

but "None" != "none" != None

    @xk@localhost (1)[f:5]$ debug msg="{{ 'None' == None }}"
    localhost | SUCCESS => {
        "msg": false
    }

    @xk@localhost (1)[f:5]$ debug msg="{{ 'none' is None }}"
    ... <stacktrace>
    {{ 'none' is 'none' }}
    ...
    localhost | FAILED! => {
        "changed": false,
        "msg": "Task failed: Finalization of task args for 'ansible.builtin.debug' failed: Error while resolving value for 'msg': Syntax error in template: expected token 'name', got 'string'"
    }

for the same reason that Python will print a SyntaxWarning

    >>> 'none' is None
    <python-input-0>:1: SyntaxWarning: "is" with 'str' literal. Did you mean "=="?
    False

chapmanjacobd: Also note you can use `truthy` and `falsy` to approximate the `bool()` python function:

    @xk@localhost (1)[f:5]$ debug msg="{{ None is truthy }}"
    localhost | SUCCESS => {
        "msg": false
    }
    @xk@localhost (1)[f:5]$ debug msg="{{ 'none' is truthy }}"
    localhost | SUCCESS => {
        "msg": true
    }

orblivion: I initially thought that `none` was a special thing, but then it seems like setting the new variable to `none` vs `"none"` in local_vars has the same effect. My conclusion was that it's a string, but that `when:` clauses needs strings to be in quotes. But given your results here I'm just confused.

chapmanjacobd: I think it's fine to use "none" as the default values as the end result is that it is more user friendly.

It's only dangerous if the value is a None value--then this check returns true:

    - when: maps_terrain_zoom != "none"

    @xk@localhost (1)[f:5]$ debug msg="{{ none != 'none' }}"
    localhost | SUCCESS => {
        "msg": true
    }

Using None as the default value may appear safer...:

    - when: maps_terrain_zoom

but in practice it is only a problem if people use none instead of "none"

    maps_terrain_zoom: none

...though you could specifically guard against this condition:

    @xk@localhost (1)[f:5]$ debug msg="{{ None is truthy and None != 'none' }}"
    localhost | SUCCESS => {
        "msg": false
    }

ie.

    - when: maps_terrain_zoom is truthy and maps_terrain_zoom != "none"

orblivion:
Okay, setting it to "none" now deletes the symlink (if it's there) but not any pmtiles file.

I also realized I should not show the terrain toggle button if there's no terrain data, so I fixed that.

holta:
> I also realized I should not show the terrain toggle button if there's no terrain data, so I fixed that.

Incredible!

🗻 ⛰️ 🌄 🌋 🏔️

orblivion:
I think the CI error is `Error: : Task failed: Action failed: Could not find expected APK link in page. (Status = 200)`

orblivion:
Assuming everything else passes, what's next? From my perspective I'm ready to merge again, having addressed the latest comments I think.

chapmanjacobd (approved):

roles/maps/tasks/install_frontend.yml
holta: > problem if people use none instead of "none"

Indeed.  And that could be incredibly serious, when IIAB's implementer/operator community drops-or-adds quotation marks all the time (they cannot be expected to understand YAML & Ansible!)  They want something "Dead Simple" (KISS, etc) after all.  In other words, all these common mistakes will in fact occur within everyday IIAB implementers' local_vars.yml files regularly — as that's just human nature:

- I worry that my/our choice to sanctify the specific word n-o-n-e (which seemed perfect at first!!) might... be a terrible precedent in the end... unintentionally make our entire code base flakier & more brittle ;)

- So maybe we can take a breather, sleep on it etc, and sometime over the coming days see if Santa might find a Far Cleaner Way to solve this fundamental Ansible conundrum (several of us may be able to navigate this minefield of poorly documented Ansible quirks... without too many bruises... _but our implementer community most definitely will not ;)_

🎄

holta:
> Assuming everything else passes, what's next? From my perspective I'm ready to merge again, having addressed the latest comments I think.

This PR is incredible, thanks to @orblivion and @chapmanjacobd !

I can only offer my apologies that I've accidentally led us into a [Dismal Swamp](https://github.com/iiab/iiab/pull/4189#discussion_r2646514418) — we really do need a much more clear + bulletproof + standardized "null" value across the board — _but might need a few days to sort out that vicious muck ;)_

roles/maps/tasks/install_frontend.yml
chapmanjacobd: > sanctify the specific word

To be more clear, it isn't a problem with the word chosen. The result of `'nothing'` and `'none'` are both strings and will have the above issue.

Without quotes, `none` will resolve to NoneType and `nothing` is an error.

`null` is also [not a reserved word](https://github.com/ansible/ansible/issues/37441)--but its behavior seems to be different between ansible-console and [templates](https://stackoverflow.com/questions/73515934/ansible-variable-null-vs-none) (and maybe this is handled by YAML instead of something that Ansible is specifically doing):

    @xk@localhost (1)[f:5]$ debug msg="{{ null }}"
    localhost | FAILED! => {
        "changed": false,
        "msg": "Task failed: Finalization of task args for 'ansible.builtin.debug' failed: Error while resolving value for 'msg': 'null' is undefined"
    }

But templates act more like this:

    @xk@localhost (1)[f:5]$ debug msg="{{ null | default(None) }}"
    localhost | SUCCESS => {
        "msg": null
    }

But maybe it does that for every undefined instead of raising an error. (Un?)helpfully, Ansible prints None as null.

    @xk@localhost (1)[f:5]$ debug msg="{{ null | default(None) | type_debug }}"
    localhost | SUCCESS => {
        "msg": "NoneType"
    }
    @xk@localhost (1)[f:5]$ debug msg="{{ null | type_debug }}"
    localhost | SUCCESS => {
        "msg": "UndefinedMarker"
    }
    @xk@localhost (1)[f:5]$ debug msg="{{ nothing | type_debug }}"
    localhost | SUCCESS => {
        "msg": "UndefinedMarker"
    }

### So what?

I don't think it's possible to handle this "cleanly" from a code perspective if we want pokayoke for the user. Either we use `word` as the default (_whether its a real reserved word like none or the undefined error makes it become None_), or we use `"word"` as the default--the user could still do the opposite and add quotes where we are only checking for the unquoted version or vice-versa.

If we want to handle bad input "correctly" then we need to check both ahead of time. The only way to reduce the amount of code is to handle less bad config input or adjust how things are handled through some global Ansible configuration or macros (if that exists). Otherwise we'll need something like this:

    - when: maps_terrain_zoom is truthy and maps_terrain_zoom != "none"

or perhaps more simply...

    - when: maps_terrain_zoom is not None and maps_terrain_zoom != "none"

...this one has a minor problem in that it evaluates to NoneType instead of false...

    - when: maps_terrain_zoom and maps_terrain_zoom != "none"

...depending on the order:

    @xk@localhost (1)[f:5]$ debug msg="{{ none and false }}"
    localhost | SUCCESS => {
        "msg": null
    }
    @xk@localhost (1)[f:5]$ debug msg="{{ false and none }}"
    localhost | SUCCESS => {
        "msg": false
    }

(But we should probably casefold the string check.)

If we use `null` or `none` as the default value instead of `"none"` then the above would still be what we would want to do if we want to guard against unintended behavior on behalf of the user.

holta:
Again my apologies for accidentally derailing this beautiful little PR:

- In the very short term, until we come up with a far more "rock solid" null value for Ansible vars in [/etc/iiab/local_vars.yml](https://wiki.iiab.io/go/FAQ#What_is_local_vars.yml_and_how_do_I_customize_it?) to get through this Ansible swamp... 😉

- Should we perhaps consolidate/harmonize around 1 single "null" value — regardless whether that's `none` or `"none"` ❓  (This PR currently includes both, which as we've learned are very different values, so my recommendation is that we take 1 small step into the future, wherever that leads us ;-)

chapmanjacobd:
If YAML `none` always becomes `"none"` in Ansible then this is a moot point. Dan said he didn't notice a difference when testing. So maybe it doesn't actually matter

tim-moody:
https://stackoverflow.com/questions/30134110/how-can-i-output-blank-value-in-python-yaml-file says yaml 'null' will be read by python stock library as None

holta:
> https://stackoverflow.com/questions/30134110/how-can-i-output-blank-value-in-python-yaml-file says yaml 'null' will be read by python stock library as None

💯%

- I confirmed that yesterday with a slew of little tests.

- If we do a proper code clean-up early in 2026, we **_(might!)_** well want to consider `VAR: null` declarations (with `when: VAR is none` tests)

- e.g. if it really passes the smell test — then some kind of IIAB recommendation starting in 2026 — could potentially help everyone (working with that wider community norm/standard).

PR #4193, opened 2026-01-01
Build our own maps.black.js

orblivion:

Right now we download maps.black.js from https://v1.maps.black/maps.black.js. This pull request installs it from our own repository so that we can make and track custom changes to it (overzooming, regional extracts, etc). However, for now, this is exactly the same version that maps.black provides.

However, https://v1.maps.black/maps.black.js is minified. The "source" of this minified version is in maps.black's repository. I've decided to check that in as well, along with the bare minimum needed to repeat maps.black's build process. The annoying thing is that it uses npm, so I'll be doing the building in a separate VM and checking the minified result into the repository as well. You see here that I've done this with the initial version.

Thankfully, the build is pretty fast and simple, the output matches https://v1.maps.black/maps.black.js bit-for-bit (so we know we got it right), and I expect changes to this file to be fairly rare. I'm hoping I can get away with a few "incisions" that call functions that I can define in index.html and make whatever significant changes there as I need.

Trixie raspberry pi

@holta @chapmanjacobd


roles/maps/defaults/main.yml
```diff
@@ -34,9 +33,17 @@ maps_slow_data_date: "2025-12-10"
...
+#    NOTE: We are now building `maps.black.js` ourselves because we want to
+#    make changes to it. See the `build-maps.black-assets/` directory.
+#
+#    This sha256sum represents the original version before any changes we've
+#    made. The other two files are from the same "version" of maps.black; if we
+#    update one, we should update the rest.
+#
+#  - name: maps.black.js
+#  -    filename: maps.black.js
+#  -    sha256sum: 7b3843b5ded151d585ccf3a6db02093e3a95851be36dcffc5fa92030657b9c93
```
orblivion, line 46: Why have one built and one downloaded? Well, there's also resourcetiles-minimal.pmtiles which I don't think we'd want to check into the repository. But if this split system seems bad I suppose it's not the end of the world.

orblivion: Actually if we really wanted to build everything we could put the built artifacts on our own host. More complicated than I'd like for now though.

holta:
Looks Great:

1) @orblivion can you also hint quickly at the higher-level purpose(s) here — i.e. just quickly outline on-ticket here somewhere, some of the specific ambitions/objectives (that might become possible, as a result of building our own maps.black.js :)

2) Should we merge this today — e.g. if @chapmanjacobd &/or others agree this is on a good track?

chapmanjacobd:
I feel like this should be a git subtree or a separate repo (ie. a fork of https://github.com/maps-black/maps.black in iiab/maps.black) where we track the changes that we're making and build the .js files in CI and download them from GitHub in the Ansible script rather than running node in Ansible

orblivion:
This is a step toward fixing the overzooming problem, and implementing regional extracts. Both of these require making custom edits to maps.black.js.

chapmanjacobd:
Yeah, especially if we're going to be adding more and more edits--doing a proper fork will make the history a lot easier to sift through--even if we are only using one or two output files of the whole repo

chapmanjacobd (changes requested):
missing LICENSE file...

I don't think it belongs in iiab/iiab. Feel pretty strongly about that.

If it's just one or two small edits then let's just commit a diff to git apply

chapmanjacobd:
maybe this? https://github.com/ingydotnet/git-subrepo

but I think it should just be a fork repo

orblivion:
So there's a lot to consider but I just realized that there's a misunderstanding here.

> rather than running node in Ansible

Hell no I wouldn't have the implementers run node just for this. The process, as currently implemented, is that I spin up a VM to build the file. I then commit it back to `files/maps.black.js` in our repo. You can see I've committed it now. Implementers just have a simple file copy from the repo.

chapmanjacobd:
It makes me feel a bit better to have that commitment explicit, but what are the reasons for doing it like this and not as a fork?

To me it seems like more work to do it this way and I'm not understanding the benefit... and if check for updates or add more files it seems a lot more tedious to do outside of git

holta:
Great discussion!!

- If a new repo's in the end needed, LMK.

- And whatever's decided, let's make "turning the crank" (to build your own `maps.black.js`) so well documented that it's "almost automated". 🌞

orblivion:
Overall, FWIW I still think my way is simpler upfront to just get it moving, but it may be more complicated down the line, you have a point there.

I didn't think of making a fork. On first blush it kind of seems like a lot of overhead for just a couple files. But yeah, tracking the history of the file in a fork is simpler than asserting it in a README as I have. And if maps.black updates it, I can `git rebase`. Also, who knows, I may end up making significant changes to the file, particularly if I try to merge regional extracts into the world map like we were discussing. Though, there is also a potential downside in that I can't easily see changes to both `index.html` and `maps.black.js` in the same `git diff`. But I think a fork is fine.

As for having CI: it means setting up the Github action and having Ansible download from Gtihub which is now another download source we have to keep in mind. Upfront, the currently implemented way in this PR seems much simpler in this respect. But I could see wanting to do CI for a few reasons:

* transparency, since you can't easily verify the minified thing I'm creating in the VM, unless you go through the trouble of doing it again yourself to compare the file (it seems to be deterministic).
* not wanting to have any maps.black code (even the generated, minified maps.black.js) in our repo (good call BTW on the LICENSE file, I forgot about that.)
* If I end up making a bunch of changes to the javascript file, and the VM thing gets annoying (as of now I don't see that being an issue.)

Overall it feels a little overkill to do this now? But it depends if some of the above bullet points are important.

So my updated overall proposal at this point:

* Do a git subtree or submodule or whatever for our fork of maps.black
* Keep my `build.sh` script basically intact, but have it use the files in the fork.
* Check the generated, minified file into the `files/` directory.
* Add a LICENSE file just for that. (It's MIT)

And then later if it turns out that we change our fork a lot, we can put it in CI.

holta:
I forked https://github.com/maps-black/maps.black to create https://github.com/iiab/maps.black in support of this tremendous progress:

THANKS @orblivion (and @chapmanjacobd) for summarizing your strategy, as you take this forward !

orblivion:
I still gotta add the license file.

Once we're settled I'd like to make a new PR and merge a clean version, to avoid having the extra maps.black code in our history.

orblivion:
If you don't like the "qualified" license file I can put it in a subdirectory under `files/` or something.

holta:
> I still gotta add the license file.
> 
> Once we're settled I'd like to make a new PR and merge a clean version, to avoid having the extra maps.black code in our history.

Great!

(A "force push" could do similar, but let's avoid brute force [and git's force pushes] wherever we can ;)

https://matthew-b-groves.medium.com/matts-tidbits-23-why-to-avoid-force-pushing-691d3abbab48 😅

holta:
Just FYI `git subtree` (which embeds an external repo and its history) and `git submodule` were both considered:

But it sounds like... both are overkill / sub-optimal / etc... certainly for now?! 🤔

chapmanjacobd:
Yes, I feel like the git fork approach is better than subtree/subrepo. 

git submodule is a nice idea in theory but in practice it's not much different than documenting in a script or readme where the other repo exists and what commits/tags we are using. The problem with git submodule is that it adds extra hoops for everyone else who clones the repo and it seems better to just have things loosely linked for now.

We may want to fully connect CI builds into Ansible via the GitHub Releases API, but that also seems like overkill for now. Let's wait until (if/when) we are needing to update/patch this more than a couple times per month at least.

orblivion:
> Yes, I feel like the git fork approach is better than subtree/subrepo.

One thing I actually didn't quite understand, though it's just an informational aside: wouldn't a subtree/submodule/(subrepo?) require a fork anyway?

Seems like it's ready for my new PR so I will close.

chapmanjacobd:
> wouldn't a ... require a fork anyway?

submodule yes. subtree/subrepo no

orblivion:
Oh that's right you're just merging the whole history of that other repo into your repo.

PR #4195, opened 2026-01-02
Rename ansible var: maps_sat_zoom -> maps_satellite_zoom

orblivion:
*UPDATE YOUR `local_vars.yml` accordingly!*

### Fixes bug:

Confusing variable name

Rename ansible var: maps_sat_zoom -> maps_satellite_zoom

Raspi Trixie

@holta

### DIFF
--- a/roles/maps/README.md
+++ b/roles/maps/README.md
@@ -12,21 +12,21 @@ To configure your map, set the following variables (for the option your choose!)
    maps_enabled: True
 
    maps_vector_quality: ne
-   maps_sat_zoom: 7
+   maps_satellite_zoom: 7
    ```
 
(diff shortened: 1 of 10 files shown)


orblivion:
If existing users take this update without updating their local_vars, they will go back to satellite zoom level 7. Not the end of the world but it may be confusing or may slip by. Would it be worth perhaps adding a warning (that I can remove in a couple months) in Ansible that will fail *if `maps_sat_zoom` is defined*?

holta:
> If existing users take this update without updating their local_vars, they will go back to satellite zoom level 7. Not the end of the world but it may be confusing or may slip by. Would it be worth perhaps adding a warning (that I can remove in a couple months) in Ansible that will fail _if `maps_sat_zoom` is defined_?

Not required!

(Later in 2026 we can start to invest in backwards compatibility — if/where that's genuinely needed — but NOT a priority today ;-)

PR #4196, opened 2026-01-02
Build our own maps.black.js (Cleaned up)

orblivion:
This is a cleanup of #4193. In that version I had checked some code in that I don't want to be in the history.

Build maps.black.js from a fork of the maps.black repo, check the built version back into this repo. Use that built version during our Ansible installs instead of downloading from maps.black. The build should be done by the developer (not in Ansible) on the (probably rare) occasion that we need to edit maps.black.js. (If it becomes less rare we can move this to CI).

Note that this built version perfectly matches https://v1.maps.black/maps.black.js, at least as of the time of this commit.

The purpose of this is to facilitate a few upcoming features, including fixing overzooming and regional extracts.

Trixie raspberry pi

@holta @chapmanjacobd


chapmanjacobd (approved):

roles/maps/build-maps.black-assets/build.sh
```diff
@@ -0,0 +1,19 @@
...
+cd $SCRIPT_DIR
+
+set -euo pipefail
+
+curl -fsSL https://deb.nodesource.com/setup_24.x | sudo bash - &&\
+sudo apt install -y nodejs
+
+git clone https://github.com/iiab/maps.black
+cd maps.black
+git checkout 12205728ceb0672682381088c6c1e7ad05fdc4f5
```
jvonau: This could be handed like [calibre-web](https://github.com/iiab/iiab/blob/master/roles/calibre-web/tasks/install.yml#L92-L98) or [sugarizer](https://github.com/iiab/iiab/blob/master/roles/sugarizer/tasks/install.yml#L37-L43) git wise? Might have to search the right ansible argument to use a checkout and the git hash could be in defaults/main.yml

orblivion:
I'm ready to merge, but since it hasn't been yet I snuck in one tweak to the license file.

PR #4202, opened 2026-01-03
Implement overzooming for osm, s2maps, and terrain

orblivion:
### Fixes bug:

If the implementer chooses anything but "full quality" for vector, satellite, or terrain, and the user then zooms past the highest zoom level for one of those layers, the layer will act strangely. For vector it will simply disappear. For satellite it will not disappear but if you zoom in far and reload the page, it will disappear. For terrain, it will *sometimes* act like satellite though sometimes act fine.

My current solution has been to just stop the user from zooming in that far. That logic is kind of annoying, and it doesn't really help us with terrain.

This will allow us to safely "overzoom". Here's an explanation:

The "maxzoom" setting of the different layers (satellite, vector, terrain) tells maplibre what the highest zoom level of the layer's pmtiles file is. maps.black has those values hard-coded according to the files that it provides. If the user zooms in past that, maplibre knows to just keep reusing that last zoom level. This is what we've been calling overzooming.

For IIAB maps, we've been cutting off those high zoom levels, so maps.black's hard-coded values no longer apply. maplibre doesn't know about this so it has the above error behavior if the user zooms in past the last zoom level that we have.

So here, we edit maps.black.js to tell it that the "maxzoom" is a value that we supply, matching the truncated files that we are using. So if I'm using a zoom-0-7 satellite file, I can safely zoom to 8 or 9. And the map won't disappear on page reload, it'll just keep showing 7.

Here is the corresponding change to maps.black.js: https://github.com/iiab/maps.black/commit/c1baf6d07f181ff62f9c272060056640dbec9c7f

And my reasons for this:

* We wanted this feature long ago at some point, particularly since the previous IIAB maps allowed this.
* It makes terrain less buggy (and I can call it no longer "experimental")
* I needed to make maps.black.js editable for regional extracts anyway. This was a pretty simple change once that was set.
* It's simpler and less buggy than my current "stop from zooming" trick.


roles/maps/templates/index.html.j2
```diff
@@ -324,49 +353,6 @@
                 })
                 const scaleControl = new maplibregl.ScaleControl()
                 let setUpMapTimeout = null
-                function setMaxZoom() {
```
orblivion, line 327: So keep in mind that now, users will be able to zoom in "all the way" (I think zoom level 22). This will look pretty bad for medium and low quality layers, but I figure they can decide for themselves.

But, if at some point you think it looks more polished to stop them somewhere, like it could be something like "5 levels after the maximum zoom level among the visible layers", let me know.

holta:
> Smoke-tested on which OS or OS's:

@orblivion What testing does this PR need?

(Or is that already done!)

roles/maps/templates/index.html.j2
```diff
@@ -45,6 +45,35 @@
             let mb = null
```
chapmanjacobd, line 45: `mb` is locally scoped and globally scoped (line 412). Is this assignment needed here? maybe we should remove the let part and prefix with `window.` to be more clear

chapmanjacobd (approved):

roles/maps/templates/index.html.j2
orblivion: Is this locally scoped? It's at the top level of the `<script>`. I don't mind a `window.` if it's more clear in any case. I know this whole thing is a bit of a mess.

But this is unrelated to this PR right? Can I do this in a followup?

orblivion:
I've tried it out but maybe we should get some opinions on how it looks and behaves with some different pmtiles max zoom settings.

orblivion:
(I also think I haven't tried this *specific version* of the code so I should do that at very least one last time.)

roles/maps/templates/index.html.j2
chapmanjacobd: You're right it is globally scoped in both assignments.

I'm just calling it out because it seemed strange. I already approved the PR in my review so it was more of a nitpick anyway

orblivion: Yeah there could be this and a bunch of other things to do in a better way. When it's the right time I hope to clean this up.

Maybe you're right that the `let` doesn't belong here. It's referred to in plenty of lines after the `let` and before it's assigned at the bottom, but those lines don't run until after the assignment. However, I wonder if it would instead be weird then to refer to a variable in all this code that isn't even defined until the bottom of the file.

chapmanjacobd: > This doesn't mean let, etc. are not defined in global, they just don't also exist as properties on the global object. This is possible because global scope has two buckets: an object record (global object and its properties) and a declarative record (declarations not in global object). Definitions in both contribute to those values which are globally available.
>
> https://old.reddit.com/r/learnjavascript/comments/mrbt1q/let_vs_var_global_scope_vs_local_scope/gulbwmk/

I think it doesn't matter that much... but it can be a bit confusing

holta:
> I've tried it out but maybe we should get some opinions on how it looks and behaves with some different pmtiles max zoom settings.

LMK if merging tonight is best?  (Or any other suggestion!)

orblivion:
I think it's worth doing a little QA on this. Here's a test plan:

## Start installing IIAB with this PR

On your testing device, run: `curl iiab.io/install.txt | bash -s 4202`

## Edit `/etc/iiab/local_vars.yml` a few seconds later (when IIAB installer asks you to do that!)

This test is about zoom levels. I have given a few options for different zoom levels to try out for satellite and terrain. Please **make one choice** among those options (I am leaving the choice to you so that you might try something I am not anticipating):

    maps_install: True
    maps_enabled: True
    maps_vector_quality: "osm-z9"
    maps_satellite_zoom: (choose 7, 9, or 11)
    maps_terrain_zoom: (choose 7, 8, or 9)

Leave other map settings at their default.

## Wrap up the install:

`sudo iiab`

## Test:

We want to test that, on "lower quality" data files, the user can zoom in all the way and it still works smoothly. (For this PR I am removing previous restrictions on zooming in on "low quality" maps.).

**Note**:

* If you notice bugs, try to do a "hard reload" the browser page. There may be a browser caching issue.

**Things to try**:

* Zoom in and out as much as you can
* Try out different map "styles" (hybrid, natural, etc)
* Try turning terrain on and off (there's a new "terrain" button under the "globe" button) - You should see terrain when you zoom in on a mountain and then tilt the view (ctrl + mouse drag)
* Try turning globe on and off
* Try reloading the browser page in various locations, zooms, and globe/terrain configurations - Making sure you see the same thing again after reloading.

**Bugs to look for**:

* Missing data
* Other glitches

**Things that are OKAY and EXPECTED**:

* Taking a couple seconds for data to load sometimes
* Satellite data looking pixellated when very zoomed in
* Minor streets and street names missing (since this is not a full quality vector map)

holta:
Thanks @deldesir, @EMG70, @chapmanjacobd for having helped @orblivion push on these difficult UI/UX design questions!!

PR #4206, opened 2026-01-05
Add more retries to ansible install

chapmanjacobd:
### Fixes bug:

https://github.com/iiab/iiab/discussions/4205

Adds more retries and [resume-retries](https://github.com/pypa/pip/pull/12991) to the Ansible install

@Ark74 @holta

### DIFF
--- a/scripts/ansible
+++ b/scripts/ansible
@@ -288,7 +288,7 @@ EOF
     # messes -- and obviating the need for these 2: (above, both commented out)
     # - 'apt -y install libffi-dev python3-dev'
     # - painstaking pinning of cryptography or cffi (etc) to older version #s
-    /usr/local/ansible/bin/python3 -m pip install --prefer-binary --upgrade ansible-core
+    /usr/local/ansible/bin/python3 -m pip install --retries 50 --resume-retries 50 --prefer-binary --upgrade ansible-core
     echo -e "\nCreate symlinks /usr/local/bin/ansible* -> /usr/local/ansible/bin/ansible*"
     cd /usr/local/ansible/bin
     for bin in ansible*; do
--- a/scripts/ansible
+++ b/scripts/ansible
@@ -220,10 +220,11 @@ retry_pip_install() {
     local venv="$1"
     shift
     local pkgs=("$@")
+    local has_resume="no"
 
     for i in $(seq 1 3); do
         # Check for resume-retries support
-        local has_resume=$($venv/bin/python3 -m pip install --help | grep -q "resume-retries" && echo "yes" || echo "no")
+        has_resume=$($venv/bin/python3 -m pip install --help | grep -q "resume-retries" && echo "yes" || echo "no")
 
         # Attempt pip upgrade if flag is missing
         if [ "$has_resume" == "no" ]; then
--- a/scripts/ansible
+++ b/scripts/ansible
@@ -239,7 +239,7 @@ retry_pip_install() {
     done
 
     # Build command
-    local cmd=("$venv/bin/python3" "-m" "pip" "install" "--retries" "30" "--prefer-binary" "--upgrade")
+    local cmd=("$venv/bin/python3" "-m" "pip" "install" "--retries" "10" "--prefer-binary" "--upgrade")
     [[ "$has_resume" == "yes" ]] && cmd+=("--resume-retries" "150")
     cmd+=("${pkgs[@]}")
 
--- a/scripts/ansible
+++ b/scripts/ansible
@@ -221,7 +221,7 @@ retry_pip_install() {
     shift
     local pkgs=("$@")
 
-    for i in $(seq 1 10); do
+    for i in $(seq 1 3); do
         # Check for resume-retries support
         local has_resume=$($venv/bin/python3 -m pip install --help | grep -q "resume-retries" && echo "yes" || echo "no")
 
@@ -233,20 +233,19 @@ retry_pip_install() {
             has_resume=$($venv/bin/python3 -m pip install --help | grep -q "resume-retries" && echo "yes" || echo "no")
         fi
 
-        # Build command
-        local cmd=("$venv/bin/python3" "-m" "pip" "install" "--retries" "3" "--prefer-binary" "--upgrade")
-        [[ "$has_resume" == "yes" ]] && cmd+=("--resume-retries" "30")
-        cmd+=("${pkgs[@]}")
-
-        # Execute
-        if "${cmd[@]}"; then
-            return 0
-        fi
-
         echo "Attempt $i failed. Retrying in 2s..."
         sleep 2
     done
 
+    # Build command
+    local cmd=("$venv/bin/python3" "-m" "pip" "install" "--retries" "30" "--prefer-binary" "--upgrade")
+    [[ "$has_resume" == "yes" ]] && cmd+=("--resume-retries" "150")
+    cmd+=("${pkgs[@]}")
+
+    # Execute
+    if "${cmd[@]}"; then
+        return 0
+    fi
     return 1
 }
 
--- a/scripts/ansible
+++ b/scripts/ansible
@@ -216,6 +216,40 @@ echo -e "\napt update; apt install python3-venv"
 $APT_PATH/apt update
 $APT_PATH/apt -y install python3-venv
 
+retry_pip_install() {
+    local venv="$1"
+    shift
+    local pkgs=("$@")
+
+    for i in $(seq 1 10); do
+        # Check for resume-retries support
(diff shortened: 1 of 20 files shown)


chapmanjacobd:
hmm maybe we need to upgrade pip inside the venv to get full resilience as the tests are failing from `--resume-retries` being a recent feature starting from pip 25.1

chapmanjacobd:
It is kinda funny that `--retries` uses exponential backoff but `--resume-retries` doesn't add any delay between retries... :/ Maybe we should add some sleep-and-retry ourselves

holta:
@chapmanjacobd do you know what this CI failure (`no such option: --resume-retries`) is all about?

```
Usage:   
  /usr/local/ansible/bin/python3 -m pip install [options] <requirement specifier> [package-index-options] ...
  /usr/local/ansible/bin/python3 -m pip install [options] -r <requirements file> [package-index-options] ...
  /usr/local/ansible/bin/python3 -m pip install [options] [-e] <vcs project url> ...
  /usr/local/ansible/bin/python3 -m pip install [options] [-e] <local project path> ...
  /usr/local/ansible/bin/python3 -m pip install [options] <archive url/path> ...

no such option: --resume-retries
Error: Process completed with exit code 2.
```

https://github.com/iiab/iiab/actions/runs/20729224223/job/59512655498?pr=4206

chapmanjacobd:
The version of pip in CI must be older than pip 25.1 (Apr 2025)

holta:
Curious that 1 of the 4 CI workflows did _NOT_ have this failure!

https://github.com/iiab/iiab/actions/runs/20729224224/job/59512655598?pr=4206

chapmanjacobd:
That one has 

python3-venv 3.13.5-1

Compare to a failing one:

python3-venv 3.12.3-0ubuntu2.1

Looks like Python 3.13.5-1 in Debian Trixie/Ubuntu includes pip version 25.1.1+dfsg-1: https://old.reddit.com/r/ps2/comments/1ptkqa0/im_stuck_trying_to_install_psbbn_please_help/

chapmanjacobd:
I explored function reuse but it seems like most of the other places that pip install are called via Ansible and so it makes more sense to do something like this if/where that is needed:

```yaml
- tasks:
    - name: Install a Python package with retries
      pip:
        name: some-python-package
        state: present
      register: pip_result
      retries: 5
      delay: 10
      until: pip_result is not failed
```

jvonau:
https://stackoverflow.com/questions/59796680/disable-pip-install-timeout-for-slow-connections perhaps changing timeout is enough?

chapmanjacobd:
@jvonau that setting is the delay timeout between successful read(). I've modified the large file get_urls to reduce that timeout because the current setting is likely a bit excessive but also I've added retries

holta:
@chapmanjacobd I don't yet understand...

```
    timeout: "{{ download_timeout // 2 }}"
```

If 50 seconds is much better than 100 seconds (what ae the reasons if so!) ...would it not be better to change it "across the board" below... ?

https://github.com/iiab/iiab/blob/26a713ceea7d5e76a22942f31d47af36ebb6c83a/vars/default_vars.yml#L51

jvonau:
> @chapmanjacobd I don't yet understand...
> 
> ```
>     timeout: "{{ download_timeout // 2 }}"
> ```
> 
> If 50 seconds is much better than 100 seconds (what ae the reasons if so!) ...would it not be better to change it "across the board" below... ?
> 
> https://github.com/iiab/iiab/blob/26a713ceea7d5e76a22942f31d47af36ebb6c83a/vars/default_vars.yml#L51

I'd expose 'retries:' 'delay:' to be in default_vars.yml and make the 'timeout:' adjustment there.

scripts/ansible
```diff
@@ -288,7 +322,7 @@ EOF
     # messes -- and obviating the need for these 2: (above, both commented out)
     # - 'apt -y install libffi-dev python3-dev'
     # - painstaking pinning of cryptography or cffi (etc) to older version #s
-    /usr/local/ansible/bin/python3 -m pip install --prefer-binary --upgrade ansible-core
+    retry_pip_install /usr/local/ansible ansible-core
```
jvonau: Why not just upgrade pip across the board and drop the function? This is a virtual environment after all.

holta:
Changing IIAB's `download_timeout` from 100s to 60s — alongside softcoding `download_retries` (10) and `download_retry_delay` (2s) ...as part of https://github.com/iiab/iiab/pull/4206/commits/f08839f9306060d7aefdb0f79ab613c6fc57817a — certainly works for me (and hopefully for others, in countries/regions with lousy ISPs especially :)

(Thanks @chapmanjacobd!)

scripts/ansible
chapmanjacobd: `retry_pip_install` does run `$venv/bin/python3 -m pip install --upgrade pip` so this will attempt to update pip if it doesn't support `--resume-retries` up to ten times. Maybe ten times is a bit excessive. I'll move some things around

holta: @chapmanjacobd explain more your recommendation, if you can?

chapmanjacobd: Well I'm assuming `pip` in pypi isn't as flaky as installing other packages. I've never personally had `pip install --upgrade pip` fail but I have had pypi give me random errors for misc packages, especially less popular ones. Maybe this intuition is misleading and ten retries is reasonable though?

We don't want to DDoS anyone but retries are pretty cheap: trading idle computer time and user head-scratching for more reliability

scripts/ansible
```diff
@@ -216,6 +216,40 @@ echo -e "\napt update; apt install python3-venv"
 $APT_PATH/apt update
 $APT_PATH/apt -y install python3-venv
 
+retry_pip_install() {
```
jvonau: I'd call the function check_pip_version

scripts/ansible
```diff
@@ -216,6 +216,40 @@ echo -e "\napt update; apt install python3-venv"
...
+            echo "Missing --resume-retries. Attempting pip upgrade..."
+            $venv/bin/python3 -m pip install --upgrade pip
+            # Re-check after upgrade attempt
+            has_resume=$($venv/bin/python3 -m pip install --help | grep -q "resume-retries" && echo "yes" || echo "no")
+        fi
+
+        echo "Attempt $i failed. Retrying in 2s..."
+        sleep 2
+    done
+
```
jvonau: And end the function here.

scripts/ansible
jvonau: check_pip_version
/usr/local/ansible/bin/python3 -m pip install --retries 10  --resume-retries 150 --prefer-binary --upgrade ansible-core

jvonau:
I'm starting to like this AI thing... lol
```
AI Overview

The --resume-retries option in pip controls the maximum number of attempts to resume or restart an incomplete package download. This feature is particularly useful for users with poor or flaky internet connections. 
Usage
You can use the --resume-retries option with pip install or pip download commands, specifying the desired number of retries: 
bash

pip install <package_name> --resume-retries <number_of_retries>
# Example:
pip install torch --resume-retries 10

Key Points

    Availability: This is an experimental feature available in pip versions 25.1 and later.
    Default Value: In recent versions of pip, the default value for --resume-retries is 5 attempts.
    Functionality: When a download is interrupted, pip attempts to resume it from where it stopped. If the remote server does not support partial downloads, pip will restart the download entirely. The process repeats until the maximum number of retries is reached or the download completes.
    Environment Variable: The retry limit can also be set using the environment variable PIP_RESUME_RETRIES.
    Difference from --retries: The --resume-retries option is distinct from the older --retries option.
        --resume-retries handles incomplete downloads (e.g., connection termination mid-stream).
        --retries handles initial connection failures (e.g., server not reachab
```

Given --resume-retries is what you want to use and comes with pip 25.1 see the above suggestions about function name and the ansible-core call

jvonau:
Thanks for e45183f looks good to me.

holta:
Thanks @jvonau.  For helping @chapmanjacobd to get this solid.  Ideally with a more solid diagnosis as to what's really happening with #4205 before weekend (if the stars align hopefully, i.e. I'm asking him to give me a call to see if that helps us understand better...)

jvonau:
Side question, where does the documentation for the below come from? Those options are not listed on https://docs.ansible.com/projects/ansible/latest/collections/ansible/builtin/get_url_module.html

register:
until:
retries:
delay:

chapmanjacobd:
They are at the task level not under `get_url`:

https://docs.ansible.com/projects/ansible/latest/reference_appendices/playbooks_keywords.html#task

chapmanjacobd:
closing this approach in favor of #4210 which seems more flexible

PR #4207, opened 2026-01-06
[proot] add fallback install.txt url option, and remove PR old stuff.

Ark74:
### Fixes bug:
iiab.io not being reachable, adds retry options.

### Description of changes proposed in this pull request:
Adds fallback to iiab-factory install.txt file.
Removes discovery PR 4122 logic, as it's no longer necessary as it's now merged.

### Smoke-tested on which OS or OS's:
Debian (proot)

### Mention a team member @username e.g. to help with code review:
@holta

PR #4208, opened 2026-01-06
[proot] update roles/proot_services & README, pointing to new repo iiab-android (was: overhaul into a termux-setup)

Ark74:
### Fixes bug:
Creates a robust termux setup suite, capable to integrate with Termux-API in order to interact with Android 12-16 current settings for later setup IIAB on Android on a simple interaction.

Replace initial 0_termux-setup.sh with a termux-setup (v2) suite.
This suite uses termux and termux-api apps both from F-Droid, unifying on a single environment and simplifying the setup process. This integration deprecates the need of Shizuku to a single setup suite running on Termux.

All the modules bundle themselves into a single `0_termux-setup.sh` file to ease parsing when pulling from Android devices.

Help output:
```
Usage:
  ./0_termux-setup.sh
    -> Termux baseline + Debian bootstrap (idempotent). No ADB prompts.

  ./0_termux-setup.sh --with-adb
    -> Termux baseline + Debian bootstrap + ADB pair/connect if needed (skips if already connected).

  ./0_termux-setup.sh  --adb-only [--connect-port PORT]
    -> Only ADB pair/connect if needed (no Debian; skips if already connected).
       Tip: --connect-port skips the CONNECT PORT prompt (you’ll still be asked for PAIR PORT + PAIR CODE).

  ./0_termux-setup.sh --connect-only [CONNECT_PORT]
    -> Connect-only (no pairing). Use this after the device was already paired before.

  ./0_termux-setup.sh --ppk-only
    -> Set PPK only: max_phantom_processes=256 (requires ADB already connected).
       Android 14-16 usually achieve this via "Disable child process restrictions" in Developer Options.

  ./0_termux-setup.sh --check
    -> Check readiness: developer options flag (if readable),
       (Android 14+) "Disable child process restrictions" proxy flag, and (Android 12-13) PPK effective value.

  ./0_termux-setup.sh --all
    -> baseline + Debian + ADB pair/connect if needed + (Android 12-13 only) apply --ppk + run --check.

  Optional:
    --connect-port 41313    (5 digits) Skip CONNECT PORT prompt used with --adb-only
    --timeout 180           Seconds to wait per prompt
    --reset-debian          Reset (reinstall) Debian in proot-distro
    --no-log                Disable logging
    --log-file /path/file   Write logs to a specific file
    --debug                 Extra logs

Notes:
- ADB prompts require: `pkg install termux-api` + Termux:API app installed + notification permission.
- Wireless debugging must be enabled.
- This script never uses adb root.
```

This integration will need to rewrite current README.

#### Test this out,

1) Get an Android Phone
2) Install F-Droid
3) Install
    - Termux
    - Termux-API
4) Enable Developer Mode
5) Open Termux
6) Download the latest script: 
```
URL=https://raw.githubusercontent.com/Ark74/iiab/refs/heads/add_termux-setup_suitev2/roles/proot_services/0_termux-setup.sh
curl -Lo 0_termux-setup_v2.sh $URL
```
7) Run  ` bash 0_termux-setup.sh`  (--all , --check, etc.)
8) Follow the process to setup ADB, enlarge PPK value (A12-13) or disable child process (A14-16).
9) You are all set to continue with #4207 
```
IIAB_ANDROID_INSTALL="https://raw.githubusercontent.com/iiab/iiab/refs/heads/master/roles/proot_services/1_iiab-on-android.sh"
curl $IIAB_ANDROID_INSTALL  | bash
```

Debian (proot)

@holta @chapmanjacobd


holta:
@chapmanjacobd would you be able to review this, e.g. today or tomorrow if you can?

chapmanjacobd:
It's still marked as draft but looks amazing so far!! LMK when it is ready to review

Ark74:
I think is 99% complete, I did some heavy testing, and fits all the goals I've comme up while doing so, now I would appreciate some fresh eyes to see if I'm off by far or actually it makes sense as much as it makes sense to me, the idea would be to be as intuitive as possible and the script should provide as much info to know what to do.

Hopefully not spamming info everywhere.
Once it gets to 99.9% ready then something will come up and be 98% ready xD, that's the only reason is in draft.

It can be tested so far as if it were kind of production, while I'm writing, I'm thinking of a nice to have / missing piece :P
So, maybe after the next commit, it would be easier to test.

chapmanjacobd:
Also any tips for running in an Android emulator would be helpful :) 

or is that not possible; you mostly use `scrcpy`?

Ark74:
> Also any tips for running in an Android emulator would be helpful :)

Emulators work, but I have not found one that works and don't ask for $200USD/month.
So yeah, I mostly use real hardware, I pretty much prefer `kdeconnect` over `scrcpy`, it's much more simple and gives better integration, mainly the **clipboard**.

Ark74:
Ok, so a simple curl and variable can simplify things already. So let's set it for review.

holta:
@chapmanjacobd if you have time to review this, Fantastic!

roles/proot_services/termux-setup/50_mod_adb.sh
```diff
@@ -0,0 +1,272 @@
...
+# -------------------------
+
+# Local stamp so we can detect "connect-only" misuse after reinstall/clear-data.
+ADB_PAIRED_STAMP="${ADB_STATE_DIR}/stamp.adb_paired"
+
+adb_hostkey_fingerprint() {
+  # Returns a stable fingerprint for THIS Termux install's adb host key.
+  local pub="${HOME}/.android/adbkey.pub"
+  [[ -r "$pub" ]] || return 1
+  if have sha256sum; then
```
chapmanjacobd: If the user only has md5sum before then this will be different when checking later... maybe better to just do `pkg install` beforehand and rip out some of these fallbacks to have more consistent behavior

chapmanjacobd: I think `sha256sum` should always be included in Termux (it’s in coreutils and busybox)

roles/proot_services/termux-setup/99_main.sh
```diff
@@ -0,0 +1,358 @@
...
+# Self-check
+# -------------------------
+self_check() {
+  log "Self-check summary:"
+  log " Android release=${ANDROID_REL:-?} sdk=${ANDROID_SDK:-?}"
+
+  if have proot-distro; then
+    log " proot-distro: present"
+    log " proot-distro list:"
+    proot-distro list 2>/dev/null | sed 's/^/ /' || true
```
chapmanjacobd: > sed 's/^/ /'

nitpick: this formatting stuff should probably go into the thing that's calling self_check()

roles/proot_services/termux-setup/README.md
```diff
@@ -0,0 +1,27 @@
+# termux-setup modules
+
+This project is maintained as multiple Bash "modules" that are bundled into a single script:
+`0_termux-setup_v2.sh` (the file served for: `curl ... | bash`).
```
chapmanjacobd: why `_v2`? seems like clutter that could be removed. If the history is not in git, v1 could be an earlier commit that is replaced by this file

roles/proot_services/termux-setup/README.md
```diff
@@ -0,0 +1,27 @@
...
+
+## Rules
```
chapmanjacobd: These dev notes are really useful but it seems like it might be helpful to have a bit more for users at the very top, even just a link to the place they should look for the actual install instructions (in case they accidentally end up here somehow) and then make it clear that the rest of the file are notes for developers

roles/proot_services/termux-setup/10_mod_logging.sh
```diff
@@ -0,0 +1,76 @@
...
+  # If logging is disabled, still allow --debug to trace to console.
+  if [[ "${LOG_ENABLED:-1}" -ne 1 ]]; then
+    if [[ "${DEBUG:-0}" -eq 1 ]]; then
+      set -x
+      ok "Debug trace enabled (bash -x) -> console (logging disabled)"
+    fi
+    return 0
+  fi
+
+  mkdir -p "$LOG_DIR" 2>/dev/null || true
```
chapmanjacobd: I don't think this `2>/dev/null || true` is wise. Makes sense on `rm` but for `mkdir` and `chmod` it will be helpful to know if there are any errors. `mkdir -p` rarely prints an error. 

`chmod --quiet` might be fair game but those tradeoffs should be carefully evaluated

roles/proot_services/termux-setup/50_mod_adb.sh
```diff
@@ -0,0 +1,272 @@
...
+    [[ -n "${serial:-}" ]] || continue
+    [[ "$serial" == ${HOST}:* ]] || continue
+    [[ "$state" == "offline" ]] || continue
+    [[ "$serial" == "$keep_serial" ]] && continue
+    adb disconnect "$serial" >/dev/null 2>&1 || true
+  done < <(adb devices 2>/dev/null | tail -n +2 | sed '/^[[:space:]]*$/d')
+}
+
+adb_pair_connect() {
+  need adb || die "Missing adb. Install: pkg install android-tools"
```
chapmanjacobd: maybe we could just install it for the user?

roles/proot_services/termux-setup/20_mod_termux_base.sh
```diff
@@ -0,0 +1,115 @@
+# shellcheck shell=bash
+# Module file (no shebang). Bundled by build_bundle.sh
+
+# Termux apt options (avoid conffile prompts)
+TERMUX_APT_OPTS=( "-y" "-o" "Dpkg::Options::=--force-confdef" "-o" "Dpkg::Options::=--force-confold" )
```
chapmanjacobd: ahh I recognize this magic code! Very good.

But it also makes me wonder... maybe we are not covering every Termux specific edge-case. The source code for the [pkg](https://github.com/termux/termux-tools/blob/master/scripts/pkg.in) command is strange and, in some places, bizarre but it is worth a peek

Do we need to run [termux-setup-package-manager](https://github.com/termux/termux-tools/blob/master/scripts/termux-setup-package-manager.in)? I guess it is not doing much actually.

This is probably fine AS-IS

roles/proot_services/termux-setup/10_mod_logging.sh
```diff
@@ -0,0 +1,76 @@
...
+rotate_logs() {
+  [[ -d "${LOG_DIR:-}" ]] || return 0
+  local n=$((LOG_KEEP + 1))
+  while IFS= read -r f; do
+    [[ -n "${f:-}" ]] || continue
+    [[ -n "${LOG_FILE:-}" && "$f" == "$LOG_FILE" ]] && continue
+    rm -f -- "$f" 2>/dev/null || true
+  done < <(ls -1t "$LOG_DIR"/*.log 2>/dev/null | tail -n +"$n" || true)
+}
+
```
chapmanjacobd: I'm worried this might do more harm than good... do we really need log rotation? maybe just rename old logs and keep 2 or 3 most recent versions

It's probably fine but it gives me a slight pause

roles/proot_services/termux-setup/20_mod_termux_base.sh
```diff
@@ -0,0 +1,115 @@
...
+    if ! read -r ans < /dev/tty; then
+      warn "No interactive TTY available; skipping mirror selection (run script directly to be prompted)."
+      return 0
+    fi
+    ans="${ans:-Y}"
+    if [[ "$ans" =~ ^[Yy]$ ]]; then
+      # Logging redirects stdout/stderr to pipes, which can break the UI.
+      # Run it using /dev/tty and the original console fds (3/4).
+      if [[ -r /dev/tty ]]; then
+        termux-change-repo </dev/tty >&3 2>&4 || true
```
chapmanjacobd: has this been tested? I'm curious how the TUI logs look after ANSI is stripped. Just an artistic curiosity. But also this doesn't break the TUI?? This is pretty neat.

But if it doesn't work it's not the end of the world to skip logging this part... somehow...

roles/proot_services/termux-setup/20_mod_termux_base.sh
```diff
@@ -0,0 +1,115 @@
...
+    warn "termux-change-repo not found; skipping mirror selection."
+    return 0
+  fi
+
+  if [[ -r /dev/tty ]]; then
+    printf "\n${YEL}[iiab] One-time setup:${RST} Select a nearby Termux repository mirror for faster downloads.\n" >&2
+    local ans="Y"
+    printf "[iiab] Launch termux-change-repo now? [Y/n]: " > /dev/tty
+    if ! read -r ans < /dev/tty; then
+      warn "No interactive TTY available; skipping mirror selection (run script directly to be prompted)."
```
chapmanjacobd: we should probably say which script to run. ie. `termux-change-repo`

roles/proot_services/termux-setup/10_mod_logging.sh
```diff
@@ -0,0 +1,76 @@
...
+  exec \
+    > >(tee >(sed -E 's/\x1B\[[0-9;]*[ -/]*[@-~]//g' >>"$LOG_FILE")) \
+    2> >(tee >(sed -E 's/\x1B\[[0-9;]*[ -/]*[@-~]//g' >>"$LOG_FILE") >&2)
+
+  ok "Logging to: $LOG_FILE"
+
+  # If --debug, send xtrace only to log
+  if [[ "${DEBUG:-0}" -eq 1 ]]; then
+    exec 9>>"$LOG_FILE"
+    export BASH_XTRACEFD=9
```
chapmanjacobd: I did not know that this existed! Very interesting

roles/proot_services/termux-setup/99_main.sh
```diff
@@ -0,0 +1,358 @@
...
+      step_debian_bootstrap_default
+      adb_pair_connect_if_needed
+
+      # Android 12-13 only (SDK 31-33): apply PPK tuning automatically
+      if [[ "${ANDROID_SDK:-}" =~ ^[0-9]+$ ]] && (( ANDROID_SDK >= 31 && ANDROID_SDK <= 33 )); then
+        log "Android SDK=${ANDROID_SDK} detected -> applying --ppk automatically (12-13 rule)."
+        ppk_fix_via_adb || true
+      else
+        log "Android SDK=${ANDROID_SDK:-?} -> skipping auto-PPK (only for Android 12-13)."
+      fi
```
chapmanjacobd: It seems like this should maybe run as soon as possible; before the heavy stuff

roles/proot_services/termux-setup/20_mod_termux_base.sh
```diff
@@ -0,0 +1,115 @@
...
+    curl \
+    coreutils \
+    grep \
+    sed \
+    gawk \
+    openssh \
+    proot proot-distro \
+    android-tools \
+    termux-api \
+    || true
```
chapmanjacobd: What sort of exit code errors are you expecting here? it seems like it might be better to know when something fails here rather than later--especially when providing tech support to other people when the error message might be less clear or have less information as to the root cause

chapmanjacobd (approved):

chapmanjacobd:
Things look pretty good. My comments are mostly nitpicks. I have some good advice but nothing is _blocking_ from this being merged. Things can be improved later

roles/proot_services/termux-setup/README.md
Ark74: There is already a script `0_termux-setup.sh` this one is initially based on it, but with massive changes, so I don't know if we want to overwrite the initial one. If so, great. That's why I mention naming is optional.

chapmanjacobd: Are there any cases where it is more useful than this new version? If not, I would say keeping around the older file can be a bit confusing. If it is useful, then I would rename the existing `0_termux-setup.sh` to something that describes the situation in which it is useful

roles/proot_services/termux-setup/README.md
holta: >might be helpful to have a bit more for users at the very top, even just a link to the place they should look for the actual install instructions (in case they accidentally end up here somehow)

@Ark74 is this reasonable?

Ark74: Ok, makes sense.

roles/proot_services/termux-setup/50_mod_adb.sh
Ark74: If you run the installer that's what it does.
But if you only run tools like --check, --adb-only should we have install instructions spread on every tool?

I was thinking more like "I'm just looking thanks, not buying"

roles/proot_services/termux-setup/20_mod_termux_base.sh
Ark74: Yeah I'm borrowing that :) apt makes more sense to me. I promise I'll start blending more and more into the termux ecosystem, I'm sure there are more than one or two things to learn from there.

roles/proot_services/termux-setup/10_mod_logging.sh
Ark74: Fair, I'll work on simplify that

roles/proot_services/termux-setup/50_mod_adb.sh
chapmanjacobd: ah I glossed over that. No need to change this

deldesir (commented):
Would it be useful to add the following to README?

Ensure you install android-tools (`pkg install android-tools`)

When you switch apps to answer the Termux notification, Android closes the pairing dialog in Settings, which changes the pairing code and port immediately. This makes the code you entered invalid by the time the script tries to use it. So use Split Screen if your devices support it.

You must keep the "Pair with device" Settings screen open and visible while interacting with Termux.

Open Settings -> Wireless Debugging -> Pair device with pairing code.
Enter Split Screen mode (usually by tapping the app icon in the Recent Apps view and selecting "Split top" or similar).
Open Termux in the other half of the screen.
Run the script again.
When the notification asks for the code, answer it without leaving the split screen view.
This ensures the pairing session stays active.

holta:
@Ark74 do you agree with @deldesir's suggestion [above](https://github.com/iiab/iiab/pull/4208#pullrequestreview-3637544093) ?

Ark74:
@deldesir the current behavior documented on the published README and this PR are different things.

This PR tries to become a much more robust "suite" that to actually test it, you need to follow the steps on the description (not the ones on the current README).
You are very well invited to help us test it, as it seems to be the the path to go in the near future :+1:

About android-tools is [here](https://github.com/iiab/iiab/pull/4208/files#diff-74327cee4af12ead3df7ba79f692bed7968ace8c881ea091a7199995227ef199R103)

holta:
> This PR tries to become a much more robust "suite" 

> to actually test it, you need to follow the steps on the description (not the ones on the current README).

@deldesir

I think @Ark74 is referring to the 9 Steps under "Test this out," at the top of this ticket here:

https://github.com/iiab/iiab/pull/4209#issuecomment-3722845283

roles/proot_services/termux-setup/99_main.sh
Ark74: uff, I got this one late. I'll check this one a bit later on. It shouldn't affect the behavior in case of testing.

roles/proot_services/termux-setup/README.md
Ark74: Renamed! ;)

roles/proot_services/termux-setup/README.md
Ark74: More on this one a bit later :)

roles/proot_services/termux-setup/50_mod_adb.sh
Ark74: It's available by default and also checks to make sure.

roles/proot_services/termux-setup/99_main.sh
Ark74: Now it is a function at 00_lib_common.sh

roles/proot_services/termux-setup/10_mod_logging.sh
Ark74: removed!

roles/proot_services/termux-setup/20_mod_termux_base.sh
Ark74: Yeah, everything has been tested at least once before published. It's working both in logs and TUI ok. It's not infallible under some very specific conditions it may break.

roles/proot_services/termux-setup/20_mod_termux_base.sh
Ark74: Completely, done ;)

roles/proot_services/termux-setup/20_mod_termux_base.sh
Ark74: There are now warnings pointing to something failing, and enforce meeting the requirements to continue. Hopefully that's not overkill.

Ark74:
@chapmanjacobd thank you for your help reviewing, I really appreciate the time you put into it :)

These last changes restore the original name, so it will overwrite the current script.
```
0_termux-setup.sh
```

Luckily y you'll find the timestamp on the header useful to identify which one you are running.

```
#!/data/data/com.termux/files/usr/bin/bash
set -euo pipefail

# -----------------------------------------------------------------------------
# GENERATED FILE: 2026-01-08T13:41:54-06:00 - do not edit directly.
# Source modules: termux-setup/*.sh + manifest.sh
# Rebuild: (cd termux-setup && bash build_bundle.sh)
# -----------------------------------------------------------------------------
```

You are invited to test this PR on actual phones, as it is at least as mature of what we have published, and with improved support and simpler workflow (F-Droid/Termux-API).
Help us test it, and let us know if you find it useful and where can we improve it.

I'll try to make some companion visuals, videos to demo the working features, but that will take a bit more of time.

Regards o/

chapmanjacobd (approved):
yeah the latest changes look great

> LOG_KEEP=20          # keep last N logs

sorry I made you rewrite that a bit--I wasn't fully awake when reading--but I agree the setting of 3 is probably better than 20

> mapfile -t miss < <(baseline_missing_prereqs || true)

my eyes start melting at BASH 4.0 code but I trust that this works. converting the lines into a file and then reading into an array

maybe it could be simplified to something like this

```bash
baseline_missing_prereqs() {
  for b in proot-distro adb termux-notification termux-dialog; do
    have "$b" || echo "$b"
  done
  have sha256sum || echo "sha256sum (coreutils)"
}
```

but your mapfile _looks_ faster

> sh -s <<'EOF'

great choice

> device_config

where does this program come from? AOSP? I don't see it when I search `apt-file`. Sorry I didn't catch this earlier

everything else looks good

deldesir:
> > This PR tries to become a much more robust "suite"
> 
> > to actually test it, you need to follow the steps on the description (not the ones on the current README).
> 
> @deldesir
> 
> I think @Ark74 is referring to the 9 Steps under "Test this out," at the top of this ticket here:
> 
> [#4209 (comment)](https://github.com/iiab/iiab/pull/4209#issuecomment-3722845283)

I am confused. I wanted to point ou I had to use split screen on my device to provide ports and pairing code for the adb process, as going back and forth changed the pairing code.

Ark74:
> I am confused. I wanted to point ou I had to use split screen on my device to provide ports and pairing code for the adb process, as going back and forth changed the pairing code.

Oh, ok. I don't think we need that with this PR, let me record a quick video to share and we can check.

https://ark.switnet.org/vid/A12_termux-setup_demo_hb.mp4

Ark74:
> my eyes start melting at BASH 4.0 code but I trust that this works. converting the lines into a file and then reading into an array
> 
> maybe it could be simplified to something like this
> 
> ```shell
> baseline_missing_prereqs() {
>   for b in proot-distro adb termux-notification termux-dialog; do
>     have "$b" || echo "$b"
>   done
>   have sha256sum || echo "sha256sum (coreutils)"
> }
> ```

Sure, let's keep readiness first when possible.

> where does this program come from? AOSP? I don't see it when I search `apt-file`. Sorry I didn't catch this earlier

Indeed, it's an adb command for most Android versions ~10 and later from what I can tell comes by default,
```
~ $ adb shell command -v device_config
/system/bin/device_config
```

Ark74:
@deldesir any comments on the video?

deldesir:
> @deldesir any comments on the video?

Thanks @Ark74. This was exactly what I did, but if you were using my smarphone, it would fail at 8:39. The simple action of switching window as you did in 8:39 would change both port and pairing code. Looks like the issue is similar to the one reported [here](https://github.com/tytydraco/LADB/issues/46#issuecomment-981985112)

Ark74:
@deldesir that's, interesting :thinking: 

could you retry and share the log? maybe a small screen video just to get an idea on it.
What brand is you phone? Android version? Is it a custom rom or default the vendor shipped with?

PS. Thank you very much!

deldesir:
Ok, I am doing it. Will update you once done

chapmanjacobd:
> FOR SAMSUNG ONE UI CORE SERIES
> Do not change the process limit, because setting phantom process to false is already good enough

> And for some reason, setting/changing device config somehow makes phantom process stuck at enable only and adb shell acts goofy

> https://old.reddit.com/r/termux/comments/15l8cop/hey_i_hope_everyone_is_doing_well_i_just/jva7eo7/

Also I just noticed we're only using get and not put?

    adb shell settings put global settings_enable_monitor_phantom_procs false

roles/proot_services/termux-setup/README.md
Ark74: Thinking about pointing out to the right place to look for information, maybe for further PRs, we can add a "android" folder and remove from proot_services what is not part of the role:

Today:
```
iiab/roles/proot_service/
├── defaults
├── files
│   └── pdsm
│       ├── etc_profile.d
│       ├── usr_local_bin
│       └── usr_local_pdsm
│           ├── lib
│           └── services-available
├── tasks
└── termux-setup
    └── dist

12 directories

```

in some near future:

```
iiab/roles/proot_services/
├── defaults
├── files
│   └── pdsm
│       ├── etc_profile.d
│       ├── usr_local_bin
│       └── usr_local_pdsm
│           ├── lib
│           └── services-available
└── tasks

10 directories

iiab/android/
└── termux-setup
└── docs
    └── etc

3 directories

```

So we can aim to have specific android related information.
@holta @chapmanjacobd

Ark74: Update: 0c2d4a9b67e6f250f70ec78bcbb8e7c86404aad0

chapmanjacobd: Initially I was against this idea but I like how you've separated it out.

We may even want it to have two separate repos: one iiab-android (for the tooling like your iiab/android folder) and one for pdsm (which could support IIAB and non-IIAB use-cases). The pdsm role within IIAB would mostly just pull/install from a release in that other repo

How do you feel about that Luis? 

It may also be the case that these things don't cleanly separate into different repos in which case I think you're proposal is probably the way to go for the time being and we can evaluate this again when adding support for HarmonyOS, etc (if that ever happens)

roles/proot_services/termux-setup/99_main.sh
Ark74: Fixed

roles/proot_services/termux-setup/README.md
Ark74: > How do you feel about that Luis?

Oh!, yeah that could be a good way to split them and let them specialize in their main spaces.
Letting proot_services live in current IIAB.

Ark74: @holta what do you think from this proposal of splitting into iiab-android maybe? ^

Ark74:
> Also I just noticed we're only using get and not put?
> 
> ```
> adb shell settings put global settings_enable_monitor_phantom_procs false
> ```

For A14-16 if adb is available we only check / verify if monitor_phantom is set as needed to confirm "OK", if we can't we only give proper advice as in https://github.com/iiab/iiab/pull/4208/commits/861d9dfa02fb552d26eb3e68caecc2e6afd0fced

The main idea so far is to simplify setup when possible, for Android 14-16 adb is not strictly necessary, the pairing / connecting complicates setup a bit, so adb might be desired but  not actually a requirement. A12-13 is a complete different story, adb is a requirement to being able to complete the installation.

Now, I suppose that if adb is connected on A14-16, we might like not only to check but to set it up too if required :thinking: 
Maybe further down the road we would have reasons to ask for adb ready on 14-16, but as of right now I think that passing 2 ports + 1 code to avoid 1 tap/click may be overkill.

Would it be ok to wait for some feedback before steering the wheel? What do you think?
@chapmanjacobd @holta @deldesir

Ark74:
Other than splitting Android related stuff into a folder or a dedicated repo, I think we've gone over all the details.
Maybe getting closer to actually merge?

roles/proot_services/termux-setup/README.md
holta: Sounds reasonable but I'm not yet fully grasping the big picture:

Are you and @chapmanjacobd in favor of a new repo called something like `iiab-android` ?

holta:
@deldesir @chapmanjacobd @Ark74 do you think this should be merged very soon, e.g. today?

chapmanjacobd:
> wait for some feedback before steering the wheel

That seems very reasonable

> skipping adb for A14-16

I thought device_config only works inside adb? So we're no longer doing that? ok

I think this PR is ready to merge

holta:
Great that this PR is in a good place. ✅

➡️ However before merging this PR, let's stabilize overall Ansible operations first (facts gathering especially) — using the following PR or similar...

- PR #4219

roles/proot_services/termux-setup/README.md
Ark74: Thank you all, this is a reality now:
https://github.com/iiab/iiab-android

Ark74:
@holta this PR is now just a README update.

holta:
@Ark74 can you help everyone better understand the overall purpose[*] of new repo https://github.com/iiab/iiab-android — helping me/us write this up in proot_services/README.md ?

[*] Current or evolving purpose, Thanks!!

Ark74:
Is **IIAB on Android** a role?

The work to expand the IIAB installer to be _proot-distro_ aware has been done on different fronts.

`proot_services` is exactly what the name implies: small scripts to manage services so apps can work without systemd, due to proot-distro limitations. It's an OS facing component and gets to become one more custom IIAB role.

We had been using `proot_services` to document the whole "IIAB on Android" effort, but I suggested that’s not what we want. `proot_services` isn't really the right place for that, because "IIAB on Android" is broader than the role itself.

IIAB on Android includes:
- the `0_termux-setup` _"suite"_ (current Android specific setup in there).
- the `1_iiab_on_android` wrapper (pre-populating `local_vars_android.yml` and launching the actual IIAB installer)
- the layer of changes made across existing IIAB roles (via PR #4122), i.e., core IIAB,
- and of course the proot-distro service manager (pdsm / `proot_services`).

I think `proot_services` should remain an IIAB role and continue to evolve along with the broader IIAB on Android effort, but stay focused on what it specifically does.

Meanwhile, the "IIAB on Android" concept is more like a _"mode"_ that IIAB can run in,  and it isn’t confined to a single role.
So everything that's not `proot_service` should have its own place to actually point the users and developers.

TLDR;
> IMO, IIAB on Android is not a role, but a "mode" that current IIAB can run in, proving it's amazing portability.

As for the new repo: iiab-android it will gives us room to grow into more specialized development for Android related matters. The project is evolving quickly, and splitting it out helps spread the load and focus on Android without disrupting IIAB's current nature.

holta:
@Ark74 thanks for your above writeup !

➡️ Let's work on summarizing key points and intuitions — to appear prominently e.g. maybe within https://github.com/iiab/iiab-android/blob/main/README.md and/or in GitHub's description of the [iiab/iiab-android](https://github.com/iiab/iiab-android) repo itself? 🗺️

➡️ Please help us keep prominent links between all 3 READMEs, helping new people who want to understand quickly how this all works: 🏒

1. https://github.com/iiab/iiab-android/blob/main/README.md
2. https://github.com/iiab/iiab-android/blob/main/termux-setup/README.md
3. https://github.com/iiab/iiab/blob/master/roles/proot_services/README.md

tim-moody:
Please reread #4220

PR #4210, opened 2026-01-08
flaky net: use cURL instead

chapmanjacobd:
### Fixes bug:

https://github.com/iiab/iiab/discussions/4205

This replaces all the trivial downloads that use get_url to get_curl

CI

@holta


holta:
@Ark74 would you have time to review this?

jvonau:
Just spitballing but given get_curl is replacing get_url and /usr/share/ansible/plugins/modules does not exist on my install would it be better to have get_curl.py live in the same directory as get_url.py in /usr/local/ansible/lib/python3.13/site-packages/ansible/modules/ and omit the change to ansible.cfg? IE copy get_curl.py to /usr/local/ansible/lib/python3.13/site-packages/ansible/modules/ after the creation of the virtual environment.

chapmanjacobd:
It's definitely reasonable that we fine-tune the location. I'll remove the `/usr/share/ansible/plugins/modules` as that is obviously not needed. If we have to move `./library` somewhere I guess `/usr/local/ansible/library` would be a good spot 

But if we can keep it relative to `ansible.cfg` I feel like that would be better because it ensures more conformity with checking out iiab git branches like `iiab -s N` etc. If you don't like the name library we could have it be something else like "custom", "modules", "actions", or "ansible", or something like that

I'll rename to modules as I feel like that is more descriptive but feel free to nudge again

chapmanjacobd:
@Ark74 while evaluating this you can add `"-w" ,"%{num_retries} retries\n",` to modules/get_curl.py:42 cmd[] to see number of retries printed out

modules/get_curl.py
```diff
@@ -0,0 +1,68 @@
...
+        force=dict(type='bool', default=True),
+        timeout=dict(type='int', default=60),
+        retries=dict(type='int', default=10),
+    )
+
+    module = AnsibleModule(argument_spec=argument_spec, supports_check_mode=True)
+    args = module.params
+
+    # Handle directory-to-filename logic similar to get_url
+    # We rely on curl -J -O to determine the filename in the dest dir case
```
orblivion: The man page shows using -J and -O together as well but they seem to contradict each other. -J uses the Content-Disposition header (which seems bad for us since we have less control when downloading from 3rd party sources) but -O sounds like it uses the filename from the URL. Could you clarify?

roles/nodered/tasks/settings.yml
```diff
@@ -42,22 +42,18 @@
 # 2022-10-13: These 9 lines (OLD WAY) might be removable by ~2024?  PR #3402
 - name: "OLD WAY: Copy /usr/lib/node_modules/node-red/settings.js to /home/{{ nodered_linux_user }}/.node-red/settings.js"
   copy:
-    remote_src: yes
```
orblivion, line 45: Could you explain this change? We want to copy `settings.js` from the controller's npm?

roles/phpmyadmin/tasks/install.yml
```diff
@@ -4,21 +4,15 @@
 
 
 - name: Download {{ phpmyadmin_dl_url }} to {{ downloads_dir }}
-  get_url:
+  get_curl:
     url: "{{ phpmyadmin_dl_url }}"    # e.g. https://files.phpmyadmin.net/phpMyAdmin/5.0.4/phpMyAdmin-5.0.4-all-languages.zip
-    dest: "{{ downloads_dir }}"    # /opt/iiab/downloads
+    dest: "{{ downloads_dir }}/phpmyadmin.zip"    # /opt/iiab/downloads
```
orblivion: The comment implies it's a zip but `phpmyadmin_dl_url` is actually a `tar.xz`

modules/get_curl.py
```diff
@@ -0,0 +1,68 @@
...
+
+    try:
+        process = subprocess.Popen(cmd, stdout=subprocess.PIPE, stderr=subprocess.PIPE, text=True)
+        stdout, stderr = process.communicate()
+
+        if process.returncode != 0:
+            module.fail_json(msg="curl failed", rc=process.returncode, stdout=stdout, stderr=stderr)
+
+        module.exit_json(changed=True, dest=args['dest'], rc=0)
+    except Exception as e:
```
orblivion: Do you want to delete the dest file here if it exists? What if there's some sort of error and an empty or unfinished file is left sticking around? The next run you'll say `changed=False` and assume the file is done downloading.

modules/get_curl.py
chapmanjacobd: Yep. curl will actually error out if those are combined `curl: --continue-at and --remote-header-name cannot be combined` which I initially unintentionally did. I'll remove the misleading comment

roles/nodered/tasks/settings.yml
chapmanjacobd: remote_src doesn't make sense here since we are using Ansible with localhost. Here's an example for how to do it in a configuration where you are running Ansible on remote machines:

```
- name: Copy a "sudoers" file on the remote machine for editing
  ansible.builtin.copy:
    src: /etc/sudoers
    dest: /etc/sudoers.edit
    remote_src: yes
    validate: /usr/sbin/visudo -csf %s
```

It doesn't make sense to use in our use cases where the target machine's filesystem is the same as the source. It's all local src.

roles/phpmyadmin/tasks/install.yml
chapmanjacobd: good catch. I'll fix and update the misleading comment

modules/get_curl.py
chapmanjacobd: actually you are thinking of `-o`. `-O` is different. I'll change to use long names so this is more clear

modules/get_curl.py
chapmanjacobd: Interesting question and maybe one we'll have to come back to and change later but for now I think it's not needed.

With `fail_json()`, Ansible assumes the task has failed. By definition, a failed task does not successfully complete a "change" to the system state. So either the task will fail or it will succeed and changed=True.

If `dest` is a file then curl will only exit zero if is able to use Range requests to finish downloading the file or if Range requests are not possible it will start over from byte 0 overwriting the file.

If `dest` is a folder then it will overwrite the existing file that it comes up with (determined by URL path basename and Content-Disposition) (unfortunately curl doesn't support combining --continue-at and --remote-header-name)

So in both cases we will get similar or better behavior by not deleting the file prematurely. And it helps with debugging to keep things on disk when it fails.

orblivion: > Interesting question and maybe one we'll have to come back to and change later but for now I think it's not needed.

Well now that you deleted the `changed=False`, this is no longer an issue.

orblivion: (Assuming curl can figure out the status of the download, as you say)

orblivion: Though now `force: False` doesn't do anything.

But that aside, why is `force: True` the default?

chapmanjacobd: ah! I deleted that earlier so I didn't quite understand what you were talking about when I hit Ctrl-F. But I'm glad we arrived at the same conclusion and that you pointed it out

chapmanjacobd: > why is force: True the default

because that is the behavior that makes the most sense for something like Ansible--I removed the force option stub. `get_url` behaves similarly as it will always overwrite files when the dest is a folder

We should use task level options like `when:` or `args: creates:` if we ever want to skip the task when the file already exists as that is a bit more clear

orblivion: > it will always overwrite files when the dest is a folder

I guess that explains why it keeps re-downloading the small maps.black assets every time :facepalm: Thanks.

Well okay, but dest isn't always a folder, right? Just looking down the line, the Gitea checksum, the APK, etc. That's now going to be downloaded every run. Did it do that before, and/or do we want that now? (This is a side effect from deleting the `changed=False` thing).

Also, I think simply adding `args: creates:` to the `get_curl` will be tantamount to the `changed=False` thing. I.e. incomplete files would be considered done before curl gets a chance to look into it. Though `when:` or `args: creates:` on a location that the download is _moved_ to would work.

I think this is hairy because `get_url` (as I understand) already does this under the hood. It downloads to an unspecified temp location and then moves over the result. That way it's unambiguous when it's done. The download thing that we made with `aria2c` effectively does the same thing. So I'm not sure this is a drop-in replacement with the same behavior.

chapmanjacobd: `changed=False` only happened when `force=False` which none of the code was using.

> Though when: or args: creates: on a location that the download is moved to would work.

> get_url (as I understand) already does this under the hood. It downloads to an unspecified temp location and then moves over the result ... That way it's unambiguous when it's done

That seems fair. I'll add this change

https://github.com/ansible/ansible/blob/devel/lib/ansible/modules/get_url.py

ASIDE: I've been looking at url_get but actually it is `main()` that is the `run_module()` function! I thought that was just tests or something

chapmanjacobd: okay I've added back in the force option and it means the same thing as `get_url`: 

We don't download if the `dest` is a file and already exists but if `dest` is a folder then we have to download it to know what the filename will be--Ansible has a bit less of an excuse because they could parse the response headers and stop the download to the temporary dest but we have even more of an excuse because even if we make a separate HEAD request, we don't know _exactly_ what `curl` will choose

modules/get_curl.py
```diff
@@ -0,0 +1,146 @@
...
+    if tmp_dest:
+        if not os.path.isdir(tmp_dest):
+            module.fail_json(msg=f"tmp_dest must be a directory: {tmp_dest}", dest=dest, url=url)
+        temp_dir = tmp_dest
+    else:
+        temp_dir = module.tmpdir
+
+    # Create temporary file/directory for download
+    try:
+        temp_path = tempfile.mkdtemp(dir=temp_dir)
```
orblivion, line 64: I'm thinking about the issue I heard about a couple months back where there was unnecessary copying going on during a `mv` because something ended up in a different file system. I don't recall the details but I think you may even have been involved in the fix @chapmanjacobd .

So with that in mind, do we know where `get_url` puts its temp directories? `module.tmpdir` https://github.com/ansible/ansible/blob/devel/lib/ansible/modules/get_url.py#L427C10-L427C33 If it's in `/tmp/` then at least `tempfile.mkdtemp` matches it.

orblivion: But if let's say `get_url` defaults to the same file system and `/tmp` is a different one (and not a tmpfs), maybe we'd be adding additional writes here.

orblivion: Sorry, to be explicit: Python's `tempfile` makes a directory under `/tmp/`. I'm wondering if Ansible has been doing the same thing.

roles/luanti/tasks/install.yml
```diff
@@ -34,9 +34,15 @@
...
 
 - name: Download default game https://content.luanti.org/packages/Luanti/minetest_game/ (2025-03-02 VERSION FOR NOW) -- if luanti_default_game == "minetest" AND (RasPiOS >= 13 or Debian >= 13 or Ubuntu >= 24.10)
+  get_curl:
+    url: https://content.luanti.org/packages/Luanti/minetest_game/releases/29922/download/
+    dest: "{{ downloads_dir }}/game.zip"
+    timeout: "{{ download_timeout }}"
+  register: luanti_game_dl
+
+- name: Unarchive luanti default game
   unarchive:
```
orblivion, line 45: Seems like the point is to use curl instead of the built-in `unarchive` downloader. Makes sense. I don't see anything about deleting the archive after unarchiving. Not deleting the archive would be good actually, in that `get_curl` wouldn't download it on subsequent runs. (There is the minor downside of space taken up but I'm assuming this stuff won't be of significant size).

But I'm curious if `unarchive` deletes the archive when it's done? The documentation for it doesn't say anything about it.

modules/get_curl.py
chapmanjacobd: Most recent OSes use `tmpfs` but you're right it's something that we need to keep in mind.

`get_url` works the same way and I think it makes sense to follow that pattern. In both `get_curl` and `get_url` you can override this by passing in a `tmp_dest` [parameter](https://github.com/ansible/ansible/blob/52cd605887e592a906d316eb11d6507b5100b36c/lib/ansible/modules/get_url.py#L524) if you need to download something larger than can fit in memory.

Re additional writes for older OSes: as it is just a one-time setup I would imagine it is mostly inconsequential--it matters more if something is writing over and over every day or every hour

roles/matomo/tasks/install.yml
```diff
@@ -107,11 +107,17 @@
         - localhost
         - "{{ tcp_host }}"
 
-- name: Download and Extract Matomo (~3 min)
+- name: Download Matomo (2 min)
+  get_curl:
+    url: "{{ matomo_dl_url }}"    # e.g. https://builds.matomo.org/matomo.tar.gz
+    dest: "{{ downloads_dir }}/matomo.tar.gz"
```
orblivion, line 113: @holta What's the difference between `/opt/iiab/downloads` (aka `downloads_dir`), `/library/working` and `/library/downloads`? The latter two you told me to use for temporary map download stuff, but it seems like `downloads_dir` is used for installing other software?

roles/nextcloud/tasks/install.yml
```diff
@@ -136,10 +136,15 @@
 #    nextcloud_dl_url: https://download.nextcloud.com/server/releases/latest-25.tar.bz2
 #  when: php_version is version('8.0', '<')
 
-- name: Unarchive {{ nextcloud_dl_url }} (~261 MB) to {{ nextcloud_root_dir }} (~820 MB initially, sometimes ~875 MB later, {{ apache_user }}:{{ apache_user }})
+- name: Download {{ nextcloud_dl_url }} (~261 MB) to {{ nextcloud_root_dir }} (~820 MB initially, sometimes ~875 MB later, {{ apache_user }}:{{ apache_user }})
```
orblivion, line 139: nitpick: If I'm understanding correctly, it seems like for this `name` you can leave it at `Download {{ nextcloud_dl_url }} (~261 MB)`, and for the next `name` you can leave out `(~261 MB)`.

nitpick: I downloaded the file and it's closer to 138M? Not worth fixing now unless it's trivial to sort out.

roles/nodered/tasks/settings.yml
chapmanjacobd: Actually you're right. This is doing something different than I thought, however, reading through the code it seems like it's just doing the same work twice--but I'll revert this change in case the original ownership is important

roles/phpmyadmin/tasks/install.yml
```diff
@@ -4,21 +4,15 @@
...
-    path: "{{ downloads_dir }}/{{ phpmyadmin_name_zip }}"    # e.g. /opt/iiab/downloads/phpMyAdmin-5.0.4-all-languages.zip
   register: phpmyadmin_dl
 
-
-- name: Unarchive {{ phpmyadmin_dl_url }} (7+ MB) to /opt (60+ MB)
+- name: Unarchive {{ phpmyadmin_dl.dest }} (7+ MB) to /opt (60+ MB)
   unarchive:
-    remote_src: yes    # Overwrite even if "already exists on the target"
-    src: "{{ phpmyadmin_dl_url }}"    # e.g. https://files.phpmyadmin.net/phpMyAdmin/5.1.1/phpMyAdmin-5.1.1-all-languages.tar.xv
+    src: "{{ phpmyadmin_dl.dest }}"    # e.g. https://files.phpmyadmin.net/phpMyAdmin/5.1.1/phpMyAdmin-5.1.1-all-languages.tar.xv
```
orblivion, line 15: super nitpick: Seems like `xv` is another old typeo, though at least it's clear that it's a typo.

roles/luanti/tasks/install.yml
chapmanjacobd: That's a good point. unarchive calls `urls.fetch_file` which saves the archive to a temp directory and cleans it up later via `module.add_cleanup_file`

roles/phpmyadmin/tasks/install.yml
chapmanjacobd: yeah should be xz

roles/luanti/tasks/install.yml
orblivion: But what if it's in the filesystem to begin with? If it doesn't delete the file if the source is in the fs I think we're good here.

chapmanjacobd: yeah unarchive never deletes

roles/nodered/tasks/settings.yml
orblivion: Funny, you convinced me and now you'll have to explain me how I was right.

modules/get_curl.py
orblivion: I appreciate it, it's actually easier for me to follow the logic of all of the calls to `get_curl` knowing that it works roughly the same as before.

> ASIDE: I've been looking at url_get but actually it is main() that is the run_module() function! I thought that was just tests or something

Was this just an interesting observation or did you mean that you intended to change `run_module` to `main`? I don't mind either way I just want to make sure you didn't forget if you intended to do so.

> if dest is a folder then we have to download it to know what the filename will be

Is this why you've added an explicit file path for a bunch (though not all?) of them? Otherwise we're going to be downloading over and over? But you're saying that Ansible has been doing this anyway so we're not any worse off for the ones where `dest` is still a directory?

(As a side note, it's odd to me that Ansible would rely on headers to choose a file name. For a browser I understand since a human can deal with the situation.)

roles/nodered/tasks/settings.yml
chapmanjacobd: Well it's a boolean that means that the src file already exists on the remote system and doesn't need to be copied over--which is the opposite of what it sounds like to me.

But in the code Ansible copies it over anyway to `/tmp` in case the copy function needs it... *shrug*

And actually in the nodered role we explicit set the owner and group so all of this is moot unless the default 644 permissions are not desired...

So I'll just revert the change in case it is doing something that I don't understand

scripts/ansible
```diff
@@ -216,6 +216,31 @@ echo -e "\napt update; apt install python3-venv"
...
+
+        # Check for resume-retries support
+        local has_resume=$($venv/bin/python3 -m pip install --help | grep -q "resume-retries" && echo "yes" || echo "no")
+        if [ "$has_resume" == "yes" ]; then
+            return 0
+        fi
+
+        echo "Attempt $i failed. Retrying in 2s..."
+        sleep 2
+    done
```
orblivion, line 241: Do you want to `return` non-zero here with an error message saying you couldn't install it after 3 attempts? Or does bash do this automatically? Otherwise I don't know the point of `return 0`.

modules/get_curl.py
chapmanjacobd: > did you mean that you intended to change run_module to main?

no I mean the Ansible modules are main(). eg. `get_url` is get_url.py:main() which initially I initially thought was some integration test because it seemed rather long. I will keep run_module() for get_curl as that seems clearer to me

---

yeah I think the reason they `extract_filename_from_headers` is because it's what browsers do so it is what people expect to happen. And you expect `get_curl` to behave the same way as `get_url` right?

It's better to just provide the filename that you want rather than the download directory because that eliminates all of this

scripts/ansible
chapmanjacobd: it's just an early return so that it doesn't print attempt failed when it succeeded already

but sure I'll add return 1 at the end. that makes sense to have--there will already be a really good error message provided by pip if this fails

modules/get_curl.py
```diff
@@ -0,0 +1,149 @@
...
+                module.fail_json(
+                    msg="curl failed", rc=process.returncode, stdout=stdout, stderr=stderr, dest=dest, url=url
+                )
+        except Exception as e:
+            module.fail_json(msg=f"Failed to execute curl: {to_native(e)}", dest=dest, url=url)
+
+        if dest_is_dir:
+            # Find the downloaded file(s) in temp directory
+            try:
+                downloaded_files = [entry.path for entry in os.scandir(temp_path)]
```
orblivion: In what situation would there be more than one file?

roles/luanti/tasks/install.yml
```diff
@@ -34,9 +34,15 @@
...
-    remote_src: yes
-    src: https://content.luanti.org/packages/Luanti/minetest_game/releases/29922/download/
+    src: "{{ luanti_game_dl.dest }}"
     dest: "{{ luanti_working_dir }}/games/"
   when: luanti_default_game == "minetest" and ((is_raspbian and os_ver is version('raspbian-13', '>=')) or (is_debian and not is_raspbian and os_ver is version('debian-13', '>=')) or (is_ubuntu and os_ver is version('ubuntu-2410', '>=')))    # Gratuitous parens for clarity
```
orblivion, line 48: I just noticed that you copied the `when:` clause over for mongodb. Perhaps the same here?

modules/get_curl.py
chapmanjacobd: It's mostly defensive programming but curl can definitely download more than one file for one URL when dest=folder which is another reason to encourage the use of dest=filename

roles/nodered/tasks/os-integration.yml
```diff
@@ -2,10 +2,9 @@
 
 # 2022-06-09: Crucially, the 3 "dest" dirs (below) exist on all mainline OS's
 - name: "Download 3 Node-RED enhancements: Node-RED icon, start menu item, /etc/logrotate.d/nodered"
-  get_url:
+  get_curl:
     url: "{{ item.url }}"
     dest: "{{ item.dest }}"
-    force: yes
```
orblivion, line 8: Is this something you forgot to undo or do you not want to force here? Perhaps the intention was to always overwrite with the latest version?

roles/osm-vector-maps/tasks/install.yml
```diff
@@ -178,10 +178,9 @@
 
 
 - name: Force Download redirect {{ osm_repo_url }}/{{ maps_branch }}/osm-source/pages/viewer/installer-index.redirect to test page {{ vector_map_path }}/maplist/index.html
-  get_url:
+  get_curl:
     url: "{{ osm_repo_url }}/{{ maps_branch }}/osm-source/pages/viewer/installer-index.redirect"
     dest: "{{ vector_map_path }}/maplist/index.html"
-    force: yes
```
orblivion, line 184: On the other hand no worries on this one, I'm pretty sure it won't be updated.

roles/luanti/tasks/install.yml
chapmanjacobd: huh?

    git diff master...HEAD | grep -i when

oh. good idea

roles/nodered/tasks/os-integration.yml
chapmanjacobd: yeah okay I'll add that back

roles/osm-vector-maps/tasks/install.yml
chapmanjacobd: eh. If it needs to overwrite the file... I added it back in. We could remove the task in a separate PR if its not needed at all

modules/get_curl.py
orblivion: Would this ever be desired behavior? Did `get_url` do this? Would it make sense to just fail in this case? Just seems weird to me.

(This will be the last point. Everything else is a nitpick and I don't want to drag on any further.)

chapmanjacobd: Okay I'll just fail it if there is more than one file

roles/matomo/tasks/install.yml
holta: @orblivion for content (a.k.a. EduPacks, learning materials) let's download to
`/library/downloads`

And if any of these need to be unpacked or processed, we can do that within:
`/library/working`

Whereas that 3rd one is much less important, used occasionally by IIAB Ansible roles to download programs:
`/opt/iiab/downloads` (aka `downloads_dir`)

(Hope that helps!?)

orblivion (approved):

holta:
@chapmanjacobd should we give this a shot, merging now?

chapmanjacobd:
Sure!

holta:
@EMG70 would you be able to do some post-merge testing — e.g. of LARGE-sized IIAB install(s) on the latest RPiOS ?

(If possible anyway!!)

EMG70:
Sudo iiab-diagnostics https://paste.centos.org/view/83a70289

IIAB large size installed on Debian 13.03 .
All working OK ,no issues found .

tim-moody:
How do you test the changes to the role phpmyadmin?

PR #4212, opened 2026-01-10
Don't download map js/style assets over and over

orblivion:
### Fixes bug:

maps.black and maplibre assets were downloading every time. `resourcetiles-minimal.pmtiles` in particular takes about 5 seconds.

I couldn't figure out why before and I gave up. It turns out the reason is that the `force` parameter is as-if true if the destination is a directory. I had no idea. (seems dumb to me but maybe there's a good reason I'm not thinking of).

Specify a file path for the `dest` of the `get_url`s in the maps role. @chapmanjacobd your PR #4210 does not update these with `get_curl`. If you plan to do so very soon I can hold off on this.

Trixie raspi

@chapmanjacobd


chapmanjacobd (approved):
> If you plan to do so very soon I can hold off on this

No I don't plan on supporting the checksum parameter at this time

PR #4213, opened 2026-01-12
Cleanup: Remove no longer used `phpmyadmin_name_zip`

orblivion:

This variable is no longer used as of the `get_curl` PR: https://github.com/iiab/iiab/pull/4210/files#diff-d1a1702f3dc7f2318ff77da03bbab7e0f786b7c8e11312430242a9727467e00fL7

I didn't want to add more nitpicks to hold it up the review, but I thought it might be worth cleaning this up to avoid future confusion. The only uses of `phpmyadmin_name_zip` were the definition and a comment, and I think the comment was a typo based on looking at other code.

@holta

### DIFF
--- a/roles/phpmyadmin/defaults/main.yml
+++ b/roles/phpmyadmin/defaults/main.yml
@@ -7,4 +7,3 @@
 phpmyadmin_version: 5.2.3
 phpmyadmin_name: "phpMyAdmin-{{ phpmyadmin_version }}-all-languages"
 phpmyadmin_dl_url: "https://files.phpmyadmin.net/phpMyAdmin/{{ phpmyadmin_version }}/{{ phpmyadmin_name }}.tar.xz"
-phpmyadmin_name_zip: "{{ phpmyadmin_version }}/{{ phpmyadmin_name }}.tar.xz"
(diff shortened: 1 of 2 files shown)


holta:
@orblivion is this "smoke tested" ?

(Granted roles/phpmyadmin is not used much at all FWIW — still basic "smoke testing" is IIAB's usual/minimum request with all PRs.)

tim-moody:
To be clear, a role is not smoke tested if _install is False. This comment also applies to #4210

orblivion:
To be clear, this variable whose definition I deleted is no longer used anywhere in this codebase. Unless it's used in some other repository that's pulled in during installation, I see no way this change would cause a change in behavior.

tim-moody:
>To be clear, this variable whose definition I deleted is no longer used anywhere in this codebase. Unless it's used in some other repository that's pulled in during installation, I see no way this change would cause a change in behavior.

True. The real change was in #4210 where I asked the same question and got no response.

holta:
Thanks everyone for clarifying!

Original question was here:

- https://github.com/iiab/iiab/pull/4210#issuecomment-3741100104

PR #4215, opened 2026-01-13
[proot] fix missing minimal default facts for Android 16

Ark74:
### Fixes bug:
Android 16 is more prone to trigger security restrictions when gathering all subsets.
Since the first step is to gather facts: is_proot or not is_proot are not yet available.

### Description of changes proposed in this pull request:
This change fix missing gather files routine not patched for Android 16, so it defaults to minimal and grow from there if required. Matching working run-one-role.yml.

### Smoke-tested on which OS or OS's:
Debian 13 - Android 16 (Samsung - A165MUBS6CYL1)

### Mention a team member @username e.g. to help with code review:
@holta

PR #4219, opened 2026-01-14
[proot] fix insufficient minimal facts bootstrap

### LINKED ISSUE #4218: PR #4122 appears to prevent PostgreSQL from installing
PR #4122 seems to have broken regular LARGE-sized IIAB installs.  Here's an example of the repeated failure, as the LARGE-sized IIAB install tried to install Moodle and hence PostgreSQL:

```
TASK [postgresql : Install /library/pgsql-iiab/postgresql.conf from template (postgres:postgres, 0640)] *************************************************************************************************************************************
[ERROR]: Task failed: Error rendering template: '<' not supported between instances of '_AnsibleTaggedStr' and 'int'

Task failed.
Origin: /opt/iiab/iiab/roles/postgresql/tasks/install.yml:72:3

70     msg: "{{ ansible_memtotal_mb | type_debug }}"
71
72 - name: Install /library/pgsql-iiab/postgresql.conf from template (postgres:postgres, 0640)
     ^ column 3

<<< caused by >>>

Error rendering template: '<' not supported between instances of '_AnsibleTaggedStr' and 'int'
Origin: /opt/iiab/iiab/roles/postgresql/templates/postgresql.conf.j2

fatal: [127.0.0.1]: FAILED! => {"changed": false, "msg": "Task failed: Error rendering template: '<' not supported between instances of '_AnsibleTaggedStr' and 'int'"}
```

@Ark74 the immediate fix (please confirm, and submit a "quickfix" PR to repair this if so!?) ...would appear to be changing `iiab-stage.yml` **and** `run-on-role.yml` from...

```
            ansible_memtotal_mb: "{{ memswap.stdout.split()[0] }}"
            ansible_swaptotal_mb: "{{ memswap.stdout.split()[1] }}"
```

...to...

```
            ansible_memtotal_mb: "{{ memswap.stdout.split()[0] | int }}"
            ansible_swaptotal_mb: "{{ memswap.stdout.split()[1] | int }}"
```

(...so that variable type, for both above variables, is properly restored from type `str` to type `int` !)

Ark74:
### Fixes bug:
Fix #4218 
Fix missing #4215 

As suggested previously this PR enables a follow up gather facts normal on non-android devices to  complete missing values.

**Ubuntu 24.04 (VM)**
- Medium : :ok: 
    - install log: https://termbin.com/76h0
    - diagnostics: https://paste.centos.org/view/51091644
- Large: :ok: 
    - install log: https://termbin.com/nenq
    - diagnostics: https://paste.centos.org/view/c56181dc

**RPiOS - RPi4** :ok:
- Large: 
    - install log: https://termbin.com/4sjj
    - diagnostics: https://paste.centos.org/view/052beb01
@holta

### DIFF
--- a/iiab-stages.yml
+++ b/iiab-stages.yml
@@ -12,7 +12,7 @@
     - /etc/iiab/iiab_state.yml
 
   pre_tasks:
-    - name: Gather minimal facts default
+    - name: Gather minimal facts bootstrap
       setup:
         gather_subset:
           - 'all'
@@ -35,13 +35,13 @@
           register: memswap
           changed_when: false
 
-        - name: Set fallback memtotal and swaptotal facts
+        - name: Set fallback memtotal and swaptotal facts (proot)
           set_fact:
             ansible_memtotal_mb: "{{ memswap.stdout.split()[0] | int }}"
             ansible_swaptotal_mb: "{{ memswap.stdout.split()[1] | int }}"
 
   tasks:
-    - name: Complete gathering facts from initial light set.
+    - name: Gather default facts after minimal bootstrap (non-proot)
       setup:
       when: not is_proot
 
--- a/iiab-stages.yml
+++ b/iiab-stages.yml
@@ -11,13 +11,36 @@
     - /etc/iiab/local_vars.yml
     - /etc/iiab/iiab_state.yml
 
-  pre_tasks:
-    - name: Gather minimal facts bootstrap
+  tasks:
+    - name: Detect proot environment early
+      shell: |
+        if [ -n "${ANDROID_ROOT}" ]; then
+          echo true
+        else
+          echo false
+        fi
+      register: proot_probe
+      changed_when: false
+      failed_when: false
+
+    - debug:
+        var: proot_probe
+
+    - name: Set is_proot_initial
+      set_fact:
+        is_proot_initial: "{{ (proot_probe.stdout | trim) == 'true' }}"
+
+    - name: Gather minimal facts bootstrap (proot-safe)
+      when: is_proot_initial
       setup:
         gather_subset:
           - 'all'
           - '!hardware'
           - '!network'
+
+    - name: Gather default facts after minimal bootstrap (non-proot)
+      when: not is_proot_initial
+      setup:
 
     - name: Compute memory and swap facts when hardware subset is missing
       when: ansible_swaptotal_mb is not defined or ansible_memtotal_mb is not defined
@@ -39,11 +62,6 @@
           set_fact:
             ansible_memtotal_mb: "{{ memswap.stdout.split()[0] | int }}"
             ansible_swaptotal_mb: "{{ memswap.stdout.split()[1] | int }}"
-
-  tasks:
-    - name: Gather default facts after minimal bootstrap (non-proot)
-      setup:
-      when: not is_proot
 
     - name: 0-init
       include_role:
--- a/iiab-stages.yml
+++ b/iiab-stages.yml
@@ -38,12 +38,12 @@
           - '!hardware'
           - '!network'
 
-    - name: Gather default facts after minimal bootstrap (non-proot)
+    - name: Gather default facts
       when: not is_proot_initial
       setup:
 
     - name: Compute memory and swap facts when hardware subset is missing
-      when: ansible_swaptotal_mb is not defined or ansible_memtotal_mb is not defined
+      when: ansible_facts['swaptotal_mb'] is not defined or ansible_facts['memtotal_mb'] is not defined
       block:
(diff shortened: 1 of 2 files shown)


run-one-role.yml
```diff
@@ -34,10 +34,14 @@
...
           changed_when: false
 
-        - name: Set fallback memtotal and swaptotal facts
+        - name: Set fallback memtotal and swaptotal facts (proot)
           set_fact:
-            ansible_memtotal_mb: "{{ memswap.stdout.split()[0] }}"
-            ansible_swaptotal_mb: "{{ memswap.stdout.split()[1] }}"
+            ansible_memtotal_mb: "{{ memswap.stdout.split()[0] | int }}"
+            ansible_swaptotal_mb: "{{ memswap.stdout.split()[1] | int }}"
+  tasks:
```
holta:
```suggestion

  tasks:
```

iiab-stages.yml
```diff
@@ -12,7 +12,7 @@
     - /etc/iiab/iiab_state.yml
 
   pre_tasks:
-    - name: Gather minimal facts default
+    - name: Gather minimal facts bootstrap
```
holta: @Ark74 can we speed this up?  So that facts aren't being gathered twice (a slow process), on non-proot environments.

My suggestion would basically be (if possible) a big if/else clause: (assuming `is_proot` is already defined at this point hopefully!?)

1. If `is_proot` then... gather facts one way

2. If `not is_proot` then... gather facts another way

(Can you make something like this work?)

iiab-stages.yml
```diff
@@ -35,12 +35,16 @@
...
           set_fact:
-            ansible_memtotal_mb: "{{ memswap.stdout.split()[0] }}"
-            ansible_swaptotal_mb: "{{ memswap.stdout.split()[1] }}"
+            ansible_memtotal_mb: "{{ memswap.stdout.split()[0] | int }}"
+            ansible_swaptotal_mb: "{{ memswap.stdout.split()[1] | int }}"
 
   tasks:
+    - name: Gather default facts after minimal bootstrap (non-proot)
+      setup:
+      when: not is_proot
```
holta: @Ark74 can you consider moving all facts gathering into the `pre_tasks:` section?

(To be cleaner / more modular?)

holta:
CLARIF: I don't mean to be critical!!

- Am just asking if we can (somehow!) speed this up substantially (:

- i.e. so facts aren't being gathered twice 🙏

iiab-stages.yml
jvonau: Was setup like that prior to #4215, think the problem is with the value for is_proot in [default_vars.yml](https://github.com/iiab/iiab/blob/master/vars/default_vars.yml#L240-L242) is detected based on the gathered facts but that can't occur until the facts are gathered. Perhaps that should be "False" and 'is_proot: True' be added to [local_vars_android.yml](https://github.com/iiab/iiab/blob/master/vars/local_vars_android.yml). The is_proot variable would then be available in the main play as /etc/iiab/local_vars.yml is read before the gather facts stage.

holta: @Ark74 does @jvonau's suggestion (above) have potential?

(Or anything like that maybe?!)

Ark74: Indeed, we don't have is_proot available until the first 'bootstraping' fact gathering.
If we want to avoid double facts gathering we would need to define a proot environment identification inside both iiab-stages.yml and run-one-role.yml

So we can then say if proot, minimal gathering if not proot, regular gathering.

EDIT: Wait, it just sink in. @jvonau is not a bad idea. Let me look that approach.

iiab-stages.yml
Ark74: Sure

holta: Or if there's a practical reason to move all "facts gathering" from `pre-tasks:` to `tasks:` ?

(Goal: anything clean / modular — i.e. very understandable and maintainable!)

holta: For example if `is_proot` is already set (as Ansible invokes it within default_vars.yml) prior to the `tasks:` section being run — then the `pre-tasks:` section could potentially go away completely?

(Just an idea, that we might consider, if `tasks:` is practically/pragmatically better than `pre-tasks:` ?!)

iiab-stages.yml
Ark74: Ok, no.
I think current PR is the most clean / compact way to prevent breaking any host.

UPDATE: I take back what I said earlier. Actually seeing #4221 in action I think it's a better approach. I was not catching some details.

Ark74:
I've tested #4221 on the problematic A16 and it works. 
No need to keep this one.

As result of the testing some updates had been made to termux-setup.
Thank you @holta @jvonau

PR #4217, opened 2026-01-14
Ansible 0-init cleanup

jvonau:
### Fixes bug:
work towards #4138, kind of tired of viewing the warnings. 
### Description of changes proposed in this pull request:
adopt newer facts naming scheme
### Smoke-tested on which OS or OS's:
CI
### Mention a team member @username e.g. to help with code review:
related: #4142

PR #4221, opened 2026-01-14
POC [proposal to hard code `is_proot: True` in local_vars_android.yml]

jvonau:
### Fixes bug:
alternate method https://github.com/iiab/iiab/pull/4219#discussion_r2690524729

### DIFF
--- a/vars/local_vars_android.yml
+++ b/vars/local_vars_android.yml
@@ -3,6 +3,7 @@
 ## on the planet, Android Phones.
 ## This is a WIP project and it's evolving fast.
 
+# Default hard-coded variable for android / proot-distro environments; please don't edit.
 is_proot: True
 
 ##############################
--- a/vars/local_vars_android.yml
+++ b/vars/local_vars_android.yml
@@ -3,6 +3,8 @@
 ## on the planet, Android Phones.
 ## This is a WIP project and it's evolving fast.
 
+is_proot: True
+
 ##############################
 # Currently it targets 5 main apps
 # 1 Kiwix          localhost:8085/kiwix
--- a/iiab-stages.yml
+++ b/iiab-stages.yml
@@ -12,15 +12,20 @@
     - /etc/iiab/iiab_state.yml
 
   pre_tasks:
-    - name: Gather minimal facts default
+    - name: Gather minimal facts for proot
       setup:
         gather_subset:
           - 'all'
           - '!hardware'
           - '!network'
+      when: is_proot
+
+    - name: Gather full facts default
+      setup:
+      when: not is_proot
 
     - name: Compute memory and swap facts when hardware subset is missing
-      when: ansible_swaptotal_mb is not defined or ansible_memtotal_mb is not defined
+      when: is_proot
       block:
         - name: Read memtotal and swaptotal from /proc/meminfo
           shell: |
--- a/run-one-role.yml
+++ b/run-one-role.yml
@@ -11,15 +11,20 @@
     - /etc/iiab/iiab_state.yml
 
   pre_tasks:
-    - name: Gather minimal facts default
+    - name: Gather minimal facts for proot
       setup:
         gather_subset:
           - 'all'
           - '!hardware'
           - '!network'
+      when: is_proot
+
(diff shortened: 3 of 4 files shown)


jvonau:
The casting can be fixed later or I can toss that in also just want to see the CI tests run first

jvonau:
https://github.com/iiab/iiab/actions/runs/21006356291/job/60389604411?pr=4221
```
PLAY [all] *********************************************************************

TASK [Gather minimal facts for proot] ******************************************
skipping: [127.0.0.1]

TASK [Gather full facts default] ***********************************************
ok: [127.0.0.1]

TASK [Read memtotal and swaptotal from /proc/meminfo] **************************
skipping: [127.0.0.1]

TASK [Set fallback memtotal and swaptotal facts] *******************************
skipping: [127.0.0.1]

TASK [0-init] ******************************************************************
included: 0-init for 127.0.0.1

TASK [0-init : ...IS BEGINNING ============================================] ***
```

holta:
@Ark74 did I understand you correctly... saying there was some kind of problem with this approach @jvonau has articulated here?

(I might have misunderstood, if so apologies!!)

- https://github.com/iiab/iiab/pull/4219#discussion_r2690943669

chapmanjacobd (approved):
Makes sense to me but Luis seemed apprehensive about this approach. I'm curious what reasons there might be?

If this is time sensitive and master is effectively broken?, I wouldn't be against merging this first and we can figure out any long term fixes that need to be made later

imho this approach seems cleaner where #4219 has a lot of moving parts? but maybe there's an edge case I'm not seeing

Ark74:
@holta well, let me test it.
I was still trying to fit everything in one layer, and failing xD

This is actually clever, using both layer of vars files plays  nice, still. From all the work and errors I've done in the last month, I've come to learn that unless tested even if it makes clearly a lot of sense, something might be hiding.

Actually kudos to @jvonau I thought I understood the first time I read the description, then I failed to see the bigger picture, What a week huh[1]

[1] [What a week huh](https://www.peppercarrot.com/rc/viewer/misc__2022-11-02_What-a-week-huh_by-David-Revoy.html) by David Revoy  − [CC-BY 4.0](https://creativecommons.org/licenses/by/4.0/deed.rc)

jvonau:
I was going to come back to flip around the order of gather_facts but @holta beat me to the change.

holta:
@Ark74 can you review this?  (Thanks if possible!)

Ark74:
@holta I think there are missing changes from #4217 but so far I did tested the current state and confirmed they work.
I closed #4219 in favor of this one. So, I think it would be the best place to continue the changes proposed here.

@jvonau would you agree?

holta:
@Ark74 what's your opinion on replacing these 2 lines... ?

```
when: ansible_swaptotal_mb is not defined or ansible_memtotal_mb is not defined
```

...is that really what you want — thinking about code safety one way or the other?  (Y/N, either way!)

Ark74:
> @Ark74 what's your opinion on replacing these 2 lines... ?

Using hardcoded values for is_proot in vars files (default/android) makes sure that we only run the minimal setup in proot-distro environments. As long as the variables (_memtotal/swaptotal_) are honored later on to be used by other roles.
Yeah, it's safe to replace the `when` condition.

holta:
@Ark74 is your recommendation that this PR be merged as is?

vars/local_vars_android.yml
```diff
@@ -3,6 +3,8 @@
 ## on the planet, Android Phones.
 ## This is a WIP project and it's evolving fast.
 
+is_proot: True
```
Ark74, line 7:
```suggestion
# Default hard-coded variable for android / proot-distro environments; please don't edit.
is_proot: True
```

Ark74 (changes requested):
One minor comment to add.

vars/local_vars_android.yml
Ark74: @holta if possible let's add this minor comment.

Ark74:
With that minor change, it should be :ok:

PR #4223, opened 2026-01-15
Download map regions [towards extracts of extracts?]

orblivion:

Adds a UI for downloading regions, with the help of CLI:  When you turn on the new `maps_region_downloader` setting in Ansible, you get a button on the top left. It lets you draw a rectangle, which gives you a popup to copy a command to paste into a terminal (connected to your IIAB).

Screenshot 2026-01-14 at 23-43-32

Screenshot 2026-01-14 at 23-43-44

You run the command and it downloads full quality vector satellite and terrain.

Screenshot from 2026-01-14 23-50-05

You then reload the page, and you can click on the region, and you enter a mode where you are looking at it in full quality. But you're barred from scrolling away. The borders are surrounded by 50% transparent dark bars to indicate that the world outside of it exists but it's not of interest right now.

Screenshot 2026-01-14 at 23-50-30

And you can see full quality everything.

Screenshot 2026-01-14 at 23-50-55
Screenshot 2026-01-14 at 23-51-22

Finally, again with the `maps_region_downloader` setting, you get a wastebasket button, when viewing a region, that brings up a popup to delete the regions. It is again a command to paste into a terminal (connected to your IIAB). You also have a "return" button on the right to bring you back to the full world map.

Screenshot 2026-01-14 at 23-51-51

Accompanying change to maps.black: https://github.com/iiab/maps.black/compare/c1baf6d07f181ff62f9c272060056640dbec9c7f...1e188ffbdc4f18a5c258598d48c64fe1dbbd9c8e

This one is a lot. Happy to be on hand answering questions for reviewers. I'll anticipate a couple questions:

* I've looked at making it switch in and out of the region view mode based on zoom and location. It works well under certain conditions but is often not great. I could work on it but I wanted a simple solid option first.
* The code is a mess. I imagine it has plenty of javascript code smell. At some point, maybe soon, I would like to use our new fork of maps.black to make a proper "map.onload" handler instead of using timeouts. I don't think the UI is unreliable yet but the code to keep it reliable could get simpler.

@holta @chapmanjacobd


chapmanjacobd:
top-left is likely better for the disable region button since the other extract related buttons are on that side. My eyes naturally looked there first (across the different screen shots) since I knew the download and delete region buttons were there

orblivion:
I think I agree but for a different reason. The other extract-related buttons are for "admins", which is a little different, and they also won't be there when the admin variable is turned off.

However, having it appear on the left would probably be more noticeable to a viewer, and also people expect "back" buttons to be on the left I think.

Also btw I'm ready for the critique that nobody will notice the button at all and feel trapped in that view. But we'll see what people say.

orblivion:
You know, maybe I could even make a whole bigger widget that combines buttons. But, in the long run hopefully the "Go back" button will go away in lieu of scrolling and zooming away.

roles/maps/templates/index.html.j2
```diff
@@ -162,7 +391,7 @@
                             true:  "raster/s2maps/2023",
                             false: "raster/s2maps/2023", // we'll have a smaller satellite file with the same name in this case
```
chapmanjacobd, line 417: does this do anything? it's fine as placeholder but I'm just curious

roles/maps/templates/index.html.j2
```diff
@@ -140,6 +333,42 @@
                     shouldSetDefaultView = false
                 }
             }
+            let drawControl, terraDrawInstance
+            let downloadedExtracts = null;
+            function setExtracts() {
+                return fetch('./extracts.json')
+                    .then((response) => {
+                        if (response.status === 200) {
```
chapmanjacobd: Does the server ever return something like HTTP status code 304 Not Modified?

Maybe we handle 404 special but in other cases try to decode as json and handle cases where that fails?

roles/maps/templates/tile-extract.py.j2
```diff
@@ -0,0 +1,168 @@
...
+
+def validate_extract_name(extract_name):
+    # Make sure the extract_name has only valid characters
+    assert set(extract_name) < set(string.ascii_letters + string.digits + "-_")
+    assert len(extract_name) > 0
+    assert len(extract_name) < 35
+
+def validate_extract_box(extract_box):
+    # `float` is just a parsing check.
+    assert len([float(n) for n in extract_box.split(",")]) == 4
```
chapmanjacobd: mostly nitpick: would be better to use raise with clear exceptions like TypeError, ValueError, or RuntimeError

Asserts can be skipped with `-O` but nobody really uses that. If you want to use assert it's good to add a reason why like `, "Extract name too long"` then you can catch that and print it and exit nonzero in main()

chapmanjacobd (commented):
Is there a clear distinction between "region" and "extract"? Should we only use one word or the other?

chapmanjacobd:
Everything else looks pretty solid!

chapmanjacobd (approved):

roles/maps/templates/index.html.j2
orblivion: If you choose Natural Earth instead of OSM for vector and you're not on an extract, `MAPS_VECTOR_IS_OSM === false` and you use all of these `false` things. So if you're looking at a satellite view, this is used. The comment is maybe outdated now though.

Did I understand your question right?

BTW this is an example of something I can delete if I make a low-zoom OSM file instead of NE.

roles/maps/templates/tile-extract.py.j2
orblivion: I did not know there was a mode that skips asserts!

orblivion:
Thinking a bit about this...

I could probably stand to clear up my usage of "region", "extract", even "download". In fact, I should probably do away with "extract" other than the process of getting the download. "Extract" probably shouldn't ever hit the UI.

The problem with "region" is that it's a thing that exists outside of this feature. "Downloaded region" is perfect but people (myself included) would probably tend to collapse this down to "region".

orblivion:
Wait, everything is downloaded. So maybe that's not a great term either. "Full Quality Region"?

roles/maps/templates/index.html.j2
chapmanjacobd: ah the key is visibility? I guess I'm just confused by the maplibre interface here

chapmanjacobd:
> Full Quality Region

Yeah that's pretty good. Maybe "Zoomable Region" or "Full-Zoom Region" as an alternative

roles/maps/templates/index.html.j2
orblivion: Still not sure if I'm following you, but maybe I'll explain this code a bit more and it'll answer your question.

The output of this dict-of-dicts is a maps.black "mapstyle", which I think is their own concept. Each mapstyle implies a maplibre style, and one or more data sources. For satellite there's only one data source. But for the "liberty" style, the data could be either OSM or Natural Earth, though both in the OpenMapTiles schema.

The input to the dict-of-dicts is twofold. Frist is "mapStyleChoice", which is *our* own concept. It can be "satellite", "hybrid", "natural" or "political". Second is whether the vector data is OSM (true) or Natural Earth (false). Both of these together determine the appropriate maps.black "mapstyle".

So for instance if in the UI I pick "natural", and in my IIAB settings I choose an OSM vector map, I'd end up with `"openstreetmap-openmaptiles/openfreemap/liberty"`. If I switch to a Natural Earth vector map, I'd end up with `"naturalearth-openmaptiles/openfreemap/liberty"`. If I switch to "satellite" then the mapstyle is `"raster/s2maps/2023"` either way, I don't care about the vector map.

orblivion: It probably would be more readable if it was some nested if statements.

orblivion:
Actually how about "High-Def Region"/"Hi-Def Region"? Is that an internationally recognized term?

roles/maps/templates/index.html.j2
chapmanjacobd: no need for nested if statements but just a comment explaining the true/false OSM/naturalearth mapping would make it more clear

chapmanjacobd:
No that probably doesn't translate well. Although "Full Resolution Region" is likely fine... or "Full Data Region"...

"Full Quality Region" is probably the best one so far

holta:
"High-resolution" (or maybe even "Hi-res") are well understood terms if nec?

holta:
> "Full Quality Region" is probably the best one so far

+1

chapmanjacobd:
Yeah "High Resolution Region" also works. I quite like that one--even better than "Full Quality Region". But both are good

My main worry is saying something like "High Quality Region" (which we haven't iterated to yet) implies the other regions are "low quality" lol

orblivion:
Okay cool. "Full Quality Region" then.

jvonau:
tile-extract.py.j2 could make better use of soft-coding available in defaults/main.yml in place of hard-coding
ie
 HOSTING_DIR = "/library/www/osm/maps" 
could be:
 HOSTING_DIR = "{{ maps_serve_path }}"

roles/maps/tasks/install_tile_extract.yml
```diff
@@ -0,0 +1,36 @@
...
+- name: Unarchive pmtiles archive
+  unarchive:
+    src: "{{ tile_extract.pmtiles_archive_tmp_file }}"
+    dest: "{{ tile_extract.pmtiles_archive_tmp_dir }}"
+    remote_src: yes
+    creates: "{{ tile_extract.pmtiles_archive_tmp_dir }}/pmtiles"
+
+- name: Copy pmtiles executable into working dir
+  # We cp instead of mv because we want the above command to not repeat unnecessarily.
+  command: "cp {{ tile_extract.pmtiles_archive_tmp_dir }}/pmtiles {{ tile_extract.working_dir }}"
```
chapmanjacobd: I guess this should be `cp -p` to preserve exec bit

I wonder if you could use a conditional creates to check for two different locations--

or set working_dir as the dest in unarchive instead??

jvonau: Or maybe try the ansible way with ansible.builtin.file in place of command? You can set the mode with that module.

orblivion: Yeah it was initially `mv` and I switched to `cp` just to make it idempotent. I got lazy here about using the `file` directive but I'll do that.

This possibly explains an error that a tester got.

orblivion: Do you mean the `copy` directive? With `file` I could soft or hard link, which I guess would be fine too.

chapmanjacobd: yeah you can set mode at the same time with the copy directive

orblivion:
@EMG70 Okay I hopefully fixed the error you were getting. Try updating to the latest version:

```
cd /opt/iiab/iiab
sudo git pull https://github.com/orblivion/iiab download-regions-demo --no-rebase --no-edit
```

And then run the role again:

```
sudo ./runrole maps --reinstall
```

Hopefully that fixes the bug.

holta:
Huge thanks to @orblivion, @EMG70, @chapmanjacobd & EVERYONE for helping here!!  Question:

- Should this PR be merged very soon, possibly in the next few hours, and then iterated / refined later?
- Or does it need more battle-testing (front end, back end) first?

holta:
Tiny Update:

- @orblivion is revising this PR to auto-install an Intel/AMD versus ARM binary, depending on which 64-bit architecture is needed by any particular IIAB machine.

- @chapmanjacobd recommends [GitHub's Releases API](https://docs.github.com/en/rest/releases/releases), e.g. if one of their two binaries/executables needs to be downloaded automatically from: https://github.com/protomaps/go-pmtiles/releases

roles/maps/tasks/install_tile_extract.yml
```diff
@@ -31,12 +31,12 @@
     src: "{{ tile_extract.pmtiles_archive_tmp_file }}"
     dest: "{{ tile_extract.pmtiles_archive_tmp_dir }}"
     remote_src: yes
-    creates: "{{ tile_extract.pmtiles_archive_tmp_dir }}/pmtiles"
 
 - name: Copy pmtiles executable into working dir
   copy:
     src: "{{ tile_extract.pmtiles_archive_tmp_dir }}/pmtiles"
     dest: "{{ tile_extract.working_dir }}/pmtiles"
+    force: true
```
chapmanjacobd: this is the default, actually. (similar to the `cp` command will overwrite by default)

https://github.com/ansible/ansible/blob/28927a70b43a43ecd99d5926ef7860cd6a35d12e/lib/ansible/modules/copy.py

holta:
@orblivion should this be merged now?  (Or maybe tomorrow-or-so if you prefer?)

To allow for a wider community to give us even more diverse/broad refinement feedback here!

orblivion:
The most recent push was in response to you saying that regions may be hard to find. My trick is using two colors (turquoise and black).

Screenshot 2026-01-19 at 18-15-40
Screenshot 2026-01-19 at 18-16-07

holta:
> regions may be hard to find. My trick is using two colors (turquoise and black).

Interesting!  Might be too garish for some geography educators, but either way they'll let us know ;)

orblivion:
Options on this are wide open. (We could even make it configurable) I just thought it was hard to see before, and you wanted to err on the side of obnoxious so long as it was visible.

holta:
> Options on this are wide open. (We could even make it configurable)

Yeah, let's make it configurable later on, after we gather further opinions!  With an FAQ entry somewhat like this old advice...

image

( Above screenshot is from: https://github.com/iiab/iiab/wiki/IIAB-Maps#customizations )

chapmanjacobd:
FWIW I think it looks pretty good as a mostly neutral default! But I wonder how well it matches the other UI components like the control buttons. Should we add some rounding?

roles/maps/templates/index.html.j2
orblivion: https://github.com/iiab/iiab/pull/4223/commits/8d8f211dfda451db1f0691c11ec485d32b44b192

I decided to just create the empty file so I don't have to think about 404s. But also got ready to add other things to this file.

orblivion:
I almost went with white which would match the controls. It's a little harder to see though.

Rounded rectangles seem like they're not a one-liner in maplibre but I could add it to future ideas.

orblivion:
Pending review of my last couple commits I'm ready to go.

chapmanjacobd:
> git format-patch --stdout --first-parent --no-merges a94ab97..HEAD

good catch adding that `force: false` on the empty extracts json
there are two typos though: `Initiatie the the extracts json`

---

one very minor nitpick:

>     except:
>        raise ValueError("extract_box is malformed: ", extract_box)

bare exceptions are a code smell because they can turn Ctrl-C into "extract_box is malformed" https://stackoverflow.com/a/54948581/697964

I would recommend changing this to `except Exception:`, but actually what you want here probably looks like this:

```py
    parts = extract_box.split(",")
    if len(parts) != 4:
        raise ValueError(f"extract_box must have 4 values: {extract_box}")

    try:
        floats = [float(n) for n in parts]
    except ValueError:
        raise ValueError(f"extract_box contains non-float values: {extract_box}")

    return floats
```

eh... no that is a bit verbose :/

just use `except Exception:` instead of `except:` and I'll be happy

AssertionError inherits Exception so it works fine: 
https://docs.python.org/3/library/exceptions.html#exception-hierarchy

chapmanjacobd (approved):

PR #4240, opened 2026-01-21
Map region file size estimates

orblivion:

1) Adds an estimated file size for the user to agree to before downloading.

I found this by poking at the [pmtiles CLI](https://docs.protomaps.com/pmtiles/cli). It sucks that we're parsing text; it's one thing that could break to look out for if we upgrade it. (Or perhaps something we could add as a json output in a fork? Or send upstream?)

I'm quite open to more elegant (and safer) ways of doing some of what I'm doing here, particularly dealing with parsing and displaying the units.

2) Use json output for getting pmtiles file bounds.

Fortunately, in one other place I was parsing text output, I found that json output was available.

3) Fix a bug in name collision checker

When I changed the json file schema I didn't update this part.

RasPi Trixie

@chapmanjacobd @holta

### DIFF
--- a/roles/maps/templates/tile-extract.py.j2
+++ b/roles/maps/templates/tile-extract.py.j2
@@ -1,6 +1,6 @@
 #!/usr/bin/python3
 
-import collections, json, glob, os, shutil, string, subprocess, sys
+import collections, json, glob, os, re, shutil, string, subprocess, sys
 
 HOSTING_DIR = "{{ maps_serve_path }}"
 EXTRACTS_JSON = os.path.join(HOSTING_DIR, "{{ tile_extract.info_file }}")
@@ -51,7 +51,7 @@ def get_extract_files(extract_name):
         },
     }
 
-def yeS_no(prompt):
+def yes_no(prompt):
     while True:
         response = input(f'{prompt} y/n ')
         if response == "y":
@@ -96,26 +96,30 @@ def approve_download_size(extract_name):
         )
         assert result.returncode == 0, "pmtiles extract --dry-run exited with returncode " + str(result.returncode)
 
-        # Parse transfer and archive sizes
+        # Parse transfer and archive sizes from the last line of --dry-run output.
+        # (Though we'll check every line to make it a bit more robust)
         #
         # NOTE: The reason this is two different sizes is due to something
         # called "overfetching" which is making fewer requests at the cost of
         # grabbing some extra data we don't need. This is actually a ratio we
         # can control and will likely look into for optimizing user experience.
-        transfer_num = transfer_unit = archive_num = archive_unit = None
+        pattern = (
+            r'\d+/\d+/\d+ \d+:\d+:\d+ extract.go:\d*: '
+            r'Extract transferred (\d*\.?\d*) (KB|MB|GB|TB) '
+            r'\(overfetch \d*\.?\d*\) '
+            r'for an archive size of (\d*\.?\d*) (KB|MB|GB|TB)'
+        )
         for line in result.stdout.split('\n'):
-            match line.split():
-                case [_date, _time, _code_line, "Extract", "transferred",
-                    transfer_num, transfer_unit, "(overfetch", _overfetch, "for",
-                    "an", "archive", "size", "of", archive_num, archive_unit]:
-                    break
-                case _:
-                    pass
+            re_match = re.match(pattern, line)
+            if re_match is not None:
+                break
 
-        if transfer_num is None:
+        if re_match is None:
             print (result.stdout)
             print ()
             return yes_no("Cannot get file size estimate due to unexpected output from pmtiles. Download anyway?")
+
+        (transfer_num, transfer_unit, archive_num, archive_unit) = re_match.groups()
 
         # Determine the sizes of all of the pmtiles files in bytes
         unit_mult = {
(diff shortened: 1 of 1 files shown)


orblivion:
Actually "estimate" may not even be appropriate here, afaik these are accurate values with rounding.

holta:
@chapmanjacobd do review this PR, if you're able?!

roles/maps/templates/tile-extract.py.j2
```diff
@@ -60,24 +60,93 @@ def delete_extract(extract_name):
...
+                    transfer_num, transfer_unit, "(overfetch", _overfetch, "for",
+                    "an", "archive", "size", "of", archive_num, archive_unit]:
+                    break
+                case _:
+                    pass
+
+        if transfer_num is None:
+            print (result.stdout)
+            print ()
+            raise RuntimeError("Cannot get file size estimate due to unexpected output.")
```
orblivion: Since I don't really know all of the behavior of `pmtiles`, maybe what should happen here is "sorry I can't give you an estimate. Download anyway?" sort of thing.

roles/maps/templates/tile-extract.py.j2
```diff
@@ -60,24 +60,93 @@ def delete_extract(extract_name):
...
+        # NOTE: The reason this is two different sizes is due to something
+        # called "overfetching" which is making fewer requests at the cost of
+        # grabbing some extra data we don't need. This is actually a ratio we
+        # can control and will likely look into for optimizing user experience.
+        transfer_num = transfer_unit = archive_num = archive_unit = None
+        for line in result.stdout.split('\n'):
+            match line.split():
+                case [_date, _time, _code_line, "Extract", "transferred",
+                    transfer_num, transfer_unit, "(overfetch", _overfetch, "for",
+                    "an", "archive", "size", "of", archive_num, archive_unit]:
```
chapmanjacobd: This is pretty difficult to read?! I would at the very minimum move the formatting stuff outside of the parsing stuff

roles/maps/templates/tile-extract.py.j2
chapmanjacobd: maybe instead of fatal error we still let people download if they want? There is a y/n check after all

roles/maps/templates/tile-extract.py.j2
```diff
@@ -60,24 +60,93 @@ def delete_extract(extract_name):
...
+            'TB': 1_000_000_000_000,
+        }
+        extract_sizes[files['source']] = {
+            "transfer": float(transfer_num) * unit_mult[transfer_unit],
+            "archive":  float(archive_num)   * unit_mult[archive_unit],
+        }
+
+    # Convert the file sizes back to the best available unit
+    transfer_total = sum(size['transfer'] for size in extract_sizes.values())
+    archive_total = sum(size['archive'] for size in extract_sizes.values())
```
chapmanjacobd: would also be good to compare with free space:

     total_b, used_b, free_b = shutil.disk_usage(path)

free_b - archive_total ?

chapmanjacobd (changes requested):

roles/maps/templates/tile-extract.py.j2
chapmanjacobd: yeah

roles/maps/templates/tile-extract.py.j2
orblivion: I agree it's hard to read. I'll think of something better. Though I'm not sure what you mean by formatting here.

chapmanjacobd: ah you're right the word strings are part of the parsing

Personally, I rarely use Python's match: case because it has a few weird footguns. I think for this I'd use regex

orblivion: I love Haskell too much :shrug: but yeah Python's version can trip you up; I can think of one way at any rate.

roles/maps/templates/tile-extract.py.j2
orblivion: A couple complications here:

1) I'm afraid of comparing to archive total because I don't know if there is temporary disk usage for all of the transferred data before it's deleted. Though I guess it would *probably* be discarded in memory after each request?

2) The file is extracted/downloaded to the /library/downloads/maps and then moved to /library/www/osm/maps/. How sure are we that they're on the same partition? Both would matter as far as file size.

Is there a simple thing we could do? I guess I could check the size on both directories and if they're identical we could safely use that data. And if they're different maybe I could take the smaller of the two?

chapmanjacobd: Could check both and print separate warnings for each

or skip printing a warning if free space is the same bytes value--pretty unlikely except it is same partition or some kind of non-fs mirroring setup

chapmanjacobd: Could check both and print separate warnings for each

or skip printing a warning if free space is the same bytes value--pretty unlikely except it is same partition or some kind of non-fs mirroring setup

tim-moody:
>How sure are we that they're on the same partition? 

The standard is for /library to be on a single partition either the same as or distinct from /
When people depart from the standard it usually means /library/zims or /library/www/html/modules is a separate partition.

I should add that downloads of zims and modules make this assumption.

orblivion:
Oh right, if they were separate partitions, the move from /library/downloads to /library/www would be a whole long copy, so the user would already be having a bad time.

orblivion:
@chapmanjacobd I think that addresses your suggestions?

chapmanjacobd (approved):
looks good!

orblivion:
I decided to delete that debug output because it might confuse the user. It might show file sizes but only of one of the three files.

PR #4242, opened 2026-01-22
maps large pmtiles download: allow-overwriting file when control file corrupted

chapmanjacobd:
### Fixes bug:

Sometimes the aria2c control file won't be successfully written to disk. Thankfully this operation seems to be atomic so the file shows up as a `.aria2__temp` file instead of `.aria2` file.

The default behavior of aria2c does not handle this situation very well because it doesn't recognize this failure state: it writes to a new file with a `.1.ext` extension; but you can use the `--allow-overwrite=true` flag to force overwrite the file and start over which is the safest option in this case.

`--allow-overwrite=true` does not disable resume logic. If a valid .aria2 control file exists, aria2 will still use it and resume/continue as usual.

Alternatively, we could use `--continue --check-integrity=true` which is still safe because aria2c will recheck the chunks that have been downloaded against the meta4 file... but there's no option to only do it when the control file exists so that will slow down other cases where the control file was correctly written. Given that this situation is itself an edge case let's keep things simple and reliable by just starting over.

I tested it out on Ed's PC and it fixed the problem

@orblivion @holta

### DIFF
--- a/roles/maps/tasks/download_large_file.yml
+++ b/roles/maps/tasks/download_large_file.yml
@@ -115,6 +115,7 @@
             --show-console-readout=false \
             --enable-http-pipelining=true \
             --seed-time=0 \
+            --allow-overwrite=true \
             "{{ iiab_map_host_url }}/{{ item }}.meta4" \
             >> "{{ log_file }}"
 


Ark74:
Good catch!

orblivion:
Curious how you replicated the problem to test it? Just put a dummy file there?

holta:
Thanks so much @chapmanjacobd, for having clarified that the problem @EMG70 ran into, while very real, is (probably) also somewhat rare:

> the aria2c process likely got SIGKILL when shutting down at one point but it's good that we found this edge case! it is very real but it still shouldn't happen at all if aria2c is running normally

orblivion:
Looks like you changed your mind on which option to take?

chapmanjacobd:
> Curious how you replicated the problem to test it? Just put a dummy file there?

The incomplete file was left intact on Ed's PC--just moved to the www/maps folder. So I copied it back to downloads/ to retry

But you can get the same behavior by ctrl-c halfway through and then rm the .aria2 control file

chapmanjacobd:
```
diff --git a/roles/maps/tasks/download_large_file.yml b/roles/maps/tasks/download_large_file.yml
index 2f8ede76a..f72cf5d62 100644
--- a/roles/maps/tasks/download_large_file.yml
+++ b/roles/maps/tasks/download_large_file.yml
@@ -100,10 +100,16 @@
         cd {{ working_dir }}
 
         date > "{{ log_file }}"
         echo "Downloading {{ item }}..." >> "{{ log_file }}"
         echo >> "{{ log_file }}"
+
+        OVERWRITE_OPTS=""
+        if [[ -f "{{ item }}" && ! -f "{{ item }}.aria2" ]]; then
+            OVERWRITE_OPTS="--continue --check-integrity=true"
+        fi
+
         aria2c \
             --async-dns=false \
             --connect-timeout="{{ download_timeout }}" \
             --log-level=warn \
             --console-log-level=warn \
@@ -113,11 +119,11 @@
             --max-connection-per-server=4 \
             --file-allocation=falloc \
             --show-console-readout=false \
             --enable-http-pipelining=true \
             --seed-time=0 \
-            --allow-overwrite=true \
+            $OVERWRITE_OPTS \
             "{{ iiab_map_host_url }}/{{ item }}.meta4" \
             >> "{{ log_file }}"
 
         chmod 644 "{{ item }}"
         mv "{{ item }}" "{{ dest_path }}"
```

> Looks like you changed your mind on which option to take?

actually I changed my mind back because I don't know 100% how safe it is but I know this small change will be safe

orblivion:
Yeah it seems safer to me. The downside is what, restarting a large download in a pretty rare case? Seems like the right call.

orblivion (approved):

PR #4243, opened 2026-01-23
"Most recently installed wins" for maps

orblivion:

This renames the "new" maps nginx file to match the "old" maps. This way, in case both are installed at the same time, the most recently installed will win by overwriting the file.

Before this PR, if both are installed with both nginx files, the new maps wins: visiting `/maps/` or `/maps` will show the new map. If I were to rename the new maps file from `maps-nginx.conf` to `pretty-good-maps-nginx.conf`, and install both, then the old maps would win. This is because in alphabetical order, "osm-vector-maps-nginx" comes after "maps-nginx", but before "pretty-good-maps-nginx".

RasPi Trixie

@holta


holta:
Awesome.

- And let's rename nginx.yml to enable-or-disable.yml very soon.
- Alongside other lint-like refactorings, e.g. standardizing around `none` &/or `"none"` as our null value for all vars.

PR #4244, opened 2026-01-23
Index html cleanup

orblivion:

Clean up a bit of my tangled frontend code and fix one probably inconsequential mistake. It looks complicated but it's actually quite simple if you look at all three commits separately:

* 543778b955e9c5a396e0147e678900653115e1fd Due to a bad copy/paste or something, I assigned the wrong timeout to `setUpMapTimeout`. I'm guessing it didn't create any actual bugs. I think it would have been in a different scope.
* 0e83293707c71b5070c2e437d55c614451b3f061 Move almost everything out of the window load handler. I had accumulated a dependency of things inside of it that referred to other things defined inside of it. I thought it would be complicated to take it out. Turns out if I just take them all out at once it's fine. I just have two functions calls in there now.
* 7ba8cc46876fcdbc2d02fb10ea8ea246179b2dfc Strictly indentation. If you add ?w=1 to the end of the URL it'll tell you it's whitespace only changes.

The benefit of this is that now I can organize my code a little better. Hopefully make it easier for others to follow. Perhaps it'll simplify some future feature implementations.

My next step in cleanup was going to be to add a map.on('load'...) handler for maps.black to get rid of some timeout loops I have, but that's looking to be a lot more of a mess than I expected so I may put that off.

RasPi Trixie

@chapmanjacobd

### DIFF
--- a/roles/maps/templates/index.html.j2
+++ b/roles/maps/templates/index.html.j2
@@ -385,595 +385,595 @@
                     }
                 })
 
-                function setStyle() {
-                    // The first level key is the `mapStyleChoice`, i.e. "natural", "political", etc, from the dropdown menu. (These are names specific to
-                    // IIAB maps.)
-                    //
-                    // The second level key is a boolean indicating whether the currently visible map is using OSM or Natural Earth data for vector data.
-                    //
-                    // Both of these together point to a maps.black "mapstyle" which implies a maplibre style and data sources.
-                    mb.mapstyle = {
-                        // Has nicely colored regions, and buildings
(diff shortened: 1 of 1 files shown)


holta:
Thanks @orblivion!  And thank you to @chapmanjacobd if you have time to review this in coming days?

chapmanjacobd (approved):
thanks for putting the indent changes to a separate commit

though it is pretty easy to see via this anyway:

    git format-patch --stdout --first-parent --no-merges origin/master...HEAD --ignore-all-space

PR #4246, opened 2026-01-24
[proot] enable php-fpm via nginx role.

Ark74:
### Fixes bug:
php-fpm is not explicitly enabled in the nginx role. In proot-distro, this leaves php-fpm disabled by default.

This PR explicitly enables php-fpm after core packages are installed.

On systemd-based environments this isn't needed because services are enabled automatically on install. In proot, that doesn't happen. Since the nginx role lacks a php-fpm entry in enable-or-disable.yml, php-fpm remained disabled and was only running on the initial run due to a restart.

The fix adds an enable step alongside the service restart, ensuring php-fpm stays enabled by default, so roles like Matomo work correctly after a "reboot".

Debian 13 (Android)

@holta @chapmanjacobd

### DIFF
--- a/roles/nginx/tasks/install.yml
+++ b/roles/nginx/tasks/install.yml
@@ -71,8 +71,7 @@
   when: not is_proot
 
 - name: Enable & Restart recently installed 'php-fpm' via pdsm (proot)
-  when:
-    - is_proot
+  when: is_proot
   block:
     - name: Enable 'php-fpm' via pdsm (proot)
       command: pdsm enable php-fpm
--- a/roles/nginx/tasks/install.yml
+++ b/roles/nginx/tasks/install.yml
@@ -70,9 +70,15 @@
     state: restarted
   when: not is_proot
 
-- name: Restart 'php-fpm' via pdsm (proot)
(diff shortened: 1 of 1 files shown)


chapmanjacobd (approved):

roles/nginx/tasks/install.yml
```diff
@@ -70,9 +70,15 @@
     state: restarted
   when: not is_proot
 
-- name: Restart 'php-fpm' via pdsm (proot)
-  command: pdsm restart php-fpm
-  when: is_proot
+- name: Enable & Restart recently installed 'php-fpm' via pdsm (proot)
+  when:
+    - is_proot
```
holta: Can we keep simple `when:` clauses on a single line?

```suggestion
  when: is_proot
```

Ark74:
I see some weird behavior with kolibri as it doesn't finish the installation started. It gets enabled so after "reboot" it works. So it may be related to this kind of issues. But I need to look closer as it doesn't seem explicit the reason of such behavior.

Once that is clean I think we are good with any weird behavior on proot_services.

It's getting to an amazing level of usability :)

PR #4249, opened 2026-01-27
[proot] disable sudo's use_pty for tty consistency

Ark74:
### Fixes bug:
Fixes interactive prompts inside proot-distro ran with "sudo" (root non-sudo commands are not affected).

Disable sudo calling pseudo tty, preventing tty mismatch which breaks interactive prompts like the use of:

```
/opt/iiab/maps/tile-extract/tile-extract.py

New regions will be XX MB downloaded, YY MB on disk. This will
leave about ZZ GB free space on this partition. Continue? y/n
```

Debian (proot)

@holta @chapmanjacobd

### DIFF
--- a/roles/proot_services/tasks/main.yml
+++ b/roles/proot_services/tasks/main.yml
@@ -18,9 +18,6 @@
       include_tasks: install.yml
       when: proot_services_installed is undefined
 
-    - name: Disable sudo use_pty in proot (fix interactive prompts)
-      include_tasks: no_use_pty_sudo.yml
-
     - name: Enable/Disable/Reload NGINX for proot_services, if nginx_enabled
       include_tasks: nginx.yml
 
--- a/roles/proot_services/tasks/main.yml
+++ b/roles/proot_services/tasks/main.yml
@@ -16,7 +16,10 @@
 
     - name: Install proot_services (pdsm) if 'proot_services_installed' not defined, e.g. in {{ iiab_state_file }}    # /etc/iiab/iiab_state.yml
       include_tasks: install.yml
-      when: proot_services_installed is undefined and is_proot
+      when: proot_services_installed is undefined
+
+    - name: Disable sudo use_pty in proot (fix interactive prompts)
+      include_tasks: no_use_pty_sudo.yml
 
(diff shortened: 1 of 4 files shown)


Ark74:
You might need to test:

```
maps_region_downloader: True
```

without this PR to actually see sudo's problematic prompt behavior on Android.

roles/proot_services/tasks/main.yml
```diff
@@ -16,7 +16,10 @@
 
     - name: Install proot_services (pdsm) if 'proot_services_installed' not defined, e.g. in {{ iiab_state_file }}    # /etc/iiab/iiab_state.yml
       include_tasks: install.yml
-      when: proot_services_installed is undefined and is_proot
+      when: proot_services_installed is undefined
+
+    - name: Disable sudo use_pty in proot (fix interactive prompts)
```
holta: @Ark74

I'd prefer it if Lines 21-22 can be placed within install.yml ?

(install.yml should generally contain "all" steps, if at all possible!)

chapmanjacobd:
Interesting! 

Other than Adam's comment, this PR looks good to merge

perhaps one alternative could be

    yes | sudo -S /opt/iiab/maps/tile-extract/tile-extract.py ...

Ark74:
> `yes | sudo -S /opt/iiab/maps/tile-extract/tile-extract.py ...`

I suppose, but the idea is to keep Android and "PC" (GNU/Linux) working on the same line and according to the main instructions, I think that means being able to choose at the prompt.

I mean for 
```
sudo /opt/iiab/maps/tile-extract/tile-extract.py extract test -97.743111793,20.257032292,-97.300938209,21.591739125
```
on proot-distro we can remove `sudo` and that would have worked correctly.

```
/opt/iiab/maps/tile-extract/tile-extract.py extract test -97.743111793,20.257032292,-97.300938209,21.591739125
```

But as mentioned before the idea is that both environments can follow along with the expected workflow.

Screenshot_20260127-105727_Fennec

roles/proot_services/tasks/main.yml
Ark74: Done :+1:

chapmanjacobd (approved):

PR #4251, opened 2026-01-29
Clean NGINX stuff in roles/maps and roles/osm-vector-maps

holta:

- Refines PR #4243 to use `maps-nginx.conf` across both the old roles/osm-vector-maps and the new roles/maps.
- Brings both maps roles in line with IIAB norms (enable-or-disable.yml).

Tested http://box/maps successfully in both directions(*) on Ubuntu 24.04

(*) Installing roles/osm-vectors-maps on top or roles/maps — and then installing (actually just re-enabling!) roles/maps on top of roles/osm-vector-maps

@orblivion


roles/maps/tasks/enable-or-disable.yml
```diff
@@ -1,21 +1,23 @@
...
-# 2026-01-23: We renamed the config file to `osm-vector-maps-nginx.conf` so that one can install
-# old maps or new maps -- even if the other is already installed: "most recently installed wins"
-- name: Remove previous config at {{ nginx_conf_dir }}/maps-nginx.conf
-  file:
-    path: "{{ nginx_conf_dir }}/maps-nginx.conf"
-    state: absent
+
+# 2026-01-28: We renamed the NGINX config file to `maps-nginx.conf` in roles/osm-vector-maps as well, so that
+# one can install old maps or new maps -- even if the other is already installed: "most recently installed wins"
+# - name: Remove previous config at {{ nginx_conf_dir }}/osm-vector-maps-nginx.conf
```
orblivion, line 16: What's the intention with having this around but commented out?

holta: Goal is to highlight the more active file deletion (and equivalent advisory) over in roles/osm-vector-maps/tasks/enable-or-disable.yml

Of course both can go away within a year or so.  But an explanation in both places seems wise for now.

orblivion: Having an explanation makes sense. But I'm wondering about the file deletion. You have it in the other role but it's commented out here.

holta: CLARIF: I'd rather not slown down here. to delete `/etc/nginx/conf.d/osm-vector-maps-nginx.conf` as I don't think that buys us much.

However that bit of slowness is a good precaution over in our older roles/osm-vector-maps/tasks/enable-or-disable.yml (even if all of roles/osm-vector-maps will probably be going away at some point, likely well within a year or so).

orblivion (approved):

PR #4254, opened 2026-02-03
Add statically hosted search

orblivion:

Add a new option for searching, which is to host the search index in a bunch of json files. This works just fine on a Raspberry Pi Zero 2W, which was the goal here. Nominatim on that hardware is a bit sluggish at best. Inspired by (and code copied from) another project, which is attributed here.

I have a list of things to clean up after this, including the fact that some cities are missing due to de-duping problems. That's going to be dealt with by fixing the database generation, probably not needing new code here. But I also have some ideas for sorting and finding sub-3-letter city names (if they exist).

I've tried this on a much bigger data set as well. It works fine so long as none of the json files are too massive. I think I can pull it off, it'll just take some more optimization.

BTW if you want to see what I used to generate the database: https://github.com/iiab/maps2/blob/master/build-databases/static/import-places.py I actually copied this from a different project, but it's my own project from a couple years ago. (So I don't think I need to attribute it?)

Trixie RasPi

@holta @chapmanjacobd

### DIFF
--- a/roles/maps/README.md
+++ b/roles/maps/README.md
@@ -55,22 +55,35 @@ To add terrain files, you can set this optional setting. You may find that when 
 
 ## Can I try out search (which is still experimental)?
 
-This is not recommended for very low-power devices such as Raspberry Pi [Zero 2 W](https://www.raspberrypi.com/products/raspberry-pi-zero-2-w/), though this might change.
+### Low-power search
 
-As of this writing, search includes only administrative regions and natural features.
+This option is good for all devices. Fast and simple, but limited features.
+
+This is a list of all cities with population 1000 or higher (**~35MB**)
+
+   ```
+   maps_search_engine: "static"
+   maps_search_static_db: "pop-1k-cities"
+   ```
+
+### High-power search
+
+These options are not recommended for very low-power devices such as Raspberry Pi [Zero 2 W](https://www.raspberrypi.com/products/raspberry-pi-zero-2-w/), though this might change.
+
+As of this writing it includes only administrative (i.e. political) regions and natural features.
 
 1. If you want **~640 MB** "small" (only California, as of this writing) search:
 
    ```
    maps_search_engine: "nominatim"
-   maps_search_full: False
+   maps_search_nominatim_db: "basic"
    ```
 
 2. If you want **~67 GB** "full" (planet-wide) search:
 
    ```
    maps_search_engine: "nominatim"
-   maps_search_full: True
+   maps_search_nominatim_db: "full"
    ```
 
 ## Full Quality Regions (experimental)
@@ -155,5 +168,7 @@ For these large file downloads:
   * https://github.com/maplibre/maplibre-gl-js
   * https://github.com/maplibre/maplibre-gl-geocoder?tab=ISC-1-ov-file#readme
   * https://github.com/watergis/maplibre-gl-terradraw
-* Search (for now): https://github.com/osm-search/Nominatim?tab=GPL-3.0-1-ov-file#readme
+* Search (for now):
+  * https://github.com/osm-search/Nominatim?tab=GPL-3.0-1-ov-file#readme
+  * https://github.com/jacopofar/static-osm-indexer/?tab=readme-ov-file#licensing-anc-crediting (some pieces and inspiration taken from this project)
 * Other credits: https://github.com/iiab/iiab/blob/master/roles/www_base/files/html/html/credits.html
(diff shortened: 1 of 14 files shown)


chapmanjacobd (approved):
nice and clean

chapmanjacobd:
re: deduping city names... geonames is kinda trash but I wouldn't dedupe things. just filter out low population cities instead

Maybe add a buffer of 1km between city names which share 70% similarity in the name

https://gist.github.com/chapmanjacobd/fba001fc5c45874455f28db0fbebe2e2

<details>
<summary>maybe like this?</summary>

```py
import argparse
import pandas as pd
import geopandas as gpd
from rapidfuzz import fuzz
from shapely.geometry import Point

def deduplicate_spatial_fuzzy(args):
    df = pd.read_csv(args.input)
    
    # Initialize GeoDataFrame
    gdf = gpd.GeoDataFrame(
        df, 
        geometry=gpd.points_from_xy(df['X'], df['Y']), 
        crs="EPSG:4326"
    )

    # Project to EASE-Grid 2.0 for consistent metric distance
    gdf_projected = gdf.to_crs("EPSG:6933")

    # Spatial join to find neighbors within the metric radius
    joined = gpd.sjoin_nearest(
        gdf_projected, 
        gdf_projected, 
        max_distance=args.max_dist, 
        distance_col="distance_m",
        lsuffix="left", 
        rsuffix="right"
    )

    # Filter out self-joins
    matches = joined[joined.index != joined['index_right']].copy()

    to_remove = set()
    for idx, row in matches.iterrows():
        if idx in to_remove or row['index_right'] in to_remove:
            continue
            
        # RapidFuzz comparison
        score = fuzz.ratio(str(row['name_left']), str(row['name_right']))
        
        if score >= args.threshold:
            # Keep the point with the lower index
            to_remove.add(max(idx, row['index_right']))

    # Drop flagged duplicates and export
    final_gdf = gdf.drop(index=list(to_remove))
    final_gdf.drop(columns='geometry').to_csv(args.output, index=False)

if __name__ == "__main__":
    parser = argparse.ArgumentParser()
    parser.add_argument('--input', type=str, default='~/test.csv')
    parser.add_argument('--output', type=str, default='~/output.csv')
    parser.add_argument('--max_dist', type=float, default=5000) # Meters
    parser.add_argument('--threshold', type=float, default=80.0) 
    
    args = parser.parse_args()
    deduplicate_spatial_fuzzy(args)
```

</details>

holta:
@orblivion please ASAP initialize to:

```
maps_search_engine: none
```

Or if you prefer:

```
maps_search_engine: "none"
```

Either way, make it consistent with others — e.g. `maps_terrain_zoom: none` as is, or `maps_terrain_zoom: "none"` if you prefer!

orblivion:
@chapmanjacobd 

> `gpd.sjoin_nearest`

So this is what a geo-enabled database is for.

I think we're talking about opposite but related problems. I think you're talking about the fact that searching for a city will return a couple nearby cities of the same name. And I did notice such things in the data. (There's also an issue with similar names in a cluster: Boston, East Boston, Boston Seaport, South Boston, VA Boston Healthcare System)

What I'm talking about is two actual different cities with the same name. So far I disambiguate by state/province and country. I'd like to [do the same for county](https://en.wikipedia.org/wiki/Franklin_Township,_Pennsylvania_(disambiguation)), and maybe more? (admin1 admin2 admin3 and admin4) = (state, county, ?, ?) in geonames parlance. Without enough disambiguation, I end up with exact same names.

The way I handle it right now, I actually just clobber such dupes, taking the one with the [higher population](https://github.com/iiab/maps2/blob/698292430961796e6ec02fa491191f785722f4bf/build-databases/static/import-places.py#L213). And that's the problem I was talking about: clobbering cities that I don't need to clobber.

With that in mind, maybe there are some cities I should continue clobber. I'll see how many are left after I do county etc. Or maybe if I use your pandas script I could find even more. It doesn't look like you're even checking for similar name though?

As for the bad quality, should we consider just using OSM data instead? This was just quicker.

chapmanjacobd:
What I've done before is if the city name is unique just use that, otherwise if the names are distinct within a country show Name (Country) otherwise use Name, State (Country)

Yeah I think just using OSM will be better

holta:
> Yeah I think just using OSM will be better

1. Will the same issues presumably arise with "Mexico" having at least 3 different meanings depending on context — e.g. city, state, country...

2. And if so, is the current approach of using the one with the highest population possibly clean enough / good enough / best for now?

3. I haven't yet this new search yet: should we offer some kind of very basic drop-down / pop-up menu showing you the ~10 most likely "Paris" contenders — with state/population shown for each — to help you find the one you want?  Not sophisticated, and not beautiful, but FWIW that's how IIAB Maps' search has worked to date and certainly it works until we come up with something better ;)

chapmanjacobd:
Yes it looks like OSM uses "Mexico City" rather than "Mexico" which seems more proper--it's "Ciudad de Mexico" not "Ciudad Mexico". Geonames is wrong here.

> highest population

I've done something before where you could have different icons depending on city size. Classification like that (single integer) doesn't add much extra bytes to each of the JSON files.

⛺ 🏡 🏘️ 🏢 🏙️ 🌃

PR #4255, opened 2026-02-03
Proper default value for map search engine

orblivion:
### Fixes bug:

Using an empty string for a default "none" value.

### Smoke-tested on which OS or OS's:

RasPi Trixie

### Mention a team member @username e.g. to help with code review:
@holta

PR #4256, opened 2026-02-04
Static search: add to marker popup

orblivion:
A couple small items to keep PRs small.

* Add the population to the search result marker popup. Right now it just shows the city name and state and country.
* Embolden the city name, since there's a few items there now.
* Make sorting by population more proper by actually subtracting numbers, instead of strings (which Javascript treats as numbers anyway, but come on).

RasPi Trixie

@chapmanjacobd

### DIFF
--- a/roles/maps/templates/index.html.j2
+++ b/roles/maps/templates/index.html.j2
@@ -707,7 +707,7 @@
                                 },
                                 place_name: result.name,
                                 properties: {
-                                    population: result.pop,
+                                    population: Number(result.pop),
                                 },
                                 text: result.name,
                                 place_type: ['place'],
@@ -760,7 +760,7 @@
                                     ${state_and_country_name}
                                 </div>
                                 <div>
-                                    Population: ${feature.properties.population}
+                                    Population: ${feature.properties.population.toLocaleString()}
                                 </div>
                             </div>
                           `
--- a/roles/maps/templates/index.html.j2
+++ b/roles/maps/templates/index.html.j2
@@ -706,7 +706,9 @@
                                     coordinates: center
                                 },
                                 place_name: result.name,
-                                properties: [],
+                                properties: {
+                                    population: result.pop,
+                                },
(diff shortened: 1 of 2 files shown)


holta:
(Building on...)

- PR #4254 
- PR #4255

roles/maps/templates/index.html.j2
```diff
@@ -743,6 +745,26 @@
                 static: {
                     api: staticGeocoderApi,
                     options: {
+                        popupRender: feature => {
+                          // Hopefully similar enough to the default popup for search results
```
chapmanjacobd: why is hope needed? Inspect element? HTML diff, [image diff](https://github.com/MozillaFoundation/ci-image-diff)??

I don't personally need those things but I find this a strange comment

roles/maps/templates/index.html.j2
```diff
@@ -743,6 +745,26 @@
...
+
+                          // Not ideal to assume that comma means this but I suspect maplibre does this too
+                          const [city_name, ...rest] = feature.place_name.split(',')
```
chapmanjacobd, line 750: You will likely find this either unnecessary or inadequate when moving to OSM--best to put all the logic for building the names into a single spot than making it work well for Geonames and then forcing OSM into that format to make it work with this....

but I'm not against merging this now

chapmanjacobd (approved):

roles/maps/templates/index.html.j2
orblivion: I inspected the element and copied what it had. The hope I guess is more about how it would handle different sorts of input. Not that it matters that much. And it'll diverge a lot over time.

roles/maps/templates/index.html.j2
orblivion: I don't even need to go as far as OSM. I can put it in different fields from geonames. It's on my list.

PR #4260, opened 2026-02-06
Deb822 format for apt repos [for Kolibri, MongoDB, Node.js, Yarn — including interim Sequoia PGP SHA-1 workaround for Kolibri & MongoDB on Debian & RPiOS Trixie]

jvonau:
### Fixes bug:
#4253
### Description of changes proposed in this pull request:
change to deb822 format for apt repos - does the same as `apt modernize-sources` 
https://share.google/aimode/fzTIprU1NSpnTNyig
### Smoke-tested on which OS or OS's:
CI & RasPiOS pending
### Mention a team member @username e.g. to help with code review:
@Ark74 @chapmanjacobd

PR #4265, opened 2026-02-09
Modernize Tailscale apt key / install routine, using deb822-formatted /etc/apt/sources.d.list/tailscale.sources

holta:
### Fixes bug:

Smoother operation on modern OS's like Debian Trixie 13+ and RPiOS Trixie 13+.

For consistency, does not use https://docs.ansible.com/projects/ansible/latest/collections/ansible/builtin/deb822_repository_module.html, using Ansible templates to drop the entire `/etc/apt/sources.d.list/tailscale.sources` into place, in keeping with:

- PR #4260

Also refines minor typos in above PR, and with more consistent file naming.

Debian 13 Trixie.  And ongoing CI workflows on other OS's to help confirm.

@jvonau @Ark74 @chapmanjacobd


chapmanjacobd (approved):

jvonau:
I did a deeper dive on tailscale there is more at play than just the curl of the keys https://github.com/iiab/iiab/compare/master...jvonau:deb822?expand=1 info in the notes.

holta:
> I did a deeper dive on tailscale there is more at play than just the curl of the keys https://github.com/iiab/iiab/compare/master...jvonau:deb822?expand=1 info in the notes.

Interesting.  Thanks for catching Tailscale's packaging weirdness.  The dark side being that Tailscale is no doubt going to mess around with its install recommendations in the coming months, as they catch up with Deb822 recommendation / requirement for modern OS's — e.g. for Debian 13 and Ubuntu 26.04 especially:

- https://github.com/iiab/iiab/compare/master...jvonau:deb822 is great to have figured out what Tailscale's packaging is up to.  But also (like everything here!?) potentially getting too far ahead of our/their skis?  Aside: [Line 19 there](https://github.com/iiab/iiab/blob/7e39fc7d339397b4b05f4be32f11a9a098feaeae/roles/tailscale/tasks/install.yml#L19) looks like it should contain `is version('13', '>=')` rather than `is search('13', '>=')`

- Not sure if the `apt -y modernize-sources` interim workaround is indeed best??  Another tempting option might possibly to leave `/usr/share/keyrings/tailscale-archive-keyring.gpg` in place "and change as little as possible" (ideally ;) until Tailscale makes up their mind as to their [official recommendation for manual installs](https://pkgs.tailscale.com/stable/#debian-trixie) as we monitor this in coming months.

jvonau:
> > I did a deeper dive on tailscale there is more at play than just the curl of the keys https://github.com/iiab/iiab/compare/master...jvonau:deb822?expand=1 info in the notes.
> 
> Interesting. Thanks for catching Tailscale's packaging weirdness. The dark side being that Tailscale is no doubt going to mess around with its install recommendations in the coming months, as they catch up with Deb822 recommendation / requirement for modern OS's — e.g. for Debian 13 and Ubuntu 26.04 especially:
> 
>     * [master...jvonau:deb822](https://github.com/iiab/iiab/compare/master...jvonau:deb822) is great to have figured out what Tailscale's packaing is up to.  But also (like everything here!?) potentially getting too far ahead of our/their skis?  Aside: [Line 19 there](https://github.com/iiab/iiab/blob/7e39fc7d339397b4b05f4be32f11a9a098feaeae/roles/tailscale/tasks/install.yml#L19) looks like it should contain `is version('13', '>=')` rather than `is search('13', '>=')`
>

I wanted to cover both debian-13 and rasbian-13 without making the when clause overly long winded. That does the right thing when run on the debian 12 and raspbian 13 CI test builds.
 >     * Not sure if the `apt -y modernize-sources` interim workaround is indeed best??  Another tempting option might possibly to leave `/usr/share/keyrings/tailscale-archive-keyring.gpg` in place "and change as little as possible" (ideally ;) until Tailscale makes up their mind as to their [official recommendation for manual installs](https://pkgs.tailscale.com/stable/#debian-trixie) as we monitor this in coming months.

I'd be in favor of using /usr/share/keyring for the key location and a revised repo template, adding tailscale-archive-keyring is defensive should the default behavior is change from including Recommends.
https://forums.raspberrypi.com/viewtopic.php?t=363521
https://unix.stackexchange.com/questions/447952/how-do-recommends-and-suggests-interact-with-apt-get-dist-upgrade-and-apt-get-au

jvonau:
https://tailscale.com/download/linux
#4096 could be used with 'modernize-sources'

If you go with the template then:
Suites: noble 
could become
Suites: "{{ ansible_distribution_release }}"

and use "{{ ansible_distribution_release }}" to grab the matching gpg key.

roles/tailscale/tasks/install.yml
```diff
@@ -3,10 +3,13 @@
   register: df1
 
 
-- name: "Set up apt source (jammy) in /etc/apt/sources.list.d/tailscale.list and its key /usr/share/keyrings/tailscale-archive-keyring.gpg, to install Tailscale"
-  shell: |
-    curl -fsSL https://pkgs.tailscale.com/stable/ubuntu/jammy.noarmor.gpg > /usr/share/keyrings/tailscale-archive-keyring.gpg
-    curl -fsSL https://pkgs.tailscale.com/stable/ubuntu/jammy.tailscale-keyring.list > /etc/apt/sources.list.d/tailscale.list
+- name: Download signing key to /etc/apt/keyrings/tailscale.gpg, to install Tailscale
+  shell: curl -fsSL https://pkgs.tailscale.com/stable/ubuntu/noble.noarmor.gpg > /etc/apt/keyrings/tailscale.gpg
+
```
jvonau, line 28: shell: curl -fsSL https://pkgs.tailscale.com/stable/{{ ansible_facts.distribution }}/{{ ansible_facts.distribution_release }}.noarmor.gpg > /usr/share/keyrings/tailscale-archive-keyring.gpg

roles/tailscale/templates/tailscale.sources
```diff
@@ -0,0 +1,5 @@
+Types: deb
+URIs: https://pkgs.tailscale.com/stable/ubuntu/
```
jvonau, line 2: URIs: https://pkgs.tailscale.com/stable/{{ ansible_facts.distribution }}

roles/tailscale/templates/tailscale.sources
```diff
@@ -0,0 +1,5 @@
...
+Suites: noble
```
jvonau, line 3: Suites: "{{ ansible_facts.distribution_release }}"

roles/tailscale/templates/tailscale.sources
```diff
@@ -0,0 +1,5 @@
...
+Components: main
+Signed-By: /etc/apt/keyrings/tailscale.gpg
```
jvonau: Signed-by: /usr/share/keyrings/tailscale-archive-keyring.gpg

roles/tailscale/templates/tailscale.sources
holta: Good suggestions all around (I'll try to test later), but this one however introduces risk, as Tailscale isn't always ready to support recent distros when people need them.

Hard-coding `ubuntu` and `noble` (and updating to the latest LTS codename/suite every few years) is probably more robust.

holta:
@jvonau No worries at all, but my point earlier was just that `is search('13', '>=')` is incorrect syntax according to:

https://docs.ansible.com/projects/ansible/latest/collections/ansible/builtin/search_test.html

jvonau:
> @jvonau No worries at all, but my point earlier was just that `is search('13', '>=')` is incorrect syntax according to:
> 
> https://docs.ansible.com/projects/ansible/latest/collections/ansible/builtin/search_test.html

Would have to wait for debian-14 to test that theory, but it worked for today. ;)

This looks like a fair compromise: 
 - moves to noble
 - deb822 format
 - didn't alter upstream's key location.

+1

PR #4266, opened 2026-02-09
maps: search bar no overlap with buttons

orblivion:
### Fixes bug:

Search bar, which is a maplibre control set on the "top-left", overlaps controls on the "top-right" when using on a small enough phone vertically.

Just set a max-width on the search bar.

overlapping
fixed

@chapmanjacobd suggested solving this by sliding down the controls on the top-right for thin screens to make room for the search bar. Just aesthetically I just prefer setting the max width to stop it from overlapping. On my own phone, and every test phone in Firefox's dev tools, it's really not that much of an overlap in the first place. It seems like shrinking the width loses 1-3 characters. I can still type more characters than on my desktop or on my phone in horizontal mode, where the search bar shrinks (maplibre's own responsive design choice).

Now, if there are phones that I'm not considering which are actually a lot smaller and that our users might use, then yeah I agree with setting a rule to move the controls down to make room for it. @holta what do you think?

RasPi Trixie

@chapmanjacobd

### DIFF
--- a/roles/maps/templates/index.html.j2
+++ b/roles/maps/templates/index.html.j2
@@ -1023,6 +1023,14 @@
                     const stylesheetText = [...mapLibreGlGeocoderCSS.cssRules, ...mapLibreGlTerradrawCSS.cssRules].reduce((accumulator, rule) => accumulator + rule.cssText, '');
                     const myStyles = new CSSStyleSheet()
                     myStyles.replace(stylesheetText)
+                    .then(() => {
+                        myStyles.insertRule(`
+                            /* Ensure the search bar doesn't stretch to hit the zoom buttons */
+                            .maplibregl-ctrl-top-left {
+                                max-width: calc(100% - 60px);
+                            }
+                        `)
+                    })
                     mb.shadowRoot.adoptedStyleSheets = [...mb.shadowRoot.adoptedStyleSheets, myStyles]
 
                     if (geocoder) {


holta:
> Now, if there are phones that I'm not considering which are actually a lot smaller and that our users might use, then yeah I agree with setting a rule to move the controls down to make room for it. @holta what do you think?

I'm ok with this!  Unless @chapmanjacobd or others have a better idea?

chapmanjacobd (approved):
wow! very tidy

I tried a bunch of things before but didn't think to try something like this

I agree it looks a bit better than jumping up and down

holta:
@orblivion is this supposed to work... ?  (Or am I pushing too hard maybe, by asking it to render on a low-income person's 2018 phone?)

image

Routine: Chrome browser -> https://lrn2.org/maps -> Ctrl-Shift-M (render as if mobile) -> Dimensions -> Edit... -> enable "JioPhone 2" (and then go back to "Dimensions" tab, to ask it to render in a JioPhone 2)

roles/maps/templates/index.html.j2
```diff
@@ -1023,6 +1023,14 @@
                     const stylesheetText = [...mapLibreGlGeocoderCSS.cssRules, ...mapLibreGlTerradrawCSS.cssRules].reduce((accumulator, rule) => accumulator + rule.cssText, '');
                     const myStyles = new CSSStyleSheet()
                     myStyles.replace(stylesheetText)
+                    .then(() => {
+                        myStyles.insertRule(`
+                            /* Ensure the search bar doesn't stretch to hit the zoom buttons */
+                            .maplibregl-ctrl-top-left {
+                                max-width: calc(100% - 60px);
```
chapmanjacobd, line 1030: maybe instead of 60px you could use 5em

chapmanjacobd: https://refine.dev/blog/rem-vs-em/#what-is-em-and-rem-in-css

which might help with Holta's example above

or

https://old.reddit.com/r/css/comments/s299ta/understanding_viewport_width_vw_versus_percentage/

> render in a JioPhone 2

You might need to refresh the page to get accurate 100% CSS calc (depending on how they implemented the Chrome Devices view)

orblivion: The width of `.maplibregl-ctrl-group button` is given in terms of px. I tried changing it with the browser inspect tool and it indeed makes the width of the controls change:

https://github.com/maplibre/maplibre-gl-js/blob/v5.7.3/src/css/maplibre-gl.css#L131

Likewise the right margin of `.maplibregl-ctrl-top-right .maplibregl-ctrl`:

https://github.com/maplibre/maplibre-gl-js/blob/v5.7.3/src/css/maplibre-gl.css#L101

So I think we should stick with px to make sure it universally matches.

orblivion:
@holt Okay this JioPhone 2 looks like an example of the kind of hypothetical small phone that I was talking about. I don't see JioPhone 2 on my Firefox so I missed it. (I don't see it on Chromium now though either. Funny enough I do see Pixel 7 and it looks different from my own Pixel 7.)

But yes when I was testing I saw the same thing you do if I pushed it below 350 px or so. If that's a real phone you think our users may be using, no worries I could stick in @chapmanjacobd 's fix somewhere around 400 px. I think I already tried it and it works.

roles/maps/templates/index.html.j2
orblivion: Regarding unit issues with @holta 's example above: I've seen the same thing at around the same number of px. There is a threshold beyond which the max-width seems to lose the battle with some other setting on a element in the search bar that has a width or min-width, so the overlap happens despite my max-width.

holt:
@orblivion I think you might have the wrong Holt 🙂

> @holt Okay this JioPhone 2 looks like an example of the kind of hypothetical small phone that I was talking about.

orblivion:
Sorry about that!

@holta ?

holta:
@orblivion @chapmanjacobd small/old phones "by the millions" [*] are definitely relevant — if further progress turns out to be possible?  (Should we wait for a subsequent PR to refine that?)

[*] A small number (!) if almost 6 Billion people use smartphones, out of 8.3 Billion, if we believe: https://www.demandsage.com/smartphone-usage-statistics/

holta:
Many thanks to @orblivion, who also appears to be working on a follow-up for tiny screens!

PR #4268, opened 2026-02-10
maps: handle search bar arrangement on very narrow phones

orblivion:
### Fixes bug:

On very narrow phones, the search bar overlaps the top-right controls. My previous fix #4266 was only good for larger phones.

This time, I implement @chapmanjacobd 's initial suggestion of sliding the top-right controls down to make room for the search bar, and I also make the search bar hug the right side of the screen to make it look cleaner. But I only do this for particularly narrow phones. For larger phones, I still think it looks nicer to squish the search bar.

I made the "larger phone" range rather small, but it applies to my phone. I think I could have made the transition closer to 300px-wide screens instead of 350px, but at that point the search bar is getting squished a fair amount and I'd rather give more room for people to type.

### DIFF
--- a/roles/maps/templates/index.html.j2
+++ b/roles/maps/templates/index.html.j2
@@ -1024,43 +1024,45 @@
                     const myStyles = new CSSStyleSheet()
                     myStyles.replace(stylesheetText)
                     .then(() => {
-                        /* If the screen gets narrow enough (such as with vertically oriented phones
-                           that are especially low resolution) setting max-width on the search bar isn't
-                           enough to stop it from overlapping the top-right controls. And anyway, we
-                           don't want it to get too small. Instead, we move the top-right controls down
-                           and let the search bar take up the whole top of the screen. */
-                        myStyles.insertRule(`
-                            @media screen and (max-width:349px) {
-                                .maplibregl-ctrl-top-right {
-                                    top: 60px !important;
-                                    transition: top 0.25s ease;
+                        if (geocoder) {
+                            /* If the screen gets narrow enough (such as with vertically oriented phones
+                               that are especially low resolution) setting max-width on the search bar isn't
+                               enough to stop it from overlapping the top-right controls. And anyway, we
+                               don't want it to get too small. Instead, we move the top-right controls down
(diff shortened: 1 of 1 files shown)


roles/maps/templates/index.html.j2
```diff
@@ -1024,10 +1024,37 @@
...
+                           don't want it to get too small. Instead, we move the top-right controls down
+                           and let the search bar take up the whole top of the screen. */
                         myStyles.insertRule(`
-                            /* Ensure the search bar doesn't stretch to hit the zoom buttons */
-                            .maplibregl-ctrl-top-left {
-                                max-width: calc(100% - 60px);
+                            @media screen and (max-width:349px) {
+                                .maplibregl-ctrl-top-right {
+                                    top: 5em !important;
+                                    transition: top 0.2s ease;
```
orblivion: 1) When I load the page on my laptop browser with a narrow test phone screen, the transition happens right away. Is this what we want? @chapmanjacobd I'm guessing you put this here for the user rotating their screen?

2) I didn't put the `top: 0` in the other media clause. As a result the transition only happens one way. If we have the transition we should do both I guess.

3) I realized I did 5em. I should figure out whether the search bar is measured in em or px.

chapmanjacobd: The default is top: 0 so it's probably fine without it in the other media query

I would increase the transition to 0.3s or whatever value the geocoder is using to match as it looks like it's taking longer when I try rotating my phone

orblivion: > The default is top: 0 so it's probably fine without it in the other media query

Again the purpose is to apply the transition to the other direction.

> whatever value the geocoder is using

Wait, using for what? Delay between stopping typing and searching?

chapmanjacobd: The geocoder/searchbar also has a transition on it when I rotate my phone

You're right that the other media query also needs that transition property (or just pull it outside of a media query)

holta:
Good to merge?!

holta:
Very awesome, thanks @orblivion & @chapmanjacobd !

PR #4271, opened 2026-02-13
Switch to iiab.switnet.org for satellite imagery

orblivion:

Download satellite imagery from switnet.org instead of maps.black, as we do for other tile data.

RasPi Trixie (Though I only full tested the regional download. For downloading the full new region I got as far as running out of space on my SD card.)

@chapmanjacobd @holta

### DIFF
--- a/roles/maps/README.md
+++ b/roles/maps/README.md
@@ -29,6 +29,8 @@ To configure your map, set the following variables (for the option your choose!)
    maps_satellite_zoom: 12
    ```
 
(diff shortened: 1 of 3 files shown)


chapmanjacobd (approved):

roles/maps/README.md
```diff
@@ -29,6 +29,8 @@ To configure your map, set the following variables (for the option your choose!)
    maps_satellite_zoom: 12
    ```
 
+See `maps_dot_black_vector_tiles` and `maps_dot_black_satellite_tiles` [here](https://github.com/iiab/iiab/blob/master/roles/maps/defaults/main.yml) for all valid values.
```
orblivion, line 32: I just added these. I had in my notes that you wanted the new option in the "fine print".

PR #4274, opened 2026-02-17
maps: don't create history every time you scroll

orblivion:
### Fixes bug:

Right now every time you move the map, it updates the "hash" so that reloading or sharing the page will leave you where you left off. However the way it was implemented, it adds to the browser history each time.

Entering the 21st century and using something called `history.replaceState(...)` instead of `window.location.hash = ...`

I'd be curious if there are unseen downsides to this. Do people actually want this in their history sometimes?

@holta Whether the /home/ link points to `/maps` or `/maps/`, no matter how many times you move on the map, this change will make it a single "back button" click to take you back home from the map.

Raspi Trixie

@chapmanjacobd


chapmanjacobd (approved):

chapmanjacobd:
> Do people actually want this in their history sometimes?

Sometimes Google Maps does this--not with moving or zooming the map but depending on the search/place details sidebar. It's pretty annoying.

I think it will be best to keep it out of history. If people are curating a list of places they could use bookmarks even with this change

Issue #4130, opened 2025-11-02
The use of fake-hwclock


jvonau:
Upstream has [removed](https://github.com/RPi-Distro/pi-gen/commit/6811c1dc41b0eb71eefb9cf2beebbd18b9d1d06e) fake-hwclock from the trixie images it looks to be replaced with systemd-timesync. Confirmed missing from the image info file and with apt on an unbooted image. 

>jvonau@kickass:/mnt/scratch/git/mk-image$ sudo ./mount-image.sh ~/Downloads/2025-10-01-raspios-trixie-arm64-lite.img
[sudo] password for jvonau: 
/home/jvonau/Downloads/2025-10-01-raspios-trixie-arm64-lite.img
loop is /dev/loop11
using qemu
ENTERING CHROOT
root@kickass:/# apt list fake*
fake-hwclock/stable,stable 0.14 all
fake/stable 1.1.11-3.1+b1 arm64
fake/stable 1.1.11-3.1 armhf
fakechroot/stable,stable 2.20.1+ds-18 all
fakemachine/stable 0.0.11-1+b2 arm64
faker-doc/stable,stable 33.3.1-1 all
faker/stable,stable 33.3.1-1 all
fakeroot/stable,now 1.37.1.1-1 arm64 [installed,automatic]
fakeroot/stable 1.37.1.1-1 armhf
faketime/stable 0.9.10+2024-06-05+gba9ed5b2-0.6 arm64
faketime/stable 0.9.10+2024-06-05+gba9ed5b2-0.6 armhf

We [install](https://github.com/iiab/iiab/blob/master/roles/1-prep/tasks/raspberry_pi.yml#L25-L28) fake-hwclock but the service doesn't start as it is masked
jvonau@pi500:~ $ `sudo systemctl status fake-hwclock`
>○ fake-hwclock.service
     Loaded: masked (Reason: Unit fake-hwclock.service is masked.)
     Active: inactive (dead)
jvonau@pi500:~ $ systemctl status systemd-timesync 

The file mentioned in the commit above exists but has 0 bytes
jvonau@pi500:~ $ `ls -la /var/lib/systemd/timesync`
>total 8
drwxr-xr-x  2 systemd-timesync systemd-timesync 4096 Oct 28 20:58 .
drwxr-xr-x 13 root             root             4096 Sep  9 03:38 ..
-rw-r--r--  1 systemd-timesync systemd-timesync    0 Nov  2 07:54 clock

[...]

Issue #4168, opened 2025-12-06
MediaWiki failing due to MariaDB (MySQL) failure to start on Raspberry Pi OS 2025-11-24 and 2025-12-04 [/var/lib/mysql/tc.log corruption triggered by boot time collision with NTP-like systemd-timesyncd & fake-hwclock-load.service ?]


holta:
MariaDB fails to launch on every boot.  OS is 64-bit RPiOS 2025-11-24 (LITE version) on RPi 5.  Needs investigation — I'll take a look in coming days.  Possibly the new RPiOS 2025-12-04 might be different (hypothetically, or maybe not!)  Ok, just a few basic diagnostics for now:

```
root@box:~# systemctl status mysql
× mariadb.service - MariaDB 11.8.3 database server
     Loaded: loaded (/usr/lib/systemd/system/mariadb.service; enabled; preset: enabled)
     Active: failed (Result: exit-code) since Sat 2025-12-06 13:36:14 EST; 8min ago
 Invocation: 9790dd36c4c94dd88f0abc3b246d662c
       Docs: man:mariadbd(8)
             https://mariadb.com/kb/en/library/systemd/
    Process: 834 ExecStartPre=/usr/bin/install -m 755 -o mysql -g root -d /var/run/mysqld (code=exited, status=0/SUCCESS)
    Process: 849 ExecStartPre=/bin/sh -c [ ! -e /usr/bin/galera_recovery ] && VAR= ||   VAR=`/usr/bin/galera_recovery`; [ $? -eq 0 ]   && echo _WSREP_START_POSITION=$VAR > /run/mysqld/wsrep-start-position || exit 1 (code=exited, status=>
    Process: 906 ExecStart=/usr/sbin/mariadbd $MYSQLD_OPTS $_WSREP_NEW_CLUSTER $_WSREP_START_POSITION (code=exited, status=1/FAILURE)
   Main PID: 906 (code=exited, status=1/FAILURE)
     Status: "MariaDB server is down"
        CPU: 124ms

Dec 06 13:36:14 box mariadbd[906]: 2025-12-06 13:36:14 0 [Note] InnoDB: log sequence number 3629297; transaction id 321
Dec 06 13:36:14 box mariadbd[906]: 2025-12-06 13:36:14 0 [Note] InnoDB: Loading buffer pool(s) from /var/lib/mysql/ib_buffer_pool
[...]

PR #3141, opened 2022-03-16
1-prep -> 2-common: Consolidate, simplify and clarify packages.yml (16 apt packages) + network.yml (12 apt packages) making useful WiFi tools available to everyone

holta:
Another incremental step forward away from chaotic spaghetti dependencies &mdash; towards making IIAB's package dependencies modularized, approachable, understandable and maintainable.  Not the final word (a more comprehensive inventory reconciliation might makes sense in future) but this PR makes progress as follows:

- Penny-wise-but-Pound-foolish complexity around economizing the 300kB total download size of {iw, rfkill, wireless-tools} is put to an end.  Instead these 3 apt packages are made mandatory (this makes some vital tools available to any IIAB field operator who suddenly inserts a USB WiFi adapter, at any time in future, e.g. commands like `iw dev`, `iw list`, `iw reg get`, `iwlist wlan0 scan` etc).

- 11-or-12 core networking packages are consolidated and explained in roles/2-common/tasks/network.yml to make things far more understandable.

- Likewise, 16 non-networking packages are consolidated with cleaner explanations in the inventory/manifest at roles/2-common/tasks/packages.yml

- roles/1-prep/tasks/hardware.html now populates a _preliminary_ answer into variable `has_wifi_device` (this can change at any time, when a USB WiFi adapter is inserted or removed).  The idea being that this kind of variable (currently unused!) sets the stage for different meaningful purposes in future.  This code can of course move around to any other better place later, whenever that proves wise and needed.

This PR is mostly about in-line docs simplifying circuitous IIAB code and making it far more readable &mdash; with testing ongoing to avoid snafus on the way.  It builds on:

- PR #2885
- PR #2889 
- PR #2893
- PR #3138
- PR #3139

PR #4277, opened 2026-02-19
Drop fake-hwclock

jvonau:
### Fixes bug:
#4130 #4168 
drop hardware.yml and raspberry_pi.yml
Pending
- [hardware.yml](https://github.com/iiab/iiab/blob/master/roles/1-prep/tasks/hardware.yml) has dead code since internal-wifi.yml went away See: 
 #3139 
 #3141
- Nothing in [raspberry_pi.yml](https://github.com/iiab/iiab/blob/master/roles/1-prep/tasks/raspberry_pi.yml) is applicable to current trixie or ubuntu as i2c-rtc paths to the files to modify are now obsolete. 
Configuring add-on rtc chips should be a prerequisite prior to installation as there would be no one size fits all approach.


chapmanjacobd (approved):

Ark74:
I recall we found something related on 24.04, in Trisquel we don't ship timesyncd for privacy reasons. So hwclock was fairly important[1]. The issue we found is not precisely that hwclock was removed, but that util-linux got split into two packages.

* util-linux
* util-linux-extra

Leaving _hwclok_ on the extra one, now back to this issue.

I'm not really familiar with the mechanism to make an informed opinion, I would test the change and confirm if it solves the issue, then again I'm not in place right now to make a test on rpis sorry :sneezing_face: 

1 https://gitlab.trisquel.org/trisquel/package-helpers/-/issues/225

holta:
This PR looks extremely promising!  I hope to be able to help test it (with others) over the coming week — and if all's good this will solve even more than just:

- #4168

jvonau:
> I recall we found something related on 24.04, in Trisquel we don't ship timesyncd for privacy reasons. 

I'd be interested in the reasoning behind that but that is really outside of the discussion on if **fake-hwclock** should not be installed going further.
> So hwclock was fairly important[1]. The issue we found is not precisely that hwclock was removed, but that util-linux got split into two packages.
> 

The other major distros are using timesyncd out of the box for NTP/time management without an RTC, how Trisquel ensures the system time is accurate is not really that important to IIAB as consumers. IIAB should not cause time to jump backwards as noted in #4168 which could happen with any NTP client with fake-hwclock active.
>     * util-linux
> 
>     * util-linux-extra
> 
> 
> Leaving _hwclock_ on the extra one, now back to this issue.
> 
> I'm not really familiar with the mechanism to make an informed opinion, I would test the change and confirm if it solves the issue, then again I'm not in place right now to make a test on rpis sorry 🤧
> 
> 1 https://gitlab.trisquel.org/trisquel/package-helpers/-/issues/225

The  **fake-hwclock** deb package is completely different from _hwclock_ that is shipped in the util-linux-extra deb package and the reason for the move of hwclock to extra is because of the widespread use of timesyncd. I'm not suggesting some other NTP client could not be used in place of timesyncd but advancing the system clock without a RTC is what timesyncd is suppose to do but fake-hwclock did that in the past in place of timesyncd when that support was not present yet.

Short story fake-hwclock with a real RTC active leads to time weirdness as noted.

holta:
@Ark74

Should we consider configuring `systemd-timesyncd` in `/etc/systemd/timesyncd.conf` or `/etc/systemd/timesyncd.conf.d/99-custom-ntp.conf` to use trusted, privacy-respecting NTP servers ?

Useful diagnostics:

- `timedatectl timesync-status`
- `timedatectl timesync-status --monitor` (to watch changes in real-time)
- `timedatectl show-timesync` (parseable for scripts)
- `systemctl status systemd-timesyncd`

EMG70:
IIAB size one installed using curl iiab.io/fast.txt | bash -s 4277
Wi-Fi hotspot & everything seem to work OK. Please ignore Maps download as it is unrelated to the test requested.


sudo iiab-diagnostics https://paste.centos.org/view/046f3035

holta:
@Ark74 and/or others:

Do you think this is pretty safe to merge ASAP?

Ark74 (approved):
Tested on a rpi5
All running but cups-web, but not related to this PR

Thanks

holta:
My hunch is to merge this (PR #4277) later this week — as-or-after concrete details are very likely to emerge at long last in the coming 48h or so, hinting at the root cause underlying this pesky little MariaDB bug — that's clearly been kicking around for many months now:

- #4168

holta:
> All running but cups-web

@Ark74 is the CUPS problem serious, do you know?

(Thanks so much for passing along your https://paste.centos.org/view/0d378f18 — which doesn't seem to show serious CUPS problems ?!)

Ark74:
> @Ark74 is the CUPS problem serious, do you know?

Not at all. As I mentioned I was browsing from a device over the network, CUPS by default is only accesible via localhost, in order to browse via lan  it is required to change the security policy. I confirmed browsing from inside the box and CUPS was there without issues.

I suppose the main use for CUPS is to use the rpi as a usb-only printer hub to share over the network :printer: 
Nice!

holta:
Great!  (Thanks @Ark74 for having clarified CUPS.)

chapmanjacobd (approved):

holta:
FYI I'm in favor of merging this experimentally in coming days.

Even if we don't (yet?) have proof that this really solves...

- #4168

(Still it stands a quite good chance!)

PR #4285, opened 2026-02-26
[maps] update vars for maps on Android

Ark74:
### Fixes bug:
Update Maps vars accordingly with latest Maps work

Update maps to stay updated.

Debian 13 (proot)

@holta

### DIFF
--- a/vars/local_vars_android.yml
+++ b/vars/local_vars_android.yml
@@ -48,10 +48,10 @@ maps_enabled: True
 maps_vector_quality: "ne"                 # See `maps_dot_black_vector_tiles` in https://github.com/iiab/iiab/blob/master/roles/maps/defaults/main.yml for details and valid values.
 maps_satellite_zoom: 7                    # See `maps_dot_black_satellite_tiles` in https://github.com/iiab/iiab/blob/master/roles/maps/defaults/main.yml for details and valid values.
 maps_terrain_zoom: none                   # See `maps_dot_black_terrain_tiles` in https://github.com/iiab/iiab/blob/master/roles/maps/defaults/main.yml for details and valid values.
-maps_search_engine: static                  # See https://github.com/iiab/iiab/blob/master/roles/maps/tasks/install.yml for valid values.
+maps_search_engine: static                # See https://github.com/iiab/iiab/blob/master/roles/maps/tasks/install.yml for valid values.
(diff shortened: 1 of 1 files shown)


holta:
@orblivion please review / revise after merge!

Thank You to @Ark74 for catching this, as we move towards keeping mainline vars files within https://github.com/iiab/iiab/tree/master/vars much more in sync (:

Ark74:
Screenshot_20260226-142057_Fennec

Search enabled :+1:

vars/local_vars_android.yml
```diff
@@ -42,13 +42,16 @@ kiwix_enabled: True
...
 maps_install: True
 maps_enabled: True
-maps_vector_quality: ne
-maps_satellite_zoom: 7
-maps_terrain_zoom: none
-maps_region_downloader: True
+maps_vector_quality: "ne"                 # See `maps_dot_black_vector_tiles` in https://github.com/iiab/iiab/blob/master/roles/maps/defaults/main.yml for details and valid values.
+maps_satellite_zoom: 7                    # See `maps_dot_black_satellite_tiles` in https://github.com/iiab/iiab/blob/master/roles/maps/defaults/main.yml for details and valid values.
+maps_terrain_zoom: none                   # See `maps_dot_black_terrain_tiles` in https://github.com/iiab/iiab/blob/master/roles/maps/defaults/main.yml for details and valid values.
+maps_search_engine: static                # See https://github.com/iiab/iiab/blob/master/roles/maps/tasks/install.yml for valid values.
```
orblivion, line 51: It looks like Android is the only one with `static` as the default. Should we do it elsewhere or is Android special for the moment?

Also we should probably put quotes around `"static"`. I think `none` is the only exception?

holta: > It looks like Android is the only one with `static` as the default. Should we do it elsewhere or is Android special for the moment?

I agree we should do it everywhere else, including default_vars.yml (until a better search comes along, that remains very fast, while better serving users!)

> Also we should probably put quotes around `"static"`.

I'd suggest going in the other direction, removing all quotation marks, and then testing.

> I think `none` is the only exception?

`"none"` in YAML should work with quotes or without.  Personally I'd remove all unnecessary quotes in YAML.

holta: > I'd suggest [...] removing all quotation marks [in local_vars.yml, for maps_* values] and then testing.

Done:

- PR #4286

PR #4291, opened 2026-02-28
Easy method to update iiab & iiab-admin-console to latest code base

jvonau:
Has #4277 also
Avoids booting using systemd-firstboot
@chapmanjacobd
https://github.com/jvonau/iiab/actions/runs/22526466495

PR #4292, opened 2026-03-01
install warning routine only if user is created

jvonau:
Less filesystem clutter when the user is not created


holta:
Apologies I'm not onboard:

Password warnings serve a valuable purpose even if the username (e.g. iiab-admin) and password (e.g. g0adm1n) appear later, no matter how they arise.

jvonau:
https://github.com/iiab/iiab/pull/4291#issuecomment-3997682777

PR #4293, opened 2026-03-02
Reactivate 40 minute Raspberry Pi Workflows

Akatama:
#### Renamed ubuntu.yml -> 30min-iiab-large-ubuntu.yml
This can be reverted, but I thought it made sense to keep with our (admittedly verbose) naming scheme for workflows that are testing.

#### 40min-iiab-none-raspios-on-zero2w.yml and 40min-iiab-unittest-deb12-on-rpi3.yml
These are now reactivated, and they run on pull_request and workflow_dispatch.

I also removed the strategy matrix for both workflows. This is because, currently, the strategy matrix doesn't do anything (as there is only one set of options). The way this is supposed to work, is that you specify multiple options it will create multiple jobs that will run in parallel, here's [an example I pulled from GitHub's documentation](https://docs.github.com/en/actions/how-tos/write-workflows/choose-what-workflows-do/run-job-variations):
```yaml
    strategy:
      matrix:
        version: [10, 12, 14]
        os: [ubuntu-latest, windows-latest]
```
The above would cause 6 jobs to run in parallel 3 "versions" for both ubuntu-latest and windows latest.

At the bottom of this you can see what ours looked like before I changed it. The multitude of commented-out options certainly didn't help readability, but when it came down to it, there is ` arch: aarch64`, `cpu: cortex-a53`, `cpu_info: cpuinfo/raspberrypi_zero2_w_arm64` and `base_image: raspios_lite_arm64:latest`. 3 of the four of these values are needed to properly run arm-runner-action, so I just moved down the hardcoded values into the arm-runner-action step.

```yaml
    strategy:
      matrix:
        arch: [aarch64] #[zero_raspbian, zero_raspios, zero2_raspios, aarch64]
        include:
        #- arch: zero_raspbian
        #  cpu: arm1176
        #  cpu_info: cpuinfo/raspberrypi_zero_w
        #  base_image: raspbian_lite:latest
        #- arch: zero_raspios
        #  cpu: arm1176
        #  cpu_info: cpuinfo/raspberrypi_zero_w
        #  base_image: raspios_lite:latest
        #- arch: zero2_raspios
        #  cpu: cortex-a7
        #  cpu_info: cpuinfo/raspberrypi_zero2_w
        #  base_image: raspios_lite:latest
        - arch: aarch64
          cpu: cortex-a53
          cpu_info: cpuinfo/raspberrypi_zero2_w_arm64
          base_image: raspios_lite_arm64:latest
```

@holta @chapmanjacobd


holta:
I much prefer recent deletions be commented out not deleted (many people worldwide use phones not large screens, and so lose critical context!!)  i.e. when git's CLI commands to dig deeper into code history... are just not realistic for many contributors.

But overall this PR seems very reasonable, e.g. especially if @chapmanjacobd can tolerate it ! (:

tim-moody:
wondering why none would take 40 minutes and large only 30.

chapmanjacobd:
Because it's testing qemu and not IIAB 🤪

I veto merging this but maybe a good compromise could be to merge and then mv back to paused in six months if it didn't help to catch any extra bugs that the other workflows did

jvonau:
Perhaps 'runs-on:' could be ubuntu-24.04-arm to avoid qemu?

jvonau:
Think '40min-iiab-unittest-deb12-on-rpi3.yml' should be renamed to reflect raspios-13 as the starting point.

holta:
> Think '40min-iiab-unittest-deb12-on-rpi3.yml' should be renamed to reflect raspios-13 as the starting point.

Excellent point.

That was part of my confusion (:

chapmanjacobd:
fwiw

> Because Arm systems differ so much and in fundamental ways, typically operating system or firmware images intended to run on one machine will not run at all on any other. This is often surprising for new users who are used to the x86 world where every system looks like a standard PC. (Once the kernel has booted, most userspace software cares much less about the detail of the hardware.)
> https://www.qemu.org/docs/master/system/target-arm.html

In other words:

- Firmware/Kernel: Must be "Hardware-Aware" (highly fragmented).
- Userspace: Only needs to be "Architecture-Aware" (highly compatible).

So maybe the workflows only need to make sure that we can boot to userspace across the different raspberry pi models?

But I don't think the Raspberry Pi Foundation compiles kernels that are only compatible with one raspberry pi model so maybe even this is not needed...

jvonau:
> But I don't think the Raspberry Pi Foundation compiles kernels that are only compatible with one raspberry pi model so maybe even this is not needed...

There are five different kernel but only 2 for 64bit userspace based on https://www.raspberrypi.com/documentation/computers/linux_kernel.html

chapmanjacobd:
I mean that they include both--you can take it out of a Pi 500 and move to a Pi Zero 2 even after updating, right? So it installs both kernels?

https://raspberrypi.stackexchange.com/questions/119509/images-with-multiple-kernel-to-share-among-pi-b-pi-3-and-pi-4

So it seems like a solved problem

jvonau:
Kernels kernel8.img and kernel_2712.img are both packaged for arm64 in the upstream kernel deb file with their respective initrds initramfs8 and initramfs_2712 both being rebuilt upon installation of an updated kernel deb file.

Akatama:
> I much prefer recent deletions be commented out not deleted (many people worldwide use phones not large screens, and so lose critical context!!) i.e. when git's CLI commands to dig deeper into code history... are just not realistic for many contributors.

It's literally unreadable with all of the commented out yaml. I am not going to say that code should never be commented out, but when we have 3 relevant lines out of 20 lines, then we need to do some clean up. This is especially relevant for people reading on phones, because they have even less space, and so their whole screen space may be taken up by commented out lines. Keep in mind we aren't even using the strategy matrix anymore, since it didn't even do anything.

>Think '40min-iiab-unittest-deb12-on-rpi3.yml' should be renamed to reflect raspios-13 as the starting point.

I think that's a good idea. Although both 40 minute workflows are on raspios 13, so I will update both of their names.

> I veto merging this but maybe a good compromise could be to merge and then mv back to paused in six months 
if it didn't help to catch any extra bugs that the other workflows did

@chapmanjacobd Adam has suggested to me that instead of waiting six months, we wait a year. Do you find that agreeable?

chapmanjacobd:
sure!

holta:
> > Think '40min-iiab-unittest-deb12-on-rpi3.yml' should be renamed to reflect raspios-13 as the starting point.
> 
> I think that's a good idea. Although both 40 minute workflows are on raspios 13, so I will update both of their names.

@Akatama would you consider "raspios" in the workflow filenames, instead of "raspios13" ?  The reason is that (as you've demonstrated!) it's not really tied to any numeric version of RasPiOS, but rather it grabs the "very latest" automatically.

jvonau:
> > > Think '40min-iiab-unittest-deb12-on-rpi3.yml' should be renamed to reflect raspios-13 as the starting point.
> > 
> > 
> > I think that's a good idea. Although both 40 minute workflows are on raspios 13, so I will update both of their names.
> 
> @Akatama would you consider "raspios" in the workflow filenames, instead of "raspios13" ? The reason is that (as you've demonstrated!) it's not really tied to any numeric version of RasPiOS, but rather it grabs the "very latest" automatically.

The name within each yaml file should reflect the OS used:
`name: '"40 min" IIAB "UNITTEST" on Debian 12 on RPi 3'`
might become
`name: '"40 min" IIAB "UNITTEST" on RasPiOS on RPi 3'`
to match the other yaml file's naming.

holta:
1. @Akatama I'd be in favor of something like "rpios-lite-latest" in workflow filenames.

   And in human-readable strings like the `name:` field, something like "latest RPiOS Lite" or "latest RPiOS Desktop".

2. Graphical desktop images are very important for newbies exploring brand new Raspberry Pi hardware — especially when it's received as a gift by younger children building an IIAB with a family member.

   In other words, "lite" or "desk" (for graphical desktop OS's) should appear in workflow and ISO-like image filenames — to keep us all honest, descriptive & accurate. 

(I hope we can get to some kind of agreement very soon in coming days on both above, in person Thursday when we meet if it's not worked out by then!)

Akatama:
@holta @jvonau 

40min-iiab-none-rpios-lite-latest-on-zero2w.yml
- workflow name: "40 min" IIAB "NONE" on RasPiOS Lite Latest on Zero 2 W

40min-iiab-unittest-rpios-lite-latest-on-rpi3.yml
- workflow name: "40 min" IIAB "UNITTEST" on RasPiOS Lite Latest on RPi 3

rpios vs raspios in the file name I think is the only remaining debate here. No matter what, it is a bit of abbreviation. I will think on it more, but I went with rpios because it is what Adam suggested.

PR #4296, opened 2026-03-04
Docs/maps

chapmanjacobd:

### DIFF
--- a/roles/maps/README.md
+++ b/roles/maps/README.md
@@ -5,10 +5,9 @@ The new IIAB Maps lets you choose a configuration that works for you. Because we
 - Street Maps (vector)
 ​- Satellite Imagery (raster)
 ​- [Terrain](#what-about-3d-terrain) (optional 3D elevation data)
+- ​[Map search](#can-i-try-out-search-which-is-still-experimental) (optional Points of Interest search)
 ​
 Need more detail in specific areas? Regardless of your global quality setting, you can download "[Full Quality Regions](#full-quality-regions-experimental)". These high-res rectangular extracts provide maximum detail for your local community without completely filling your disk full of maps. If it's a small region, it won't take up much disk space.
-
-​We also offer [map search](#can-i-try-out-search-which-is-still-experimental) in several sizes to fit your hardware.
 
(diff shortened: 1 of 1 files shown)


holta:
Most of this is really great, if @orblivion agrees!

But re-ordering the doc doesn't help newcomers trying to grasp simple concepts first:

- Search is built-in by default, and usable right away.  Search remains a global thing, certainly as of March 2026, and possibly for a very long time.  Works well higher up in the doc.

- "Full Quality Regions" are a fancy/custom option for advanced implementers.  Certainly as of March 2026.  Works well lower down in the doc.

My Opinion Anyway: I'd prefer we live in the present and focus on users number one.  Reference material that shows a technocratic view is too often counterintuitive and unhelpful to newcomers.  Examples that follow newbie user workflows are almost always much more vivid.

orblivion:
I agree the opening paragraph is better. @holta if we're ready to say that search is now a standard part of the experience, we should probably not call it "experimental" anymore (though we should move that label to Nominatim specifically). Maybe we should not mention that it's "optional" in the opening section, even if it technically is, because it's now on by default actually.

orblivion:
And if we do that, then I could see an argument for not promoting Full Quality Regions until that's a smoother experience for all types of users. But I could see it both ways.

@holta I guess I'd ask this: If you're interested in the ordering of things for the sake of the message that it sends to the user/reader, would it make sense to do that ordering in the opening paragraph? Rather than that message coming in via reading the whole document? And as it happens, this new opening paragraph mentions search first. The only thing to change would be to mention here that "Full Quality Search" is (currently) advanced.

holta:
@chapmanjacobd I'd like to merge this (or similar) promptly — e.g. today if possible — any final touch-ups?!

cc: @orblivion who helped a lot (in person) trying to finalize this.

chapmanjacobd:
For sure

The only thing that stood out to me was that "search" sometimes reads like a verb. Maybe instead of "If you want" use a smaller word like "For" so that it doesn't read like you need to search for something

holta:
A big thank you to everyone who helped @orblivion to review this, prior to his Los Angeles presentation at https://socallinuxexpo.org Saturday !

PR #4297, opened 2026-03-05
[code] update apk custom URL; refactor download mechanism

Ark74:
### Fixes bug:
Improves large APK download from custom download URL by the ADFA team.

Refactor APK download to use resilient curl command on CI environments.
- Improves response by using speed-time and retry limits.
- Avoid download complete files on every attempt, instead resumes downloads.

- Ubuntu 24.04
- Debian 13

@holta @chapmanjacobd

### DIFF
--- a/roles/code/tasks/install_apkfile.yml
+++ b/roles/code/tasks/install_apkfile.yml
@@ -6,23 +6,38 @@
 
 ### Current release - arm64 ###
 - name: Download arch arm 64-bits APK
-  get_curl:
-    url: "{{ code_download_url }}/{{ code_apk_64bit_filename }}"
-    dest: "{{ code_data_root_apk }}/{{ code_apk_64bit_filename }}"
-    timeout: "{{ download_timeout }}"
+  command: >
+    curl -fL -C -
+    --connect-timeout {{ download_timeout | default(60) }}
+    --retry 10
+    --retry-delay 5
+    --retry-all-errors
+    --speed-time 30
+    --speed-limit 1024
+    -o "{{ code_data_root_apk }}/{{ code_apk_64bit_filename }}"
+    "{{ code_download_apk_64_redirect }}"
+  args:
+    creates: "{{ code_data_root_apk }}/{{ code_apk_64bit_filename }}"
   register: apk_download_64
   retries: 5
-  delay: 5
+  delay: 10
   until: apk_download_64 is succeeded
 
 ### Future release - arm32 ###
 #- name: Download arch arm 32-bits APK
-#  get_url:
-#    url: "{{ code_download_url }}/{{ code_apk_32bit_filename }}"
-#    dest: "{{ code_data_root_apk }}/{{ code_apk_32bit_filename }}"
-#    mode: "0644"
-#    timeout: "{{ download_timeout }}"
+#  command: >
+#    curl -fL -C -
+#    --connect-timeout {{ download_timeout | default(60) }}
+#    --retry 10
+#    --retry-delay 5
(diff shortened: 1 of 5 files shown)


holta:
@Ark74 can you clarify if this is likely to (maybe!) fix situations like yesterday's 4-out-of-13 CI failures mentioning Code on the Go below... ?

- PR #4293

Ark74:
Yes, it should. The curl options cherry-picked for this PR are specifically aimed to handle possible network instability and connection drops completing CI jobs.

The only theoretical edge case for data corruption would be if the APK on the server gets replaced by a new release exactly during the 10 second delay before a retry. This is because it is using the resume flag (-C -) to avoid restarting the ~1GB download from scratch, fetching the second half of a modified file (old_apk+new_apk) would corrupt the resulting "APK".

However, the chances of a server-side update happening perfectly within that 10-second reconnection window are incredibly low. I think this tiny risk is a necessary and acceptable trade-off to ensure CI pipelines run reliably without issues.

Ark74:
As we get a fixed point for download, URL. This role becomes a much more simplified download and install task, no more parsing, no more jumping CDN blocking.
Thanks to the ADFA team to help us simplify the process.

roles/code/tasks/install_apkfile.yml
```diff
@@ -6,23 +6,38 @@
 
 ### Current release - arm64 ###
 - name: Download arch arm 64-bits APK
-  get_curl:
-    url: "{{ code_download_url }}/{{ code_apk_64bit_filename }}"
-    dest: "{{ code_data_root_apk }}/{{ code_apk_64bit_filename }}"
-    timeout: "{{ download_timeout }}"
+  command: >
+    curl -fL -C -
```
chapmanjacobd, line 10: I've encountered some strange behavior with `-C -` in the past but I can't quite remember what it was. Perhaps it is incompatible with `-J`--which you're not using so maybe it is fine. 

But googling around a bit it can be a bit buggy unfortunately:

- https://github.com/curl/curl/issues/3898
- https://serverfault.com/questions/1053995/corrupted-curl-download-when-using-the-resume-option

However, I think we should keep it and just make a note for any future weirdness that we should try removing it to see if it helps

chapmanjacobd (approved):
Looks good

roles/code/tasks/install_apkfile.yml
chapmanjacobd: You could also consider using the get_curl module https://github.com/search?q=repo%3Aiiab%2Fiiab%20get_curl&type=code

and maybe we should add some of those speed-limit options to it

Ark74: If I understood correctly get_curl will try to download from scratch on every retry, right?
Making the download for 1GB startover, the main change here from get_curl is that it fails faster, retries and resumes where it was left. Always moving forward to complete the download on each attempt.

We could try to check against corruption adding an unzip task to confirm the file structure is correct, if that's necessary, maybe retry the download again then, but it seemed a bit convoluted.

chapmanjacobd: Corrupt zips can easily pass basic integrity checks like reading the header... It would be better to check against a published hashsum. https://stackoverflow.com/a/3954546/697964

We can consolidate the code paths later if we find that one is more resilient than the other over time. Looks good to merge AS-IS

holta: Merged, thank you @Ark74 and @chapmanjacobd.

(And feel free to suggest any follow-up PRs if/as necessary!)

PR #4301, opened 2026-03-10
Fix OS vars (is_debian was erroneously being set on Mint & Trisquel)

holta:
### Fixes bug:

Fix OS detection logic, so that it parses var `os_ver` correctly, when setting our 5 `is_distro` vars.  In summary:

- is_debian should be set on Pure Debian and RPiOS.

- is_ubuntu should be set on Ubuntu, Linux Mint and Trisquel.

Rework the parsing of var `os_ver` to do the job correctly.

CI tests confirm the most common 2 distros (RPiOS and Ubuntu).  And further testing (i.e. on all 5 distros) can't hurt, ideally with LARGE-sized IIAB installs.

@chapmanjacobd @Ark74


Ark74 (approved):
Good catch, stronger definition overall. 👍

jvonau:
I was wondering when is_linuxmint & is_trisquel would be included in is_ubuntu.

PR #4302, opened 2026-03-11
[maps] enable pmtiles for arm-32bits support

Ark74:
### Fixes bug:
Related to: https://github.com/iiab/iiab-android/issues/35

Add pmtile support for maps on 32bits.

Debian 13

@orblivion @holta @jvonau

### DIFF
--- a/roles/maps/defaults/main.yml
+++ b/roles/maps/defaults/main.yml
@@ -36,6 +36,9 @@ maps_slow_data_date: "2025-12-10"
 
 tile_extract:
   pmtiles:
+    armhf:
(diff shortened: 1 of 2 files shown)


Ark74:
Working on an emulated 32bit phone. :+1:

holta:
32-bits is going away across Android and Raspberry Pi (so let's not inflate expectations!!) but still this is Ok on occasion — so long as we don't create false notions that IIAB will support 32-bit in general.

Ark74:
Indeed, this is in a sense just extending as far as possible the limits, so yeah, expectations needs to be managed. Still makes for a great news for that old device with some years of life still left.

PR #4299, opened 2026-03-09
rename workflows, remove matrix, tweak run on

chapmanjacobd:
### Fixes bug:

Align with

https://github.com/iiab/iiab/wiki/Raspberry-Pi-Images-~-Summary#iiab-image-choices

final table:

```
┌────────────────────────────────────────────────┬──────────────────────┬───────────┬────────────┐
│ Workflow File                                  │ Trigger              │ Isolation │ Output     │
├────────────────────────────────────────────────┼──────────────────────┼───────────┼────────────┤
│ 2min-ansible-lint-on-ubuntu.yml                │ push + pull_request  │ direct    │ lint only  │
│ 7min-unittest-on-ubuntu.yml                    │ push + pull_request  │ direct    │ validation │
│ 10min-unittest-IMAGE-on-rpios-lite.yml         │ push + pull_request  │ chroot    │ Image (xz) │
│ 15min-tiny-IMAGE-on-rpios-lite.yml             │ push (tags + master) │ chroot    │ Image (xz) │
│ 20min-small-IMAGE-on-rpios-lite.yml            │ push (tags + master) │ nspawn    │ Image (xz) │
│ 25min-medium-IMAGE-on-rpios-desktop.yml.paused │ push (tags + master) │ nspawn    │ Image (xz) │
│ 30min-large-on-ubuntu.yml                      │ push + pull_request  │ direct    │ validation │
│ 35min-large-on-rpios-lite.yml                  │ push + pull_request  │ nspawn    │ validation │
└────────────────────────────────────────────────┴──────────────────────┴───────────┴────────────┘
```

PR #4303, opened 2026-03-13
Add an ansible-lint workflow to improve DevEx

Akatama:
Refer to #4299 

This new workflow clones IIAB, installs Ansible (required) and ansible-lint. Then it runs ansible-lint. Takes less than 2 minutes.

#4299 Should probably be updated so that 30min-large-on-ubuntu.yml no longer lints. It's going to give the same result of this workflow, but take much longer than this one for the result to be significant.

@holta @chapmanjacobd


.github/workflows/2min-lint-on-ubuntu.yml
```diff
@@ -0,0 +1,55 @@
+name: '"2 min" Lint on Ubuntu 24.04'
+
+on:
+  push:
+    paths-ignore:
```
chapmanjacobd: could probably change to paths-include instead of ignore to target the changes to  the ansible files (is it just yaml? maybe ansible.toml could also be included)

.github/workflows/2min-ansible-lint-on-ubuntu.yml
```diff
@@ -0,0 +1,55 @@
...
+          mv $GITHUB_WORKSPACE /opt/iiab
+          mkdir $GITHUB_WORKSPACE    # OR SUBSEQUENT STEPS WILL FAIL
+      - name: Install Ansible
+        run: sudo /opt/iiab/iiab/scripts/ansible
+
+      - name: Install ansible-lint
+        run: pip install ansible-lint
+
+      - name: Run ansible-lint
+        run: ansible-lint
```
chapmanjacobd, line 44: I would swap the order of this and line 55 just to be a bit more clear

.github/workflows/2min-lint-on-ubuntu.yml
```diff
@@ -0,0 +1,55 @@
+name: '"2 min" Lint on Ubuntu 24.04'
```
chapmanjacobd: maybe "2 min" Ansible Lint ? does the OS matter here? @holta

chapmanjacobd (approved):

.github/workflows/2min-lint-on-ubuntu.yml
holta: Can we rename this thing to 1min-ansible-lint-on-ubuntu.yml ?  If it often completes in about 1 minute, Fantastic!

```suggestion
name: '"1 min" Ansible Lint on Ubuntu 24.04'
```

.github/workflows/2min-lint-on-ubuntu.yml
holta: Ansible template files (which I hope ansible-lint is also validating?!) do not use always use a consistent .j2 suffix:

Might be best to keep the `paths-ignore` general approach??

holta: Confirmed that ansible-lint also validates .j2 templates.

Hypothetically in future we (might?) also supplement ansible-lint with...

> "File Extension Rule: An opt-in rule, jinja-template-extension, is available to specifically check that files used with the ansible.builtin.template module have the recommended .j2 extension, ensuring proper editor support and clarity."

.github/workflows/2min-lint-on-ubuntu.yml
holta: True OS (distro) doesn't matter too much in this 1 instance; still I'm in favor of mentioning it for completeness.

.github/workflows/2min-lint-on-ubuntu.yml
chapmanjacobd: ah okay. although if someone adds a template they'll need to have edited an Ansible file anyway so it should still run either way!

chapmanjacobd: but I suppose someone could rename the file later from .j2 in a different PR and that wouldn't trigger the linter if they didn't also edit a yaml|yml file. (unless we also add j2 on there just to cover that case as well)

so yml|yaml|j2 should be fine

holta: Might as well run ansible-lint as broadly as it can meaningfully help, in case an individual Ansible template regresses, etc. 

(Alongside Old News... ;)

Ansible itself changes (rather often!) so linting across the board (rather often!) stands to help us all keep on track~ 🕊️

jvonau:
In the current layout if the ansible-lint test fails the job doesn't proceed to the next step with the failure showing up in about the same time as running within the revised '1-2 min' workflow. Don't think running a separate workflow gains anything as the LARGE sizes are always run with the test installs being tossed out.

holta:
Clarification:

- Not just "rapid red" signals, but also "rapid green" signals, provide high community value.

- "Rapid reds" might increasingly auto-cancel long-running jobs that are demonstrably no longer needed, if things like GitHub Actions cleanly support that.

.github/workflows/2min-lint-on-ubuntu.yml
Akatama: It takes longer than 1 minute just for ansible-lint to run. I don't think we can in good conscious call it "1 min"

.github/workflows/2min-ansible-lint-on-ubuntu.yml
Akatama: I think its more clear this way. I've always put the working-directory part AFTER the command in my workflows.

Akatama: For example, this is what you see in the 30min-iiab-large-ubuntu.yml:

```yaml
      - name: Run ansible-lint
        run: ansible-lint
        working-directory: /opt/iiab/iiab

      - name: Set up /etc/iiab/local_vars.yml
        run: |
          sudo mkdir /etc/iiab
          sudo cp /opt/iiab/iiab/vars/local_vars_large.yml /etc/iiab/local_vars.yml

      - run: sudo ./iiab-install
        working-directory: /opt/iiab/iiab
```
For the last example, I am pretty sure it was already like that before I joined. As such, I would say this is the standard for this project.

.github/workflows/2min-lint-on-ubuntu.yml
holta: Tale of the Tape... says 1 minute below (so far anyway?!)

image

Akatama: 1m 51s. Definitely not 1 minute.

.github/workflows/2min-lint-on-ubuntu.yml
holta: > Ansible template files (which I hope ansible-lint is also validating?!) do not use always use a consistent .j2 suffix:

Example of an Ansible template that uses Jinja2 templating but unfortunately never adopted the .j2 suffix convention:

https://github.com/iiab/iiab/blob/master/roles/2-common/templates/iiab-startup.sh

(No doubt there are a few others... I or someone should look through every file in every templates/ directory... when there's time!)

Akatama: Currently, we only run if a .yaml/.yml/.j2 file is change, and none of those files are in .github/workflows. The end result is, since this PR is just the workflow the linting isn't running for the PR right now.

Adam, you last comment almost makes me think you just want it to run every PR no matter what. Is that the case?

jvonau: FWIW the linting did run as part of LARGE ubuntu with [9b50e39](https://github.com/iiab/iiab/pull/4303/commits/9b50e39a132cce5ec60eefc11a7051e4157b0ea6) in https://github.com/iiab/iiab/actions/runs/23080187730/job/67047868062?pr=4303

holta: Let's keep tuning til we find the tight/timely feedback loops that QA-oriented folk thrive on~

ansible-lint only verifies {.yml, .yaml, ,j2, .ini} so maybe that's good enough for now _(if ansible-lint doesn't change behind our backs too often ;)_

holta: Let me force this PR to test itself now... by moving iiab-startup.sh to iiab-startup.sh.j2 ;)

.github/workflows/2min-lint-on-ubuntu.yml
holta: 1. Sure.  Latest runs were [1m38s](https://github.com/iiab/iiab/actions/runs/23081432991/job/67051339739), [1m47s](https://github.com/Akatama/iiab/actions/runs/23081432302/job/67051338201), [1m49s](https://github.com/Akatama/iiab/actions/runs/23081440230/job/67051356544) and [1m50s](https://github.com/iiab/iiab/actions/runs/23081441553/job/67051359336?pr=4303).

2. "Ansible Lint" in name: field and "ansible-lint" in .yml filename probably don't hurt (Jacob's suggestion) to clarify for newcomers what we're up to?

jvonau:
Just going to note that in c6d4508 the linting ran and passed but the test runs failed, shouldn't the linting catch that kind of problem?

holta:
> Just going to note that in [c6d4508](https://github.com/iiab/iiab/commit/c6d4508270ccb434a619d4f3c5156953884af5a6) the linting ran and passed but the test runs failed, shouldn't the linting catch that kind of problem?

In this case, it was my fault.  I really should have combined those 2 commits (c6d4508 and f58bbd5) into 1 single commit.  So the successful lint followed by failed CI workflows are to be expected, in this particular case.

jvonau:
> > Ansible template files (which I hope ansible-lint is also validating?!) do not use always use a consistent .j2 suffix:
>

Warning, there are some roles that might use 'template:' in place of using 'copy:' to save creating a new directory but those should lack the .j2 extension. 'We' back in [XSCE](https://github.com/XSCE/xsce) days adopted using .j2 for templates that contained replaceable variables.
> Example of an Ansible template that uses Jinja2 templating but unfortunately never adopted the .j2 suffix convention:

As noted above when IIAB 'imported' XSCE there where no replaceable variables
> https://github.com/iiab/iiab/blob/master/roles/2-common/templates/iiab-startup.sh
>

until that changed with #4071, should of moved to using .j2 back then.
> (No doubt there are a few others... I or someone should look through every file in every templates/ directory... when there's time!)

That is part of being a maintainer, happy hunting.

jvonau:
b6f36f23ea7dff93346b8458c34c7ffbbe194da6 don't you need a 'dest' to get the name change without the .j2 on the filesystem like in f4b7d2153e108f6557e84dcbe3a0728adfedfcb2?

jvonau:
https://gh.io/job-concurrency-limits

holta:
Oops that was ugly; I didn't quite realize there were so many Ansible templates remaining, to be renamed with .j2 suffixes.  

Hopefully we're at least much closer now, and the rest can be cleaned up more gracefully.

I manually canceled the resulting flood of (queued up and running, mostly superfluous) GitHub Actions workflows within iiab/iiab.  With the 9 still running ("in progress") in https://github.com/Akatama/iiab hopefully completing or timing out soon.

.github/workflows/2min-lint-on-ubuntu.yml
holta: To try to get this PR wrapped up & merged quickly, I went ahead and made that "Ansible Lint" and "ansible-lint" core idea more prominent.

And we can fine tune down the road, e.g. if GitHub Actions' RED-vs-GREEN capsule summaries / live dashboards appear too wordy / repetitive.

holta:
I think I finally nailed all of this repo's templates needing to be renamed with a .j2 suffix.

So hopefully we're very close now, and if all goes well this PR can be merged today?!

p.s. I left this final one below untouched, as the templated var is in a comment, if only as a museum piece... ;)

https://github.com/iiab/iiab/blob/master/roles/captiveportal/templates/iiab-make-cp-servers.py

chapmanjacobd (approved):
looks good to me, should we enable the opt-in .j2 stricter rule now?

https://docs.ansible.com/projects/lint/rules/jinja-template-extension/

chapmanjacobd:
You can add this to cancel other workflows when this fails:

```yaml
      - name: Cancel other workflows on failure
        if: failure()
        env:
          GH_TOKEN: ${{ github.token }}
        run: |
          CURRENT_RUN_ID="${{ github.run_id }}"
          BRANCH="${{ github.head_ref || github.ref_name }}"

          gh run list \
            --branch "$BRANCH" \
            --status in_progress \
            --json databaseId \
            --jq '.[].databaseId' |
          grep -v "$CURRENT_RUN_ID" |
          xargs -r -n1 gh run cancel
```

Akatama:
@chapmanjacobd @jvonau @holta@ At this point I will add that in a bit.

My original intention for this PR is for it to be small (just the workflow), but that hasn't happened. It is only logical that if we are using files as jinja2 templates but they aren't named .j2 then they need to be named that for this workflow to work. 

I still don't want to touch the other workflows, so Jerry/Jacob's PR #4299 can proceed.  Keep in mind that we will no longer require the 30min-large-on-ubuntu.yml to lint and it should be removed. If I made that change here, it would cause a merge conflict with that PR, which is why I am not touching it.

I think there may be a way to control workflow execution order, so that instead of canceling the other running workflows, just have this one run first. Whether that is better is part of the experiment.

holta:
>> opt-in .j2 stricter rule now?
>>
>> https://docs.ansible.com/projects/lint/rules/jinja-template-extension/

> I will add that in a bit.

Thanks @Akatama (can you do that in a small/separate PR, after I merge this now?)

Akatama:
Sure!

holta:
With a LARGE-sized install test of this PR on Ubuntu Server 24.04 LTS, I did quick functional testing logging into & doing the basics with every app on IIAB's main page (good enough!)

PR #4305, opened 2026-03-15
Host all remaining maps.black assets ourselves

orblivion:
### Fixes bug:

New Maps installation broken due to maps.black being down. (Even if it comes back, it seems we should be more resilient)

Host the remaining 2 maps.black assets ourselves and use meta4 downloader.

RasPi Trixie

@holta @chapmanjacobd


orblivion:
(Thanks for the hosting help @Ark74 !)

PR #4307, opened 2026-03-16
zstd will probably compress faster than xz--if it works

chapmanjacobd:
### Fixes bug:

Hopefully this shaves off a few minutes... lets see how it goes

https://9to5linux.com/raspberry-pi-imager-1-7-released-with-new-advanced-settings-zstd-support-and-more

### DIFF
--- a/.github/workflows/10min-unittest-IMAGE-on-rpios-lite.yml
+++ b/.github/workflows/10min-unittest-IMAGE-on-rpios-lite.yml
@@ -167,8 +167,8 @@ jobs:
 
           mv "$IMG_FILE" "${IMG_NAME}.img"
 
-          # Compress image with xz
-          xz -v -9 -T0 --stdout "${IMG_NAME}.img" > "${IMG_NAME}.img.xz"
+          # Compress image with zstd
+          zstd -T0 -6 --stdout "${IMG_NAME}.img" > "${IMG_NAME}.img.zst"
           EOF
 
       - name: Generate rpi-imager manifest
@@ -188,13 +188,13 @@ jobs:
               print("Error: IMG_NAME not set")
               sys.exit(1)
 
-          xz_name = f"{img_name}.img.xz"
-          if not os.path.exists(xz_name):
-              print(f"Error: {xz_name} not found")
+          zst_name = f"{img_name}.img.zst"
+          if not os.path.exists(zst_name):
+              print(f"Error: {zst_name} not found")
               sys.exit(1)
 
           # Get compressed file size (for download size)
-          compressed_size = os.path.getsize(xz_name)
+          compressed_size = os.path.getsize(zst_name)
           print(f"Compressed size: {compressed_size} bytes")
 
           # Get uncompressed image size and hash (for extract verification)
@@ -262,7 +262,7 @@ jobs:
                       "name": f"Internet in a Box - {os.environ.get('EDITION', 'Custom')} (64-bit)",
                       "description": f"IIAB {os.environ.get('IIAB_BASE_VERSION', '')} {os.environ.get('EDITION', '')} edition",
                       "icon": "https://raw.githubusercontent.com/iiab/iiab-factory/master/box/rpi/iiab40.png",
-                      "url": f"file:///home/YOUR_USERNAME/Downloads/{xz_name}",
+                      "url": f"file:///home/YOUR_USERNAME/Downloads/{zst_name}",
                       "image_download_size": compressed_size,
                       "extract_size": image_size,
                       "extract_sha256": extract_sha256,
@@ -280,10 +280,10 @@ jobs:
 
           python3 gen_manifest.py
 
-      - name: Upload image artifact (xz compressed)
+      - name: Upload image artifact
         uses: actions/upload-artifact@v7
         with:
-          path: nspawn-loop/${{ env.IMG_NAME }}.img.xz
+          path: nspawn-loop/${{ env.IMG_NAME }}.img.zst
           retention-days: 10
           archive: false
 
--- a/.github/workflows/15min-tiny-IMAGE-on-rpios-desktop.yml.paused
+++ b/.github/workflows/15min-tiny-IMAGE-on-rpios-desktop.yml.paused
@@ -176,8 +176,8 @@ jobs:
 
           mv "$IMG_FILE" "${IMG_NAME}.img"
 
-          # Compress image with xz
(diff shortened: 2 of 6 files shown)


holta:
Incredible!  And rpi-imager promises to work with .zst files just like it works with .xz files.

And a bit of background, if we believe Google Gemini:

> "Many Linux distributions (like Arch Linux) have shifted from .tar.xz to .tar.zst for packages because the slight increase in package size (~0.8% - 10%) is heavily outweighed by a massive improvement in installation speeds (~1300% faster)."

chapmanjacobd:
Original image size: 5.6 GiB

Before (xz): [7m 58s](https://github.com/iiab/iiab/actions/runs/23123253951/job/67161191430) 923.8 MiB

After (zstd): [40s](https://github.com/iiab/iiab/actions/runs/23123502162/job/67161913909?pr=4307) 1.2 GiB

(observation count of 1; results may vary...)

chapmanjacobd:
Lets try the unittest image in rpi-imager before merging ;-)

holta:
Should we go ahead and merge this?

(Even if .json file ingested by rpi-imager and any associated docs need a few tweaks...)

chapmanjacobd:
The main thing that makes this possible is v7 actions/upload-artifact

https://github.com/actions/upload-artifact/releases/tag/v7.0.0

ability to upload without ZIP! lol

released just in time...

chapmanjacobd:
Looks like it is working...

    sudo ~/.local/bin/rpi-imager --disable-verify --cloudinit-userdata ./user-data --cloudinit-networkconfig ./network-config Downloads/iiab-8.3-260316-raspios-unittest-58e3517.img.zst (path resolve /dev/disk/by-path/pci-0000:00:14.0-usb-0:3:1.0-scsi-0:0:0:1)

And it boots up! So I guess good to merge. Might be good to check if the rpi-imager GUI trips over itself like usual: https://github.com/iiab/iiab/actions/runs/23123502162?pr=4307

(Would also be good to test if it works on Windows--it probably does though there are some bug reports of it not working in the past https://github.com/MrSmoer/rpi-imager/commit/f28b9d0d770a92a48c5be2c7e702a82de40c74b8)

Akatama:
Python 3.14 also provides direct ZStandard support. Python 3.14 released October 2025, so support for ZStandard has been growing quickly.

holta:
Great!  Let's merge to broaden testing...

(And if we get severe complaints, we can always revert to xz.)

chapmanjacobd:
For future reference:

| zstd level | Expected size    | Compression ratio | Time estimate | Relative to xz -9      |
| ---------- | ---------------- | ----------------- | ------------- | ---------------------- |
| `-6`       | 1.3–1.7 GB       | 23–30%            | 20–40 sec     | ~10–20× faster         |
| `-8`       | 1.2–1.6 GB       | 21–28%            | 30–60 sec     | ~8–12× faster          |
| `-10`      | 1.1–1.5 GB       | 19–26%            | 45–90 sec     | ~5–8× faster           |
| `-12`      | 1.0–1.4 GB       | 18–24%            | 1.5–2.5 min   | ~3–4× faster           |
| `-15`      | 0.9–1.2 GB       | 16–21%            | 3–5 min       | ~1.5–2× faster         |
| `-17`      | 0.85–1.1 GB      | 15–19%            | 4–6 min       | similar ratio          |
| **`-18`**  | **0.82–1.0 GB**  | **14–18%**        | **5–7 min**   | closest to xz -9       |
| **`-19`**  | **0.80–0.95 GB** | **14–17%**        | **6–9 min**   | often slightly smaller |

TODO: use `zstd --long=27 -T0 -8`

chapmanjacobd:
We should probably make an official release for IIAB Release 8.3 so that the daily "What's Changed" is more concise

https://github.com/iiab/iiab/releases

holta:
1. Indeed a CI/CD glitch from https://github.com/iiab/iiab/actions/runs/23124647056/job/67165222771 suggests we might actually need a bit tighter/slower compression in the end, as GitHub Actions requires such files size be < 2GB = 2^31 bytes = 2,147,483,648 bytes):

   (Scrolling up and clicking, one can see that "Compressed size: 2195610658 bytes" just spillled over the limit...)

image

2. Interestingly, we're not limited to 10 GB (no cumulative limit at all!) on GitHub repo release pages, so long as each artifact is < 2GB.

Akatama:
It may be a better bet to have slightly bigger images that are compressed faster on the future self-hosted runner. For GitHub, we're going to run into issues with 2GB size limit no matter what, and that will probably get worse as time goes on.

holta:
> TODO: use `zstd --long=27 -T0 -8`

Agreed.  @chapmanjacobd maybe even immediately change from -6 to -8, as soon as you get the chance?

(Notwithstanding @Akatama's point, developing world implementers really do appreciate when we compress image size, so let's honor that perspective too!)

chapmanjacobd:
huh... yeah... 45.9 MiB too big

-8 might not be enough. let's go with -9 which is very slightly more time

holta:
> We should probably make an official release for IIAB Release 8.3 so that the daily "What's Changed" is more concise
> 
> https://github.com/iiab/iiab/releases

Yay the problem has since solved itself, as can be seen in these 2 subsequent pre-releases:

- https://github.com/iiab/iiab/releases/tag/daily-260316-581d46e

- https://github.com/iiab/iiab/releases/tag/daily-260316-bbd969a

holta:
Does anyone know if pre-release artifacts (e.g. `.img.zst` files) are auto-deleted by GitHub after 90 days?

That's perfectly ok — as these are intended to be relatively short-lived pre-releases after all — but still it'd be good to know what really happens :)

chapmanjacobd:
How long do you want the old releases to stick around? PR and dailies? The PR unittest Articacts have limitations but the release ones don't as far as I'm aware

Right now it's set to delete old prereleases every time we make a new prerelease--but we can configure the number of old prereleases to keep around and/or expire

Right now that is set to 7 days with no minimums

If we have a good daily release we can turn it into an official release

holta:
> it's set to delete old prereleases [older than 7 days] every time [that an RPiOS IMAGE workflow completes],

- Do pre-release pages entirely disappear?
- Or just their downloadable artifacts?
- To help with bisecting, the ideal would be something like...
   1. Keep the 10 most recent pre-releases at all times.
   2. If there happen to be more than 10 pre-releases published over the prior 10 days, keep all of those.

chapmanjacobd:
okay that should be possible

and yes the extra daily releases and tags should disappear

PR #4308, opened 2026-03-16
zstd --long=27 -T0 -9 [and experiment with "light speed" -1 compression, when generating .img artifact for unit testing]

chapmanjacobd:
### Fixes bug:

CI/CD glitch from https://github.com/iiab/iiab/actions/runs/23124647056/job/67165222771 suggests we might actually need a bit tighter/slower compression in the end, as GitHub Actions requires such files size be < 2GB = 2^31 bytes = 2,147,483,,648 bytes):

(Scrolling up and clicking, one can see that "Compressed size: 2195610658 bytes" just spillled over the limit...)

Hopefully this gets us under 2GB with a little bit of headroom

### DIFF
--- a/.github/workflows/10min-unittest-IMAGE-on-rpios-lite.yml
+++ b/.github/workflows/10min-unittest-IMAGE-on-rpios-lite.yml
@@ -168,7 +168,7 @@ jobs:
           mv "$IMG_FILE" "${IMG_NAME}.img"
 
           # Compress image with zstd
-          zstd -T0 -6 --stdout "${IMG_NAME}.img" > "${IMG_NAME}.img.zst"
+          zstd --long=27 -T0 -9 --stdout "${IMG_NAME}.img" > "${IMG_NAME}.img.zst"
           EOF
 
       - name: Generate rpi-imager manifest
--- a/.github/workflows/10min-unittest-IMAGE-on-rpios-lite.yml
+++ b/.github/workflows/10min-unittest-IMAGE-on-rpios-lite.yml
@@ -168,7 +168,7 @@ jobs:
           mv "$IMG_FILE" "${IMG_NAME}.img"
 
           # Compress image with zstd
-          zstd --long=27 -T0 -9 --stdout "${IMG_NAME}.img" > "${IMG_NAME}.img.zst"
+          zstd --long=27 -T0 -1 --stdout "${IMG_NAME}.img" > "${IMG_NAME}.img.zst"
           EOF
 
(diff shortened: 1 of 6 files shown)


.github/workflows/10min-unittest-IMAGE-on-rpios-lite.yml
```diff
@@ -168,7 +168,7 @@ jobs:
           mv "$IMG_FILE" "${IMG_NAME}.img"
 
           # Compress image with zstd
-          zstd -T0 -6 --stdout "${IMG_NAME}.img" > "${IMG_NAME}.img.zst"
+          zstd --long=27 -T0 -9 --stdout "${IMG_NAME}.img" > "${IMG_NAME}.img.zst"
```
holta: Not today but if in future we want to experiment with "light speed", for just this one CI workflow, where speed-rather-than-compression is the priority:

```suggestion
          zstd --long=27 -T0 -1 --stdout "${IMG_NAME}.img" > "${IMG_NAME}.img.zst"
```

holta: Actually what the hell... let's try the experiment... see what we learn (and revert/revise/refine later!)

chapmanjacobd: yeah zstd -1 to -3 are really fast, almost native read speed but still pretty good compression. I use -3 with btrfs for fast filesystem compression.

We will probably want to be pretty aggressive, maybe -16 or even more for popular downloads as it is a one time CPU cost that reduces bandwidth by quite a bit.

but I agree, for these unittest files, if it takes more than a few seconds longer it is probably not worth it. Still, the sweet spot might be closer to -4

chapmanjacobd: fwiw the decompression speed of all of these options are pretty similar

holta: Good ideas!  Let's try both...

- `-4` for unit test .img.zst artifacts
- `-16` for pre-release and release .img.zst artifacts (at the same time let's keep an eye on these times — if some of these consistently take "many minutes" to compress, let's look to speed up — in consideration of loyal folks who get auto-emailed announcements on each pre-release... and want installer artifacts downloadable very promptly from https://github.com/iiab/iiab/releases etc.)

https://github.com/iiab/iiab/pull/4307#issuecomment-4064558139

PR #4309, opened 2026-03-16
Make release name sort well by using [HHMM and] PR number instead of hash

chapmanjacobd:
### Fixes bug:

The pre releases don't sort well due to the use of git hash

### Description of changes proposed in this pull request:

We use PR number, if available, instead

### Mention a team member @username e.g. to help with code review:

@holta

PR #4250, opened 2026-01-27
Build RaspberryPi OS images in GitHub Actions

chapmanjacobd:
## Description of changes proposed in this pull request:

Builds and publishes images in GitHub Actions for dev use, primarily targeting `rpi-imager --cli` cloudinit workflows

A follow up PR will add support for rpi-imager integration for mainstream use

## Usage

Use rpi-imager-cli v2.0.2 cli build with the img.zst file (or v2.0.6 GUI with rpi-imager-manifest file):

```sh
sudo ~/.local/bin/rpi-imager --disable-verify \
    --cloudinit-userdata ./user-data \
    --cloudinit-networkconfig ./network-config \
    ./iiab-8.3-260223-raspios-tiny-99bef02ed.img.zst \
    /dev/sdX
```

### ./network-config

Don't try to use SHA-512 hashed SSID passkey because that won't work. Plaintext works:

```yaml
version: 2
ethernets:
  eth0:
    dhcp4: true
wifis:
  wlan0:
    dhcp4: true
    regulatory-domain: US
    access-points:
      "SSID with spaces":
        password: "tsikada"
```

### ./user-data

Use this to hash your user account passwd:

    openssl passwd -6 'my password'

```yaml
#cloud-config
hostname: box
timezone: America/Chicago
locale: en_US.UTF-8

manage_resolv_conf: false
manage_etc_hosts: true

keyboard:
  model: pc104
  layout: us
  # variant: colemak_dh

users:
  - name: xk
    passwd: "$6$NpmvAzPP.tL9Nzkn$Go...gle/INLh6mXgV39Ab0"
    lock_passwd: false
    groups: users,adm,dialout,audio,netdev,video,plugdev,cdrom,games,input,gpio,spi,i2c,render,sudo
    sudo: ['ALL=(ALL) NOPASSWD:ALL']
    shell: /bin/bash
    ssh_authorized_keys:
      - ssh-ed25519 AAAAC3NzaC1lZDI1NTE5AAAAIEIP0du7ZPjP2e9c1uB7ajnjLEYY4iJNCvEv5XQhKxD7 phone
      - ssh-ed25519 AAAAC3NzaC1lZDI1NTE5AAAAIL4RSNoe1n/mAGfuslIhJGX/N/7aM/H4BHgwmPm+OKlC lappy

enable_ssh: true
ssh_pwauth: false

package_update: true
package_upgrade: false
apt:
  preserve_sources_list: true
  conf: |
    Acquire {
      Check-Date "false";
    };

[...]

PR #4310, opened 2026-03-16
Use UTC time for releases

chapmanjacobd:
### Fixes bug:

Committer time is relative-local

### DIFF
--- a/.github/workflows/10min-unittest-IMAGE-on-rpios-lite.yml
+++ b/.github/workflows/10min-unittest-IMAGE-on-rpios-lite.yml
@@ -60,8 +60,12 @@ jobs:
           OS="rpios"
           EDITION="unittest"
           HASH=$(git -C ../iiab rev-parse --short=7 HEAD)
-          COMMIT_DATE=$(git -C ../iiab log -1 --format=%cd --date=format:'%y%m%d')
-          COMMIT_TIME=$(git -C ../iiab log -1 --format=%cd --date=format:'%H%M')
+          COMMIT_DATETIME_LOCAL=$(git -C ../iiab log -1 --format=%cd --date=iso-strict)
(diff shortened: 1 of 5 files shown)


holta:
Just as background/summary, this PR builds on things like:

- PR #4250
- PR #4291
- PR #4299 
- PR #4307 
- PR #4308 
- PR #4309

holta:
A giant thank you to @chapmanjacobd, @jvonau and also the many others who pulled together over recent months to make this happen ✅

Frequent pre-releases[*] are now starting to appear here, a few minutes after every merged PR!

https://github.com/iiab/iiab/releases

[*] Ironically pre-releases (can, really, often) be more important than so-called "final" releases in a conscientious culture of Rolling Releases... but much more about that in coming months... releases with a stronger touch of Human QA will also be coming... traditionalists of course want those too... these will not be as often naturally, but will likely appear "Fridays" (not every Friday, but many Fridays) in general... with the help of volunteers like @EMG70... before the middle of the year... but not to get ahead of ourselves when each of these efforts take many months to converge properly (more about that later!)

holta:
Example pre-release, just to make these CI/CD accomplishments more clear: (click on the "Artifacts" section!)

https://github.com/iiab/iiab/releases/tag/daily-260316-1948-PR4310

CLARIF: The current plan is to keep 10+ pre-releases at all times.  And to keep _all_ pre-releases from the prior 10 days, should a flood of PRs happen to arrive during any any short/intensive 10-day period.  In other words, when a PR lands (is merged), a new pre-release is consequently generated, and it will try to auto-delete the oldest pre-releases — while keeping at least 10 recent ones, as a kind of de facto microblog here: https://github.com/iiab/iiab/releases (also to help bisecting, if diagnosing / reverting subtle bugs proves necessary!)

PR #4313, opened 2026-03-17
Delete 2 paused & unused CI workflows

holta:
Tidying up after things like:

- PR #4250
- PR #4299 

Context: (Some!) of the focus has moved from CI-on-Ubuntu to CD-on-RPiOS, so Ubuntu-based CI workflows are (somewhat!) less needed going forward.

cc: @Akatama @chapmanjacobd


chapmanjacobd:
The good thing about Ubuntu x86_64 is that it's probably faster because it is more common but we already have a large size test and that is good enough

PR #4316, opened 2026-03-18
Remove libcamera*

chapmanjacobd:
### Fixes bug:

libcamera starts pulling in some desktop packages via depends and it might be better to remove it as most users probably aren't going to use it

Based on: https://github.com/iiab/iiab/pull/4291/changes/ccc9d0fc97b15f8aca8911e9a02d262dbf159ff9

### DIFF
--- a/.github/workflows/10min-unittest-IMAGE-on-rpios-lite.yml
+++ b/.github/workflows/10min-unittest-IMAGE-on-rpios-lite.yml
@@ -118,6 +118,7 @@ jobs:
           apt update
           mkdir -p /etc/initramfs-tools/conf.d
           echo 'MODULES=most' > /etc/initramfs-tools/conf.d/nspawn.conf
+          apt -y remove libcamera*
           apt -y full-upgrade
 
           apt install curl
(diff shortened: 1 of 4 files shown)


holta:
@jvonau @Ark74 good enough?

jvonau:
Let it rip, let see what size the desktop come in at.

Ark74:
Very straight forward

chapmanjacobd:
> Let it rip, let see what size the desktop come in at.

well desktop is paused right now actually because it wasn't in the table in PR #4299. Although IIAB tiny-desktop can fit under the 2GB limit.

but I kicked off a build with desktop medium and zstd -22 --ultra to see what happens... probably still too big:

https://github.com/chapmanjacobd/iiab/actions/runs/23224185858/job/67503272676 (oops accidentally based off master instead of this PR but the results should be pretty similar... since it is desktop already it probably has the desktop deps and removing libcamera is not going to be doing nearly as much)

chapmanjacobd:
yeah zstd -22 can get it down to 4.8GB (33% compression ratio) but not quite 2GB (13.7% compression ratio would be required...).

PR #4319, opened 2026-03-19
use iiab/iiab/releases/download URL

chapmanjacobd:
### Fixes bug:

Should allow people to use the rpi-imager-manifest file directly without downloading the zst image file

@jvonau


chapmanjacobd:
Screenshot_2026-03-19_1153 01

Seems to work! and it looks like zstd is stream-able like xz too which is a nice bonus (compared to zip which is not streamable)

jvonau:
+1 Progression of what was discussed in regards to stable url's, now just to update fast.json when there is an actual release.

holta:
This PR builds on things like:

- PR #4250
- PR #4291
- PR #4299

PR #4321, opened 2026-03-22
Fix frontend for full satellite download

orblivion:
### Fixes bug:

If satellite was set to "full" the front end wouldn't work. I was still assuming the value would be a number.

RPiOS 13 Trixie

@holta

### DIFF
--- a/roles/maps/templates/index.html.j2
+++ b/roles/maps/templates/index.html.j2
@@ -37,7 +37,7 @@
             const MAPS_SLOW_DATA_DATE = "{{ maps_slow_data_date }}"
             const MAPS_VECTOR_QUALITY = "{{ maps_vector_quality }}"
             const MAPS_VECTOR_IS_OSM = MAPS_VECTOR_QUALITY.includes("osm")
-            const MAPS_SATELLITE_ZOOM = {{ maps_satellite_zoom }} // so far just a number
+            const MAPS_SATELLITE_ZOOM = "{{ maps_satellite_zoom }}" === "full" ? 13 : {{ maps_satellite_zoom }}
             const MAPS_TERRAIN_ZOOM = "{{ maps_terrain_zoom }}" === "none" ? null : {{ maps_terrain_zoom }} // a number if not null
             const MAPS_SEARCH_ENGINE = "{{ maps_search_engine }}"
             const MAPS_REGION_DOWNLOADER = "{{ maps_region_downloader }}" === "True"
--- a/roles/maps/templates/index.html.j2
+++ b/roles/maps/templates/index.html.j2
@@ -37,8 +37,30 @@
             const MAPS_SLOW_DATA_DATE = "{{ maps_slow_data_date }}"
             const MAPS_VECTOR_QUALITY = "{{ maps_vector_quality }}"
             const MAPS_VECTOR_IS_OSM = MAPS_VECTOR_QUALITY.includes("osm")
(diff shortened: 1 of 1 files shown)


roles/maps/templates/index.html.j2
```diff
@@ -37,7 +37,7 @@
             const MAPS_SLOW_DATA_DATE = "{{ maps_slow_data_date }}"
             const MAPS_VECTOR_QUALITY = "{{ maps_vector_quality }}"
             const MAPS_VECTOR_IS_OSM = MAPS_VECTOR_QUALITY.includes("osm")
-            const MAPS_SATELLITE_ZOOM = {{ maps_satellite_zoom }} // so far just a number
+            const MAPS_SATELLITE_ZOOM = "{{ maps_satellite_zoom }}" === "full" ? 13 : {{ maps_satellite_zoom }}
```
orblivion: I will note that if it's set to `"full"` it renders as:

`const MAPS_SATELLITE_ZOOM = "full" === "full" ? 13 : full`

`full` isn't a defined name but since it's not evaluated javascript doesn't care. Same goes for terrain being set to the default of `"none"`. I just noticed this. I suppose it would be a problem if we do this with a value with a dash in it.

chapmanjacobd: interesting...

```js
> const V = "full" === "full" ? 13 : wut
undefined

> V
13

> const V2 = "NOT full" === "full" ? 13 : wut
Uncaught ReferenceError: wut is not defined
```

chapmanjacobd: so you should fix this before merging...

chapmanjacobd: unless we only support int | "full"

orblivion: Well, yes that is the idea, and it's why "none" doesn't break for terrain. Admittedly it's a bit precarious or at least confusing to read, and maybe I should clean it up anyway.

orblivion:
Okay how's that?

holta:
> Okay how's that?

And until we get CI/CD testing enacted for our "new" IIAB Maps... is this sufficiently tested?

cc: @chapmanjacobd

chapmanjacobd:
> We hope they notice and report the bug if it ever comes here!

could probably do a `console.warn(Please report this bug... ${...vars})`

chapmanjacobd:
console.warning is not a function... unfortunately

just console.warn or .error or .info or .log

chapmanjacobd (approved):
Looks perfect!

chapmanjacobd:
Just one final note: `Number("")` becomes 0 in javascript... So it might be safer to do something like this:

```js
const parsed = Number(RAW_SATELLITE_ZOOM)
if (!RAW_SATELLITE_ZOOM || Number.isNaN(parsed)) {
```

I guess mostly non-issue but it is something to keep in mind!

Also you could do something like this instead of `let` (more of a Rust flavor...):

```
const MAPS_SATELLITE_ZOOM = (() => {
    const raw = RAW_SATELLITE_ZOOM
    const parsed = Number(raw)

    if (!Number.isNaN(parsed)) return parsed
    if (raw === "full") return 13

    console.warn(...)
    return 7
})()
```

but this is 100% stylistic choice

orblivion:
That's my style too, I didn't think of it. As for blank -> 0, yeah maybe, but less likely than a bad value.

Maybe next time on both counts, I don't want to drag this on further if we agree it's low risk for now.

PR #4324, opened 2026-03-24
Reduce maps CI size

orblivion:

For CI, use data files that are much smaller than even the default installation for users:

* `naturalearth6-NE2_HR_SR_W_DR-WEBP.2025-12-10.zoom_0-04.pmtiles` which gives the "natural" style its "natural" look (zoom 0-4, down from of 0-6. I could make it smaller but it would look more broken for QA testers.)
* `openstreetmap-openmaptiles.2025-12-10.zoom_0-01.pmtiles` - vector maps (OSM zoom 0-1, down from Natural Earth 0-8)
* `s2maps-sentinel2-2023.2025-12-10.zoom_0-04.pmtiles` - satellite (zoom 0-4, down from 0-7)
* `static-search.2025-12-10.pop-100k-cities.tar.gz` - city search (all 100k+ population towns, down from all 1k+ population towns)

This brings the total down from ~297M to ~63M.

Thanks again to @Ark74 for helping with the hosting side.

RPi OS Trixie

@holta @chapmanjacobd

### DIFF
--- a/roles/maps/tasks/main.yml
+++ b/roles/maps/tasks/main.yml
@@ -49,6 +49,10 @@
           value: "{{ maps_search_nominatim_db }}"
         - option: maps_search_static_db
           value: "{{ maps_search_static_db }}"
+        - option: maps_preset_full_quality_regions
+          value: "{{ maps_preset_full_quality_regions }}"
+        - option: maps_ne6_zoom
+          value: "{{ maps_ne6_zoom }}"
 
   rescue:
 
--- a/roles/maps/templates/tile-extract.py.j2
+++ b/roles/maps/templates/tile-extract.py.j2
@@ -105,9 +105,9 @@ def approve_download_size(extract_name):
         # can control and will likely look into for optimizing user experience.
         pattern = (
             r'\d+/\d+/\d+ \d+:\d+:\d+ extract.go:\d*: '
-            r'Extract transferred (\d*\.?\d*) (KB|MB|GB|TB) '
+            r'Extract transferred (\d*\.?\d*) (kB|MB|GB|TB) '
             r'\(overfetch \d*\.?\d*\) '
-            r'for an archive size of (\d*\.?\d*) (KB|MB|GB|TB)'
+            r'for an archive size of (\d*\.?\d*) (kB|MB|GB|TB)'
         )
         for line in result.stdout.split('\n'):
             re_match = re.match(pattern, line)
@@ -121,7 +121,7 @@ def approve_download_size(extract_name):
 
         # Determine the sizes of all of the pmtiles files in bytes
         unit_mult = {
-            'KB': 1_000,
+            'kB': 1_000,
             'MB': 1_000_000,
             'GB': 1_000_000_000,
             'TB': 1_000_000_000_000,
--- a/vars/local_vars_maps_test.yml
+++ b/vars/local_vars_maps_test.yml
@@ -309,7 +309,7 @@ maps_from_internet_archive: False
 
 # "New" IIAB Maps, as of 2025 and 2026 -- INSTRUCTIONS: https://github.com/iiab/iiab/blob/master/roles/maps/README.md
 
-# This will run on CI so we want to be careful how much it downloads. As of now, maps will download (slightly different from on-disk size) roughly:
+# 2026-03-24: This will run on CI so we want to be careful how much it downloads. As of now, maps will download (slightly different from on-disk size) roughly:
 #
 # 1M   various js/css
 # 7M   naturalearth6-NE2_HR_SR_W_DR-WEBP.2025-12-10.zoom_0-04.pmtiles
--- a/vars/local_vars_maps_test.yml
+++ b/vars/local_vars_maps_test.yml
@@ -309,32 +309,34 @@ maps_from_internet_archive: False
 
 # "New" IIAB Maps, as of 2025 and 2026 -- INSTRUCTIONS: https://github.com/iiab/iiab/blob/master/roles/maps/README.md
 
-# This will run on CI so we want to be careful how much it downloads. As of now, maps will download:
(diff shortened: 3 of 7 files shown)


holta:
Looks good:

And we can fine-tune later where aspects are too small or too large or...

holta:
@orblivion can you make sure maps/README.md is updated — maybe a link to this PR at the bare minimum?

e.g. as part of this PR if it's a quick tweak — or "immediately" in another PR if you prefer?

roles/maps/tasks/main.yml
```diff
@@ -49,6 +49,10 @@
           value: "{{ maps_search_nominatim_db }}"
         - option: maps_search_static_db
           value: "{{ maps_search_static_db }}"
+        - option: maps_preset_full_quality_regions
+          value: "{{ maps_preset_full_quality_regions }}"
+        - option: maps_ne6_zoom
+          value: "{{ maps_ne6_zoom }}"
```
chapmanjacobd, line 55: maybe we use default values so it still works with older local_vars.yams?

chapmanjacobd: oh yeah I guess that is what default_vars does

chapmanjacobd (approved):

vars/local_vars_maps_test.yml
```diff
@@ -308,20 +308,35 @@ osm_vector_maps_enabled: False
 maps_from_internet_archive: False
 
 # "New" IIAB Maps, as of 2025 and 2026 -- INSTRUCTIONS: https://github.com/iiab/iiab/blob/master/roles/maps/README.md
+
+# This will run on CI so we want to be careful how much it downloads. As of now, maps will download (slightly different from on-disk size) roughly:
```
holta: Please include the date more often.  Unambiguous docs should be the goal.  Language like "As of now" is unhelpful years later, and even weeks later (:

```suggestion
# 2026-03-24: This will run on CI so we want to be careful how much it downloads. As of now, maps will download (slightly different from on-disk size) roughly:
```

orblivion:
How would I approach it in the README? Isn't the README for end users? This PR is for CI.

holta:
> Isn't the README for end users?

Personally I don't find the term "end users" to be terribly useful (:  But if we must use that term, roles/maps/README.md is definitely not for educators, learners, parents, etc.

README.md is arguably for tech-savvy implementers primarily, but even that's not the full truth, as much more importantly such README's are absolutely critical for long-term maintenance — and medium-term and short-term too — e.g. a great example is providing very pragmatic insights to testing community volunteers (needing to understand CI context, etc!)

In any case, some kind of a link from maps/README.md and CI details are needed — conveying the basic intuition of what's happening and why — i.e. at the very bare minimum (for starters) a link to this PR and a sentence clarifying what it's for?

orblivion:
I kind of forgot that you suggested just linking to this issue as a holdover. Hopefully this change is in the right ballpark.

I didn't mention that it cuts down about 200M of downloads because the README currently describes the smallest "normal" installation as "170MB". "170MB" doesn't account for the size of the default static search database and the "naturalearth6" pmtiles file (which gives the "natural" look), both of which this new "testing size" shrinks. If we want to account for all that in this README, that's a bigger task to change all of the numbers.

roles/maps/templates/tile-extract.py.j2
holta, line 1: @orblivion is there some reason we have to use "kB" with a lowercase "k" ?

(Naturally persnickety SI folk prefer that — while uppercase "KB" is the global norm in broad usage, for many decades now — so either is fine but let's use the much more common term if there's a choice?)

orblivion: Most of this is about parsing the output of `pmtiles` and we don't have a choice there (short of forking `pmtiles`). However if we're talking about the confirmation output, we can change that. Gimme a second.

orblivion: So to get `"kB"` to show up in the output, I would need a region so small that the sum of terrain vector and satellite would need to be under one MB and I'm not sure that's even possible from testing it right now. The reason it comes up here is that sometimes one of these three alone is less than one MB.

orblivion: I test the size of each file separately, and need to parse it out. For London and/or Nepal here, the regions are small enough that one of the files was less than one MB and it failed to parse.

chapmanjacobd: fwiw kB is technically correct for the SI standard of 1000 bytes and preferred over KB

https://ux.stackexchange.com/questions/13815/files-size-units-kib-vs-kb-vs-kb

- 1mB would be 1/1000 bytes (like millimeter)
- 1MB is mega (1000 1kB) bytes (like megameter)

But I agree it does look a bit weird when Microsoft makes them all the same... I used to be strongly KiB, MiB, GiB 1024 but digging into it a bit more it makes more sense to do 1000 for disks. 

If you use 1024 then people will never be able to estimate how much space they have, eg. "I have a 8GB flash drive but the computer won't let me put an 8GB file on it". But if you use 1000 then at least half the time it will always match

(but when talking about RAM it always makes sense to use 1024)

holta: I'm a proponent of the metric system (SI).

But at least on the UI/UX side (!) we should not paint ourselves into a puritanical corner — when the culturally predominant term (uppercase KB) is in fact what by the masses use — for many decades now.

holta: Where practical anyway (:

(If various upstreams mandate technocratically correct lowercase "k" in "kB", in lower-level areas especially, that occasionally bubble up, no problem :-)

holta:
Good enough to merge?

And also... ?

- PR #4326

Issue #4322, opened 2026-03-23
Multiple issues [re: Quick Picks / presets]


callofduty194-cpu:
Good afternoon!

I installed IIAB using the curl command, as well as sudo iiab after. Worthy to note I also installed global IIAB maps.

Installation seemed to go smooth without any hiccups. Upon completion, I then installed the 256GB English starter pack thru the IIAB browser admin portal.

![Image](https://github.com/user-attachments/assets/3712f4af-a44f-45c2-9cfe-1b47251e6c8a)

![Image](https://github.com/user-attachments/assets/c3fe68d8-f3ae-4565-89f7-0e3f714c982d)

Upon completion, when I try to access Project Gutenberg, it says undefined zim in a menu definition. When I try to access KA lite, it says this site cannot be reached (chrome, ERR_CONNECTION_REFUSED), and lastly when I try to open World Maps & Satellite Photos, I receive a 404 message.

I am not too technically saavy in this side of things.
Being that I have multiple errors, should I just wipe clean and attempt a fresh install? Or are these fixes relatively easy?

Thank you very much!

holta:
Hi @callofduty194-cpu 

1. What do URLs http://10.10.10.10/kiwix and http://localhost/kiwix show?

2. Can you please run `iiab-diagnostics` at the command-line and reply to us with the resulting URL?

3. Note that KA Lite is no longer possible.  This year it's being fully replaced by Kolibri instead.

4. Can you say a bit more about the exact OS, how much RAM, and what kind of hardware?

Thanks!

callofduty194-cpu:
Thank you for the quick response!

1. Being that I can only use box URL's, box/kiwix works but localhost refuses to connect! 

![Image](https://github.com/user-attachments/assets/0666ed97-e4d6-41be-9b0a-37db756cca5d)
![Image](https://github.com/user-attachments/assets/c18ed227-5a1a-4ed7-b513-085e957e2238)

2. Sure! https://paste.centos.org/view/63de29bc
3. Good to know, thank you.
4. I am running a stock raspberry pi 3B+

Thank you again for your help! Very much appreciated.

callofduty194-cpu:
![Image](https://github.com/user-attachments/assets/497335fd-24c0-4aaa-9ab6-b51b6f204f5c)

![Image](https://github.com/user-attachments/assets/0099c9a5-1f48-4e73-9489-0f1bbf36d786)

[...]

Issue #4172, opened 2025-12-11
problem with pip [during Admin Console reinstall of speedtest-cli]


tim-moody:
When installing iiab-admin-console on Intel bare metal:

```
TASK [cmdsrv : Install speedtest-cli into virtual environment] ***********************************************************************
[ERROR]: Task failed: Module failed:
:stderr: Traceback (most recent call last):
  File "/usr/local/speedtest/bin/pip3", line 5, in <module>
    from pip._internal.cli.main import main
ModuleNotFoundError: No module named 'pip'

Origin: /opt/iiab/iiab-admin-console/roles/cmdsrv/tasks/main.yml:26:3

24 #     - download2
25
26 - name: Install speedtest-cli into virtual environment
     ^ column 3

fatal: [127.0.0.1]: FAILED! => {"changed": false, "cmd": ["/usr/local/speedtest/bin/pip3", "install", "--no-cache-dir", "speedtest-cli"], "msg": "\n:stderr: Traceback (most recent call last):\n  File \"/usr/local/speedtest/bin/pip3\", line 5, in <module>\n    from pip._internal.cli.main import main\nModuleNotFoundError: No module named 'pip'\n"}
```

holta:
> When installing iiab-admin-console on Intel bare metal:

Does this happen repeatedly?  On repeated clean//fresh install too?

jvonau:
Details are a bit sparse, no OS, no hardware... Are there any other roles in iiab that use ansible's [pip:](https://github.com/iiab/iiab/pull/3952/commits/c2b634a6d2a3fc8c5e4df5c6b680ea3493e383f4) module failing? Think this was seen before? https://github.com/iiab/iiab-admin-console/issues/575

tim-moody:
Ub 24.04.3,  Intel bare metal. This is an old install that I am trying to upgrade. I had forgotten about the previous report. This is the only use of pip in iiab-admin-console. Did ansible upgrade and now require this to be ansible.builtin.pip?

holta:
Just to recap #575, here's the trick that's so often helpful in recent years, to upgrade Admin Console:

```
rm -rf /usr/local/speedtest
cd /opt/iiab/iiab-admin-console
git pull
./install
```

- Is another tweak also required here?  e.g. if it's now time to automate this workaround or similar?

Issue #4158, opened 2025-11-22
Enhancement: Remove kalite completely from IIAB


tim-moody:
Kalite no longer installs on IIAB and should be removed wherever present in iiab and iiab-admin-console.

holta:
I agree it's time.  There's a ton of documentation & similar cleanup work to do removing KA Lite; I'll get that rolling & completed well before Dec 31.

jvonau:
> completed well before Dec 31.

Little behind in that goal.

holta:
Indeed:

I'd like to get several overdue maintenance tasks like this done within the coming week.

jvonau:
Still pending

Issue #4135, opened 2025-11-06
ADVANCE WARNING: Debian 12 Bookworm and Raspberry Pi OS 12 Bookworm will very soon be desupported / deprecated by IIAB


holta:
Just FYI to keep software quality high, (1) IIAB keeps focus on recent OS's (LTS and non-LTS OS's), (2) when a new generation of an LTS OS is available for more than 15 days, IIAB will seek to deprecate (desupport) the prior generation, in order to avoid fragmenting our very limited resources, across too many OS's / distros.

Background:

- #3416 
- PR #3621 
- #3672 
- #3758 
- PR #4035

jvonau:
How does that effect the 2 Bookworm arm based test VMs that are employed?

holta:
> How does that effect the 2 Bookworm arm based test VMs that are employed?

Good Question:

https://github.com/pguyot/arm-runner-action is an incredible ARM64-on-GitHubActions emulation hack but only offers us relatively ancient [OS images from years ago,](https://github.com/pguyot/arm-runner-action#base_image).

So if the above approach doesn't modernize rather soon, we'll look at new options.

Reality Check: GitHub Actions workflows (CI) never perfectly match what's needed (!) — but the good news is experienced volunteers like @Akatama and @thotmx do a great job helping us evolve these regularly along the way — so I'm confident we'll find a good answer by/around early 2026 or so?

Possibly using [GitHub Actions' new ARM64 runners](https://github.com/actions/partner-runner-images?tab=readme-ov-file#available-images): [1] [2]

- https://github.com/iiab/iiab/blob/master/.github/workflows/40min-iiab-none-deb12-on-rpi3.yml
- https://github.com/iiab/iiab/blob/master/.github/workflows/40min-iiab-unittest-raspios-on-zero2w.yml

Backstory:

1. "Linux arm64 hosted runners now available for free in public repositories (Public Preview)" (January 16, 2025)
   https://github.blog/changelog/2025-01-16-linux-arm64-hosted-runners-now-available-for-free-in-public-repositories-public-preview/
2. "GitHub-hosted arm64 runners" (Jun 22, 2022 - Oct 16, 2025)
   https://github.com/orgs/community/discussions/19197

jvonau:
[...]

PR #4325, opened 2026-03-24
DEPRECATE kalite

jvonau:
### Fixes bug:
https://github.com/iiab/iiab/issues/4322#issuecomment-4118038749
as titled
CI


tim-moody:
At least during transition I would prefer set kalite install and enabled to false in default vars. I would personally just remove from all other _vars files rather than commenting out.

jvonau:
The transition has been in place since #3733 in 2024 while #4158 suggest that admin-console should not offer to install kalite. I'll assume no changes have occurred as 4322 shows that is possible to access the kalite role.

holta:
Looks good!

Let's try to remove & expunge KA Lite rather completely starting now.

(Of course KA Lite was supposed to go away 7 years ago, prior to Python 2's December 31, 2019 expiry date — so now it's really time — including cleaning up all docs which will take a bit longer but I'll belp.)

holta:
```
TASK [js-menu : Install js-menu config file] ***********************************
Error: : Task failed: 'kalite_server_port' is undefined
```

From:

https://github.com/iiab/iiab/actions/runs/23495692678/job/68376047566?pr=4325

tim-moody:
Our vars files have directives (install, enable) and constants (port). We should leave the constants in default vars for backwards compatibility.

holta:
My opinion is that backwards compatibility period has already passed, and that it's time for a proper spring cleaning.

Reason Being: the pragmatic truth is that KA Lite has been out of the picture for years.

tim-moody:
https://iiab.me/en-school/ is running on current ubuntu LTS

holta:
@chapmanjacobd I guess you knew that, but I didn't quite grasp until now that 35min-large-on-rpios-lite.yml tests Admin Console — whereas 30min-large-on-ubuntu.yml does not!

Corollary... this PR's current "30 min" runs succeed on Ubuntu:
https://github.com/iiab/iiab/actions/runs/23495692650/job/68376047670?pr=4325 (attempt 1)
https://github.com/iiab/iiab/actions/runs/23495692650/job/68460436527?pr=4325 (attempt 2)

Whereas its "35 min" runs fail on RPiOS:
https://github.com/iiab/iiab/actions/runs/23495692678/job/68376047566?pr=4325 (attempt 1)
https://github.com/iiab/iiab/actions/runs/23495692678/job/68390370333?pr=4325 (attempt 2)

chapmanjacobd:
yes because the rpios image uses the iiab-factory install script and the ubuntu one is based off the old medium-sized install CI test which uses the install script in-repo

Clicking on the KA Lite links doesn't seem to work:

http://iiab.me:8008/learn/khan/math/algebra-basics/

(But when installing KA Lite manually it seemed half-broken anyway... didn't dig too deeply in it though)

> KA Lite Central Server: Closed access since March 1st 2020
> https://kalite.learningequality.org/

The KA Lite / Kolibri choice is a bit confusing for a new user but leaving the constants in default vars for a bit also makes sense to me

jvonau:
The kalite install has been broken since #4085 Sept 2025 where 'is_linuxmint_22' was removed, but support for Ubuntu-2204 was removed in #3828 back in 2024, Debian-12 should be gone soon #4135

tim-moody:
There is no doubt that kalite no longer installs, so IIAB should no longer support installing kalite. But it still runs on Ubuntu 24.04 (the iiab.me links don't work because of the firewall I forgot about.) My point is that I always want Admin Console to be backwards compatible, so that an existing install, which has kalite, can upgrade to the latest Admin Console and still work.

jvonau:
> There is no doubt that kalite no longer installs, so IIAB should no longer support installing kalite. But it still runs on Ubuntu 24.04 

#3731 Maybe if this installation was upgraded to Ubuntu 24.04 from an earlier Ubuntu release with kalite already installed and python2 didn't get removed during the upgrade. Related: https://github.com/iiab/iiab/issues/4172#issuecomment-3642852332
> (the iiab.me links don't work because of the firewall I forgot about.)

Right, kalite was not part of the transition to [nginx](https://github.com/iiab/iiab/tree/master/roles/nginx) #2762, so there is no proxy and the port has to be used. The kalite support was supposed to withdrawn back then but never was.
>My point is that I always want Admin Console to be backwards compatible, so that an existing install, which has kalite, can upgrade to the latest Admin Console and still work.

In regards to https://github.com/iiab/iiab/pull/4325#issuecomment-4119480542, could be solved by adding the missing variable to:

- re-add kalite_server_port to default_vars in iiab
- adding kalite_server_port to https://github.com/iiab/iiab-admin-console/blob/master/vars/default_vars.yml
- hard-coding the port in https://github.com/iiab/iiab-admin-console/blob/master/roles/js-menu/templates/config.json.j2#L4

Think I'll just re-add kalite_server_port as https://github.com/iiab/iiab/blame/master/roles/network/templates/gateway/iiab-gen-iptables uses that value and changing anything is a PITA.

holta:
Indeed we should have removed all this legacy KA Lite cruft around 2024, or 2025 at the latest.

Lessons Learned 🙃

ADVANCE WARNING: Any remaining KA Lite code and variables will in fact be removed from the iiab/iiab repo, as a follow-up to this PR, in the coming 90 days.

jvonau:
e92754c will allow legacy local_vars files out in the wild to remain functional until a future point in time, with a quick edit to make them current when that point is reached in the future then the error message is returned to point out the problem.

holta:
Good enough!

Thank you to @jvonau.

And spring cleaning will take this further some time in the coming 90 days as that's what spring is for 🌻

holta:
Related:

- PR iiab/iiab-admin-console#639

jvonau:
Why was 'when: iiab_base_ver is version('8.4', 'ge')' removed and the code commented out?

holta:
Suggestion appreciated, but what will happen is rather complete removal of KA Lite within 90 days.  

End of story.

PR #4330, opened 2026-03-29
Specify bash for a couple shell directives

orblivion:
### Fixes bug:

On certain configurations, it seems that bash is not assumed for ansible `shell:` directives. @Akatama ran into this when downloading regions in CI. Hopefully he can confirm that this will fix it.

RasPi OS Trixie

@holta

### DIFF
--- a/roles/maps/tasks/download_large_file.yml
+++ b/roles/maps/tasks/download_large_file.yml
@@ -155,3 +155,5 @@
         echo "aria2c error:" 1>&2
         tac "{{ log_file }}" | sed '/'"Download Progress Summary as of"'/q' | tac 1>&2
         exit 1
+      args:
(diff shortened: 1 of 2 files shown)


roles/maps/tasks/download_large_file.yml
```diff
@@ -155,3 +155,5 @@
         echo "aria2c error:" 1>&2
         tac "{{ log_file }}" | sed '/'"Download Progress Summary as of"'/q' | tac 1>&2
         exit 1
+      args:
+        executable: /bin/bash
```
orblivion, line 159: We probably haven't come across this because it would only happen on certain systems and only if there was a download failure.

holta:
@Akatama can you confirm after this is merged?

PR #4331, opened 2026-03-29
Maps test CI needs region downloader tool

orblivion:
### Fixes bug:

`local_vars_maps_test.yml` fails because it doesn't install the map region downloader utility. I never caught this before because I develop on a machine with it already installed. (And, because we have not started using it for actual CI yet.)

### Notes:

`maps_region_downloader` has two functions: install the downloader tool, and show the downloader button on the map. This fix is annoying for the hypothetical user who wants to install regions via ansible vars but does not want the downloader tool button on their map. They would need to set both variables for the first run and then comment them out both out after.

However, this hypothetical user likely does not exist. The regions-via-ansible feature is intended for CI. If this proves wrong LMK.

RasPi OS Trixie

@Akatama @holta


Akatama (approved):

holta:
Tangentially related:

- PR #4330

PR #4333, opened 2026-03-30
[Lokole] use {{ apache_user }} as a group instead of {{ lokole_user }}

dahagag:
### Fixes bug:
When installing Lokole along with IIAB Admin Console, the `lokole` user gets removed from `www-data` group making nginx unable to acccess Lokole's gunicorn socket.

Add `{{ lokole_user }}` to `{{ apache_user }}` group to allow nginx to access Lokole's services/

- Ubuntu 24.04
@holta @nzola

### DIFF
--- a/roles/lokole/tasks/install.yml
+++ b/roles/lokole/tasks/install.yml
@@ -77,23 +77,16 @@
   ansible.builtin.user:
     state: present
     name: "{{ lokole_user }}"
-    #group: "{{ lokole_user }}"
+    group: "{{ apache_user }}"
     groups: dialout, dip
     system: yes
     #uid: "{{ lokole_uid }}"
     home: "{{ lokole_home_dir }}"
 
-- name: "Add user '{{ apache_user }}' to group '{{ lokole_user }}' for socket access via NGINX"
-  ansible.builtin.user:
-    name: "{{ apache_user }}"
-    groups: "{{ lokole_user }}"
-    append: yes
-
 - name: mkdir {{ lokole_run_dir }}
   file:
     state: directory
     path: "{{ lokole_run_dir }}/lokole_restarter"
-    group: "{{ lokole_user }}"
     owner: "{{ lokole_user }}"
     mode: g+rw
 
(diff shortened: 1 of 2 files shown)


holta:
Thanks @dahagag !

Is this tested enough to be merged?

dahagag:
Currently being tested.

tim-moody:
Without checking, my guess is this will allow {{ lokole_user }} to log into Admin Console.

jvonau:
> ### Fixes bug:
> 
> When installing Lokole along with IIAB Admin Console, the `lokole` user gets removed from `www-data` group making nginx unable to acccess Lokole's gunicorn socket.

https://share.google/aimode/BgmOVcFdI9K3PsBq4
https://docs.ansible.com/projects/ansible/latest/collections/ansible/builtin/user_module.html

Shouldn't https://github.com/tim-moody/iiab-admin-console/blob/master/roles/console/tasks/main.yml#L102-L105 use `append: yes` as to not clobber group memberships that were previously established?

tim-moody:
>When installing Lokole along with IIAB Admin Console, the lokole user gets removed from www-data group

>Shouldn't https://github.com/tim-moody/iiab-admin-console/blob/master/roles/console/tasks/main.yml#L102-L105 use append: yes as to not clobber group memberships that were previously established?

As I understand https://docs.ansible.com/projects/ansible/latest/collections/ansible/builtin/user_module.html, with the default append false it will remove www-data from all groups but shadow, so yes this is a problem. But it is not exactly the problem reported here, which has to do with lokole_user group membership. Not sure why this has emerged as the code goes back to 2020, unless ansible changed. 

I need to test, but I don't think this line is still needed.

tim-moody:
ran install on deb 13 with latest and only lokole install/enabled

after iiab, /etc/groups has 

sudo:x:27:timm,iiab-admin
www-data:x:33:
shadow:x:42:
users:x:100:timm
iiab-admin:x:1001:iiab-admin
lokole:x:989:www-data

after iiab-admin-console install
with shadow group lines removed

the above are all the same

login and change password in Adm Cons worked fine

tim-moody:
So, 

- I expected www-data to be in group www-data by default, so something else is killing that
- www-data has been added to lokole, rather than the other way around, which could explain the first comment

https://github.com/iiab/iiab-admin-console/pull/643 got rid of the shadow lines.

PR #4175, opened 2025-12-12
Let's try reverting to MediaWiki 1.44.3 for PHP 8.4+

holta:
### Fixes bug:

Experimental workaround to deal with Wikimedia Foundation's current inability to support MediaWiki on 2025 OS's like Debian 13 Trixie, Raspberry Pi OS Trixie, Ubuntu 25.10, etc:

- PR #4169

### Description of changes proposed in this pull request:

Hopefully reverting to MediaWiki 1.44.3 (and subsequent versions of 1.44.x) will do the job, until upstream (Wikimedia Foundation) gets Parsoid working on modern OS's with PHP 8.4, 8.5 etc.

### Smoke-tested on which OS or OS's:

Ongoing!

### Mention a team member @username e.g. to help with code review:

@jvonau

### Tangentially related issue:

- #4168

PR #4334, opened 2026-03-31
MediaWiki 1.44.4 security release

holta:
Building on:

- PR #4175


holta:
ASIDE: Just FYI WMF public communications remain ambiguous as to whether MediaWiki 1.45.2 really works on PHP 8.4 or not...

https://www.mediawiki.org/wiki/Parsoid

(Either way, MediaWiki is expected in May or June, and should hopefully finally resolve this — and if we're lucky MediaWiki 1.46.0 will also work on PHP 8.5 that was released back Nov 20, 2025?)

PR #4342, opened 2026-04-07
use is_container instead of rpi_image for container specific exclusions

chapmanjacobd:
### Fixes bug:

Right now it's difficult to exclude things which fail in containers. Moving these away from rpi_image allows for greater flexibility and control.

Container/VM detection - set to True when running inside systemd-nspawn, LXC, etc.

Used to skip kernel-level sysctl and systemd service operations that fail in containers.

@holta

### DIFF
--- a/.github/workflows/10min-unittest-IMAGE-on-rpios-lite.yml
+++ b/.github/workflows/10min-unittest-IMAGE-on-rpios-lite.yml
@@ -91,6 +91,7 @@ jobs:
           cp --preserve=mode,timestamps ../iiab/vars/local_vars_unittest.yml "$MOUNT_DIR/etc/iiab/local_vars.yml"
 
           echo "rpi_image: True" >> "$MOUNT_DIR/etc/iiab/local_vars.yml"
+          echo "is_VM: True" >> "$MOUNT_DIR/etc/iiab/local_vars.yml"
           sed -i "s/^iiab_admin_user_install: True/iiab_admin_user_install: False/" "$MOUNT_DIR/etc/iiab/local_vars.yml"
 
           if ! command -v systemd-nspawn &> /dev/null; then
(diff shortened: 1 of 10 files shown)


chapmanjacobd:
sorry for the noise. I think it's ready now

holta:
@Ark74 can you review this?

holta:
@Akatama if you happen to have time, review this quickly?

Akatama (approved):
I don't know enough about the Ansible configuration itself to say if Jacob missed any of these configurations, but at least what he has done here looks correct.

Ark74 (commented):
@chapmanjacobd so the idea is to cherry-pick some specific tasks that are causing trouble and moving them to a specific "mode", right? Because from what I can tell, both is_VM and rpi_image are applied.

So, like creating a closet on the wall to store the brooms and mops we keep tripping over, right?

chapmanjacobd:
yeah we're just moving some tasks from rpi_image to is_VM to make them more generalized.

Without this PR if I use rpi_image then it will try to install the RPi WiFi firmware on x86_64 Debian... which fails ;-)

Ark74:
Got it!

Ark74 (approved):

jvonau:
One of the side effects of using is_VM is that alters the network work detection to force ['appliance mode' ](https://github.com/iiab/iiab/blob/master/roles/network/tasks/detected_network.yml#L299-L303). The dnsmasq related [file](https://github.com/iiab/iiab/blob/master/roles/network/tasks/enable_services.yml#L52-L57) is not installed and wifi services are not [enabled](url), so the hotspot won't be available upon first boot on the generated images.

https://github.com/iiab/iiab/actions/runs/24139098917/job/70435173310#step:6:7539

chapmanjacobd:
ah okay... hmm... I'll use `is_container` instead then...

Issue #4348, opened 2026-04-11
Maps FQR size estimator failed [9 out of 11] times: what does "Double-check your Internet connection. (error code -9)" really mean?


holta:
@orblivion

cc: @Ark74 @chapmanjacobd

1. ["Full Quality Region" (FQR)](https://github.com/iiab/iiab/blob/master/roles/maps/README.md#full-quality-regions) command [tile-extract.py](https://github.com/iiab/iiab/blob/master/roles/maps/templates/tile-extract.py.j2) below failed 3 out of 5 time over an hour — let's try to find out what's really causing "Double-check your Internet connection. (error code -9)" as this error message does not appear informative or actionable (?) as https://speed.cloudflare.com continues to show 0% packet loss for the home's Internet connection.

   FWIW the failures happened 17 seconds in, 9 seconds in, and 19 seconds in.

2. Any chance that very large FQRs like this 122GB ("north_america") exacerbate the problem — compared to smaller FQRs?

   Is the "Internet connection" error message perhaps referring to some other kind of connectivity problem not visible within speed.cloudflare.com — that's needed by our FQR size estimator?

3. Until we understand this better, should we add some kind of a "please try again" for IIAB implementers, ideally with a status dump to help us figure this out too?  (e.g. should we try to auto-detect "error code -9" and/or similar, recording/logging status at that moment?)

```
root@u26-fqrs:~# time /opt/iiab/maps/tile-extract/tile-extract.py extract north_america -172.474448568,5.615988359,-10.849448568,85.034927591
Determining download size...
Traceback (most recent call last):
  File "/opt/iiab/maps/tile-extract/tile-extract.py", line 258, in <module>
    if not noninteractive and not approve_download_size(extract_name):
                                  ~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^
  File "/opt/iiab/maps/tile-extract/tile-extract.py", line 99, in approve_download_size
    raise Exception(f"Download size check failed. Double-check your Internet connection. (error code {result.returncode})")
Exception: Download size check failed. Double-check your Internet connection. (error code -9)

real    0m16.805s
user    0m11.705s
sys     0m5.782s
```

```
root@u26-fqrs:~# ping mit.edu
PING mit.edu (23.50.69.136) 56(84) bytes of data.
64 bytes from a23-50-69-136.deploy.static.akamaitechnologies.com (23.50.69.136): icmp_seq=1 ttl=56 time=8.34 ms
^C
[...]

PR #4349, opened 2026-04-11
Turn on Maps FQR (Full Quality Region) downloader by default

holta:
@orblivion will rework the download pop-up — so that IIAB implementer/operator has some kind of immediate (push-button) option to lock down this [FQR downloading](https://github.com/iiab/iiab/blob/master/roles/maps/README.md#full-quality-regions) feature — soon after the "last" FQR region is selected/specified/downloaded!

Related:

- #4348

### DIFF
--- a/roles/maps/README.md
+++ b/roles/maps/README.md
@@ -135,26 +135,26 @@ You can download rectangular "Full Quality Regions" (FQRs) to supplement your lo
 
 DETAILS: IIAB's downloadable regions (FQRs) include OpenStreetMap vector data up to [zoom level](https://wiki.openstreetmap.org/wiki/Zoom_levels) 14 (overzoomable to about zoom level 18), satellite photo data up to zoom level 13, and 3D terrain up to zoom level 10.  (As of April 2026, [Map Search data](#how-do-i-install-map-search) is not yet affected, no matter how many FQR regions you download!)
 
-### Installation
+### Prerequisites
 
-PREREQ: Confirm that at least a [minimum IIAB Maps](#whats-a-minimum-iiab-maps-install) is installed!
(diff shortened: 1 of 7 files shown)


holta:
Good enough for now?

(@chapmanjacobd @orblivion if you see problems/shortcomings, LMK for the next round!)

roles/maps/README.md
```diff
@@ -233,7 +244,7 @@ sudo git pull
 After [changing any IIAB Maps variables](#whats-a-minimum-iiab-maps-install) in `/etc/iiab/local_vars.yml`, you can now "reinstall" IIAB Maps to enact your new settings, by running:
 
 ```
-sudo ./runrole --reinstall maps
+sudo ./runrole maps --reinstall
```
chapmanjacobd, line 247: Might be good to keep using this order:

```
command [options] [arguments]
```

otherwise it seems like maps is a subcommand

chapmanjacobd (commented):
.

roles/maps/README.md
holta: Ok: might as well stick to the old POSIX / BSD standard.

In contrast, FWIW Linux Mint requires `apt dist-upgrade -y` and `apt autoremove -y`
Whereas these fail on Linux Mint: `apt -y dist-upgrade` and `apt -y autoremove`

chapmanjacobd: Those are good examples of subcommands :⁠-⁠) ie. apt install and apt search accept different flags and arguments

There definitely are commands that only work the Windows way where options are often last if not using /flag notation... But for the most part unix programs will support options before arguments and options after positionals or mixed between positions may not always be supported

Issue #4311, opened 2026-03-16
Kolibri 0.19.2 and 0.19.3 Python 3.14 Django problem on Ubuntu 26.04 ?  (OS final release ETA is April 23)


holta:
Are others experiencing the same?

```
TASK [kolibri : Enable & Start 'kolibri' systemd service, if kolibri_enabled] ***************************************************************************************************************************************************************
[ERROR]: Task failed: Module failed: Unable to start service kolibri: Job for kolibri.service failed because the control process exited with error code.
See "systemctl status kolibri.service" and "journalctl -xeu kolibri.service" for details.

Origin: /opt/iiab/iiab/roles/kolibri/tasks/enable-or-disable.yml:1:3

1 - name: Enable & Start 'kolibri' systemd service, if kolibri_enabled
    ^ column 3

fatal: [127.0.0.1]: FAILED! => {"changed": false, "msg": "Unable to start service kolibri: Job for kolibri.service failed because the control process exited with error code.\nSee \"systemctl status kolibri.service\" and \"journalctl -xeu kolibri.service\" for details.\n"}
```

```
root@box:~# python3 --version
Python 3.14.3
```

`journalctl -u kolibri` shows...

```
Mar 16 14:27:05 box systemd[1]: Starting kolibri.service - Kolibri...
Mar 16 14:27:05 box kolibri[49123]: /usr/lib/python3/dist-packages/kolibri/dist/requests/__init__.py:102: RequestsDependencyWarning: urllib3 (1.26.20) or chardet (5.2.0)/charset_normalizer (2.0.12) doesn't match a supported version!
Mar 16 14:27:05 box kolibri[49123]:   warnings.warn("urllib3 ({}) or chardet ({})/charset_normalizer ({}) doesn't match a supported "
Mar 16 14:27:05 box kolibri[49123]: /usr/lib/python3/dist-packages/kolibri/dist/magicbus/process.py:25: SyntaxWarning: "\ " is an invalid escape sequence. Such sequences will not work in the future. Did you mean "\\ "? A raw string is a>
Mar 16 14:27:05 box kolibri[49123]:   \         A
Mar 16 14:27:05 box kolibri[49123]: /usr/lib/python3/dist-packages/kolibri/dist/magicbus/plugins/servers.py:20: SyntaxWarning: "\ " is an invalid escape sequence. Such sequences will not work in the future. Did you mean "\\ "? A raw str>
[...]

PR #4351, opened 2026-04-13
django fix for U2604 [for Kolibri 0.19.3]

jvonau:
### Fixes bug:
#4311
temp workaround
2604-multipass vm

### DIFF
--- a/roles/kolibri/tasks/install.yml
+++ b/roles/kolibri/tasks/install.yml
@@ -180,6 +180,12 @@
     src: kolibri.service.j2
     dest: /etc/systemd/system/kolibri.service
 
+- name: 'Temp fix for django'
+  when: python_version is version('3.14', '>=')
+  template:
+    src: django.fix
+    dest: /usr/lib/python3/dist-packages/kolibri/dist/django_filters/__init__.py
+
 - name: Stop 'kolibri' systemd service, for Kolibri provisioning (after daemon_reload)
   systemd:
     daemon_reload: yes
--- a/roles/kolibri/templates/django.fix
+++ b/roles/kolibri/templates/django.fix
@@ -1,0 +1,30 @@
+# flake8: noqa
+import importlib.util
+
+from .filters import *
+from .filterset import FilterSet
+
+# We make the `rest_framework` module available without an additional import.
+#   If DRF is not installed, no-op.
+if importlib.util.find_spec('rest_framework') is not None:
+from . import rest_framework
+
+
+__version__ = '21.1'
+
+
+def parse_version(version):
+'''
+    '0.1.2.dev1' -> (0, 1, 2, 'dev1')
+    '0.1.2' -> (0, 1, 2)
+    '''
+v = version.split('.')
+ret = []
+for p in v:
+if p.isdigit():
+ret.append(int(p))
+else:
+ret.append(p)
+return tuple(ret)
+
+VERSION = parse_version(__version__)
--- a/roles/kolibri/templates/django.fix
+++ b/roles/kolibri/templates/django.fix
@@ -9,22 +9,21 @@ from .filterset import FilterSet
 if importlib.util.find_spec('rest_framework') is not None:
     from . import rest_framework
 
-
 __version__ = '21.1'
 
 
 def parse_version(version):
-'''
+    '''
     '0.1.2.dev1' -> (0, 1, 2, 'dev1')
     '0.1.2' -> (0, 1, 2)
     '''
-v = version.split('.')
-ret = []
-for p in v:
-if p.isdigit():
-ret.append(int(p))
-else:
-ret.append(p)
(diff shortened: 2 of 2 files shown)


holta:
Thanks @jvonau for having confirmed precisely:

- #4311

@rtibbles any idea if Kolibri 0.19.4 is expected this month?

Related:

- learningequality/kolibri#13823
- PR learningequality/kolibri#14395

cc: @Ark74

chapmanjacobd (approved):

chapmanjacobd:
yeah I think the main thing blocking py3.14 support is Kolibri's dep `django==3.2.25`. Django 3.2 LTS has been unsupported for about two years now: https://www.djangoproject.com/download/

https://github.com/search?q=repo%3Alearningequality%2Fkolibri+django+path%3Arequirements&type=code

But this workaround should fix the kolibri.service error:

    AttributeError: module 'pkgutil' has no attribute 'find_loader'

edit: ahh I see 0.19.4. yeah we should probably just wait then

holta:
> yeah I think the main thing blocking py3.14 support is Kolibri's dep `django==3.2.25`. Django 3.2 LTS has been unsupported for about two years now: https://www.djangoproject.com/download/
> 
> https://github.com/search?q=repo%3Alearningequality%2Fkolibri+django+path%3Arequirements&type=code
> 
> But this workaround should fix the kolibri.service error:
> 
> ```
> AttributeError: module 'pkgutil' has no attribute 'find_loader'
> ```
> 
> edit: ahh I see 0.19.4. yeah we should probably just wait then

@rtibbles is Kolibri 0.19.4 forthcoming soon, to obviate the need for this patch, if you know?

jvonau:
>AttributeError: module 'pkgutil' has no attribute 'find_loader'

That is just python removing deprecated code, the world moves forward.
ChatGPT:

Let’s break that error message down into plain language:

### The message

> **module 'pkgutil' has no attribute 'find_loader'**

### What it means

* **module**: In Python, a *module* is just a file (or library) that contains code you can use.
  Example: `math`, `os`, or `pkgutil`.

* **`pkgutil`**: This is a built-in Python module used for working with packages (collections of modules).

* **attribute**: An *attribute* is something inside a module—like a function or variable.
  Example: in `math.sqrt`, `sqrt` is an attribute of the `math` module.

* **`find_loader`**: This is (or rather, *was*) a function that used to exist in `pkgutil`. It helped Python locate and load modules.

---

### So what’s the problem?

Python is saying:

👉 “You’re trying to use `pkgutil.find_loader`, but that function doesn’t exist anymore.”

This usually happens because:

* The code was written for an **older version of Python**
* You’re running it on a **newer Python version** where `find_loader` was removed or changed

---

### Simple analogy

Think of `pkgutil` like a toolbox:

* Old version of the toolbox had a tool called **`find_loader`**
* New version removed that tool and replaced it with something else
* Your code is still trying to use the old tool → error

---

### Example

#### Old code (breaks now):

python
import pkgutil

loader = pkgutil.find_loader("json")
```

#### What happens:

Python says: ❌ “`find_loader` doesn’t exist”

---

### Modern alternative

In newer Python versions, you’d use something like:

python
import importlib.util

spec = importlib.util.find_spec("json")
```

This does a similar job (finding a module), but uses the newer system.

---

### Why this happens in your case (Django/Kolibri)

The page you’re on is about fixing compatibility issues. That error likely appears because:

* Kolibri or Django code is using **old Python APIs**
* Your environment is running a **newer Python (like 3.10+)**

---

### Quick summary

* `pkgutil` = a Python helper module
* `find_loader` = an old function that no longer exists
* Error = code is outdated for your Python version
* Fix = replace it with newer methods (like `importlib`)

---

rgarrison1:
@jvonau 
Ubuntu Server 26.04 LTS 
HP 14-bw065nr with 8 GB RM and 120 GB emmc 

sudo apt update && upgrade -y
sudo apt install mc
sudo curl iiab.io/install.txt | bash -s 4351 | tee install1.txt
Edited local_vars.yml to change timeout from 100 to 700
sudo iiab | tee results.txt
Select password, option 2
Failed on kolibri
#    I'll leave it in this state in case you have any guidance. (editing this post)
I'll not make any changes to local_vars, and make a second pass

see attached
[install1.txt](https://github.com/user-attachments/files/27270332/install1.txt)
[results.txt](https://github.com/user-attachments/files/27270333/results.txt)

chapmanjacobd:
@rgarrison1 I'm getting similar errors in CI https://github.com/chapmanjacobd/iiab/actions/runs/25204590640/job/73902487966?pr=4389#step:10:3297 so that might just be something that will fix itself eventually once people realize that it isn't working reliably

jvonau:
https://github.com/iiab/iiab/issues/4386#issuecomment-4367342446

PR #4354, opened 2026-04-14
use iiab-factory virt-detect instead of waiting for "photographed"

chapmanjacobd:
related to https://github.com/iiab/iiab-factory/pull/259

@holta @Ark74


holta:
@Akatama @Ark74 review after merge?  If ya can!

PR #4356, opened 2026-04-14
autoremove for Images

jvonau:
### Fixes bug:
Removes unused packages found in https://github.com/iiab/iiab/actions/runs/24397037494/job/71257366308#step:6:337
The newest RpiOS was released with the newest packages with less than ideal dependencies  https://github.com/iiab/iiab/pull/4291#issuecomment-4076354199
CI

### DIFF
--- a/.github/workflows/10min-unittest-IMAGE-on-rpios-lite.yml
+++ b/.github/workflows/10min-unittest-IMAGE-on-rpios-lite.yml
@@ -121,6 +121,7 @@ jobs:
           echo 'MODULES=most' > /etc/initramfs-tools/conf.d/nspawn.conf
           apt -y remove libcamera*
           apt -y full-upgrade
+          apt -y autoremove
 
           apt install curl
           curl -fLo /usr/sbin/iiab https://raw.githubusercontent.com/iiab/iiab-factory/master/iiab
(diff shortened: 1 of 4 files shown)


jvonau:
random failure https://github.com/iiab/iiab/actions/runs/24410774390/job/71306643406?pr=4356#step:10:4919
>Error: https://registry.yarnpkg.com/object-assign/-/object-assign-4.1.1.tgz: Request failed \"500 Internal Server Error\"

holta:
Optional: any chance we should follow Linux Mint, which requires `apt autoremove -y` instead of `apt -y autoremove` ?

cc: @chapmanjacobd

holta:
GitHub Actions is now we-running our "30min" CI to see if the error in https://github.com/iiab/iiab/actions/runs/24410774390/job/71306643406?pr=4356 recurs 4h later?

Ongoing here: https://github.com/iiab/iiab/actions/runs/24410774390/job/71348685860?pr=4356

chapmanjacobd (approved):
I guess we can always `apt-mark` if we find something breaks later...

chapmanjacobd:
> Linux Mint, which requires apt autoremove -y instead of apt -y autoremove

probably a non-issue here as these changes are only for the RPi images and it seems to run fine with either syntax.

Gemini says "In Linux Mint (and all Debian-based systems), the placement of the -y flag is flexible. [apt] supports both syntaxes, and they function identically." So maybe things have changed from before. I don't doubt that there was a time when apt didn't support "global flags" though

holta:
FWIW the `apt` command is/was quite different on Linux Mint in recent years — requiring the `-y` flag be on the right side...

(...I should recheck this year, soon if possible... and if necessary testing it again after Linux Mint 23 emerges this summer!)

PR #4357, opened 2026-04-14
Update vector data to April 2026

orblivion:

* Update osm vector data to April 1 2026 (satellite, terrain, and even naturalearth vector doesn't seem to change)
* Change zoom naming scheme from (for example) `zoom_0-07` to `z00-07`
* Change zoom naming scheme from `full` to (in the case of OSM) `z00-z14`. i.e. explicitly name the zoom levels.
* Add `osm-z11` option.

#### So now the way this works:

As we add a new vector file and update the `vector_data_date`, Ansible will download the new file and update the symlink that index.html looks for.

FQRs are managed via the file `extracts.json` which `index.html` reads to know which FQR files exist. This PR changes the format of `extracts.json` to include the date, which `index.html` now needs to find the file, because it can no longer assume the date based on `vector_data_date`. There's no telling how old the FQRs are. The FQR managing script now will add the dates to `extracts.json` when it updates it. It is also a bit more careful about deleting them, in case stray files with different dates are somehow sitting around.

RasPi OS Trixie

@holta

### DIFF
--- a/roles/maps/README.md
+++ b/roles/maps/README.md
@@ -48,7 +48,7 @@ Here are 3 examples below, to help you decide what you'll put in [/etc/iiab/loca
 3. Or if you want **~168 GB** = 78 GB vector (Higher detail, up to zoom 14 [users can overzoom to zoom 18], including 3D buildings, from OpenStreetMap) + 80 GB satellite (up to zoom 12), include:
 
    ```
-   maps_vector_quality: osm-full
+   maps_vector_quality: osm-z14
    maps_satellite_zoom: 12
    ```
 
--- a/roles/maps/defaults/main.yml
+++ b/roles/maps/defaults/main.yml
@@ -117,10 +117,15 @@ maps_dot_black_vector_tiles:
   # maps_vector_quality = "osm-z14"
   osm-z14: "openstreetmap-openmaptiles.{{ maps_vector_data_date }}.z00-z14.pmtiles"
 
-  # "medium res" osm, up to zoom level 9 (original file has 14).
+  # (old) "medium res" osm, up to zoom level 9 (original file has 14).
   # (TODO does this include colors and topography? Or is it used along with naturalearth6 above in most styles?)
   # maps_vector_quality = "osm-z9"
   osm-z9: "openstreetmap-openmaptiles.{{ maps_vector_data_date }}.z00-09.pmtiles"
+
+  # (new) "medium res" osm, up to zoom level 11 (original file has 14).
+  # (TODO does this include colors and topography? Or is it used along with naturalearth6 above in most styles?)
+  # maps_vector_quality = "osm-z11"
+  osm-z11: "openstreetmap-openmaptiles.{{ maps_vector_data_date }}.z00-11.pmtiles"
 
   # "low res" - mostly borders, rivers, country names, large roads.
   # maps_vector_quality = "ne"
--- a/roles/maps/defaults/main.yml
+++ b/roles/maps/defaults/main.yml
@@ -98,10 +98,10 @@ maps_dot_black_naturalearth6_symlink_name: "naturalearth6-NE2_HR_SR_W_DR-WEBP.pm
 # Mostly colors, topography, etc.
 maps_dot_black_naturalearth6_tiles:
   # For actual users
-  full: "naturalearth6-NE2_HR_SR_W_DR-WEBP.{{ maps_slow_data_date }}.z00-06.pmtiles"
+  full: "naturalearth6-NE2_HR_SR_W_DR-WEBP.{{ maps_slow_data_date }}.z00-z06.pmtiles"
 
   # FOR TESTING ONLY
-  ci: "naturalearth6-NE2_HR_SR_W_DR-WEBP.{{ maps_slow_data_date }}.z00-04.pmtiles"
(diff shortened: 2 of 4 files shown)


roles/maps/defaults/main.yml
```diff
@@ -220,9 +230,9 @@ static_search_symlink_name: world-map
 static_search_data:
   # Cities-only static database
   # maps_search_static_db = "pop-1k-cities"
-  pop-1k-cities: "static-search.{{ maps_data_date }}.pop-1k-cities"
+  pop-1k-cities: "static-search.{{ maps_slow_data_date }}.pop-1k-cities"
```
orblivion, line 233: Per the above comments this was supposed to be the "slow" date anyway. But I'll change it to the new "search data date" when I update search.

roles/maps/README.md
```diff
@@ -48,7 +48,7 @@ Here are 3 examples below, to help you decide what you'll put in [/etc/iiab/loca
 3. Or if you want **~168 GB** = 78 GB vector (Higher detail, up to zoom 14 [users can overzoom to zoom 18], including 3D buildings, from OpenStreetMap) + 80 GB satellite (up to zoom 12), include:
 
    ```
-   maps_vector_quality: osm-full
+   maps_vector_quality: osm-z14
```
orblivion, line 51: Don't want to update z9->z11 in the README until I we try downloading it and get the file size as reported locally.

roles/maps/templates/tile-extract.py.j2
```diff
@@ -29,24 +29,28 @@ def validate_noninteractive(noninteractive):
...
 def get_extract_files(extract_name):
-    MAPS_DATA_DATE = "{{ maps_data_date }}"
+    MAPS_VECTOR_DATA_DATE = "{{ maps_vector_data_date }}"
     MAPS_SLOW_DATA_DATE = "{{ maps_slow_data_date }}"
     IIAB_MAP_HOST_URL = "{{ iiab_map_host_url }}"
 
-    def get_paths(base):
+    VECTOR_ZOOM_RANGE = "z00-z{{ '%02d' % maps_vector_max_zoom }}"
+    SATELLITE_ZOOM_RANGE = "z00-z{{ '%02d' % maps_satellite_max_zoom }}"
+    TERRAIN_ZOOM_RANGE = "z00-z{{ '%02d' % maps_terrain_max_zoom }}"
```
orblivion, line 44: I know that we don't need zero-padding for any of the existing numbers but I want to future-proof it.

orblivion:
https://github.com/orblivion/iiab/actions/runs/24413611959/job/71316798194

At the end of Build Image:

```
Ansible quit before photograph. Exiting with failure.
Error: Process completed with exit code 1.
```

Any idea @chapmanjacobd ? I don't see any obvious actual cause.

I looked for "error" above in the logs and found this, along with a few other thing that fail the same way. I'm guessing unrelated, since it didn't stop the installation right away:

```
TASK [yarn : debug] ************************************************************
Warning: : Encountered 1 template error.
error 1 - 'yarn_installed' is undefined
Origin: /opt/iiab/iiab/roles/yarn/tasks/main.yml:26:10

24     var: yarn_enabled
25 - debug:
26     var: yarn_installed
            ^ column 10

ok: [127.0.0.1] => {
    "yarn_installed": "<< error 1 - 'yarn_installed' is undefined >>"
}
```

chapmanjacobd:
Ah I think you'll need to merge main back into this branch? although the tests are passing now? ah yes it's only on your branch--

So you don't need to do anything really... the tests pass in here which test the changes in your branch applied on top of master. If you want the error in your branch to go away you can merge master back in

holta:
I converted this PR to draft, as @orblivion is reworking FQR (full quality region) aspects, which need to fit in with our new filename scheme!

orblivion:
I think that covers it. I did a fair amount of Python and JS now for the FQR stuff so @chapmanjacobd your review would be valuable.

roles/maps/defaults/main.yml
```diff
@@ -30,7 +29,7 @@ maps_static_search_root: "{{ maps_serve_path }}/{{ maps_static_search_dir }}"
 # date designations down the line. I just want one "slow" date to set all of
 # these items *for now*.
 #
-maps_data_date: "2025-12-10"
```
chapmanjacobd, line 33: no lingering references to `maps_data_date` anywhere, right?

chapmanjacobd (approved):

chapmanjacobd:
looks good

roles/maps/defaults/main.yml
orblivion: Double checked, nothing in the whole repo.

orblivion:
Some followups I can do:

* For the `extracts.json` generator script, I give a warning if some region is incomplete, i.e. only osm and satellite but not terrain. I could add a similar warning if there are multiple dates for the same thing. And double check that it's using the latest one if so (though because of alphabetical ordering this might accidentally already be happening).

* For osm zoom 11 I can test it out and then add it to the README.

orblivion:
Testing with a couple different quality levels for world map. Not very high quality for time sake. Though I did test adding FQRs. I also tested deleting FQRs. Anything else I should try before we merge?

PR #4368, opened 2026-04-17
Make satellite optional in maps

orblivion:

* Add `none` option for satellite
    * Don't download satellite in this case
    * Don't allow user to choose hybrid or satellite styles in world map (but do allow in FQR)
* Clean up related js code a bit now that logic is getting more hairy (a bit more to come in followup PR) 
* Add warning about satellite license to README

Ras Pi OS Trixie

@holta @chapmanjacobd

### DIFF
--- a/roles/maps/tasks/install_frontend.yml
+++ b/roles/maps/tasks/install_frontend.yml
@@ -61,23 +61,25 @@
     state: link
   when: maps_vector_quality == "nat-z8"
 
-- name: "Make sure maps_satellite_zoom has a valid value"
-  assert:
-    that: maps_dot_black_satellite_tiles[maps_satellite_zoom] is defined
-    fail_msg: "Error: maps_satellite_zoom is set to an invalid value. See `maps_dot_black_satellite_tiles` in https://github.com/iiab/iiab/blob/master/roles/maps/defaults/main.yml for valid values."
-    quiet: yes
+- when: maps_satellite_zoom != "none"
+  block:
+    - name: "Make sure maps_satellite_zoom has a valid value"
+      assert:
+        that: maps_dot_black_satellite_tiles[maps_satellite_zoom] is defined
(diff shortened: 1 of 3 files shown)


holta:
@chapmanjacobd review after merge, if you can?

chapmanjacobd (approved):

roles/maps/tasks/install_frontend.yml
```diff
@@ -61,23 +61,25 @@
...
-- name: Select satellite tiles
-  file:
-    src: "{{ maps_serve_path }}/{{ maps_dot_black_satellite_tiles[maps_satellite_zoom] }}"
-    dest: "{{ maps_serve_path }}/{{ maps_dot_black_satellite_symlink_name }}"
-    state: link
+    - name: Select satellite tiles
+      file:
+        src: "{{ maps_serve_path }}/{{ maps_dot_black_satellite_tiles[maps_satellite_zoom] }}"
+        dest: "{{ maps_serve_path }}/{{ maps_dot_black_satellite_symlink_name }}"
+        state: link
```
chapmanjacobd, line 82: could add a `state: absent` if maps_satellite_zoom == "none"

orblivion: Hmm that's interesting. So it would be a signal to the sort of implementer who looks at the serving directory that the satellite isn't there. Though if they tried to add the symlink manually after deploying with "none", it wouldn't be enough to make it work again. @holta does this matter and if so which way?

orblivion: To be clear, @chapmanjacobd is suggesting that if we set satellite to "none", we not only disable it in the index.html, we also delete the symlink. It would have no effect on the behavior of the app but maybe it's more clear for implementers?

And we could do the same for terrain.

holta: Making the current state instantly visible via symlink(s) sounds great to me!

(If that makes everything a bit easier to mange & a bit easier to understand I hope, then go ahead @orblivion !?)

PR #4369, opened 2026-04-17
More maps index.html cleanup

orblivion:

A followup to #4368 with some more cleanups that make the code hopefully easier to follow:

* Delete a couple unused variables
* Rename a couple variables
    * if a function called `getMapVectorType` returns a variabled called `MAPS_VECTOR_TYPE` only *some* of the time, something is not named quite right.
    * I don't need everything to be called `MAPS_*`. Unlike in the ansible namespace, everything here is maps.
* Pull a small bit of logic into functions. Partially for repetition, partially so the function name explains what's going on as it's being called.

RasPi OS Trixie

@chapmanjacobd

### DIFF
--- a/roles/maps/templates/index.html.j2
+++ b/roles/maps/templates/index.html.j2
@@ -242,7 +242,7 @@
                 mb.globe = globe === 'g'
 
                 // contour lines is another thing that uses terrain maps, which we can consider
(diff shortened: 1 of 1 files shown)


holta:
@orblivion @chapmanjacobd looking good for a quick merge?

chapmanjacobd (approved):
nice!

PR #4372, opened 2026-04-18
Map search length and placeholder

orblivion:
#### More search results shown.

At least for now, instead of looking at user agent and trying to figure out "what counts as a phone but not a tablet?" I figure we could cut to the chase and look at the window size. (Not even screen size; if a desktop user shrinks their window we may as well accommodate that as well). For now for "small windows" I stay at 5 search results, for larger I do 10.

We may want to tweak all of this. For instance on my phone in landscape mode I only see 4 but on portrait mode I can fit all 5. But I haven't made anything worse.

#### List city count in placeholder text

In my data repository, [I made a change](https://github.com/iiab/maps2/commit/981e492ed6f10caba6584e87062659e4e4e2afd6) to add the city count to the static search database, primarily so that we can easily see it to add it to the README. It's not in the currently released database, but soon. In the mean time, so long as it's there, I thought I'd experiment with showing it in the search bar as the placeholder text, as seen here:

search

Again this number is not in the currently released database so until it's released it'll still say "Search".

I can also remove or revise this. It was an idea. I may have seen it in the previous IIAB osm-vector-maps actually, maybe I got the idea from there.

RasPi OS Trixie

@holta

### DIFF
--- a/roles/maps/files/static_search.js
+++ b/roles/maps/files/static_search.js
@@ -18,6 +18,7 @@ var AddressTextualIndex = class {
     const data = await (await this.fetcher(`${this.baseURL}/index_metadata.json`)).json();
     this.stopWords = new Set(data.stopwords);
     this.minLength = data.token_length;
+    this.numCities = data.num_cities;
   }
   async fileSearch(queryTokens) {
     let results = [];
--- a/roles/maps/files/static_search.js
+++ b/roles/maps/files/static_search.js
@@ -18,7 +18,7 @@ var AddressTextualIndex = class {
     const data = await (await this.fetcher(`${this.baseURL}/index_metadata.json`)).json();
     this.stopWords = new Set(data.stopwords);
     this.minLength = data.token_length;
-    this.numCities = data.num_cities;
+    this.numCities = data.num_cities || 163042;
   }
   async fileSearch(queryTokens) {
     let results = [];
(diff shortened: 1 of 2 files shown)


holta:
> search
> Again this number is not in the currently released database so until it's released it'll still say "Search".

Good enough!  Of course snarky urbanists / planners / geographers will tell us that a town of 1000 people is not big enough to be a city — which typically have minimum populations of (something like!) 10,000 people (:

holta:
@Ark74 and @chapmanjacobd would you be able to review quickly, so this can be merged today?

holta:
> > search
> > Again this number is not in the currently released database so until it's released it'll still say "Search".
> 
> Good enough! Of course snarky urbanists / planners / geographers will tell us that a town of 1000 people is not big enough to be a city — which typically have minimum populations of (something like!) 10,000 people (:

Or better yet, consider hard coding "Search 163,042+ places" in the short-term, maybe even in this PR?

(So we get immediate feedback, allowing this UI/UX to evolve very quickly.)

chapmanjacobd (approved):

chapmanjacobd:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/toLocaleString

holta:
Merge ASAP to get wider feedback?

cc: @Ark74 @avni

holta:
@orblivion

Is it intended for "Search 163,042 places" to disappear from the search textfield, when you click most of the buttons near the top-right?

- Enable globe / Disable globe
- Enable terrain / Disable terrain
- Select Map Style
  - Natural Map
  - Political Map
  - Satellite
  - Satellie + Map

holta:
Hardly the end of the world if "Search 163,042 places" disappears (from the search textfield) !

(But possibly the look & feel should be consistent across time?)

PR #4376, opened 2026-04-19
Set geocoder placeholder on every map refresh

orblivion:
### Fixes bug:

Custom map search placeholder (that shows the number of places to search for) disappears when you turn the globe on or off, or change the map style.

Refresh the map placeholder in an existing function that sets up the map after each "refresh". The way maps.black is designed, it sort of rebuilds the map every time you change one of these settings.

RasPiOS Trixie

@holta


holta:
Original issue described at the bottom of this PR here:

- https://github.com/iiab/iiab/pull/4372#issuecomment-4275159248

PR #4382, opened 2026-04-22
New static search database

orblivion:
### Fixes bug:

A couple thousand cities were not reachable in the static search database. I made tests to narrow them down and then fixed the problems. Some of it was special characters that we had to ignore. Some of it was cities with names shorter than the usual minimum token length we assume.

Draft until the new database is uploaded to our host, after which I'll test it front to back.

#### Major Items

* New database file
* New code that uses it.
* A bunch of [tests](https://github.com/iiab/maps2/blob/static-search-2026-04-22.1/build-databases/static/test-db/index.js) to confirm the improvements, and which can help us carefully add more.

#### Search Improvements

* All cities are now actually searchable
* Better sorting: Now in addition to population, we consider:
    * Exact match (primarily of city name, secondarily of state and country)
    * Proximity to center of map view
* Separate city, state, country fields - for now used in sorting but it could be useful for UI

Note: The new database is not compatible with the old `static_search.js`, and vice versa. For now we're assuming that users will install via ansible. When users are swapping databases manually we have to be more careful.

RasPiOS Trixie

@holta @chapmanjacobd

### DIFF
--- a/roles/maps/defaults/main.yml
+++ b/roles/maps/defaults/main.yml
@@ -219,6 +219,8 @@ nominatim_data:
   # maps_search_nominatim_db = "full"
   full: "nominatim.{{ maps_search_data_date }}.full.sqlite"
 
+static_search_symlink_name: world-map
+
 static_search_data:
   # Cities-only static database
   # maps_search_static_db = "pop-1k-cities"
--- a/roles/maps/defaults/main.yml
+++ b/roles/maps/defaults/main.yml
@@ -208,16 +208,17 @@ maps_terrain_max_zoom: 10
 
 nominatim_db_symlink_name: nominatim.sqlite
 
+# Keeping nominatim on maps_slow_data_date until we actually update it again
 nominatim_data:
   # Basic nominatim database
   # maps_search_nominatim_db = "basic"
-  basic: "nominatim.{{ maps_search_data_date }}.basic.sqlite"
+  basic: "nominatim.{{ maps_slow_data_date }}.basic.sqlite"
     # California admin+natural for now. TODO make a small worldwide one. (unless we go to the frontend-only one)
     # TODO - Make a basic small whole-world map, at least as good as previous maps
 
   # Full nominatim database
   # maps_search_nominatim_db = "full"
-  full: "nominatim.{{ maps_search_data_date }}.full.sqlite"
+  full: "nominatim.{{ maps_slow_data_date }}.full.sqlite"
 
 static_search_symlink_name: world-map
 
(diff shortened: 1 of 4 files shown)


roles/maps/defaults/main.yml
```diff
@@ -210,22 +211,20 @@ nominatim_db_symlink_name: nominatim.sqlite
 nominatim_data:
   # Basic nominatim database
   # maps_search_nominatim_db = "basic"
-  basic: "nominatim.{{ maps_slow_data_date }}.basic.sqlite"
+  basic: "nominatim.{{ maps_search_data_date }}.basic.sqlite"
```
orblivion: The Nominatim dbs are going to be a bit of an albatross. I guess we just have a different date for everything.

orblivion: (I'm not updating nominatim here)

roles/maps/tasks/install_static_search.yml
```diff
@@ -15,10 +15,3 @@
     dest_base_path: "{{ maps_static_search_root }}"
     item: "{{ static_search_data[maps_search_static_db] }}"
     is_archive: true
-
```
orblivion, line 18: So here is where I remove the search database symlink for the sake of cache busting. Maybe we already decided against this, I forgot where (or if) the conversation on this concluded.

I'll add one more argument in favor: In the case of search in particular, an old cache won't lead to search not working at all. It'll be search sort of working. That could be confusing for us as we develop it. The transition from the first db to this one will be more obviously broken because we're adding admin1 and country fields. But later on if there are more subtle fixes it won't be so obvious. At least that's my prediction.

roles/maps/defaults/main.yml
orblivion: Just keeping it on "slow" data date until we actually update it.

roles/maps/tasks/install_static_search.yml
orblivion: Okay I've now brought the symlink back. I've instead added a cache busting GET param.

I just tried playing around with it by manually pointing the symlink to a different static search database, as an implementer might. At least on Firefox, the cache here seems to be extra stubborn. Sometimes a hard reload doesn't even work (though in another instance it stopped caching altogether. hard to debug). So I would warn that implementers may get frustrating results trying to manually hack this thing.

orblivion:
The databases have been uploaded to our host and I've tested them.

roles/maps/tasks/install_static_search.yml
orblivion: To be clear, the reason the implementer would have the problem despite the cache busting GET param is that I set the param to the name of the latest search database, rather than randomly at each page load. I don't want to remove all the benefits of caching.

chapmanjacobd (approved):
looks solid

Issue #4379, opened 2026-04-21
Unable to load an item into my menu


rgarrison1:
I tried to add 'Post Disaster ...' to my local system.
Task:
/usr/bin/wget -c --progress=dot:giga https://download.kiwix.org/zim/other/zimgit-post-disaster_en_2024-05.zim -O /library/working/zims/zimgit-post-disaster_en_2024-05.zim
Succeeded
Result:
2026-04-20 21:53:17 (11.4 MB/s) - ?/library/working/zims/zimgit-post-disaster_en_2024-05.zim? saved [644665513/644665513]

Task:
scripts/zim_install_move.sh zimgit-post-disaster_en_2024-05.zim
Succeeded
Result:
Traceback (most recent call last):
File "/opt/admin/cmdsrv//scripts/iiab_update_menus.py", line 45, in
main()
File "/opt/admin/cmdsrv//scripts/iiab_update_menus.py", line 33, in main
adm.put_kiwix_enabled_into_menu_json()
File "/usr/lib/python3/dist-packages/iiab/adm_lib.py", line 639, in put_kiwix_enabled_into_menu_json
expected_name = zim_perma_ref_iso2[perma_ref] + '-' + perma_ref.replace('.','_')
~~~~~~~~~~~~~~~~~~^^^^^^^^^^^
KeyError: 'wikipedia_en_100_maxi'
Finished running iiab_update_menus.py

No addition to my home menu can be located.
'The Great Library of Alexandria' does show the content.
It loads as
http://boxy/kiwix/viewer#zimgit-post-disaster_en_2024-05/home#list--0
It appears to work correctly

I try to locate the specific menu item to manually load.
<CTRL>-F 
Post    (0 hits)
Disaster (0 hits)

I have uninstalled the content and reinstalled the content, each time I verify 'Check to Have the Menu Updated Automatically' is selected.

Running on Orange Pi 4 Pro, 4 GB, Ubuntu Server (Noble)

NOTE: The comment regarding 'wikipedia_en_100_maxi' may be a clue; I also cannot find that in the selectable menu, it is also included in 'The Great Library of Alexandria', and it does behave correctly when I select it in there. I suspect both (Disaster prep and WikiPedia 100) are having related issues.

tim-moody:
Admin Console does not support manual download of zims, only zims downloaded through the Admin Console.

[...]

PR #4385, opened 2026-04-24
fix zim perma ref calc

tim-moody:
### Fixes bug:
#4379

- there are multiple functions that calculate permaref with different algorithms
- consolidate into single function and make regex more date specific

deb 13 vm


holta:
Thank you @rgarrison1 & @tim-moody

PR #4387, opened 2026-04-30
Pretend the Natural Earth pmtiles file is just another OSM vector tileset

orblivion:

Instead of having two different "naturalearth" vs "openstreetmap" modes for vector tiles, treat (`nat-z8`) almost as if it's just another openstreetmap tileset.

#### Implications

* Retire the `naturalearth-openmaptiles.pmtiles` symlink. Instead, if I choose `nat-z8`, point the `openstreetmap-openmaptiles.pmtiles` symlink to `naturalearth-openmaptiles.2025-12-10.z00-z08.pmtiles`, just like it would for any openstreetmap pmtiles file.
* Allow the "political" style to work for Natural Earth
* Most importantly: allow Natural Earth to work with upcoming seamless FQRs

#### Details

Turns out it works okay enough with no effort. I stick in the naturalearth file and tell maps.black that it's the openstreetmap file. I no longer set the maps.black mapstyle to NE-specific styles such as `"naturalearth-openmaptiles/openfreemap/liberty"`. I set it to OSM-specific styles such as `""openstreetmap-openmaptiles/openfreemap/liberty"`.

So how does this work? Both the OSM and NE pmtiles files are in the openmaptiles schema. This means that conceivably, they could use all the same styles. So I just do it, and it looks just about the same on "natural" and "hybrid" as it did before. It even looks okay on "political" (called "qwant/basic" on maps.black, which did not make it officially available for Natural Earth for some reason), though fewer political boundaries show up.

That said, the Natural Earth "openfreemap/liberty" and "maps.black/hybrid" styles are not exactly the same as their OpenStreetMap counterparts. They have different minzoom and maxzoom on many of the layers. Again, it's dealing with the same schema, but perhaps someone decided that the data available on NE was more well suited to show up at different zoom levels than for OSM. I didn't immediately notice a degradation, but just in case - I have it adjust to match how it was before. So, as far as I know, it won't look any different.

#### Some stuff under the hood:

I patch `maps.black.js` differently. It's now a smaller diff from upstream, though it works on more data. I needed this for hacking the naturalearth zooms, and it'll get me part of the way to seamless FQRs for later.

Also: while I did add those massive functions to patch the osm style to work like the old ne style, that should be the extent of any special casing. So I did get to finally simplify some of the other code.

In addition to the code changes in this PR getting me part of the way to seamless FQRs, pretending that NE is OSM is actually necessary for seamless FQRs. This is because I will need the world map and the FQRs to be using the same maps.black mapstyle at any given time.

RasPi OS Trixie

@holta @chapmanjacobd


orblivion:
I might not be explaining myself optimally, but it's late. Maybe I can make some edits to my PR description tomorrow.

chapmanjacobd:
Makes sense to me... OMT schema. Does Samoan still show up as the secondary language?

actually that's probably a bug in one of these files:

- https://github.com/maps-black/maps.black/blob/main/naturalearth-vector/download_wikidata_translations.js
- https://github.com/maps-black/maps.black/blob/main/naturalearth-vector/build.sh

edit: 

it looks like download_wikidata_translations.js is always overwriting earlier `p[id]['name']` (line 84 and 94)

maybe sorting beforehand would be better? maybe something like this?

```js
const PREFERRED_LANGS = ['en', 'fr', 'de', 'es']

translations = nativeLabels.results.bindings.reduce((p,c) => {
  const id = c.id.value.replace('http://www.wikidata.org/entity/', '')
  const lang = c.nativeLabel['xml:lang']
  const value = c.nativeLabel.value
  if (!p[id]) p[id] = {}
  
  const existingLang = p[id]['name_lang']
  const newIsBetter = !existingLang || 
    PREFERRED_LANGS.indexOf(lang) < PREFERRED_LANGS.indexOf(existingLang)
  
  if (newIsBetter) {
    p[id]['name'] = value
    p[id]['name_lang'] = lang
  }
  return p
}, translations)
```

if we could (dynamically) prefer the original country's name that might also be cool

orblivion:
It looks like you're talking about fixing MDB's data generator? Actually we could PR it upstream if so.

The same name issue happens. Again I think the only difference vis a vis the OSM style is the min and max zooms. Actually the name issue shows up in the "political" style as well.

I see some stuff about names in the style. I'll try to poke at it. But I think it could just as well be a followup PR.

roles/maps/build-maps.black-assets/build.sh
```diff
@@ -11,7 +11,7 @@ sudo apt install -y nodejs
 
 git clone https://github.com/iiab/maps.black
 cd maps.black
-git checkout 1e188ffbdc4f18a5c258598d48c64fe1dbbd9c8e
+git checkout 5d76b905cc595684f99296a13e652668407c03df
```
orblivion, line 14: Here's what the fork looks like now: https://github.com/iiab/maps.black/compare/main...5d76b905cc595684f99296a13e652668407c03df

roles/maps/templates/index.html.j2
```diff
@@ -152,27 +200,13 @@
...
-                            // So, we'll give it some valid value.
-                            return 1
-                    }
+                    return WORLD_MAP_VECTOR_MAX_ZOOM
                 },
 
-                getPmtilesName: (baseName) => {
-                    // baseName and fname won't include ".pmtiles". ".pmtiles" gets added by the calling function.
-
+                getPmtilesName: (baseName, url) => {
```
orblivion, line 206: This function will go away with seamless FQRs.

roles/maps/templates/index.html.j2
```diff
@@ -183,17 +217,244 @@
...
+                // but it's apparently created with different zoom ranges in mind so the styles
+                // are a bit different. This function adjust the OpenStreetMap style to behave
+                // like the Natural Earth style (for the world map. FQRs should all be actual
+                // OpenStreetMap) This function is generated by:
+                //
+                // https://github.com/iiab/maps2/blob/master/build-tiles/nat-to-osm
+                //
+                // It also makes sure that these zoom levels are the only differences between
+                // these the styles.
+                setLibertyForOsmToNaturalEarth: (style) => {
```
orblivion, line 245: If I can figure out a way to request the naturalearth style and merge it in dynamically, I could do away with this crap. Didn't spend a ton of time, figured this was the fastest path. maplibre seems to be built to have one active style at a time, even if we can hack it dynamically.

roles/maps/templates/index.html.j2
```diff
@@ -127,6 +137,44 @@
...
+                            if (mapStyleChoice === STYLE_HYBRID) {
+                                iiabMapsDotBlackPatch.setHybridForOsmToNaturalEarth(style)
+                            }
+                        }
+
+                        style.iiabPatched = true
+                    }
+                    return handlerResult
+                },
                 getMaxZoomTerrariumPmtiles: () => {
```
orblivion, line 178: These functions are no longer directly named in our maps.black fork (linked above) but I use them in the above function as helpers. With seamless FQRs, the logic in them will get simpler and I will likely just inline a lot of it.

PR #4395, opened 2026-05-10
Fix & modernize {netwarn pop-up, chromium} graphical autostarts + CLI pwdwarn -- for RPiOS -- whereas graphical pwdwarn is going away

holta:
### Fixes bug:

3 of these 4 warrnings / autostarts are now fixed and enhanced.

On the not-so-happy side of things, 1 of the original 4 triggers is going away: as auto-detection of published passwords is no longer practical in that graphical desktop case — certainly it's no longer easy with [labwc](https://labwc.github.io/)'s autostart since April 2026, when Raspberry Pi OS new hardened sudo's security began asking for a sudo password by default: https://www.raspberrypi.com/news/a-security-update-for-raspberry-pi-os/

- 2 of these (autostarting visually on graphical desktop) are not only fixed, but also now work much better: `/usr/local/sbin/iiab-netwarn &` and `/usr/bin/chromium --hide-crash-restore-bubble http://box/home &` — will each be triggered by the OS's presumed primary graphical desktop user's `/home/USER/.config/labwc/autostart` file going forward.  To keep things very simple, the primary graphical desktop user is presumed to have OS UID 1000 — which is usually the case, and certainly a way better approach than our past assumption (to date) that `iiab_admin_user: iiab-admin` would really correspond to the OS's primary graphical desktop user.  REASON: IIAB implementers often install their OS / distro with a very different username than that, as can be customized in [/etc/iiab/local_vars.yml](https://wiki.iiab.io/go/FAQ#What_is_local_vars.yml_and_how_do_I_customize_it?) (but often that var's left untouched!)

- The 3rd trigger is also refined (CLI warnings about published passwords, using `/etc/profile.d/iiab-pwdwarn-profile.sh`) to remind/alert the IIAB implementer about weak passwords for both `iiab-admin` _and_ also for the username associated with OS UID 1000 — when/if one or both are confirmed to have a password hash corresponding to Ansible var `iiab_admin_published_pwd`.

Raspberry Pi OS with desktop 2026-04-21

@ZachLiebl

### DIFF
--- a/roles/iiab-admin/tasks/pwd-warnings.yml
+++ b/roles/iiab-admin/tasks/pwd-warnings.yml
@@ -2,7 +2,7 @@
 #                    AND roles/network/tasks/netwarn.yml  FOR iiab-network
 
 
-- name: Get username for UID 1000 via shell -- typical graphical desktop login on Raspberry Pi OS
+- name: Get username for UID 1000 via shell -- typical primary login on Raspberry Pi OS
   command: id -nu 1000
   register: uid_1000_username
   ignore_errors: true
--- a/roles/iiab-admin/tasks/pwd-warnings.yml
+++ b/roles/iiab-admin/tasks/pwd-warnings.yml
@@ -14,9 +14,9 @@
     mode: '0644'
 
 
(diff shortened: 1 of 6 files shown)


holta:
Thanks @ZachLiebl for having reviewed this!

PR #4396, opened 2026-05-10
BRIGHT green Ansible map file download explanations can be concise?

holta:
@orblivion can we try out more concise (and even brighter green!) map file download tips?

My hunch is that the blank lines are now necessary in the end, as this message is so bright (GREEN!) and Ansible's pause during the download draws the implementer's eye to this message regardless.


holta:
Thanks @orblivion, for having given this PR a quick lookover!

PR #4401, opened 2026-05-14
Seamless Full Quality Regions

orblivion:
Seamless full quality regions. Instead of clicking on a region to look at it, you can just zoom right in. This breaks a handful of assumptions so it's a pretty big PR. I'll make comments inline.

Performance is worse because more is going on on the screen at once. I was hoping that maplibre would be smart enough to avoid this but I'm going to have to put in some optimizations as a followup.

Check it out on https://lrn2.org/maps/. That one I have naturalearth, which is particularly weird because the switch from zoom 6 to zoom 7 changes from NE to OSM. Try searching for Kilimanjaro and turning on terrain.

RasPi OS Trixie

@holta @chapmanjacobd

### DIFF
--- a/roles/maps/README.md
+++ b/roles/maps/README.md
@@ -57,11 +57,12 @@ Here are 3 examples below, to help you decide what you'll put in [/etc/iiab/loca
 
 See `maps_dot_black_vector_tiles` and `maps_dot_black_satellite_tiles` [here](https://github.com/iiab/iiab/blob/master/roles/maps/defaults/main.yml) for all valid values.
 
-*NOTE: The satellite data is licensed "NonCommercial" under [CC BY-NC-SA 4.0](https://creativecommons.org/licenses/by-nc-sa/4.0/).  To skip satellite imagery, set:*
+*NOTE: The satellite data is licensed "NonCommercial" under [CC BY-NC-SA 4.0](https://creativecommons.org/licenses/by-nc-sa/4.0/).  To skip worldwide satellite imagery, set:*
 
   ```
   maps_satellite_zoom: none
-  ```
+
+*Full Quality Regions will still download satellite data, though it will not be visible in the UI.*
 
 ## How do I install 3D terrain?
 
--- a/roles/maps/defaults/main.yml
+++ b/roles/maps/defaults/main.yml
@@ -102,7 +102,6 @@ maps_dot_black_naturalearth6_tiles:
   # FOR TESTING ONLY
   ci: "naturalearth6-NE2_HR_SR_W_DR-WEBP.{{ maps_slow_data_date }}.z00-z04.pmtiles"
 
-# Two different symlinks because the front end could requset openstreetmap or naturalearth
 maps_dot_black_osm_symlink_name: openstreetmap-openmaptiles.pmtiles
 
 maps_dot_black_vector_tiles:
@@ -119,6 +118,14 @@ maps_dot_black_vector_tiles:
   # "low res" - mostly borders, rivers, country names, large roads.
   # maps_vector_quality = "nat-z8"
   # (nat-z8 = "Natural Earth")
+  #
+  # NOTE: We will pass this into maps.black as if it's the OpenStreetMap data, even though
+  # it's Natural Earth. They're both in the OpenMapTiles schema. The OSM and NE variants of
+  # the "Natural" style we use are compatible, with just some zoom range differences (which
+  # makes no difference that I notice). This will fail to show "naturalearth" in attributions
+  # ("naturalearth6" is separate), even in "generous" attribution mode. However maps.black
+  # and the naturalearth website say that crediting authors is unnecessary. It's not worth
+  # the time to fix just for consistency.
   nat-z8: "naturalearth-openmaptiles.{{ maps_slow_data_date }}.z00-z08.pmtiles"
 
   # FOR TESTING ONLY
@@ -186,6 +193,10 @@ maps_dot_black_terrain_tiles:
   # them out to be sure.
   5: "terrarium.{{ maps_slow_data_date }}.z00-z05.pmtiles"
   6: "terrarium.{{ maps_slow_data_date }}.z00-z06.pmtiles"
+
+  # A "dummy" maxzoom=0 world map terrain file to fill a role that maps.black/maplibre
+  # needs if we have FQRs and the user enables terrain.
+  none: "terrarium-none.pmtiles"
 
 maps_terrain_max_zoom: 10
 
--- a/roles/maps/tasks/install_frontend.yml
+++ b/roles/maps/tasks/install_frontend.yml
@@ -79,31 +79,24 @@
     path: "{{ maps_serve_path }}/{{ maps_dot_black_satellite_symlink_name }}"
     state: absent
 
-- when: maps_terrain_zoom != "none"
-  block:
-    - name: "Make sure maps_terrain_zoom has a valid value"
-      assert:
-        that: maps_dot_black_terrain_tiles[maps_terrain_zoom] is defined
-        fail_msg: "Error: maps_terrain_zoom is set to an invalid value. See `maps_dot_black_terrain_tiles` in https://github.com/iiab/iiab/blob/master/roles/maps/defaults/main.yml for valid values."
-        quiet: yes
+- name: "Make sure maps_terrain_zoom has a valid value"
+  assert:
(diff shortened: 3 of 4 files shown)


roles/maps/defaults/main.yml
```diff
@@ -119,6 +118,14 @@ maps_dot_black_vector_tiles:
...
   # maps_vector_quality = "nat-z8"
   # (nat-z8 = "Natural Earth")
+  #
+  # NOTE: We will pass this into maps.black as if it's the OpenStreetMap data, even though
+  # it's Natural Earth. They're both in the OpenMapTiles schema. The OSM and NE variants of
+  # the "Natural" style we use are compatible, with just some zoom range differences (which
+  # makes no difference that I notice). This will fail to show "naturalearth" in attributions
+  # ("naturalearth6" is separate), even in "generous" attribution mode. However maps.black
+  # and the naturalearth website say that crediting authors is unnecessary. It's not worth
+  # the time to fix just for consistency.
```
orblivion, line 128: I'm accumulating bulletpoints as to why we should get rid of Natural Earth, so I can make a pretty strong case for it. We just need something to replace the small data tier.

roles/maps/tasks/install_frontend.yml
```diff
@@ -127,14 +120,6 @@
     executable: /bin/bash
   with_items: "{{ maps_preset_full_quality_regions }}"
 
-# If the user goes back to "none", we want to delete the symlink that gives them terrain.
```
orblivion, line 130: This `absent` for the terrain was a dupe. Actually it was the original I think.

roles/maps/templates/index.html.j2
```diff
@@ -126,336 +129,141 @@
...
-                            if (mapStyleChoice === STYLE_HYBRID) {
-                                iiabMapsDotBlackPatch.setHybridForOsmToNaturalEarth(style)
-                            }
-                        }
-
                         style.iiabPatched = true
                     }
                     return handlerResult
                 },
-                getMaxZoomTerrariumPmtiles: () => {
```
orblivion, line 178: These helper functions are reduced to a single constant now, so I just use the constant directly instead.

roles/maps/templates/index.html.j2
```diff
@@ -126,336 +129,141 @@
...
-                getMaxZoomOSM: () => {
-                    // We are looking at a full quality region. Let's make sure we can zoom all the way in.
-                    if (!!activeFQRegionName) {
-                        return 14
-                    }
-
-                    return WORLD_MAP_VECTOR_MAX_ZOOM
-                },
-
-                getPmtilesName: (baseName, url) => {
```
orblivion, line 206: When we were switching to different FQRs we were switching which pmtiles files that maps.black was assigning to the different maplibre "sources". Now, we have sources for the world map and all FQRs created and present at once. That's what's going on above.

roles/maps/templates/index.html.j2
```diff
@@ -126,336 +129,141 @@
...
-                // but it's apparently created with different zoom ranges in mind so the styles
-                // are a bit different. This function adjust the OpenStreetMap style to behave
-                // like the Natural Earth style (for the world map. FQRs should all be actual
-                // OpenStreetMap) This function is generated by:
-                //
-                // https://github.com/iiab/maps2/blob/master/build-tiles/nat-to-osm
-                //
-                // It also makes sure that these zoom levels are the only differences between
-                // these the styles.
-                setLibertyForOsmToNaturalEarth: (style) => {
```
orblivion, line 245: All this stuff lead to more considerations I needed to make. I decided that it doesn't matter. The min and max zooms are different but I can't tell the difference anyway. And again I want to get rid of NE anyway.

roles/maps/templates/index.html.j2
```diff
@@ -690,8 +431,91 @@
                     // Don't affect draw mode though.
                     fqRegionsControl.endDeleteMode()
                 })
+                mb.map.fqRegionsShown = true // attach to `mb.map` so that it clears if we reset the map
+            }
+
+            async function setFQRegionOutlinesVisibility() {
```
orblivion, line 437: If you zoom in far enough, the FQR lines disappear. This was an attempt to make terrain less buggy. Doesn't do much for the bug I was thinking of but it got rid of some weird green artifacts (the border lines showing up in a weird way) in the mountains.

roles/maps/templates/index.html.j2
```diff
@@ -690,8 +431,91 @@
...
+                const lineVisibility = mb.map.getZoom() > 10 ? "none" : "visible"
+                for(regionName in downloadedFQRegions.regions) {
+                    // TODO - fade instead? or in addition?
+                    const regionId = "region-" + regionName
+                    mb.map.setLayoutProperty(regionId + "~line", "visibility", lineVisibility)
+                    mb.map.setLayoutProperty(regionId + "~line-2", "visibility", lineVisibility)
+                }
             }
 
+            async function setBestTerrainSource() {
```
orblivion, line 449: For vector and satellite, I just piled FQR layers on top of world layers. Terrain doesn't use layers (though hillshade does). Instead, I made a system for choosing which terrain source to switch to dynamically as you scroll around. It actually works quite well I'd say.

roles/maps/templates/index.html.j2
```diff
@@ -1630,23 +1427,30 @@
                     }
                     mb.map.addControl(scaleControl);
 
+                    // in case of nat-z8 and no FQRs, get rid of OSM attribution.
```
orblivion, line 1430: Tangential fix. Didn't think of it before.

roles/maps/README.md
```diff
@@ -57,11 +57,12 @@ Here are 3 examples below, to help you decide what you'll put in [/etc/iiab/loca
...
 See `maps_dot_black_vector_tiles` and `maps_dot_black_satellite_tiles` [here](https://github.com/iiab/iiab/blob/master/roles/maps/defaults/main.yml) for all valid values.
 
-*NOTE: The satellite data is licensed "NonCommercial" under [CC BY-NC-SA 4.0](https://creativecommons.org/licenses/by-nc-sa/4.0/).  To skip satellite imagery, set:*
+*NOTE: The satellite data is licensed "NonCommercial" under [CC BY-NC-SA 4.0](https://creativecommons.org/licenses/by-nc-sa/4.0/).  To skip worldwide satellite imagery, set:*
 
   ```
   maps_satellite_zoom: none
-  ```
+
+*Full Quality Regions will still download satellite data, though it will not be visible in the UI.*
```
holta, line 65: 1. Can this be fixed?  In a subsequent PR if it's not a very quick fix?  (FQRs need to display satellite photos regardless.)

2. FQRs' disk usage estimate (vector GB : satellite GB : terrain GB) needs to be very prominent in README.md — can we please put that right on top — making much more intuitive what an FQR really is?  (Almost from the [very 1st moment](https://github.com/iiab/iiab/tree/master/roles/maps#full-quality-regions) that FQRs are introduced/explained!)

   I strongly suggest spelling it out with 2 or 3 examples: e.g. put the actual number of (vector GB, satellite GB, terrain GB) that would be downloaded in the FQR bounding box encapsulating... A. New York City, B. Africa, C. all of Planet Earth ?

orblivion: It's possible but probably slow enough that I should make it a followup. It's a decision I had to make. I chose this because it was the faster option, and anyway seemed in line with our discussions of why someone would opt to not use satellite globally.

The README request seems orthogonal to all of this so I'll make it a follow up as well?

holta:
@chapmanjacobd would you have time to review this PR?

roles/maps/templates/index.html.j2
```diff
@@ -98,7 +101,7 @@
                 window.staticSearch = {
                     init: async function () {
                         await until(() => (mb && mb.map))
-                        const searchDb = activeFQRegionName || "world-map"
+                        const searchDb = "world-map"
```
chapmanjacobd, line 104: did FQR search DB ever work? why remove?

chapmanjacobd (approved):
nice work! the code is now a lot cleaner too

roles/maps/templates/index.html.j2
orblivion: We never even got to the point where we were trying to have FQR search databases. Our current world search index is tiny. I was jumping the gun by even having this line.

PR #4404, opened 2026-05-18
Remove 404 on map load

orblivion:
### Fixes bug:

Maybe half the time you load IIAB maps, there's a 404 (and a bunch of console noise) because maps.black loads its default style before we get a chance to set our style. It's a race condition of sorts. It becomes a 404 because the default style looks for the protomaps pmtiles file, whereas we use the openmaptiles one.

Determine the style from the hash and use that to construct the `<maps-black>` tag, so that its initial load is the style we want. That way there's not the tiny amount of time for maps.black to try to load its default style.

We also get to simplify our code a bit as well (but also add a tiny bit of complication vis a vis terrain).

RasPi OS Trixie

@holta @chapmanjacobd

### DIFF
--- a/roles/maps/templates/index.html.j2
+++ b/roles/maps/templates/index.html.j2
@@ -280,44 +280,57 @@
                 history.replaceState({}, "", `#${lat}/${lon}/${pitch}/${bearing}/${zoom}/${globe}/${terrain}/${mapStyleChoice}`)
             }
 
-            // Why two different functions for reading the URL hash?
-            //
-            // The mb orientation params seem to need to be set *before* window load, or else an
-            // intermittent 404 in a resource (pmtiles, even js? just watch the console) will
-            // (somehow) cause mb to freak out or something, defaulting to zero values for all
-            // of them. Calling `readHash()` early on seems to be close enough to setting the html tag
-            // parameters on <maps-black> I guess? Which it seems to hold onto more robustly.
-            //
-            // Strangely, if we set the `globe` parameter here, it seems to make `mb.globe = ...`
-            // not work later on, which makes the `GlobeControl` no longer work. So, we call
-            // `readHashGlobeAndTerrain()` after window load. (We include terrain for uniformity).
-            //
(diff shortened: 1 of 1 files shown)


roles/maps/templates/index.html.j2
```diff
@@ -280,44 +280,57 @@
...
+            async function correctTerrainSetting() {
+                // it would be faster with mapStylePresent but that doesn't seem
+                // to avoid an error
+                await until(mapStyleLoaded)
 
-                // contour lines is another thing that uses terrain maps, which we can consider
-                mb.terrain = mb.hillshade = (terrain === 't') && (await terrainAvailable())
+                if (mb.terrain && !(await terrainAvailable())) {
+                    mb.terrain = false
+                    setUpMap() // set it up again after this change
```
chapmanjacobd, line 330: we call setUpMap() shortly after where correctTerrainSetting is called too... the flow of this feels a bit kludgy -- probably fine but feels like it could be refactored or written in a more straightforward way? The rest of the PR looks good though so just wanted to call this out specifically as being one thing which could potentially be improved

chapmanjacobd (approved):
looks good!

roles/maps/templates/index.html.j2
orblivion: Yeah I figured you might notice that. Via experimentation it seems fine. Ideally I'd like to somehow abort the previous `setUpMap` but that seems complicated.

Actually, I should perhaps pass a `map` object into `setUpMap`. That way if it gets called again, the first call sets up the "old" map and the second call sets up the "new" map, so we don't get duplicate controls on screen or something. Only problem would be if the "old" map object gets deleted, we'd see errors in the console when trying to add controls to it.

holta: Should this PR be merged now?

orblivion: I'm okay with it. I think that even if @chapmanjacobd 's concerns come to some sort of bug, it would only affect the rare case I described in the comment of this function.

PR #4407, opened 2026-05-19
Fallback for no satellite but with FQRs

orblivion:

Allow users to choose satellite if they have FQRs even if they don't have world satellite.

Just like with the analogous situation with terrain, maplibre needed a "dummy" file for the world map. But also, it would be good to fill it in with something. So I just reused the naturalearth6 raster images. Looks a bit weird but it's at least clear what part of the world you're looking at:

fqr-only-sat

Second option that does look a little better is 2016 satellite (clipped at maxzoom=0). It doesn't have the non-commercial designation that 2023 has (so I'd feel comfortable downloading it by default even if the user doesn't want satellite), but it still requires attribution. Plus we'd have to upload a new file. I didn't want to take the time to bother with all that because I wasn't sure if it's important enough. If it is it's worth the time to make this particular mode (FQRs but no world satellite) look better, I can do it.

**Edit:** Third and current (as of this writing) option: use the zoom=0 tile from any one of the FQRs. That's the whole planet:

RasPi OS Trixie
@holta @chapmanjacobd

### DIFF
--- a/roles/maps/templates/index.html.j2
+++ b/roles/maps/templates/index.html.j2
@@ -250,6 +250,17 @@
                         setMaxZoom('openstreetmap-openmaptiles', WORLD_MAP_VECTOR_MAX_ZOOM)
                         setMaxZoom('s2maps-sentinel2-2023', WORLD_MAP_SATELLITE_MAX_ZOOM)
 
+                        if (WORLD_MAP_SATELLITE_MAX_ZOOM === 0 && fqRegionsExist() && "s2maps-sentinel2-2023" in style.sources) {
+                            // Pick any FQR, use it as the satellite source. It has a zoom-0 full world image.
+                            const someFqr = Object.keys(downloadedFQRegions.regions)[0]
+                            const satelliteDate = downloadedFQRegions.regions[someFqr].dates.satellite
+
+                            style.sources['s2maps-sentinel2-2023'].url = style.sources['s2maps-sentinel2-2023'].url.replace(
+                              '/s2maps-sentinel2-2023.pmtiles',
+                              `/s2maps-sentinel2-2023.${satelliteDate}.full-region.${someFqr}.pmtiles`,
+                            )
+                        }
+
                         // Sources that use the terrarium file. We don't necessarily even use all of them
                         // but we'll change them here in case we do.
                         setMaxZoom('terrarium', WORLD_MAP_TERRAIN_MAX_ZOOM)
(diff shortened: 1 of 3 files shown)


chapmanjacobd:
This is a pretty niche option... a "1x1 Transparent Pixel" style fix might match user expectations a bit more (hiding the Satellite option when `maps_dot_black_satellite_tiles: none`).

But this is good enough for now... 

ah I understand now, we're still downloading satellite for FQRs even if the basemap is `maps_dot_black_satellite_tiles: none`. Okay I guess that makes sense. I wonder if we should make this a toggle and allow the user to choose partial FQRs (no satellite). But the size difference really isn't that much more--maybe 2x--so I think few people will actually do something like this

chapmanjacobd:
Let's merge the main idea of this PR AS-IS and we can update the maxzoom=0 world map pmtiles in a couple months when we have our own "public domain" satellite layer where we don't need complicated attribution-or-not logic and it will look better

orblivion:
> This is a pretty niche option... a "1x1 Transparent Pixel" style fix might match user expectations a bit more (hiding the Satellite option when maps_dot_black_satellite_tiles: none).

I'm not sure if I understand this. Is this about the style picker menu? Right now it's dynamically generated. The satellite and hybrid styles show up if `maps_dot_black_satellite_tiles` is not none or if any FQRs exist.

> I wonder if we should make this a toggle and allow the user to choose partial FQRs (no satellite).

As I understand, this use case doesn't really exist. If someone gets an FQR they want satellite.

roles/maps/templates/index.html.j2
```diff
@@ -241,6 +245,21 @@
                             })
                         })
 
+                        // We've copied all of the world layers to FQRs, in the same order. Now it's time to take all of the
+                        // satellite layers and move them to the beginning.
```
orblivion, line 249: Turns out to be relevant if the world vector layer has more zoom levels than the world satellite layer. I just didn't notice it before (though I did consider it in an earlier version of my code that I never pushed).

orblivion:
See new screenshots in "third and current" option in the description.

chapmanjacobd:
The naturalearth version is less blurry... how big would the 2016 s2maps be maxzoom=1? might be worth trying that. It's probably less than 10MB and a lot more usable

orblivion:
Z4 looks decent and is only 2M. If the blurriness is too much I'll go that direction.

holta:
Thanks to @orblivion & @chapmanjacobd who worked hard on validating design assumptions here today:

Let's reassess Monday, to make sure this is all converging as it should!

holta:
Should this PR be closed?

(If an improved version below was merged a week ago... ??)

- PR #4413

holta:
Oops my apologies if I misunderstood and this PR is still valuable!  It sounds like this PR is still evolving and working towards important resiliency optimizations/refactorings — e.g. to better handle "satellite" data contained within an FQR that's added to an IIAB that happens to be missing most-or-all satellite data.  (Anyway a work in progress, Thanks to @orblivion!)

PR #4413, opened 2026-05-25
A few followups (fix and optimize) for seamless FQRs

orblivion:
### Fixes bug:

* Fix layer ordering to fix bug in Hybrid mode where vector doesn't always show up in an FQR. I think that it has to do with the satellite FQR layer showing up over the world map vector layer in certain cases, depending on maxzoom of different layers, and how overzooming works. This change puts all of the satellite layers underneath all of the vector layers, which I believe fixes any such issue.

* Hide world map OSM if I'm looking at an FQR zoomed in enough. I don't want doubled up roads etc. Especially if world map "OSM" is actually Natural Earth and the roads aren't exactly the same. Also don't double up hillshade during terrain mode. It makes things darker and maybe slower.
* Leave behind possible implementation of hiding off-screen FQRs. I came close to using it but it needs more debugging that's not worth the time now. It may actually be fine as-is (the debugging has been confusing and time consuming) so I want to keep it around for now. But it's also just not that imminently necessary. It has to do with slowdowns inherent in having too many FQRs on screen at once, and most people will only have one.

RasPi OS Trixie

@holta @chapmanjacobd

### DIFF
--- a/roles/maps/templates/index.html.j2
+++ b/roles/maps/templates/index.html.j2
@@ -128,6 +128,9 @@
             // we want style choice to survive the map reset that happens when a maps.black
             // feature or style changes, so we rely on this instead of the select tag value
             let mapStyleChoice
+            // keep track of which layer ids refer to OSM layers. At some points we want to
+            // show or hide them.
+            let worldMapOSMLayers = new Set()
 
             // set immediately after the tag is defined
             let mb = null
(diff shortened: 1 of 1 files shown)


holta:
Building on things like:

- PR #4401

roles/maps/templates/index.html.j2
```diff
@@ -239,6 +248,22 @@
                             .forEach(fqrLayer => {
                                 style.layers.push(fqrLayer)
                             })
+
+                        })
+
+                        // We've copied all of the world layers to FQRs, in the same order. Now it's time to take all of the
+                        // satellite layers and move them to the beginning.
+                        for (const [index, layer] of style.layers.entries()) {
+                            layer.__sortOrder = index
```
chapmanjacobd, line 257: huh! interesting

chapmanjacobd (approved):

roles/maps/templates/index.html.j2
orblivion: To be clear, I invented `__sortOrder`, it's not any sort of standard field. I wanted a stable sort and I didn't know a better way.

chapmanjacobd: In modern JavaScript (ES2019 and later), the built-in Array.prototype.sort() method is guaranteed to be stable

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/sort#browser_compatibility (see "Stable sorting")

orblivion: Oh that's right. I think I may have shied away from that to not have to worry about old phone browsers.

PR #4416, opened 2026-05-29
Handle FQRs that cross longitude of 180/-180 (aka the antimeridian)

orblivion:
### Fixes bug:

Attempts to get FQRs that cross the antimeridian end up with the region past cut off at that line.

* Use a multipolygon to represent a two-part region, one part after -180 and one part before 180, instead of a single region that extends outside of -180 to 180.
* Inspect the tiles in the file (instead of relying on the header), for FQRs that cross this line, to find the appropriate bounds for drawing the rectangle
* `mapStyleLoaded` now looks at `mb.map.style._loaded` which hopefully is a good enough check. It happens to be faster. But it's necessary to fix a bug with displaying rectangles for regions that cross 180/-180
* split fqr's "bbox" into "ui_bounds" (rectangle, clickable for deletion) and "render_bounds" (reported to pmtiles/maplibre to display the region).
* also fix issues related to users trying to download a region while zoomed way out, getting longitude values well above 180 or below -180. basically it does a module 360 to get it into sensible equivalents that pmtiles can deal with.

Inspecting the file takes a few seconds, and this will happen every time extracts.json is updated (i.e. whenever a new FQR is downloaded, deleted, etc).

RasPi OS Trixie

@holt @chapmanjacobd

### DIFF
--- a/roles/maps/templates/index.html.j2
+++ b/roles/maps/templates/index.html.j2
@@ -182,14 +182,14 @@
                                     style.sources[s2FqrId] = {
                                         ...style.sources[s2Id],
                                         url: style.sources[s2Id].url.replace(s2Id, s2FqrId),
-                                        bounds: downloadedFQRegions.regions[fqrName].bbox,
+                                        bounds: downloadedFQRegions.regions[fqrName].render_bounds,
                                         minzoom: WORLD_MAP_SATELLITE_MAX_ZOOM + 1
                                     }
                                 } else if (worldSourceId === osmId) {
                                     style.sources[osmFqrId] = {
                                         ...style.sources[osmId],
                                         url: style.sources[osmId].url.replace(osmId, osmFqrId),
-                                        bounds: downloadedFQRegions.regions[fqrName].bbox,
+                                        bounds: downloadedFQRegions.regions[fqrName].render_bounds,
                                         minzoom: WORLD_MAP_VECTOR_MAX_ZOOM + 1
                                     }
                                 } else if (worldSourceId === terrariumId) {
@@ -198,7 +198,7 @@
                                         url: style.sources[terrariumId].url
                                             .replace(terrariumId, terrariumFqrId)
                                             .replace(/\-z0\-z10\.pmtiles$/, '.pmtiles'),
-                                        bounds: downloadedFQRegions.regions[fqrName].bbox,
+                                        bounds: downloadedFQRegions.regions[fqrName].render_bounds,
                                         minzoom: WORLD_MAP_TERRAIN_MAX_ZOOM + 1
                                     }
                                 } else if (worldSourceId === naturalEarth6Id) {
@@ -382,7 +382,23 @@
             }
 
             const mapStylePresent = () => (mb && mb.map && mb.map.style)
-            const mapStyleLoaded = () => (mapStylePresent() && mb.map.style.loaded())
+
+            // The more proper thing would be mb.map.style.loaded() or even
+            // mb.map.isStyleLoaded(), but this version seems to work for what
+            // we need. .loaded() checks a few things, including _loaded and
+            // that a bunch of source caches (whatever those are) are loaded
+            // as well. For whatever reason, it seems that if we are looking
+            // at an FQR that crosses the antimeridian, that FQR's source
+            // cache is not loaded (it even goes from loaded to not loaded
+            // when you start looking at it). But you can see the FQR. I'm
+            // not sure what's going on with that and I'm not going to try
+            // to figure that out right now. It seems to work.
+            //
+            // Also, this version is generally much faster and makes initial
+            // page loads nicer for FQR rectangles and setting FQR terrain.
+            // Hopefully I'm not wrong that this is good enough and we won't
+            // see errors show up.
+            const mapStyleLoaded = () => (mapStylePresent() && mb.map.style._loaded)
 
             // Show the outlines of the full quality regions and make them interactive.
             async function drawFQRegionOutlines() {
@@ -393,7 +409,7 @@
                 for (regionName_ in downloadedFQRegions.regions) {
                     const regionName = regionName_ // fix variable capturing in callbacks below
                     const regionId = "region-" + regionName
-                    const [min_lon, min_lat, max_lon, max_lat] = downloadedFQRegions.regions[regionName].bbox
+                    const [min_lon, min_lat, max_lon, max_lat] = downloadedFQRegions.regions[regionName].ui_bounds
 
                     if (mb.map.getSource(regionId)) {
                         // Looks like we already ran this. Quit without setting a timeout.
@@ -558,7 +574,7 @@
 
                 // Does SOME region contain EVERY corner of the map bounds?
                 return regionsOnScreen
-                .map(regionName => new maplibregl.LngLatBounds(downloadedFQRegions.regions[regionName].bbox))
+                .map(regionName => new maplibregl.LngLatBounds(downloadedFQRegions.regions[regionName].ui_bounds))
                 .some(regionBounds => mapCorners.every(lngLat => regionBounds.contains(lngLat)))
             }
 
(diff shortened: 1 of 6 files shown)


roles/maps/templates/tile-extract.py.j2
```diff
@@ -1,8 +1,107 @@
...
 
+sys.path.insert(1, "{{ tile_extract.working_dir }}/{{ tile_extract.pmtiles_python.dir_name }}/python/pmtiles")
+
+from pmtiles.reader import MmapSource, all_tiles
+
+# Any pmtiles file for a region that crosses 180/-180 longitude (aka the
+# antimeridian) will say its min and max longitudes are -180 and 180.
+# Technically true, but there's of course the massive gap in the middle.
+# So here we inspect the file to find where that hole is.
+def get_bounds(full_path):
```
orblivion, line 14: This is the part where we already have pmtiles files, and we want to figure out what goes in extracts.json for the front end to use to draw the rectangle.

roles/maps/templates/tile-extract.py.j2
```diff
@@ -124,73 +232,160 @@ def delete_extract(extract_name):
         if os.path.dirname(path) != HOSTING_DIR or f".{extract_name}." not in path:
             # Last-minute sanity check before deleting files in case some other
             # error is introduced before we get to this point.
-            raise Exception(f"Unknown error. Stopping before we delete the wrong file!", path)
+            raise Exception("Unknown error. Stopping before we delete the wrong file!", path)
         os.remove(path)
 
+def set_region_geojson(extract_box, geojson_file):
```
orblivion, line 245: This is the part where the user supplies a rectangle and we want to construct a multipolygon for the pmtiles tool to extract and create the FQR file.

roles/maps/templates/tile-extract.py.j2
```diff
@@ -124,73 +232,160 @@ def delete_extract(extract_name):
...
-            'kB': 1_000,
-            'MB': 1_000_000,
-            'GB': 1_000_000_000,
-            'TB': 1_000_000_000_000,
-        }
-        extract_sizes[paths['source']] = {
-            "transfer": float(transfer_num) * unit_mult[transfer_unit],
-            "archive":  float(archive_num)   * unit_mult[archive_unit],
-        }
+    with tempfile.NamedTemporaryFile(delete_on_close=False, mode="w") as geojson_file:
```
orblivion, line 348: Ignore-whitespace may help a little here

holta:
Really great news:

@chapmanjacobd would you have time to review today?

roles/maps/templates/tile-extract.py.j2
```diff
@@ -124,73 +232,160 @@ def delete_extract(extract_name):
...
+              [min_lon, min_lat],
+            ],
+          ],
+          [
+            [
+              # The eastern rectangle is on the western part of the -180 to 180 map.
+              # We want the part that starts at -180 and ends at max_lon. Since
+              # max_lon > 180, we subtract 360 to get the equivalent that's within
+              # -180 and 180.
+              [min_lon, min_lat],
```
chapmanjacobd: I think this line should be removed? the polygon doesn't end where it starts

roles/maps/templates/tile-extract.py.j2
```diff
@@ -21,8 +120,17 @@ def validate_extract_name(extract_name):
...
-        # `float` is just a parsing check.
-        assert len([float(n) for n in extract_box.split(",")]) == 4
+        # `float` is important here even if just as a parsing check.
+        min_lon, min_lat, max_lon, max_lat = [float(n) for n in extract_box.split(",")]
+
+        # I am concerned about longitudes in particular because -180 and 180 refer to the same place.
+        # I expect to get all all positives or all negatives for longitudes (see the big comment in
+        # index.html, inside `terraDrawInstance.on("finish"` for details) but this is just based on
+        # observation. I would like to confirm it here.
+        assert min_lon < max_lon, "longintudes seem to be out of order"
```
chapmanjacobd: typo /s/longintudes/longitudes/

roles/maps/templates/index.html.j2
```diff
@@ -382,7 +382,23 @@
...
+            // cache is not loaded (it even goes from loaded to not loaded
+            // when you start looking at it). But you can see the FQR. I'm
+            // not sure what's going on with that and I'm not going to try
+            // to figure that out right now. It seems to work.
+            //
+            // Also, this version is generally much faster and makes initial
+            // page loads nicer for FQR rectangles and setting FQR terrain.
+            // Hopefully I'm not wrong that this is good enough and we won't
+            // see errors show up.
+            const mapStyleLoaded = () => (mapStylePresent() && mb.map.style._loaded)
```
chapmanjacobd, line 401: this is probably fine as it works like you say but if we see glitches later or different behavior in older browsers we might try hooking in to MapLibre event handlers:

https://maplibre.org/maplibre-gl-js/docs/API/type-aliases/MapEventType/#load

roles/maps/templates/tile-extract.py.j2
```diff
@@ -1,8 +1,107 @@
...
+
+from pmtiles.reader import MmapSource, all_tiles
+
+# Any pmtiles file for a region that crosses 180/-180 longitude (aka the
+# antimeridian) will say its min and max longitudes are -180 and 180.
+# Technically true, but there's of course the massive gap in the middle.
+# So here we inspect the file to find where that hole is.
+def get_bounds(full_path):
+    output = subprocess.getoutput(PMTILES + " show " + full_path + " --header-json")
+    render_bounds = [float(coord) for coord in json.loads(output)["bounds"]]
```
chapmanjacobd: maybe we could wrap this in a try block for a better error message if pmtiles/this fails incl. printing output and/or stderr from pmtiles show...

roles/maps/templates/tile-extract.py.j2
```diff
@@ -1,8 +1,107 @@
...
+    # Tiles have z: zoom, x: horizontal index, and y: vertical index.
+    # For z=0, x is always 0. For z=1, x can be 0-1. For z=2, x can be 0-3. And so on.
+    # x = 0 represents the tiles that are -180 on their western edge.
+    # x = (2 ** z - 1) represents the tiles that are 180 on their eastern edge.
+    # For all the tiles in this file at our desired zoom level, let's first put their
+    # x values in order, removing duplicates. This will tell us where we have actual
+    # tile data:
+    x_values = sorted({tile[1] for tile in tiles if tile[0] == zoom})
+    if len(x_values) == 0:
+        # Not sure how we got here but we don't have any tiles!
```
chapmanjacobd, line 62: maybe we should log an warning or info/debug here?

chapmanjacobd (changes requested):
Line 299 seems incorrect

roles/maps/templates/tile-extract.py.j2
```diff
@@ -124,73 +232,160 @@ def delete_extract(extract_name):
...
+              [-180, max_lat],
+              [max_lon - 360, max_lat],
+              [max_lon - 360, min_lat],
+              [-180, min_lat],
+            ]
+          ]
+        ]
+
+    region = {
+      "type": "MultiPolygon",
```
chapmanjacobd, line 316: technically you could use `Polygon` for the normal case but this is probably not that important

roles/maps/templates/tile-extract.py.j2
```diff
@@ -1,8 +1,114 @@
...
+# Any pmtiles file for a region that crosses 180/-180 longitude (aka the
+# antimeridian) will say its min and max longitudes are -180 and 180.
+# Technically true, but there's of course the massive gap in the middle.
+# So here we inspect the file to find where that hole is.
+def get_bounds(full_path):
+    output = subprocess.getoutput(PMTILES + " show " + full_path + " --header-json")
+    try:
+        render_bounds = [float(coord) for coord in json.loads(output)["bounds"]]
+    except:
+        print (f"Failed to get bounds from pmtiles header for {full_path}. Got this output from pmtiles:\n", output)
```
orblivion, line 19: As suggested, but I'm also making this a warning instead of an error, since I've been doing more of that here.

I didn't put the subprocess.getoutput in the try block because firstly if it throws an error we won't have any `output` set for the warning message. But also, output captures both stdout and stderr (I just tried it) even for non-zero return codes, so errors in the command are already handled. If subprocess.getoutput still throws an error we should probably just see whatever Python has to show us.

chapmanjacobd (approved):

PR #4418, opened 2026-06-02
Maps: turn off globe mode if we try to turn on FQR selector.

orblivion:
### Fixes bug:

When using the FQR region selector with "globe" mode turned on, the rectangle would sometimes take the "long route" and wrap around the planet.

If the user turns on the FQR selector with globe mode turned on, it first turns off globe mode, refreshes the map (as is necessary), and loads the FQR selector tool ready to go.

RasPi OS Trixie

@chapmanjacobd @holta


chapmanjacobd (approved):
looks good!

PR #4433, opened 2026-06-30
Map data catalog

orblivion:
This is something we've been discussing. Making a requestable catalog instead of hard coding versions. FQR downloader doesn't successfully use it yet (or at least I haven't tested it yet).


holta:
Looks very promising:

Can we move towards [SSOT](https://en.wikipedia.org/wiki/Single_source_of_truth) ("Don't Repeat Yourself") please, and only store the dates in one single place like...

`roles/maps-catalog.json`

Not that `/opt/iiab/iiab/roles/maps/maps-catalog.json` should ever actually be used directly!  Instead, https should *always* be used to verify the very latest dates herein:

https://github.com/iiab/iiab/blob/master/roles/maps/maps-catalog.json

Above file can "regularly" be cached here if a local copy is really necessary...

`/etc/iiab/maps-catalog.json`

In other words: can we remove all hard-coded dates from `roles/maps/defaults/main.yml` ?

CLARIF: We should be open to renaming "catalog" to "manifest" (or some such) if that later turns out to better describe what's really needed, and what's really happening — e.g. we may discover that P2P / grassroots communities worldwide each & all venerating one (single point of failure, cultural monotone) catalog might NOT be the healthiest way forward...

orblivion:
Sure thing. I guess it means that generating the json file in the repository should be interactive, wherein we add the dates, because we'll remove the dates from main.yml.

I think we can get away with not having the /etc/ file just yet because the dates will be in the file names.

roles/maps/generate-catalog.py
```diff
@@ -0,0 +1,34 @@
+#!/usr/bin/python3
+import requests, jinja2, json, yaml
+
+# If you are installing or configuring IIAB Maps, *you should ignore `/opt/iiab/iiab/maps/maps-catalog.json`*.
```
holta:
```suggestion
# If you are installing or configuring IIAB Maps, *you should ignore `/opt/iiab/iiab/roles/maps/maps-catalog.json`*.
```

roles/maps/templates/tile-extract.py.j2
```diff
@@ -151,28 +151,27 @@ def get_new_extract_paths(extract_name):
     from the server. Existing FQRs on the file system may have different dates,
     so we can't rely on this function for getting an inventory or deletion.
     """
-    MAPS_VECTOR_DATA_DATE = "{{ maps_vector_data_date }}"
-    MAPS_SLOW_DATA_DATE = "{{ maps_slow_data_date }}"
-    IIAB_MAP_HOST_URL = "{{ iiab_map_host_url }}"
+    catalog = requests.get("https://raw.githubusercontent.com/iiab/iiab/refs/heads/master/roles/maps/catalog/catalog.json")
```
holta:
```suggestion
    catalog = requests.get("https://raw.githubusercontent.com/iiab/iiab/refs/heads/master/roles/maps/maps-catalog.json")
```

roles/maps/maps-catalog.json
```diff
@@ -0,0 +1,29 @@
...
+        "4": "https://iiab.switnet.org/maps/2/s2maps-sentinel2-2023.2025-12-10.z00-z04.pmtiles",
+        "7": "https://iiab.switnet.org/maps/2/s2maps-sentinel2-2023.2025-12-10.z00-z07.pmtiles",
+        "9": "https://iiab.switnet.org/maps/2/s2maps-sentinel2-2023.2025-12-10.z00-z09.pmtiles"
+    },
+    "terrain": {
+        "10": "https://iiab.switnet.org/maps/2/terrarium.2025-12-10.z00-z10.pmtiles",
+        "7": "https://iiab.switnet.org/maps/2/terrarium.2025-12-10.z00-z07.pmtiles",
+        "8": "https://iiab.switnet.org/maps/2/terrarium.2025-12-10.z00-z08.pmtiles",
+        "9": "https://iiab.switnet.org/maps/2/terrarium.2025-12-10.z00-z09.pmtiles",
+        "none": "https://iiab.switnet.org/maps/2/terrarium-none.pmtiles"
```
holta: Not that it really matters, but numerical order is much more readable if possible!

```suggestion
        "4": "https://iiab.switnet.org/maps/2/s2maps-sentinel2-2023.2025-12-10.z00-z04.pmtiles",
        "7": "https://iiab.switnet.org/maps/2/s2maps-sentinel2-2023.2025-12-10.z00-z07.pmtiles",
        "9": "https://iiab.switnet.org/maps/2/s2maps-sentinel2-2023.2025-12-10.z00-z09.pmtiles"
        "11": "https://iiab.switnet.org/maps/2/s2maps-sentinel2-2023.2025-12-10.z00-z11.pmtiles",
        "12": "https://iiab.switnet.org/maps/2/s2maps-sentinel2-2023.2025-12-10.z00-z12.pmtiles",
        "13": "https://iiab.switnet.org/maps/2/s2maps-sentinel2-2023.2025-12-10.z00-z13.pmtiles",
    },
    "terrain": {
        "none": "https://iiab.switnet.org/maps/2/terrarium-none.pmtiles"
        "7": "https://iiab.switnet.org/maps/2/terrarium.2025-12-10.z00-z07.pmtiles",
        "8": "https://iiab.switnet.org/maps/2/terrarium.2025-12-10.z00-z08.pmtiles",
        "9": "https://iiab.switnet.org/maps/2/terrarium.2025-12-10.z00-z09.pmtiles",
        "10": "https://iiab.switnet.org/maps/2/terrarium.2025-12-10.z00-z10.pmtiles",
```

orblivion: At a certain point it's easier to edit this manually. Not just for ordering. With the yml file we have the benefit of adding caveats to everything. "1" is only for CI, etc.

orblivion:
I removed the date files from main.yml. All of the downloads with dates come from the catalog.

Now, I've ended up just moving those dates explicitly to the catalog json generator script instead. So it looks somewhat like main.yml used to look. But maybe that's okay? It's easier to keep track of changes if we keep it there (if perhaps in a different repository).

https://github.com/iiab/iiab/pull/4433/changes#diff-ce6b6f7b758558f38327383ee564ee90031955812eca2b370af8da9fb709a42fR17

So I think everything works, but it's done in a rough manner. I'll need to clean things up and maybe you'll want some things done differently.

roles/maps/maps-catalog.json
orblivion: Nevermind I fixed this in the json generator. We can pick the order from there.

roles/maps/tasks/install.yml
```diff
@@ -11,6 +11,18 @@
       - python3-xmltodict
       - python3-requests
 
+- name: Download latest map-catalog.json
```
orblivion: I ended up using `/etc/iiab/maps-catalog.json` after all. It works well with ansible.

Should I have the FQR downloader update this file? Supposing a user installs world maps, then later gets an FQR and the catalog is updated. And then there's a bug. When debugging, would you prefer to see the catalog version during the world map download or the version during the FQR download?

roles/maps/tasks/install_frontend.yml
```diff
@@ -1,3 +1,5 @@
+# TODO - What if the catalog upgrades, but people re-run and don't want to upgrade global data? Maybe stop if they have data, even if it's old? Add a "quick tip" to delete it first, to allow upgrade?
```
orblivion, line 1: Supposing I install maps. Then the catalog updates. Then I runrole maps again (for an unrelated reason). Should I get the new vector map?

Maybe it shouldn't unless the user deletes the old vector map file? A new vector map could take them by surprise and they'd run out of space. Unless it deletes the old one first, but that's not safe. Deleting the old one after is only helpful if they don't run out of space.

roles/maps/tasks/install_frontend.yml
```diff
@@ -23,53 +25,63 @@
     dest_base_path: "{{ maps_serve_path }}"
   with_items: "{{ maps_dot_black_js_and_styles }}"
 
+- name: "Make sure maps_ne6_zoom has a valid value"
```
orblivion: Added this as well as the search databases to the catalog because you wanted all of the data dates removed from `main.yml`. LMK if I took you too literally.

roles/maps/tasks/install_nominatim.yml
```diff
@@ -32,15 +32,22 @@
...
+  assert:
+    that: maps_catalog.nominatim[maps_search_nominatim_db | string] is defined
+    fail_msg: "Error: maps_search_nominatim_db is set to an invalid value. See `\"nominatim\"` in https://github.com/iiab/iiab/blob/master/roles/maps/maps-catalog.json for valid values."
+    quiet: yes
+
 - name: Download nominatim database
   include_tasks: download_large_file.yml
   vars:
     dest_base_path: "{{ maps_data_root }}"
-  with_items: "{{ nominatim_data[maps_search_nominatim_db] }}"
```
orblivion, line 39: Not sure why it was `with_items` in the first place (or why it works anyway).

roles/maps/tasks/main.yml
```diff
@@ -37,8 +37,8 @@
           value: "{{ maps_enabled }}"
         - option: maps_region_downloader
           value: "{{ maps_region_downloader }}"
-        - option: maps_vector_quality
-          value: "{{ maps_vector_quality }}"
+        - option: maps_vector_zoom
```
orblivion, line 40: I renamed this to match the other variables. This okay?

Also I changed the zoom values in the catalog to just be numbers (instead of "osm-z11" etc) to match satellite and terrain. I don't want to change the dict keys later so I'd rather start off with numbers. I kept "nat-z8" because it's hopefully on its way out.

roles/maps/defaults/main.yml
```diff
@@ -98,117 +84,28 @@ maplibre_js_and_styles:
...
-  nat-z8: "naturalearth-openmaptiles.{{ maps_slow_data_date }}.z00-z08.pmtiles"
-
-  # FOR TESTING ONLY
-  # "medium res" osm, up to zoom level 1 (original file has 14).
-  # maps_vector_quality = "osm-z1"
-  osm-z1: "openstreetmap-openmaptiles.{{ maps_vector_data_date }}.z00-z01.pmtiles"
-
-maps_vector_max_zoom: 14
+# Store max zoom with the code rather than the catalog because we may want to warn users
+# before FQRs suddenly become larger, give them options, etc.
```
orblivion, line 99: How do you feel about this? If we leave the maximum to the catalog, users may get bigger FQRs than they're expecting should we ever increase the maximum zoom level.

roles/maps/tasks/install_frontend.yml
holta: Debatable?

Likewise we should probably decide if...
```
 /usr/bin/iiab-update-fqr <FQR name>
```
...installs any "missing" PMTiles files or not — e.g. if any of those 3 files are mussing?!

roles/maps/tasks/install.yml
chapmanjacobd: We should probably use `/etc/iiab/maps-catalog.json` for FQRs and have a separate command for updating the catalog similar to the difference between `apt update` and `apt upgrade`. 

Let the user have control over the `/etc/iiab/maps-catalog.json` file and eg. replace the URLs with a local custom server, etc

roles/maps/defaults/main.yml
```diff
@@ -15,21 +15,7 @@ nominatim_venv: "{{ nominatim_home }}/nominatim-venv"
...
-# time the vector data changes. This is why we have the "slow" designation.
-# Some of it may never change. Some of it, like search, will probably start
-# changing much faster once we're ready for it. We can even decide on multiple
-# date designations down the line. I just want one "slow" date to set all of
-# these items *for now*.
-#
-maps_vector_data_date: "2026-07-01"
-maps_search_data_date: "2026-04-22"
-maps_slow_data_date: "2025-12-10"
+maps_catalog_url: "https://raw.githubusercontent.com/orblivion/iiab/refs/heads/catalog/roles/maps/maps-catalog.json"
```
chapmanjacobd: should be iiab/iiab?

roles/maps/defaults/main.yml
chapmanjacobd: I agree this is a good idea to keep in the Ansible role

roles/maps/defaults/main.yml
orblivion: I added a comment

roles/maps/tasks/install.yml
orblivion: > similar to the difference between apt update and apt upgrade

In that case an FQR would be the equivalent of installing a Trixie package on top of Bookworm.

orblivion:
So in the latest version I no longer save the file as `/etc/iiab/maps-catalog.json` but I append to a log file instead. Log has timestamp and "downloaded/downloading world/fqr [url]". @holta does this cover debugging?

roles/maps/tasks/install.yml
orblivion: Oh I see, so you're talking about a situation where they want our world maps but they have their own FQR sources. Well, I've now added the custom catalog URL option, but I could make it a different one for FQRs. Though, maybe not until we've decided it's actually in demand. But it should be simple to split it now.

orblivion:
I've confirmed that everything in this draft PR has been picked out as separate smaller PRs, with #4470 being the last one.

PR #4436, opened 2026-07-04
Maps: use mirrors for Full Quality Regions

orblivion:

For Full Quality Regions, inspect the meta4 file to find all mirrors for a given pmtiles file. The full globe downloader is already doing this via `aria2c`, probably getting different chunks of the same file from different servers. In this case we use `pmtiles` which has its own special downloader so we probably can't do that, but at very least we can randomly pick a mirror for the whole file.

Ubuntu 26.04 on Multipass

@Ark74 @holta @chapmanjacobd

### DIFF
--- a/roles/maps/templates/tile-extract.py.j2
+++ b/roles/maps/templates/tile-extract.py.j2
@@ -147,8 +147,8 @@ def validate_noninteractive(noninteractive):
 
 def get_mirror(path):
     try:
-        meta4 = xmltodict.parse(requests.get(path + ".meta4").content)
-    except xml.parsers.expat.ExpatError:
+        meta4 = xmltodict.parse(requests.get(path + ".meta4").content, force_list=('url',))
+    except (xml.parsers.expat.ExpatError, requests.RequestException):
         # If something goes wrong just return the path we put in
         print("error finding mirror (xml parse error), defaulting to", path)
         return path
--- a/roles/maps/templates/tile-extract.py.j2
+++ b/roles/maps/templates/tile-extract.py.j2
@@ -344,9 +344,7 @@ def set_region_geojson(extract_box, geojson_file):
     geojson_file.write(json.dumps(region))
 
 
-def create_extract(extract_name, extract_box):
-    extract_paths = get_new_extract_paths(extract_name)
-
+def create_extract(extract_box, extract_paths):
     with tempfile.NamedTemporaryFile(delete_on_close=False, mode="w") as geojson_file:
         set_region_geojson(extract_box, geojson_file)
         geojson_file.close()
@@ -371,8 +369,7 @@ UNIT_MULT = {
 }
 
 
-def get_estimates(extract_box):
-    extract_paths = get_new_extract_paths("test")
+def get_estimates(extract_box, extract_paths):
     extract_sizes = {}
 
     with tempfile.NamedTemporaryFile(delete_on_close=False, mode="w") as geojson_file:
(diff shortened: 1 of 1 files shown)


holta:
Does it choose a random mirror?

If that mirror is inaccessible, does it have the ability to recover/persist on its own? (e.g. try another mirror)

orblivion:
Yeah it shuffles the whole list and tries them (max once per URL) until one works or it's tried them all.

chapmanjacobd (approved):

roles/maps/templates/tile-extract.py.j2
```diff
@@ -145,6 +145,28 @@ def validate_extract_box(extract_box):
 def validate_noninteractive(noninteractive):
     assert noninteractive in ("", "noninteractive"), f"`noninteractive` (optional) should only be set to \"noninteractive\". Got {noninteractive}"
 
+def get_mirror(path):
+    try:
+        meta4 = xmltodict.parse(requests.get(path + ".meta4").content)
+    except xml.parsers.expat.ExpatError:
```
chapmanjacobd: maybe (requests.RequestException, xml...) would be better to handle 404 errors

except Exception is also fine

roles/maps/templates/tile-extract.py.j2
```diff
@@ -145,6 +145,28 @@ def validate_extract_box(extract_box):
 def validate_noninteractive(noninteractive):
     assert noninteractive in ("", "noninteractive"), f"`noninteractive` (optional) should only be set to \"noninteractive\". Got {noninteractive}"
 
+def get_mirror(path):
+    try:
+        meta4 = xmltodict.parse(requests.get(path + ".meta4").content)
```
chapmanjacobd: , force_list=True

or 

force_list=('url',)

would probably be better

If there is only one URL tag it will be parsed as a dict otherwise it is a list....

---

ASIDE: actually I kinda hate xmltodict because of the footguns like these. 

I made this class XMLDict which is a bit more intuitive perhaps https://github.com/chapmanjacobd/syncweb-py/blob/main/syncweb/config.py

Feel free to look into it if you're ever doing a bunch of xml manipulation or construction. But overkill for now

roles/maps/templates/tile-extract.py.j2
```diff
@@ -145,6 +145,28 @@ def validate_extract_box(extract_box):
...
+    except xml.parsers.expat.ExpatError:
+        # If something goes wrong just return the path we put in
+        print("error finding mirror (xml parse error), defaulting to", path)
+        return path
+
+    try:
+        urls = [url['#text'] for url in meta4['metalink']['file']['url']]
+    except (KeyError, TypeError):
+        # If something goes wrong just return the path we put in
+        print("error finding mirror (xml schema error), defaulting to", path + ".meta4")
```
chapmanjacobd: should probably remove the + .meta4 in this log

roles/maps/templates/tile-extract.py.j2
orblivion: I'll add `requests.RequestException`. It will help for connectivity issues. 404 isn't an exception, so I lean on failure to parse xml for that.

roles/maps/templates/tile-extract.py.j2
orblivion: The reason I ended up with xmltodict is that I was failing to get the normal xml parsing to work, and this is pretty simple and it works now so I didn't care to spend more time (despite someone on Reddit warning not to use this library, without giving a reason). But I see your point with `force_list`.

chapmanjacobd:
lgtm

PR #4442, opened 2026-07-11
[kolibri] update proot patch: trust system CA so online import works

Ark74:
### Fixes bug:
Under proot on Android, Kolibri's "Kolibri Studio (online)" import source was shown disabled even when the device had working internet.

Screenshot_20260710-173607_Knowledge To Go
Screenshot_20260710-174927_Knowledge To Go

Other apps and a plain `curl` to the content server succeeded, only Kolibri reported it offline.

This is because Kolibri Studio option performs an HTTPS GET to the central content server via NetworkClient/SameHostSession. It requests uses its own bundled CA store (certifi), which under proot can be stale or
absent, so certificate verification fails and the whole check is treated as "offline".

So both SameHostSession and  kolibri_studio_status adds a reachability fallback: if NetworkClient fails, it retries a plain request using the system CA,  so the online option is not wrongly disabled.

Debian 13 (proot)

@holta

### DIFF
--- a/roles/proot_services/files/kolibri-00-prevent_ip_error_out_due_A15_restrictions.patch
+++ b/roles/proot_services/files/kolibri-00-prevent_ip_error_out_due_A15_restrictions.patch
@@ -1,9 +1,10 @@
 This patch is necessary on,
 
+* Android 16 - yes
 * Android 15 - yes
 * Android 14 - untested
 * Android 13 - untested
-* Android 12 - no
+* Android 12 - untested
 * Android 11 - untested
 * Andoird 10 - untested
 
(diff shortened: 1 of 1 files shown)


chapmanjacobd (approved):

jvonau:
> It requests uses its own bundled CA store (certifi), which under proot can be stale or
absent, so certificate verification fails and the whole check is treated as "offline".

Seems more like proot might be missing python3-certifi from the base install and might need to be installed? How that come about as python3-requests has a declared dependency on python3-certifi not sure. Given the whole kolibri install is based on deb packaging are they shipping python requests bundled and not relying on the system providing the library?

Curious, I haven't looked yet but does the issue with the option grayed occur on a stock iiab install of kolibri?

holta:
> Seems more like proot might be missing python3-certifi from the base install and might need to be installed?

> How that come about as python3-requests has a declared dependency on python3-certifi not sure. Given the whole kolibri install is based on deb packaging are they shipping python requests bundled and not relying on the system providing the library?

@Ark74 do you have a conclusion?  Or immediate recommendation?

luisguzman-adfa:
I'm checking deeper with a full debugging environment, and seems like the current PR is touching the wrong layer. And not fixing the underlying problem, so this might end up closed in favor of a complete analysis and deeper fix.

luisguzman-adfa:
```
Android blocks netlink (PermissionError 13)
└─ ZeroConf cannot enumerate interfaces (InterfaceChoice.All)
  └─ ZeroConfPlugin doesn't start; there's never "Registering ourselves to zeroconf network"
    └─ broadcast_id isn't generated; reset_connection_states never fires
      └─ _refresh_reserved_locations() never runs (the only place Studio creates the row)
        └─ NetworkLocation table is empty
          └─ known_location_for_address(Studio); None; NetworkLocationNotFound
            └─ kolibri_studio_status; "offline"; radio button grayed out
```

Ark74:
Sorry for the noise, I was missing a complete environment to test deeper, now #4444 fixes the actual reason, and while building it #4445 showed up to.

Regards

PR #4456, opened 2026-07-26
add missing zim lang codes

tim-moody:
dagbani and other language codes are missing. These have to been added manually, but copilot did it quickly.


holta:
Related:

- PR iiab/iiab-admin-console#666

tim-moody:
copied to existing install as test and works

PR #4458, opened 2026-07-30
maps-update.py to recover from deleted pmtiles files

orblivion:

Add a utility called `maps-update.py` that factors some aspects of map data set management out of our ansible code. We'll work on factoring out a bit more over time. The reason is that implementers can now call it outside of ansible as well. It allows them to delete large files they don't want without having to re-run ansible. It will eventually allow us to add more features as well, such as handling user-supplied data files, which is a pain to deal with in Ansible.

Right now it recovers from the implementer deleting large vector, satellite, or terrain files by pointing the symlinks to a small "fallback" file, and then putting the "maxzoom"s in a js file imported by the front end so it can display the newly pointed-to files properly. It doesn't recover from every footgun (like deleting the fallbacks) but it's a start.

This change involves having ansible always download the "fallback" files (osm z1, terrain z0, satellite z7) for `maps-update.py` to point the symlinks to. Right now I just put them in the usual www directory. I probably should move them to a special `fallback/` directory but that would add a little complication and I'd rather save that for a new PR.

This change also includes making `maps_region_downloader` only affect the front end. It uses the `pmtiles` tool, so I figure we should just always install the backend FQR stuff which includes it.

Various small unrelated items while I was at it.

Ubuntu LTS via Multipass

@holta @chapmanjacobd

### DIFF
--- a/roles/maps/templates/index.html.j2
+++ b/roles/maps/templates/index.html.j2
@@ -41,7 +41,6 @@
              *    configs   *
              ****************/
 
-            const WORLD_MAP_IS_NATURAL_EARTH = "{{ maps_vector_quality }}".startsWith("nat-")
             const MAP_SEARCH_ENGINE = "{{ maps_search_engine }}"
             const REGION_DOWNLOADER_ENABLED = "{{ maps_region_downloader }}" === "True"
 
--- a/roles/maps/templates/index.html.j2
+++ b/roles/maps/templates/index.html.j2
@@ -37,62 +37,17 @@
             const STYLE_SATELLITE = "satellite"
             const STYLE_HYBRID = "hybrid"
 
-            const VECTOR_TYPE_OSM = "osm"
-            const VECTOR_TYPE_NATURAL_EARTH = "nat"
-
             /****************
              *    configs   *
              ****************/
 
-            const WORLD_MAP_VECTOR_MAX_ZOOM = (num => {
-                if (Number.isNaN(num)) {
-                    // Use a default that is hopefully always available.
-                    // We hope they notice and report the bug if it ever comes here!
-                    console.warn("Unexpected value for maps_vector_quality")
-                    return 1
-                }
-                return num
-            })(Number(
-                // "osm-z11" -> "11", "nat-z8" -> "8", "somethingelse" -> undefined, etc
-                "{{ maps_vector_quality }}".split("-z")[1]
-            ))
-            // VECTOR_TYPE_OSM or VECTOR_TYPE_NATURAL_EARTH
-            const WORLD_MAP_VECTOR_TYPE = "{{ maps_vector_quality }}".split("-z")[0]
-
-            const WORLD_MAP_SATELLITE_MAX_ZOOM = (num => {
-                if (Number.isNaN(num)) {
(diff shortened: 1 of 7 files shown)


roles/maps/templates/index.html.j2
```diff
@@ -141,12 +95,6 @@
                 patchResource: async handlerResult => {
                     const style = handlerResult && handlerResult.data
                     if (style && style.sources && !style.iiabPatched) {
-                        // Cut down on noise. Sometimes a protomaps style is set before our chosen
```
orblivion, line 144: Unrelated. Just code I noticed was unnecessary, since I fixed the `openstreetmap-protomaps` problem once and for all.

roles/maps/templates/maps-update.py.j2
```diff
@@ -0,0 +1,76 @@
...
+vector_symlink = maps_serve_path / '{{ maps_dot_black_osm_symlink_name }}'
+satellite_symlink = maps_serve_path / '{{ maps_dot_black_satellite_symlink_name }}'
+terrain_symlink = maps_serve_path / '{{ maps_dot_black_terrain_symlink_name }}'
+
+vector_fallback = maps_fallback_path / '{{ maps_dot_black_vector_tiles["osm-z1"] }}'
+satellite_fallback = maps_fallback_path / '{{ maps_dot_black_satellite_tiles[7] }}'
+terrain_fallback = maps_fallback_path / '{{ maps_dot_black_terrain_tiles["none"] }}'
+
+def handle_fallback_file(symlink, fallback, required=True):
+    if not symlink.exists():
```
chapmanjacobd: nit: could unindent the function body one level by switching to early return

```py
if symlink.exists():
    return
```

roles/maps/templates/maps-update.py.j2
```diff
@@ -0,0 +1,76 @@
...
+# files and set up maps-config.js accordingly. If the user changes the files, this
+# script should just as well set up maps-config.js accordingly.
+
+vector_symlink = maps_serve_path / '{{ maps_dot_black_osm_symlink_name }}'
+satellite_symlink = maps_serve_path / '{{ maps_dot_black_satellite_symlink_name }}'
+terrain_symlink = maps_serve_path / '{{ maps_dot_black_terrain_symlink_name }}'
+
+vector_fallback = maps_fallback_path / '{{ maps_dot_black_vector_tiles["osm-z1"] }}'
+satellite_fallback = maps_fallback_path / '{{ maps_dot_black_satellite_tiles[7] }}'
+terrain_fallback = maps_fallback_path / '{{ maps_dot_black_terrain_tiles["none"] }}'
```
chapmanjacobd: should these reference the ansible values? eg.

```py
satellite_fallback = maps_fallback_path / '{{ maps_dot_black_satellite_tiles[maps_fallback_satellite_zoom] }}'
terrain_fallback = maps_fallback_path / '{{ maps_dot_black_terrain_tiles[maps_fallback_terrain_zoom] }}'
```

chapmanjacobd (approved):
looks good!

orblivion:
Good call on both fronts, should be addressed.

chapmanjacobd:
yes that looks correct

PR #4460, opened 2026-08-03
Move pmtiles executable

orblivion:

Rearrange variables related to map tools (pmtiles, FQR extract script), and then move the go-pmtiles executable to a different path.

The reason is that we use the pmtiles executable in maps-update.py now so it doesn't make sense to keep it under the "extract" tools directory. And I want the variables to be named more accurately.

Ubuntu 26.04 on Multipass

@holta


holta:

chapmanjacobd (approved):

PR #4461, opened 2026-08-03
maps: normalize vector map related setting names and values

orblivion:

Make the vector zoom setting more in line with satellite and terrain. Just have them be numbers for the most part.

* `maps_vector_quality` -> `maps_vector_zoom`
* `osm-z11` -> `11`
* etc

Leaving `nat-z8` for now. Hopefully we get rid of Natural Earth eventually anyway.

This is pulled from #4433. I am still mulling over some other questions regarding keys. For instance, the value `ci-1`/`1-ci` that I considered in #4433 doesn't make sense anymore if 1 is also the fallback zoom for vector. However those questions shake out, I think the change in this PR is something we can get out of the way.

Ubuntu 26.04 on Multipass

@holta

### DIFF
--- a/roles/maps/README.md
+++ b/roles/maps/README.md
@@ -28,7 +28,7 @@ Here are 3 examples below, to help you decide what you'll put in [/etc/iiab/loca
    maps_install: True
    maps_enabled: True
 
-   maps_vector_quality: nat-z8
+   maps_vector_zoom: nat-z8
    maps_satellite_zoom: 7
    ```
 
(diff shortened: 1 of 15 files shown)


holta:

chapmanjacobd:
we could optionally add something like this

```yaml
- name: Warn on deprecated maps_vector_quality
  debug:
    msg: "'maps_vector_quality' has been renamed to 'maps_vector_zoom'. Please update your configuration."
  when: maps_vector_quality is defined
```

or even fail at the beginning if we wanted people to really see it... maybe log all the deprecated settings and fail early if any match

but otherwise looks good!

orblivion:
It makes sense to have for a month or something because we don't want to waste people's time with silent errors, but I'm more inclined to make it an assert. Will people even notice a debug statement in the middle of a passing ansible run?

orblivion:
Okay I just noticed that you said assert would be fine too. I'll put that in and see what @holta says.

PR #4462, opened 2026-08-06
Move map options from main.yml to catalog.json and change some keys

orblivion:

Partial implementation pulled out from #4433

*Draft for now because I wanted to test it more and self-review more carefully. However it seems to work, and unlike #4433 I'm basically ready for this to be merged.*

#### What this does

Functionally this ought to make no change. It moves the options and comments out of `main.yml` and into `maps-catalog.json`. This doesn't download from github, it reads directly from the local repository. It makes a smaller first step. It also allows us to change the format of `maps-catalog.json` without breaking anything (which I will probably do to add `_comments` fields, see below).

It makes a bunch of the necessary changes to the install code and FQR downloading code. It gives us a chance to go over all of the nuances this already brings up before we have to think about downloading from github, caching, when to update or not update, etc.

One such nuance is what to do with the goofy settings/catalog keys I've created. I will highlight a comment where I lay out the dilemma.

#### Followup PRs to expect:

* Implement `download_from_catalog.yml` from #4433 which should clean up the `install_*` files a bit.
* Move documentation comments (the same ones I just moved out of `main.yml`) out of `generate-catalog.py` to `maps-catalog.json` using `"_comments"` fields.
* The big one: download `maps-catalog.json` from github and save to `/library/www/maps/`

(incomplete) Ubuntu 26.04 on Multipass

@holta @chapmanjacobd

### DIFF
--- a/roles/maps/README.md
+++ b/roles/maps/README.md
@@ -55,7 +55,7 @@ Here are 3 examples below, to help you decide what you'll put in [/etc/iiab/loca
 ![Vector OSM Full Zoom](README-assets/vector/vector-osm-z14.png)
 ![Satellite Full Zoom](README-assets/satellite/satellite-z12.png)
 
-See `maps_dot_black_vector_tiles` and `maps_dot_black_satellite_tiles` [here](https://github.com/iiab/iiab/blob/master/roles/maps/defaults/main.yml) for all valid values.
+See `vector` and `satellite` [here](https://github.com/iiab/iiab/blob/master/roles/maps/generate-catalog.py) for all valid values.
 
 *NOTE: The satellite data is licensed "NonCommercial" under [CC BY-NC-SA 4.0](https://creativecommons.org/licenses/by-nc-sa/4.0/).  To skip worldwide satellite imagery, set:*
 
@@ -99,7 +99,7 @@ PREREQ: Confirm that at least a [minimum IIAB Maps](#whats-a-minimum-iiab-maps-i
 
 ![Terrain Full Zoom](README-assets/terrain/terrain-z10.png)
 
-See `maps_dot_black_terrain_tiles` [here](https://github.com/iiab/iiab/blob/master/roles/maps/defaults/main.yml) for all valid values.
+See `terrain` [here](https://github.com/iiab/iiab/blob/master/roles/maps/generate-catalog.py) for all valid values.
 
 ### How do I view 3D Terrain?
 
@@ -243,10 +243,10 @@ If you are installing IIAB Maps for testing purposes (QA, CI, etc), there are "u
 [*] Grand total disk usage is [~66 MB instead of the ~312 MB delivered by default_vars.yml](https://github.com/iiab/iiab/pull/4324), as of March 2026.
 
    ```
-   maps_ne6_zoom: ci
+   maps_ne6_zoom: 4-ci
 
-   maps_vector_zoom: 1
-   maps_satellite_zoom: 4
+   maps_vector_zoom: 1-ci
+   maps_satellite_zoom: 4-ci
 
    maps_search_engine: static
    maps_search_static_db: pop-100k-cities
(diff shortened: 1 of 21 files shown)


roles/maps/generate-catalog.py
```diff
@@ -0,0 +1,186 @@
...
+  10: f"{iiab_map_host_url}/terrarium.{maps_slow_data_date}.z00-z10.pmtiles",
+
+  # maps_terrain_zoom = 0-none
+  # A "dummy" maxzoom=0 world map terrain file to fill a role that maps.black/maplibre
+  # needs if we have FQRs and the user enables terrain.
+  "0-none": f"{iiab_map_host_url}/terrarium-none.pmtiles",
+}, [7, 8, 9, 10, "0-none"])
+# TODO - let's sort this out:
+#
+# The question is the relationship between user settings and catalog keys
```
orblivion: Here's the dilemma I see with config/catalog keys

roles/maps/tasks/download_large_file.yml
```diff
@@ -35,10 +35,17 @@
 #
 # TODO - files should be owned by www-data
 
+- name: "Make sure archives end in .tar.gz as expected"
```
orblivion, line 38: I could have sworn I was already doing this

roles/maps/README.md
```diff
@@ -243,10 +243,10 @@ If you are installing IIAB Maps for testing purposes (QA, CI, etc), there are "u
 [*] Grand total disk usage is [~66 MB instead of the ~312 MB delivered by default_vars.yml](https://github.com/iiab/iiab/pull/4324), as of March 2026.
 
    ```
-   maps_ne6_zoom: ci
+   maps_ne6_zoom: 4-ci
 
-   maps_vector_zoom: 1
-   maps_satellite_zoom: 4
+   maps_vector_zoom: 1-ci
+   maps_satellite_zoom: 4-ci
```
chapmanjacobd, line 249: could we put the -ci thing into it's own variable? what is that part doing? 

or is it like '4-max' ?

roles/maps/maps-catalog.json
holta, line 1: 1. Can we please keep files in order by size?  Names like `0-none`, `1-ci`, `4-ci`, should be on top.

2. I'm not excited about repeating https://iiab.switnet.org/maps/2/ ~20 times but can live with that for now if we must (:

orblivion: 1. Do you want `maps-catalog.json` to be the eventual place that the implementer looks for documentation? I figured that if we're doing that, then we want it ordered in a way that's useful to the implementer. Wouldn't it be better to avoid having ci or testing values listed before the "useful" ones? Maybe we could solve this by doing it in reverse size order?
2. "For now" - well maybe we should change it before users end up downloading this file (could be a followup PR). Would you prefer a separated "host" entry (like `main.yml` had)?

roles/maps/README.md
orblivion: `1-ci` is a string that's a key into the catalog. In Ansible, you don't need quotes for strings if you use it like this. I'm just renaming the key to clarify what it's for. It sucks but it highlights the contradiction in trying to make `maps-catalog.json` a catalog of files and also documentation. See the comment here: https://github.com/iiab/iiab/pull/4462/changes#diff-ce6b6f7b758558f38327383ee564ee90031955812eca2b370af8da9fb709a42fR109

roles/maps/generate-catalog.py
chapmanjacobd: ah okay... 

My gut impression (which may be disregarded):

yeah I guess I would prefer Option 2. I think we should determine whether something is CI data or not by generating/pointing to a different catalog.json file... or a custom one if they want a mix of stable and CI data

I can see the benefits of only using a single maps-catalog.json file for everyone though. but I also feel like the Option 3 Annotated Keys idea might be mixing concerns

orblivion: I think your gut impression makes sense. Option 2 is the least weird. It just means that `maps-catalog.json` might not be the best place to send implementers as a reference.

Though I do plan to add `_comment` fields to it in a followup PR, so all those comments I moved from `main.yml` to `generate-catalog.py` will be moved there. The question is whether they'll read those or just look at the json keys in the catalog.

Actually, maybe I could make the catalog keys something different altogether, like maybe the full `z00-z11`? That way it's at least never the same as what goes into the config, so it doesn't get confused.

Or maybe we point people someplace else for documentation. Perhaps it could even be auto-generated by generate-catalog.py?

But, perhaps Option 2 with added `_comment` fields would be sufficient. Curious what @holta has to say.

roles/maps/maps-catalog.json
holta: Strict numerical order (ordered by size) is certainly infinitely more readable for me:

- Names can be tweaked to make meaning as intuitive & vivid as possible — e.g. if things like `7-fallback` (or whatever) are soon desired for vector & satellite?

- If a clueless implementer insists on using "content free" stubs like `1-ci`, that really is their problem — also help them be as clueful as possible with a comment on top of file maps-catalog.json, making Crystal Clear that real files (including tiny "bureaucratic" files like `1-ci` ;) are at:

  https://iiab.switnet.org/maps/2/

orblivion:
@holta I decided to remove the README field altogether for this PR. Since this version is not downloading the catalog from over the Internet, there's not much to warn about. In my next PR I'll start adding "comments" into the json file and I can add the README then.

This is only in the interest of merging this sooner. We want to get the wording right, which I think means a lot of back-and-forth fine tuning wording, and I'd rather do that after this PR is finally behind us.

But if it bothers you that there's no README, we can try to find a passable version for now.

orblivion:
Otherwise this is a "real" PR now.

chapmanjacobd:
LGTM

chapmanjacobd (approved):

PR #4466, opened 2026-08-10
Maps catalog json documentation

orblivion:

Adds documentation to `maps-catalog.json`.

Ubuntu LTS on Multipass

@holta

### DIFF
--- a/roles/maps/generate-catalog.py
+++ b/roles/maps/generate-catalog.py
@@ -135,8 +135,9 @@ maps_dot_black_satellite_tiles = dict_with_order({
     "DETAILS": json_comment("""
       maps_satellite_zoom: none
 
-      Disable satellite. (There is no URL associated with this option
-      because it doesn't download anything.)
+      Disable satellite. There is no URL associated with this option because it
+      doesn't download anything. However it will NOT delete satellite files
+      you have downloaded previously.
     """)
   },
 
--- a/roles/maps/generate-catalog.py
+++ b/roles/maps/generate-catalog.py
@@ -26,20 +26,8 @@ def fix_multiline_spacing(s):
     return '\n'.join([line.strip() for line in s.strip().split('\n')])
 
 def json_comment(s):
-    "Take a multi-line string and return a list of strings that is formatted nicely for json."
-
     assert '"' not in s, f"remove the double quotation mark (escaping it looks ugly) from:\n\n{s}"
-
-    lines = fix_multiline_spacing(s).split('\n')
-
-    longest_line = max(len(line) for line in lines)
-    return [
-        # Space before the line, and enough spaces after the line to make all
-        # of them uniform, with a space after the longest line.
-
-        " " + line + " " * (longest_line + 1 - len(line))
-        for line in lines
-    ]
+    return fix_multiline_spacing(s).split('\n')
 
 def url_only(zooms):
     return {
@@ -241,10 +229,9 @@ nominatim_data = dict_with_order({
 }, ["basic", "full"])
 
 JSON_README = """
-This is a catalog of the latest data available for IIAB Maps. The only truly valid version of
-maps-catalog.json is: https://github.com/iiab/iiab/blob/master/roles/maps/maps-catalog.json
-ASSUME ALL OTHER COPIES (INCLUDING THE ONE ON YOUR IIAB) ARE STALE (OUT OF DATE!)
-
+Catalog of the latest available IIAB Maps data:
+https://github.com/iiab/iiab/blob/master/roles/maps/maps-catalog.json
+ASSUME ALL OTHER COPIES of maps-catalog.json (INCLUDING THE ONE ON YOUR IIAB) ARE STALE AND OUT OF DATE!
 Catalog Guide: https://github.com/iiab/iiab/blob/master/roles/maps/MAPS_CATALOG_DETAILS.md
 Full Documentation: https://github.com/iiab/iiab/blob/master/roles/maps/README.md
 Raw file listing: https://iiab.switnet.org/maps/2/
--- a/roles/maps/generate-catalog.py
+++ b/roles/maps/generate-catalog.py
@@ -238,7 +238,7 @@ nominatim_data = dict_with_order({
 JSON_README = """
 This is a catalog of the latest data available for IIAB Maps. The only truly valid version of
 maps-catalog.json is: https://github.com/iiab/iiab/blob/master/roles/maps/maps-catalog.json
-ASSUME ALL OTHER COPIES ARE STALE (OUT OF DATE!)
+ASSUME ALL OTHER COPIES (INCLUDING THE ONE ON YOUR IIAB) ARE STALE (OUT OF DATE!)
 
 Catalog Guide: https://github.com/iiab/iiab/blob/master/roles/maps/CATALOG_DETAILS.md
 Full Documentation: https://github.com/iiab/iiab/blob/master/roles/maps/README.md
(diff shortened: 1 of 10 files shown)


orblivion:
I could probably stand to touch up some of the entries but I'm mostly curious what you think of the format.

roles/maps/maps-catalog.json
```diff
@@ -1,35 +1,209 @@
 {
+    "README": [
+        " This is a catalog of the latest data available for IIAB Maps.                         ",
+        "                                                                                       ",
+        " IMPORTANT: The copy of this file found at /opt/iiab/iiab/roles/maps/maps-catalog.json ",
+        " may not be the version that your IIAB will use for downloading map data. It may be    ",
+        " out of date, and any changes made here will have no effect.                           ",
```
orblivion: "Will have no effect" not actually true yet, but when we make it actually download soon it will be. I just didn't want to forget to put it and it's no benefit encouraging people to edit this in the mean time.

roles/maps/maps-catalog.json
holta, line 1: My initial vote:

- Tighten up "README" to be very succinct if possible?  e.g. < 10 lines maximum ideally!  And if you want hardcore brevity, maybe say something like... "The only truly valid version of maps-catalog.json is https://github.com/iiab/iiab/blob/master/roles/maps/maps-catalog.json — ASSUME ALL OTHER COPIES ARE STALE (OUT OF DATE!)  Documentation: https://github.com/iiab/iiab/blob/master/roles/maps/README.md"

- Can we make `maps-catalog.json` much more readable by moving "DETAILS" info another doc?  (And if "DETAILS" lines absolutely must stay in maps-catalog.json... [not my preference, makes it harder to read] then keep each "DETAILS" explanation to 1-line absolute maximum?)

orblivion: Sure thing the DETAILS don't need to be on this file at all if you don't want. I just thought you wanted it all in one place is all. I guess I could have the README point to this other documentation file.

holta: My impression of maps-catalog.json:

- A spreadsheet-like perspective appears unintuitive (very hard to read, and make sense of) if alternate lines "break the flow" making it hard to see patterns.

- Interpretive / Explanatory documentation (written or audio or video) should probably be "off to the side" (or maybe in a different file altogether) as a result ?!

orblivion: Okay the README is now 10 lines, 4 of which are blank.

I removed the details from the json and put them in this file, what are your impressions?

https://github.com/orblivion/iiab/blob/catalog-comments/roles/maps/CATALOG_DETAILS.md

roles/maps/tasks/install_frontend.yml
```diff
@@ -41,23 +41,23 @@
     that: maps_vector_quality is not defined
     msg:
       - "'maps_vector_quality' has been renamed to 'maps_vector_zoom' with new valid values. Please update your"
-      - "/etc/iiab/local_vars.yml and see https://github.com/iiab/iiab/blob/master/roles/maps/defaults/main.yml for valid values."
+      - "/etc/iiab/local_vars.yml and see https://github.com/iiab/iiab/blob/master/roles/maps/defaults/MAPS_CATALOG_DETAILS.md for valid values."
```
chapmanjacobd: /defaults/?

roles/maps/generate-catalog.py
```diff
@@ -22,120 +22,244 @@ def dict_with_order(d, ordered_keys):
...
+    "url": f"{iiab_map_host_url}/s2maps-sentinel2-2023.{maps_satellite_data_date}.z00-z04.pmtiles",
+    "details": fix_multiline_spacing("""
+      FOR TESTING ONLY
 
-  # FOR TESTING ONLY
-  # Super-low quality satellite, up to zoom level 4 (original file has 13)
-  # maps_satellite_zoom = 4
-  "4-ci": f"{iiab_map_host_url}/s2maps-sentinel2-2023.{maps_satellite_data_date}.z00-z04.pmtiles",
-}, ["4-ci", 7, 9, 11, 12, 13])
+      Super-low quality satellite, up to zoom level 4 (original file has 13)
```
chapmanjacobd, line 139: The "original file has 13" seems a bit repetitive. Maybe it would be better to put valid parameters near the heading like

# satellite

Valid parameters: 5-13, 4-ci

or

Minimum zoom: 2
Max zoom: 13

chapmanjacobd: Likewise for terrain and vector. Having a high level succinct "Valid values: X-Y" at the top of the section makes it a lot more readable

roles/maps/generate-catalog.py
```diff
@@ -22,120 +22,244 @@ def dict_with_order(d, ordered_keys):
...
+    "details": fix_multiline_spacing("""
+      Full nominatim database
+    """)
+  },
 }, ["basic", "full"])
 
+JSON_README = """
+This is a catalog of the latest data available for IIAB Maps. The only truly valid version of
+maps-catalog.json is: https://github.com/iiab/iiab/blob/master/roles/maps/maps-catalog.json
+ASSUME ALL OTHER COPIES (INCLUDING THE ONE ON YOUR IIAB) ARE STALE (OUT OF DATE!)
```
chapmanjacobd: maybe simplify to 

    ASSUME THIS COPY IS OUT OF DATE

chapmanjacobd (approved):
could be tidied a bit but nothing blocking a merge

roles/maps/generate-catalog.py
orblivion: Keep in mind that all copies, including the one that is not out of date, will carry this disclaimer. So "THIS COPY" may not be out of date. It depends on whether you're getting it from that URL.

roles/maps/generate-catalog.py
orblivion: Since it's not a blocker, I'd rather put this one off until next PR just because it's a structural change to the README.

roles/maps/generate-catalog.py
chapmanjacobd: yes I realize that. Still think the short version intent is more clear

but maybe you could just elide the word stale so that you don't need to put parentheses around "out of date"

orblivion: The "STALE" part was copying @holta 's suggestion. I added the "(INCLUDING THE ONE...)" part; removing that would make it much shorter.

* `ASSUME ALL OTHER COPIES ARE OUT OF DATE!`
* `ASSUME ALL OTHER COPIES (INCLUDING THE ONE ON YOUR IIAB) ARE OUT OF DATE!`
* `ASSUME ALL OTHER COPIES ARE STALE (OUT OF DATE!)`

I'd be fine with any of these.

chapmanjacobd: yeah the `ASSUME ALL OTHER COPIES ARE OUT OF DATE!` sounds the best to me but let's defer to Adam

holta: > ASSUME ALL OTHER COPIES ARE OUT OF DATE!

I slightly prefer ASSUME ALL OTHER COPIES ARE STALE (OUT OF DATE!) for emphasis, but above is fine too (:

> This is a catalog of the latest

Technically the very top words on top of maps-catalog.json are often false 🙃

Nitpicky yes, but maybe we can be more precise + more unambiguously clear?

As the word "this" is confusing at best here — when "this" file (staring one in the face!) is very often *NOT* the latest version of the file ;)

PR #4467, opened 2026-08-13
improve pip network resilience

chapmanjacobd:
### Fixes bug:

https://github.com/iiab/iiab/actions/runs/31703842597/job/94459309779?pr=4466

Retry failed network requests instead of giving up early

CI

@holta

### DIFF
--- a/roles/2-common/tasks/main.yml
+++ b/roles/2-common/tasks/main.yml
@@ -47,6 +47,18 @@
 #    src: zzz_iiab.sh
 
 
+- name: "Add pip network-resilience settings to /usr/local/ansible/pip.conf"
+  ini_file:
+    path: /usr/local/ansible/pip.conf
+    section: global
+    option: "{{ item.option }}"
+    value: "{{ item.value }}"
+    create: yes
+  with_items:
+    - { option: retries, value: '10' }
+    - { option: timeout, value: '60' }
+    - { option: resume-retries, value: '150' }
+
 - name: Recording STAGE 2 HAS COMPLETED ==========================
   lineinfile:
     path: "{{ iiab_env_file }}"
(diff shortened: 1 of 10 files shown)


holta:
Cool!

> ### Fixes bug:
> https://github.com/iiab/iiab/actions/runs/31703842597/job/94459309779?pr=4466

How often does this happen??

Related:

- PR #4466

chapmanjacobd:
first time that I've seen it in this spot

this fix should make it less likely in other places too

jvonau:
https://pip.pypa.io/en/stable/topics/configuration/#location

Given ansible is installed into a virtual environment should 'pip.conf' actually be installed to /usr/local/ansible/pip.conf to pick up these newer settings(resume-retries) that are available once ensure_pip is met? After all altering the system wide setting via /etc/pip.conf might cause the system version of pip not to recognize the newer settings potentially causing a failure outside of ansible. Given /usr/local/ansible/pip.conf would effectively be introduced to each role under ansible would the individual role additions be redundant at that point?

Side note, given there has been a shift to 64-bit RasPiOS why is piwheels even being added to the install if there is a message at the top of [piwheels](https://piwheels.org/) saying no 64-bit support?

chapmanjacobd:
Good points!

I made a few adjustments so this is mostly using the global --timeout and --retries options which were added in pip 1.0 (released April 2011).

I also minimized the use of `--resume-retries` to only Ansible's pip as the setting is stable starting at pip 25.2 (released July 2025). I tested pip 24.3.1 (October 2024) and it safely ignores the resume-retries setting (which it doesn't know about) but pip is updated in `scripts/ansible:ensure_pip` so this shouldn't be a problem anyway

PR #4468, opened 2026-08-14
Download maps-catalog.json from Github

orblivion:

Download the catalog from github instead of using the local version in the repository. For global map, it's cached in `/library/www/maps/maps-catalog-global-orig.json` and won't be overwritten with a new version if they runrole again.

Based largely on part of #4433

Ubuntu 26.04 on Multipass

@holta @chapmanjacobd

### DIFF
--- a/roles/maps/README.md
+++ b/roles/maps/README.md
@@ -263,16 +263,22 @@ For large file downloads:
 
 ## How to change or upgrade your IIAB Maps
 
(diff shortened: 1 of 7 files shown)


chapmanjacobd (approved):
makes sense to me! 

I wonder if we should support referencing local files but I guess it doesn't make a big difference as people would need to have a server hosting the maps data--but I suppose if they wanted to mix and match they'd need their own but that's a pretty limited edge case I'd imagine

PR #4470, opened 2026-08-14
Refactor repetitive parts of maps catalog downloads

orblivion:

Move a few repetitive tasks (check for valid config value, download map item, point symlink) into a new .yml file. Hopefully more readable and fewer errors. (In fact I found a case or two where I skipped the check for valid config value).

As I put this together I realized that I could clean it up further but I'll save it for a followup PR.

This is part of #4433 Hopefully the last unmerged part but I have to double check.

Ubuntu 26.04 on Multipass

@holta @chapmanjacobd

### DIFF
--- a/roles/maps/tasks/download_from_catalog.yml
+++ b/roles/maps/tasks/download_from_catalog.yml
@@ -26,8 +26,10 @@
       include_tasks: download_large_file.yml
       vars:
         item: "{{ catalog_file_url }}"
+
     - name: "Select {{ catalog_item }}"
       file:
         src: "{{ dest_base_path }}/{{ dest_name }}"
         path: "{{ dest_base_path }}/{{ symlink_name }}"
         state: link
+        force: yes
--- a/roles/maps/tasks/download_from_catalog.yml
+++ b/roles/maps/tasks/download_from_catalog.yml
@@ -1,0 +1,33 @@
+- vars:
+    # NOTE: `catalog_item_config_var` is the *name* of the config variable (i.e. "maps_vector_zoom") whereas
+    # `catalog_item_variant` is its *value* (i.e. "11"). We pass in the former to generate a useful error message.
+    catalog_item_variant: "{{ lookup('vars', catalog_item_config_var) }}"
+
+    catalog_file_url: "{{ maps_catalog.data[catalog_item][catalog_item_variant | string] | default('') }}"
+    catalog_file_name: "{{ catalog_file_url | split('/') | last }}"
+
+    # How it's saved. Either a file or a directory.
+    dest_name: "{{ catalog_file_name[:-7] if expand_archive | default(false) else catalog_file_name }}"
+
+  block:
+    - name: "Make sure archives end in .tar.gz as expected"
+      assert:
+        that: (not (expand_archive | default(false))) or catalog_file_name[-7:] == ".tar.gz"
+        fail_msg: "Error: expect download file name to end in .tar.gz when expand_archive=true"
+        quiet: yes
+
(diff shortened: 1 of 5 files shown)


roles/maps/tasks/download_from_catalog.yml
```diff
@@ -0,0 +1,33 @@
...
+    - name: "Make sure {{ catalog_item_config_var }} has a valid value"
+      assert:
+        that: catalog_file_url != ""
+        fail_msg: "Error: {{ catalog_item_config_var }} is set to an invalid value (\"{{ catalog_item_variant }}\"). See `\"{{ catalog_item_config_var }}\"` in https://github.com/iiab/iiab/blob/master/roles/maps/MAPS_CATALOG_DETAILS.md for valid values."
+        quiet: yes
+
+    - name: "Download {{ catalog_item }} from catalog"
+      include_tasks: download_large_file.yml
+      vars:
+        item: "{{ catalog_file_url }}"
```
chapmanjacobd, line 28: download_from_catalog.yml has `expand_archive` but do we need to pass that var to download_large_file.yml?

roles/maps/tasks/download_from_catalog.yml
```diff
@@ -0,0 +1,33 @@
...
+    - name: "Select {{ catalog_item }}"
+      file:
+        src: "{{ dest_base_path }}/{{ dest_name }}"
+        path: "{{ dest_base_path }}/{{ symlink_name }}"
+        state: link
```
chapmanjacobd, line 34: probably add `force: yes` like `ln -sf` so that if it is a dangling symlink Ansible won't fail?

roles/maps/tasks/download_from_catalog.yml
orblivion: It seems like it automatically gets passed in because it's a variable in the scope at that point. Should I pass it in explicitly?

chapmanjacobd: huh! TIL... well if it is working that's fine

chapmanjacobd (approved):

roles/maps/tasks/download_from_catalog.yml
holta: @orblivion is additional force safer?

orblivion: Make sense. Just as I have here in the cleanup script: https://github.com/iiab/iiab/blob/master/roles/maps/templates/maps-update.py.j2#L56

orblivion: @holta Hypothetically. I don't think I've ever encountered this particular error.

roles/maps/tasks/download_from_catalog.yml
orblivion: Yeah it's weird. Not sure I like it.

roles/maps/tasks/download_from_catalog.yml
orblivion: Only downside I can think of is if the implementer has placed a file there that they wouldn't want to accidentally delete?

holta: If there's a very real + common risk of self-sabotage (I highly doubt it??) ...then document that danger—

Or... are we all set to merge this as is?

orblivion: Seems like it won't matter much either way. Sure let's merge.

(Plus, I think we were talking about offloading the symlinking to `maps-update.py` before too long anyway, and that script will only delete it if it's a symlink.)

PR #4471, opened 2026-08-19
Pmtiles native antimeridian

orblivion:
### Fixes bug:

Stop using MultiPolygons to handle FQRs that cross the antimeridian. Instead I switch `pmtiles` to version `1.31.1` which includes @chapmanjacobd 's change that allows for this.

The reason I put this in the "Fixes bug" section is that I also change the way I think about coordinates. As I create extracts, I now think in terms of -180 to 180. Previously I'd have the eastmost longitude > westmost longitude, and as such eastmost is greater than 180 for FQRs that cross the antimeridian. Eventually we want to save a bbox explicitly alongside the pmtiles for the FQR, and we determined that longitudes should be -180 to 180. So while this is an implementation detail for now, it gets me toward thinking about coordinates correctly for the upcoming feature.

I'll add a caveat: There is a separate part of the code that "reads" the `pmtiles` file to pass info to the front end, both to display the FQR and to display the rectangle around it. This part of the code still inspects the file for location of tiles in case of crossing the antimeridian. I will need to do this until we finally save the bbox explicitly, as mentioned above. And in this part of the code, I still return coordinates the "old" way, i.e. eastmost longitude > westmost longitude for the benefit of the front end. In a future PR, I'll just move the coordinate translation logic to the front end, since it's only relevant to that drawing tool.

Still a draft ONLY because we should prepare the 32 bit `pmtiles` executable of version `1.31.1`.

Ubuntu 26.04 on Multipass

@holta @chapmanjacobd 

I would recommend looking at each commit separately to make it easier to follow.

### DIFF
--- a/roles/maps/templates/tile-extract.py.j2
+++ b/roles/maps/templates/tile-extract.py.j2
@@ -30,14 +30,14 @@ def get_bounds(full_path):
     # At this point we have a pmtiles file that reports to span -180 to 180, i.e. the entire
     # logical width of the map. While this might actually be the case, it's more likely that
     # it's a normal sized region that crosses the antimeridian, i.e. starts before 180 and
-    # wraps around to end after -180. But pmtiles doesn't have a concept of wrapping around
-    # the map. We were only able to create this region in the first place using a multipolygon
-    # with one part of the region after -180 and another part of the region before 180.
-    # For instance, if we created a region with longitudes from 165 to 190, we would turn it
-    # into a multipolygon from 165 to 180 and from -180 to -170. This spans from -180 to 180,
-    # and that's what gets recorded in the header (and we just got it with render_bounds).
-    # For ui_bounds, we want 165 and 190 so maplibre can draw the rectangle. And to find those
-    # numbers, we need to inspect the file to see where all of the tiles are.
+    # wraps around to end after -180. But pmtiles doesn't like the concept of wrapping
+    # around the map. We're able to create such regions by asking pmtiles to make an extract
+    # across the antimeridian, and it grabs the right tiles for the requested bbox, but in
+    # the header it still says -180 to 180 (and we just got it with render_bounds).
+    #
+    # For ui_bounds, we want the coordinates of the actual tiles (something like 165 to 190)
+    # so maplibre can draw the rectangle. And to find those numbers, we need to inspect the
+    # file to see where all of the tiles are.
 
     with open(full_path, "r+b") as f:
         source = MmapSource(f)
--- a/roles/maps/templates/tile-extract.py.j2
+++ b/roles/maps/templates/tile-extract.py.j2
@@ -1,6 +1,6 @@
 #!/usr/bin/python3
 
-import json, glob, os, random, re, requests, shutil, string, subprocess, sys, tempfile, xmltodict, xml
+import json, glob, os, random, re, requests, shutil, string, subprocess, sys, xmltodict, xml
 from pprint import pprint
 from collections import defaultdict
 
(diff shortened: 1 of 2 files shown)


roles/maps/templates/tile-extract.py.j2
```diff
@@ -466,21 +392,19 @@ def check_for_existing_name(extract_name, noninteractive):
         return True
 
 def check_for_overlaps(new_box):
```
chapmanjacobd, line 394: are we passing a pre-split string into here? I didn't notice an extra .split in the diff

roles/maps/templates/tile-extract.py.j2
```diff
@@ -141,7 +141,19 @@ def validate_extract_box(extract_box):
...
+
+    def lon_modulo(lon):
+        # We want the longitudes to be from -180 to 180. We want to perform a modulo 360
+        # to give it an equivalent longitude, but starting at -180 instead of 180
+        return (
+          (                 # Start off assuming that lon "ought to be" in the range of -180 < x < 180
+            (lon + 180)     # Thus, lon + 180 "ought to be" in the range of 0 < x < 360
+            % 360           # Modulo 360 will *in fact* change our range to 0 < x < 360
+          )
+        ) - 180             # Finally subtracting 180 will *in fact* set our range to -180 < x < 180
```
chapmanjacobd: if lon = 180 we return -180... is that correct? should it be 180 or -180? idk

maybe unrealistic edge case

chapmanjacobd (approved):

roles/maps/templates/tile-extract.py.j2
chapmanjacobd: I'm pretty sure this is fine but might be good to play around with pmtiles to see what it does

holta: > if lon = 180 we return -180... is that correct?

Sounds correct!  (But @orblivion can explain his recommendation!)

chapmanjacobd: maybe the comments could be more clear... it can't be both `-180 <= x <= 180` so it is probably `-180 <= x < 180`

roles/maps/templates/tile-extract.py.j2
orblivion: Not passing a string at all. In changing from `validate_extract_box` to `clean_extract_box`, I return the transformed integers and then pass those around. See `extract_box_str` vs `extract_box` below.

roles/maps/templates/tile-extract.py.j2
orblivion: I did not think of edge cases. I recall 180 being an interesting question. How do we represent the entire planet? (-180) to (180 - epsilon), as you say here, seemed to be one of our answers. So yeah I'll change the comment.

FWIW in Python:

```
>>> 179.9999999999999
179.9999999999999
>>> 179.99999999999999
180.0
```

Same in my browser. But I think that's enough precision. The numbers coming out of the UI rectangle tool have fewer decimal points.

roles/maps/templates/tile-extract.py.j2
orblivion: Again if it's a little hard to follow try looking at the individual commits. I tried organizing them in a way that's a little easier to follow than all at once.

orblivion: Also ignore whitespace though maybe it's a given at this point.

chapmanjacobd (approved):

PR #4472, opened 2026-08-19
MediaWiki: download from git tag instead of release tarball

chapmanjacobd:
### Fixes bug:

https://github.com/iiab/iiab/actions/runs/32272209759/job/96157191370?pr=4471

hopefully improves the mediawiki flaky downloads

### Smoke-tested on which OS or OS's:

CI

### Mention a team member @username e.g. to help with code review:

@holta

PR #4473, opened 2026-08-20
matomo: use more retries and timeouts to wait for status page

chapmanjacobd:
### Fixes bug:

flaky matomo failures

add more retries and parse the error messages instead of dumping the whole HTML page

CI

@holta

### DIFF
--- a/roles/matomo/tasks/install.yml
+++ b/roles/matomo/tasks/install.yml
@@ -139,7 +139,7 @@
 
 - name: Show Matomo Welcome page HTTP status
   debug:
-    msg: "Matomo welcome page returned HTTP {{ matomo_welcome.status }} for {{ matomo_welcome.url }}"
+    msg: "Matomo welcome page returned HTTP {{ matomo_welcome.status }} for {{ matomo_welcome.url }} (after {{ matomo_welcome.attempts }} attempt{{ 's' if matomo_welcome.attempts != 1 else '' }})"
 
 - name: Set a variable for the MATOMO_SESSID cookie
   set_fact:
--- a/roles/matomo/tasks/install.yml
+++ b/roles/matomo/tasks/install.yml
@@ -131,10 +131,15 @@
     url: "{{ matomo_full_url }}index.php?action=welcome"    # e.g. http://box.lan/matomo
     method: GET
     status_code: 200
+    timeout: 120
   register: matomo_welcome
+  until: matomo_welcome.status == 200
+  retries: 3
+  delay: 5
 
-- debug:
-    var: matomo_welcome
+- name: Show Matomo Welcome page HTTP status
+  debug:
+    msg: "Matomo welcome page returned HTTP {{ matomo_welcome.status }} for {{ matomo_welcome.url }}"
 
 - name: Set a variable for the MATOMO_SESSID cookie
   set_fact:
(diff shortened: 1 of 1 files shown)


holta:
Wow!  If this-or-similar is confirmed to solve the pesky-yet-intermittent problem!

chapmanjacobd:
should be safe to merge now. just retries, delays, and cleaner logging

roles/matomo/tasks/install.yml
```diff
@@ -131,10 +131,15 @@
...
   register: matomo_welcome
+  until: matomo_welcome.status == 200
+  retries: 3
+  delay: 5
 
-- debug:
-    var: matomo_welcome
+- name: Show Matomo Welcome page HTTP status
+  debug:
+    msg: "Matomo welcome page returned HTTP {{ matomo_welcome.status }} for {{ matomo_welcome.url }}"
```
orblivion: This is barely even a suggestion, more of an idea in case it sounds useful to you:

Just looking up how `until` works - it looks like there's an `attempts` key in the registered result that says how many times it tried. Perhaps that'd be cool to have in the debug output.

chapmanjacobd: Great idea!

roles/matomo/tasks/install.yml
```diff
@@ -154,9 +159,9 @@
     timeout: 120
     status_code: 200
   register: matomo_system_check
-
-- debug:
-    var: matomo_system_check
+  until: matomo_system_check.status == 200
+  retries: 3
+  delay: 5
```
orblivion, line 164: Deliberately no longer debugging this one?

orblivion: (and others below)

chapmanjacobd: Logging you mean? Because it's like 1MB of raw HTML so it's not very useful....

orblivion (approved):

roles/matomo/tasks/install.yml
orblivion: Well I mean `matomo_system_check` but that's probably what you mean, that's a whole HTTP response.

orblivion (approved):

roles/matomo/tasks/install.yml
chapmanjacobd: yeah they're all really big HTML pages so it just dumps a bunch of text when it fails... (Ansible does not truncate `matomo_system_check.content`) but I can add them back if Adam wants it! I just think it doesn't help much

holta: @orblivion are these ~1MB html chunks useful enough to be revealing?

chapmanjacobd: here's an example: https://github.com/iiab/iiab/actions/runs/32298679317/job/96215966207#step:5:8191

page-down from there to the bottom of the page ;-)

orblivion: @holta I brought this up because I wanted to make sure that the debugs weren't deleted by accident. I've never debugged Matomo so I will defer to @chapmanjacobd 's opinion.

holta:
Related PR: (to stabilize CI workflows, which had become erratic)

- PR #4472

Akatama:
@holta @chapmanjacobd Yes, this looks great!

PR #4477, opened 2026-08-24
generate images on manual release trigger

chapmanjacobd:

@holta @Akatama


Akatama (approved):
Looks good to me.

holta:
Let's give it a shot!

(Might need to be fine-tuned later, e.g. if timestamp as version number might prove insufficient.)

chapmanjacobd:
> timestamp as version number might prove insufficient

If the release tag starts with a number then it will use that... ah i see it's sometimes #.#-preview and sometimes release-#.#

yeah should be pretty simple to fine-tune this--just change the TAG_NAME comparison in the workflows

PR #4479, opened 2026-08-25
Maps: Put Full Quality Regions into directories

orblivion:

Changes fqr file layout to use `<name>.fqr` directories. Changes some of the CLI related to FQR deleting and name collisions.

Does not add the bbox/meta file yet. That is the next significant step. For now this still uses pmtiles file inspection for that.

Again this is somewhat involved, so I recommend that you **review individual commits in forward order**. I did my best to make commits that are self contained and comprehensible.

Ubuntu 26.04 on Multipass

@chapmanjacobd @holta

### DIFF
--- a/roles/maps/NAMING_CONVENTION.md
+++ b/roles/maps/NAMING_CONVENTION.md
@@ -10,6 +10,7 @@ How map data files are named
 | s2maps-sentinel2-2023      . 2025-12-10 . z00-z07       . pmtiles
 | naturalearth-openmaptiles  . 2025-12-10 . z00-z08       . pmtiles
 | openstreetmap-openmaptiles . 2026-07-01 . z00-z14       . pmtiles
+| terrarium                  . 2025-12-10 . z00-z08       . pmtiles
 | static-search              . 2026-04-22 . pop-1k-cities . tar.gz (*)
 | static-search              . 2026-04-22 . pop-1k-cities
 | nominatim                  . 2025-12-10 . basic         . sqlite
@@ -24,13 +25,15 @@ How map data files are named
 ## Full Quality Regions:
 
 ```
-type                       | date       | region | ...         | extension (for files)
+region | ... | type                       | date       | extension (for files)
 ---------------------------------------------------------------------------
-openstreetmap-openmaptiles . 2026-07-01 . africa . full-region . pmtiles
+africa . fqr / openstreetmap-openmaptiles . 2026-07-01 . pmtiles
+africa . fqr / s2maps-sentinel2-2023      . 2025-12-10 . pmtiles
+africa . fqr / terrarium                  . 2025-12-10 . pmtiles
 ```
 
 ```
-[type].[date].full-region.[region].[extension (for files)]
+[region].fqr/[type].[date].[extension (for files)]
 ```
 
 # Key
--- a/roles/maps/templates/index.html.j2
+++ b/roles/maps/templates/index.html.j2
@@ -79,7 +79,8 @@
 
         <script src="maps.black-component.js" type="module"></script>
         <script>
-            // source id includes a slash for convenience/laziness of repurposing it for the source url
+	    // MapLibre sources have IDs and URLs. For FQRs we use the ID to generate the URL,
+	    // which is why it looks sort of like a URL path.
             // Make sure this matches get_extract_dir and get_new_extract_paths in tile-extract.py
             makeFqrSourceId = (fqrName, baseId, dataDate) => `${fqrName}.fqr/${baseId}.${dataDate}`
 
--- a/roles/maps/templates/index.html.j2
+++ b/roles/maps/templates/index.html.j2
@@ -79,8 +79,9 @@
 
         <script src="maps.black-component.js" type="module"></script>
         <script>
-            // Make sure this matches get_new_extract_paths in tile-extract.py
-            makeFqrSourceId = (fqrName, baseId, dataDate) => `${baseId}.${dataDate}.full-region.${fqrName}`
+            // source id includes a slash for convenience/laziness of repurposing it for the source url
+            // Make sure this matches get_extract_dir and get_new_extract_paths in tile-extract.py
+            makeFqrSourceId = (fqrName, baseId, dataDate) => `${fqrName}.fqr/${baseId}.${dataDate}`
 
             // we want style choice to survive the map reset that happens when a maps.black
             // feature or style changes, so we rely on this instead of the select tag value
--- a/roles/maps/templates/index.html.j2
+++ b/roles/maps/templates/index.html.j2
@@ -79,6 +79,9 @@
 
         <script src="maps.black-component.js" type="module"></script>
         <script>
+            // Make sure this matches get_new_extract_paths in tile-extract.py
+            makeFqrSourceId = (fqrName, baseId, dataDate) => `${baseId}.${dataDate}.full-region.${fqrName}`
+
             // we want style choice to survive the map reset that happens when a maps.black
             // feature or style changes, so we rely on this instead of the select tag value
             let mapStyleChoice
@@ -114,13 +117,13 @@
                             const pmtilesDates = downloadedFQRegions.regions[fqrName].dates
 
                             const s2Id = "s2maps-sentinel2-2023"
-                            const s2FqrId = `${s2Id}.${pmtilesDates.satellite}.full-region.${fqrName}`
+                            const s2FqrId = makeFqrSourceId(fqrName, s2Id, pmtilesDates.satellite)
                             const osmId = "openstreetmap-openmaptiles"
-                            const osmFqrId = `${osmId}.${pmtilesDates.vector}.full-region.${fqrName}`
+                            const osmFqrId = makeFqrSourceId(fqrName, osmId, pmtilesDates.vector)
                             const terrariumId = "terrarium"
-                            const terrariumFqrId = `${terrariumId}.${pmtilesDates.terrain}.full-region.${fqrName}`
+                            const terrariumFqrId = makeFqrSourceId(fqrName, terrariumId, pmtilesDates.terrain)
                             const hillshadeId = "hillshade"
-                            const hillshadeFqrId = `${hillshadeId}.${pmtilesDates.terrain}.full-region.${fqrName}`
+                            const hillshadeFqrId = makeFqrSourceId(fqrName, hillshadeId, pmtilesDates.terrain)
                             const naturalEarth6Id = "naturalearth6-NE2_HR_SR_W_DR-WEBP"
                             const backgroundLayerId = "background"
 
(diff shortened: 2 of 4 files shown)


roles/maps/templates/index.html.j2
```diff
@@ -79,6 +79,10 @@
 
         <script src="maps.black-component.js" type="module"></script>
         <script>
+            // source id includes a slash for convenience/laziness of repurposing it for the source url
```
chapmanjacobd: could be more clear like "source id can include a slash" rather than describing the history ?

ie. what are the actual concerns about slashes? why is the user concerned here?? maybe this comment just leads to the user's confusion

chapmanjacobd: or be more declarative like "FQR source ids include the parent directory name"

roles/maps/templates/tile-extract.py.j2
```diff
@@ -249,56 +290,79 @@ def yes_no(prompt, noninteractive=False, noninteractive_response=None):
...
+    # below. If they are adding directories it's a them problem, not worth
+    # handling. Just give them a warning to fix it. (Adding unexpected files
+    # may be more plausibly normal behavior than adding directories)
+
+    # The user might have messed around with the files. Either there's more
+    # than one of satellite, terrain, or vector, or there are other unknown
+    # files.
+    if has_region_extract_type_dupes or has_extra_files:
+        print (f"Warning: there are unexpected files for the region '{extract_name}'")
+        print ("(expected at most one 'openstreetmap', one 's2maps', and one 'terranium', and nothing else)")
```
chapmanjacobd: I just noticed this terranium typo -- should probably be `terrarium`

roles/maps/templates/tile-extract.py.j2
```diff
@@ -1,7 +1,7 @@
 #!/usr/bin/python3
 
-import json, glob, os, random, re, requests, shutil, string, subprocess, sys, xmltodict, xml
-from pprint import pprint
+import json, os, random, re, requests, shutil, string, subprocess, sys, xmltodict, xml
+from pathlib import Path # trying to incrementally switch to this
```
chapmanjacobd: I like a lot of things about pathlib but I still use os.path occasionally too. I don't think it's bad practice to mix and match both as some things are much simpler to write in pathlib and some things are simpler using os.path.

I'd remove the comment though because it reflects temporary working context/personal notes rather than software behavior

roles/maps/templates/tile-extract.py.j2
```diff
@@ -13,7 +13,7 @@ from pmtiles.reader import MmapSource, all_tiles
 # Technically true, but there's of course the massive gap in the middle.
 # So here we inspect the file to find where that hole is.
 def get_bounds(full_path):
-    output = subprocess.getoutput(PMTILES + " show " + full_path + " --header-json")
+    output = subprocess.getoutput(f"{PMTILES} show {full_path} --header-json")
```
chapmanjacobd, line 16: `subprocess.getoutput` will use `shell=True` which... is fine when you need to use it but in general it's always better to do something like this:

    output = subprocess.run([PMTILES, "show", full_path, "--header-json"], capture_output=True, text=True)

> If the shell is invoked explicitly, via shell=True, it is the application’s responsibility to ensure that all whitespace and metacharacters are quoted appropriately to avoid shell injection vulnerabilities.
>
> https://docs.python.org/3/library/subprocess.html#security-considerations

roles/maps/templates/tile-extract.py.j2
```diff
@@ -222,17 +230,50 @@ def get_new_extract_paths(extract_name):
...
+def get_fqrs():
+    """
+    List FQRs based on directory names.
+
+    Directories with .fqr extension that are otherwise invalid are skipped with a warning.
     """
-    Get an inventory of exsting FQRs on the file system.
+    fqrs = {}
+    for fqr in Path(HOSTING_DIR).glob("*.fqr"):
+        match (fqr.name.split('.')):
```
chapmanjacobd, line 241: might be safe to do `fqr.name.rsplit('.', 1)` in case someone adds a dot in their folder name

roles/maps/templates/tile-extract.py.j2
```diff
@@ -222,17 +230,50 @@ def get_new_extract_paths(extract_name):
...
+            case [extract_name, "fqr"]:
```
chapmanjacobd, line 242: you could also do `case [*parts, "fqr"] if parts:`

could do both... then if someone is really crazy and does `my.fqr.fqr` it still works! Or just do one or the other...

roles/maps/templates/tile-extract.py.j2
```diff
@@ -222,17 +230,50 @@ def get_new_extract_paths(extract_name):
...
+                # TODO - validate extract_name?
+                fqrs[extract_name] = fqr
+            case _:
+                print(f"Warning: can't parse possible fqr dir name: {fqr}")
```
chapmanjacobd, line 246: or just keep it simple?

```py
if fqr.suffix == '.fqr':
    fqrs[fqr.stem] = fqr
else:
    print(f"Warning: can't parse possible fqr dir name: {fqr}")
```

roles/maps/templates/tile-extract.py.j2
```diff
@@ -222,17 +230,50 @@ def get_new_extract_paths(extract_name):
...
-    for full_path in glob.glob(os.path.join(HOSTING_DIR, "*.full-region.*.pmtiles")):
-        match (os.path.basename(full_path).split('.')):
-            case [region_extract_type, date, "full-region", name, "pmtiles"]:
-                pass
+    # TODO these values are repeated in the code, factor it out somewhere
+    VALID_REGION_EXTRACT_TYPES = {"s2maps-sentinel2-2023", "terrarium", "openstreetmap-openmaptiles"}
+
+    for full_path in fqr.glob("*.pmtiles"):
+        match (full_path.name.split('.')):
+            case [region_extract_type, date, "pmtiles"]:
```
chapmanjacobd, line 260: more complex but you could do a similar thing here:

```py
    case [*prefix_parts, date, "pmtiles"] if len(prefix_parts) >= 1:
        region_extract_type = '.'.join(prefix_parts)
```

roles/maps/templates/tile-extract.py.j2
```diff
@@ -249,56 +290,79 @@ def yes_no(prompt, noninteractive=False, noninteractive_response=None):
...
 
-    for path in all_paths:
-        if os.path.dirname(path) != HOSTING_DIR or f".{extract_name}." not in path:
+    for path in all_file_paths:
+        if path.parent.parent != Path(HOSTING_DIR) or path.parent.name != f"{extract_name}.fqr":
             # Last-minute sanity check before deleting files in case some other
             # error is introduced before we get to this point.
             raise Exception("Unknown error. Stopping before we delete the wrong file!", path)
-        os.remove(path)
+        path.unlink()
```
chapmanjacobd, line 350: seems like a good situation to use `shutil.rmtree(all_file_paths)` (validate all files before unlinking any)

roles/maps/NAMING_CONVENTION.md
```diff
@@ -24,13 +25,15 @@ How map data files are named
 ## Full Quality Regions:
 
 ```
-type                       | date       | region | ...         | extension (for files)
+region | ... | type                       | date       | extension (for files)
 ---------------------------------------------------------------------------
-openstreetmap-openmaptiles . 2026-07-01 . africa . full-region . pmtiles
+africa . fqr / openstreetmap-openmaptiles . 2026-07-01 . pmtiles
+africa . fqr / s2maps-sentinel2-2023      . 2025-12-10 . pmtiles
+africa . fqr / terrarium                  . 2025-12-10 . pmtiles
```
chapmanjacobd, line 32: It doesn't render as a markdown table but using the filename dots as column separators is pretty clever! maybe I'll steal that someday

chapmanjacobd (approved):
looks good overall. can merge AS-IS

keep in mind the `shell=True` thing though... I definitely recommend using the default `shell=False` as much as possible

roles/maps/templates/tile-extract.py.j2
chapmanjacobd: ASIDE: just one of the many subprocess API footguns. Try to stick with subprocess.run (and when you need something more customized: subprocess.Popen or os.spawnvpe)

chapmanjacobd: I guess one of the convenient things about getoutput is that it combines stdout and stderr into one stream. You can also do that with subprocess.run but it is a bit more verbose:

```py
result = subprocess.run(
    ...,
    stdout=subprocess.PIPE,
    stderr=subprocess.STDOUT,
    text=True,
    # check=True
)

output = result.stdout
```

roles/maps/templates/index.html.j2
orblivion: The comment isn't for users it's for fellow developers.

"MapLibre sources have IDs and URLs. For FQRs we use the ID to generate the URL which is why it looks sort of like a URL path."

roles/maps/templates/tile-extract.py.j2
orblivion: Should `subprocess.getoutput` not exist?

I'd like to put this one off. Firstly it's not a regression, it's just formatting (unless you think I've added a new injection risk). Furthermore, in my next PR I plan to delete this particular line anyway in favor of the bbox file in the fqr directory.

chapmanjacobd: yeah I've approved the merge so this is a non-blocking comment--just a strong recommendation for the future

> should subprocess.getoutput not exist?

IMHO it should be deprecated. But it's been around since Python 2. It's a useful shortcut to get specific behavior but the whole shell=True thing is a big footgun and subprocess.getoutput hides that it is using that

roles/maps/templates/tile-extract.py.j2
orblivion: If they did that it would fail to match `[extract_name, "fqr"]` which at this time is what I want. For that matter I'm planning to validate `extract_name` though I'm putting it off for a followup PR. (I should also add the character restrictions to `NAMING_CONVENTIONS.md`, I didn't think to do that.)

As for allowing users who share directories to add dots to the name, that could change things. Perhaps we can start restrictive as-is and discuss making it less restrictive over time. Or I could just put a better error message when it's clear that there's an attempt. ("Please name your directory <fqr_name>.fqr where fqr_name is alphanumeric. See `github.com/...../NAMING_CONVENTIONS.md` for more info"). cc @holta

roles/maps/templates/tile-extract.py.j2
orblivion: Cool. Then I wouldn't worry about checking whether I'm deleting files that aren't under that subdirectory. Maybe a followup though.

PR #4482, opened 2026-08-26
Fqr format followups

orblivion:
### Fixes bug:
Various followups from #4479 

* chdir to a sane place for tile-extract.py - I found that if you were in an fqr directory while deleting it you'd get weird errors. 
* Validate date in FQR pmtiles file names when inspecting the filesystem. skip bad ones with warning.
* Don't warn on deleting empty dirs (though this may be irrelevant in the future if the bbox file is a requirement for it to even be considered an fqr)
* Factor out repeated constants in tile-extract.py
* Maps: Validate FQR names when inspecting the file system. skip bad ones with warning.
* Delete FQRs with shutil.rmtree
* Give more complete warning for deletion (whole tree listing, albeit with maximum length), and allow for deleting a whole tree of stuff if the user put it in there

Ubuntu 26.04 on Multipass

@chapmanjacobd

### DIFF
--- a/roles/maps/templates/tile-extract.py.j2
+++ b/roles/maps/templates/tile-extract.py.j2
@@ -356,11 +356,12 @@ def delete_extract(extract_name):
 
     all_file_paths = list(fqr.glob("*"))
 
-    has_region_extract_type_dupes = {
+    # At most one of each type, or nothing at all
+    has_region_extract_type_dupes = fqr_region_extract_types and ({
         fqr_region_extract_types.count(region_extract_type)
         for region_extract_type
         in fqr_region_extract_types
-    } != {1}
+    } != {1})
     has_extra_files = bool(set(all_file_paths) - set(fqr_file_paths))
 
     # The user might have messed around with the files. Either there's more
--- a/roles/maps/templates/tile-extract.py.j2
+++ b/roles/maps/templates/tile-extract.py.j2
@@ -147,7 +147,7 @@ def check_extract_name_error(extract_name):
     if not (set(extract_name) < set(string.ascii_letters + string.digits + "-_")):
         return "name contains invalid characters"
     if len(extract_name) == 0:
-        return "no name provided"
+        return "name is missing"
     if len(extract_name) >= 35:
         return "name is too long"
 
--- a/roles/maps/templates/tile-extract.py.j2
+++ b/roles/maps/templates/tile-extract.py.j2
@@ -286,6 +286,26 @@ def yes_no(prompt, noninteractive=False, noninteractive_response=None):
             return False
         print ("Invalid response.")
 
+def list_files(root):
+    """
+    Walk the `root` directory and list all of the files up to `MAX_LEN`
+    (defined below) and give count of unlisted files.
+    """
+
+    listing = []
+    for directory, _, files in os.walk(root):
+        listing.append(f"* {directory}")
+        for file in files:
+            listing.append(f"* {directory}/{file}")
+
+    MAX_LEN = 20
+    print('\n'.join(listing[:MAX_LEN]))
+    unlisted = listing[MAX_LEN:]
+
+    if unlisted:
+        print('...')
+        print(f'(and {len(unlisted)} more)')
+
 # Main functions
 
 def delete_extract(extract_name):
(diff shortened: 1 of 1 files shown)


roles/maps/templates/tile-extract.py.j2
```diff
@@ -401,7 +447,7 @@ def get_estimates(extract_box, extract_paths):
             r'for an archive size of (\d*\.?\d*) (kB|MB|GB|TB)'
         )
         for line in result.stdout.split('\n'):
-            re_match = re.match(pattern, line)
+            re_match = re.match(pattern, line, re.ASCII)
```
orblivion, line 450: want \d to mean just [0-9]. If you think this would be a problem I can change the \ds to [0-9]

roles/maps/templates/tile-extract.py.j2
```diff
@@ -8,6 +8,16 @@ sys.path.insert(1, "{{ maps_tools.tile_extract.working_dir }}/pmtiles-python/pyt
...
 
+CATALOG_URL = "{{ maps_catalog_url }}"
+HOSTING_DIR = "{{ maps_serve_path }}"
+EXTRACTS_JSON = os.path.join(HOSTING_DIR, "{{ maps_tools.tile_extract.info_file }}")
+PMTILES = "{{ maps_tools.pmtiles.executable_path }}"
+TMP_DIR = "/library/downloads/maps"
+
+# If you delete an fqr directory while you're inside that same
+# directory, you end up with weird errors
+os.chdir(HOSTING_DIR)
```
chapmanjacobd, line 19: while this should be mostly fine, as a user I'd prefer it if this only happened if the current working directory is inside the HOSTING_DIR (and first also check that HOSTING_DIR exists, or move the mkdir call up prior to this chdir? I guess it's already created by Ansible?)

chapmanjacobd (commented):
looks good so far

roles/maps/templates/tile-extract.py.j2
chapmanjacobd: the weird errors are from the inode not existing anymore (even if you create the folder again it will be a different inode) and shells don't like invalid inodes

orblivion: > the weird errors are from the inode not existing anymore 

Yeah though it might not be where you'd expect. The error comes in the post-delete scan of the remaining fqrs. For whatever reason it calls `getcwd()` when scanning them.

orblivion: > while this should be mostly fine, as a user I'd prefer it if this only happened if the current working directory is inside the HOSTING_DIR

I could also chdir to the user's home? Or root? I don't care where, just a place the user has access to and that I am not potentially about to delete.

Otherwise - I could check if cwd is under `HOSTING_DIR` and only chdir to `HOSTING_DIR` in that case. That covers your concern and also covers all the fqrs.

Though, I'm curious why as a user would you prefer that?

orblivion: For the purposes of a bug in case the directory doesn't even exist, I think that's a fair point. Hypothetically their home might not exist either. I'll tentatively just go with the second option.

chapmanjacobd: >  I could check if cwd is under HOSTING_DIR and only chdir to HOSTING_DIR in that case.

Yeah that's what I mean. Limit the scope of the chdir

orblivion: Actually, the `HOSTING_DIR` is made by ansible and pretty much everything in this script would break if it didn't exist. Also since you mentioned user experience - just to be clear, after the script is done, the user is left in the directory they started with (even if it's one that the script deleted). `chdir` only affects the script as it's running.

chapmanjacobd: ahh you're right. subprocess can't change the parent shell CWD. the user will end up with a broken shell but that's their own fault. Something like this would be a bit too nanny

    shell = os.environ.get('SHELL', '/bin/sh')
    os.execl(shell, shell)

I'm good with the code AS-IS

chapmanjacobd (approved):

PR #4485, opened 2026-09-01
Clean up pmtiles installer

orblivion:
### Fixes bug:

Remnants of archive of `pmtiles` CLI tool was being left behind during installation. I left it behind using the sha256sum of the archive was a convenient way to do versioning.

But we don't want junk laying around. So instead I run `pmtiles version` and search for the git commit hash to see if we need to update. I chose to use that instead of the version number because it's less error prone. The version number might have a subset or something of another version number.

I guess I'm relying on that version number appearing consistently across the versions for all three architectures. If that fails, it will download each time. @Ark74 could you double check that running `pmtiles version` on the 32 bit version gives you `a3e4951ea6a0477b784c27c1dcbfd9c130878c5a` as part of the output?

@holta @chapmanjacobd

### DIFF
--- a/roles/maps/tasks/install_tools.yml
+++ b/roles/maps/tasks/install_tools.yml
@@ -22,6 +22,9 @@
     fail_msg: "Error: unsupported architecture for pmtiles: "
     quiet: yes
 
+# NOTE: This will download every time despite the checksum because we clean up
+# this archive after installing the executable. We could find another way to
+# prevent repeated downloads but it's not worth the complexity.
 - name: Download pmtiles archive for {{ dpkg_arch.stdout }}
   get_url:
     url: "{{ maps_tools.pmtiles.releases[dpkg_arch.stdout].src_url }}"
--- a/roles/maps/tasks/install_tools.yml
+++ b/roles/maps/tasks/install_tools.yml
@@ -42,6 +42,11 @@
     dest: "{{ maps_tools.pmtiles.executable_path }}"
     mode: "0744"
 
+- name: Clean up pmtiles unarchiving dir
+  file:
+    state: absent
+    path: "{{ maps_tools.pmtiles.archive_tmp_dir }}"
+
 # Getting the Python tool, which is not usually recommended over the Go tool,
 # so that we can write code with it that we don't need to build.
 - name: Get PMTiles (Python tool)
--- a/roles/maps/tasks/install_tools.yml
+++ b/roles/maps/tasks/install_tools.yml
@@ -4,14 +4,10 @@
     path: "{{ maps_tools.tile_extract.working_dir }}"
 
 - name: Check what version of pmtiles is installed
-  shell: |
-    sudo {{ maps_tools.pmtiles.executable_path }} version || echo
-  args:
-    executable: /bin/bash
-
-  # The existence of console output in download_file_details
-  # implies that the download should happen.
+  command: "{{ maps_tools.pmtiles.executable_path }} version"
   register: pmtiles_version
+  changed_when: false
+  failed_when: false
 
 - name: Install pmtiles tool
   when: maps_tools.pmtiles.version not in pmtiles_version.stdout
@@ -53,7 +49,7 @@
       copy:
         src: "{{ maps_tools.pmtiles.archive_tmp_dir }}/pmtiles"
         dest: "{{ maps_tools.pmtiles.executable_path }}"
-        mode: "0744"
+        mode: "0755"
 
     - name: Clean up pmtiles unarchiving dir
       file:
--- a/roles/maps/tasks/install_tools.yml
+++ b/roles/maps/tasks/install_tools.yml
@@ -3,58 +3,49 @@
     state: directory
     path: "{{ maps_tools.tile_extract.working_dir }}"
 
-- name: Check what version of pmtiles is installed
-  command: "{{ maps_tools.pmtiles.executable_path }} version"
-  register: pmtiles_version
-  changed_when: false
-  failed_when: false
+- name: Set up pmtiles unarchiving dir
+  file:
+    state: directory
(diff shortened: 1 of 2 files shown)


orblivion:
Whitespace-~only~ignore is useful here

roles/maps/tasks/install_tools.yml
```diff
@@ -3,44 +3,62 @@
...
-    path: "{{ maps_tools.pmtiles.archive_tmp_dir }}"
+- name: Check what version of pmtiles is installed
+  shell: |
+    sudo {{ maps_tools.pmtiles.executable_path }} version || echo
+  args:
+    executable: /bin/bash
+
+  # The existence of console output in download_file_details
+  # implies that the download should happen.
+  register: pmtiles_version
```
chapmanjacobd: does it really need sudo just to check the version?

maybe something like

```yaml
- name: Check what version of pmtiles is installed
  command: "{{ maps_tools.pmtiles.executable_path }} version"
  register: pmtiles_version
  changed_when: false
  failed_when: false
```

chapmanjacobd (approved):

roles/maps/tasks/install_tools.yml
orblivion: Only root can execute it. But I could change the permission if that's a bad look.

chapmanjacobd: I suppose it's fine then

holta: @orblivion

1) sudo might block without a password, e.g. given recent security hardening of RPiOS etc: https://www.raspberrypi.com/news/a-security-update-for-raspberry-pi-os/

2) With Ansible generally running as root, isn't sudo not needed here regardless?

3) Can we avoid "shell: |" when the shell command fits on a single line?

chapmanjacobd: probably should fix the permissions anyway to make it more normal (755, `-rwxr-xr-x`))

But if you don't want to change it you could use Ansible `become` instead of `sudo` https://docs.ansible.com/projects/ansible/latest/playbook_guide/playbooks_privilege_escalation.html

holta: Both sound plausible, but @orblivion please confirm the facts when ya can!

orblivion:
Okay updated based on @chapmanjacobd 's initial recommendation, and also set the chmod to 755.

Before merging @Ark74 could you confirm that the commit shows up as expected on the 32 bit version? Particularly of interest since you build from source. Run:

`sudo /opt/iiab/maps/pmtiles version | grep a3e4951ea6a0477b784c27c1dcbfd9c130878c5a`

See if the line with the hash shows up as expected? Thanks.

orblivion:
@Ark74 You don't necessarily need to update to this PR. Just have to have the same version of `pmtiles` installed as this. I just want to make sure that `pmtiles` shows the expected commit hash.

orblivion:
@Ark74 nevermind, @chapmanjacobd walked me through running it myself.

So this is going to be a bit of a problem. The commit and version number don't show up for `pmtiles version` in the built version. This means that, as written, it will download every time you runrole on ansible for the 32 bit platform.

Considering all of the bad options, I think what I'll do is record and check the hash of the executable (for each platform).

holta:
> Considering all of the bad options, I think what I'll do is record and check the hash of the executable (for each platform).

Question: How much of this is overkill?  (e.g. is version checking perhaps entirely unnecessary?)

In other words, should we consider "brute force" install of the latest/correct version of pmtiles every time an implementer / operator insists on running roles/maps/tasks/install.yml ?  

(Idea Would Be: Even if that occasionally means overwriting the pmtiles executable with the exact same bits, waiting a brief minute in this very rare scenario is no great loss.)

chapmanjacobd:
> How much of this is overkill

maybe a simpler way could be to use st_mtime and set a date in the default_vars.yml file and if we ever need people to update to the latest again we can increase the date

orblivion:
I blew past the other hypothetical answers to this issue to spare you my thinking.

One of the things I considered was to **always** update it, as you say. However I assumed you didn't want to spend the extra bandwidth. Happy to do it anyway and move on if this was a bad assumption. BTW keep in mind that we install it for everybody now, even those without the map downloader setting. (We use it for the pmtiles cleanup script with the "fallbacks" and stuff)

Another thing I considered is to **never** update it. I.e. always install from scratch if you want a working version of IIAB maps, because we don't fully support smooth upgrading yet. But after the point that we do, we'd need to remember to deal with this the first time we update pmtiles. On the other hand, I think we've only ever needed to upgrade this thing once, maybe twice. If we're lucky this won't even matter, but it's certainly not a guarantee. I suppose it does at least gives us a chance to just kick the can down the road.

(FWIW I think I'm basically done with the "overkill" change already)

holta:
> didn't want to spend the extra bandwidth

IIAB implementers have fast Internet in 2026+, in almost every country, so that's not a concern.  This is a tiny amount of MB like 16 or so (:

> we don't fully support smooth upgrading

True!  But blasting away old roles/maps (residue, no matter how old) to install whatever new roles/maps is available at that time ..is something we want to work 😇

> I think I'm basically done with the "overkill"

If it's rock solid and extremely future proof, ok!

(Conversely however, a versioned package management system feels out of scope to me, if it requires ongoing maintenance?  If that's the case, let's stick to the most robust/minimal solution possible!)

> maybe a simpler way could be to use st_mtime and set a date in the default_vars.yml file and if we ever need people to update to the latest again we can increase the date

I'd rather not?  If a dead simple solution is possible, that's much preferable to me anyway (: but @orblivion should make the decision (one way or another!)

orblivion:
Diff is now pretty dead simple

chapmanjacobd (approved):
:+1:

Issue #4483, opened 2026-08-29
local_vars.yml still has osm-vector-maps-install: True


rgarrison1:
I've tried a fresh install on Orange Pi 3b (Not a Raspberry), using Armbian minimal.
Initially the default is OSM vector maps; I found that the install would pause, download content was minimal, and then it timed out.

I was unable to complete installation; failed on maps download multiple times.

Refreshing the drive, clean install of Armbian, changed local_vars.yml setting osm_vector_maps_install: False and osm_vector_maps_enabled: False, then changed maps_install: True and maps_enabled: True

Failed on maps.black-component.js

[results.txt](https://github.com/user-attachments/files/31602161/results.txt)
[results2.txt](https://github.com/user-attachments/files/31602162/results2.txt)

jvonau:
> --connect-timeout=\"700\"
Exception: [AbstractOptionHandler.cc:69] errorCode=28 We encountered a problem while processing the option '--connect-timeout'.\n  -> [OptionHandlerImpl.cc:184] errorCode=1 connect-timeout must be between 1 and 600.\n

https://github.com/iiab/iiab/blob/master/roles/maps/tasks/download_large_file.yml#L113

Seems user induced, I recall from one of @rgarrison1 other reports that  "{{ download_timeout }}" was purposely set to 700 in local_vars.yml.

rgarrison1:
Probably true. I've had three different install instances succeed without related errors since, and yes, I automatically change the timeout to 700 in each instance.
My reasoning for the timeout change is that I'm not worried about the wait increasing, and since I'm usually using wlan rather than eth0 for the install, I thought the increase was warranted.
I agree this was probably a 1-off instance, not a bug. I had noticed that the OSM vector maps was listed as being being replaced in 2026, so I assumed I caught an issue when the change hadn't ocurred elegantly.

tim-moody:
I encountered a similar problem, not timeout but source not found:

```
TASK [osm-vector-maps : Record (initial) disk space used] **********************
changed: [127.0.0.1]

TASK [osm-vector-maps : Install packages for map installation: python3-geojson, python3-pil, python3-wget, php8.4-sqlite3 (can also be installed by www_base/tasks/php-stem.yml)] ***
[...]

PR #4486, opened 2026-09-02
default to apt-get update during package install

chapmanjacobd:
### Fixes bug:

Will likely help Tim's 404 here https://github.com/iiab/iiab/issues/4483#issuecomment-5513242653

@holta

### DIFF
--- a/iiab-from-cmdline.yml
+++ b/iiab-from-cmdline.yml
@@ -3,9 +3,6 @@
   become: yes
 
   module_defaults:
-    ansible.builtin.package:
-      update_cache: yes
-      cache_valid_time: 3600
     ansible.builtin.apt:
       update_cache: yes
       cache_valid_time: 3600
--- a/iiab-from-cmdline.yml
+++ b/iiab-from-cmdline.yml
@@ -2,6 +2,14 @@
 - hosts: all
   become: yes
 
+  module_defaults:
+    ansible.builtin.package:
+      update_cache: yes
+      cache_valid_time: 3600
+    ansible.builtin.apt:
+      update_cache: yes
+      cache_valid_time: 3600
+
   vars_files:
     - vars/default_vars.yml
     # ansible-core 2.19 disallows (thankfully no longer needed!)
--- a/iiab-from-console.yml
+++ b/iiab-from-console.yml
@@ -3,9 +3,6 @@
   become: yes
 
   module_defaults:
-    ansible.builtin.package:
-      update_cache: yes
-      cache_valid_time: 3600
     ansible.builtin.apt:
       update_cache: yes
       cache_valid_time: 3600
--- a/iiab-from-console.yml
+++ b/iiab-from-console.yml
@@ -2,6 +2,14 @@
 - hosts: all
   become: yes
 
+  module_defaults:
+    ansible.builtin.package:
+      update_cache: yes
+      cache_valid_time: 3600
+    ansible.builtin.apt:
+      update_cache: yes
+      cache_valid_time: 3600
+
   vars_files:
     - vars/default_vars.yml
     # ansible-core 2.19 disallows (thankfully no longer needed!)
--- a/iiab-network.yml
+++ b/iiab-network.yml
@@ -3,9 +3,6 @@
   become: yes
 
   module_defaults:
-    ansible.builtin.package:
-      update_cache: yes
-      cache_valid_time: 3600
     ansible.builtin.apt:
       update_cache: yes
       cache_valid_time: 3600
--- a/iiab-network.yml
+++ b/iiab-network.yml
@@ -2,6 +2,14 @@
 - hosts: all
   become: yes
 
+  module_defaults:
+    ansible.builtin.package:
+      update_cache: yes
+      cache_valid_time: 3600
+    ansible.builtin.apt:
+      update_cache: yes
+      cache_valid_time: 3600
+
   vars_files:
     - vars/default_vars.yml
     # ansible-core 2.19 disallows (thankfully no longer needed!)
--- a/iiab-stages.yml
+++ b/iiab-stages.yml
@@ -3,9 +3,6 @@
   become: yes
 
   module_defaults:
-    ansible.builtin.package:
-      update_cache: yes
-      cache_valid_time: 3600
     ansible.builtin.apt:
       update_cache: yes
       cache_valid_time: 3600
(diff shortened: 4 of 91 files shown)


holta:
> Will likely help Tim's 404 here https://github.com/iiab/iiab/issues/4483#issuecomment-5513242653

Is this confirmed?

cc: @tim-moody

holta:
Looks reasonable:

@jvonau would you agree?

jvonau:
What about turning installed roles on/off when offline?

chapmanjacobd:
ah great point!

maybe a module_defaults approach? 

maybe 7200 is better? but if apt updates once or twice over a long run I suppose that's not such a bad thing...

tim-moody:
FYI, Admin Console knows which tasks require a connection and will not run ansible if the server is not connected to the internet. I realize this is a special case, but since iiab install expects to run shortly after the OS was installed and does not expect to be repeated, it is an Adm Cons run that is likely to have a stale package. In the test I just ran apt update reported 103 packages that could be updated. I am wondering whether an ansible run should always be preceded with an apt update.

jvonau:
Rather than alter a bunch of config files and roles could the new defaults be shoved into ansible.cfg? This is more important for admin-console, where the chance of the cache being stale is more likely.

chapmanjacobd:
yeah repo-wide ansible.cfg doesn't work because the module_defaults mechanism is playbook-scoped, and Ansible does not expose a corresponding ansible.cfg  configuration setting:

    ansible-config list | grep -i module_defaults

jvonau:
> In the test I just ran apt update reported 103 packages that could be updated. I am wondering whether an ansible run should always be preceded with an apt update.

At that point I'll bet some of the roles may be stale also.

jvonau:
> yeah repo-wide ansible.cfg doesn't work because the module_defaults mechanism is playbook-scoped, and Ansible does not expose a corresponding ansible.cfg  configuration setting:
> 
> ```
> ansible-config list | grep -i module_defaults
> ```

Too bad, would of been way cleaner. 

I feel the original 'file not found' issue comes down to a transient network error where the apt repo was not reachable for whatever reason, adding a forced cache refresh only moves the failure to fetching the new cache when the repo is down.

chapmanjacobd:
> only moves the failure to fetching the new cache when the repo is down

maybe in some cases--but the good thing (at least with this module defaults approach) is that the error message will be in the same spot

but it definitely seems useful in some other cases: https://github.com/iiab/iiab/issues/4483#issuecomment-5514206096

holta:
90 files changed??  But maybe worth it 😁

jvonau:
Hard no!! 

Using state 'latest' instead of 'present' would force the cache updating see: 
https://docs.ansible.com/projects/ansible/latest/collections/ansible/builtin/package_module.html

Yet update_cache seems to work in https://github.com/iiab/iiab/blob/master/roles/tailscale/tasks/install.yml#L34-L42 otherwise the new repo introduced above would not be recognized. 

Other 'experts' sounding off https://stackoverflow.com/questions/49087220/how-to-update-package-cache-when-using-module-package-from-ansible

jvonau:
From the snicker side of me, why would fedora based plays be changed to use apt?

chapmanjacobd:
> fedora based plays

That is an interesting point. From what I understand adding any of the cache options silently changes it under the hood to use the apt module. So merging any parts of this PR might make apt work better for IIAB but it would also make Fedora installs not work.

Though the repo is already using `apt:` and `update_cache:` for a few roles....

No doubt the behavior of Ansible has changed over the years as that SO link implies

jvonau:
So doe admin-console get the same rewrite?
https://github.com/search?q=repo%3Aiiab%2Fiiab-admin-console+package&type=code

chapmanjacobd:
> state 'latest' instead of 'present' would force the cache updating

looks like that depends too:

>> You can use other states like V(latest) ONLY if they are supported by the underlying package module(s) executed.
>
> https://github.com/ansible/ansible/blob/devel/lib/ansible/modules/package.py

It looks like dnf supports update_cache but the actual behavior is completely different. Worse, `state: latest` is also different between them... apt won't update the cache without `update_cache: true`: https://github.com/ansible/ansible/blob/3827d66bfbdc568f58cce57b113131598c46cd5a/lib/ansible/modules/apt.py#L1448 so it isn't helpful in this situation... maybe we could make our own module wrapper that satisfies both of these needs though with dnf using no extra kwargs (the out of the box behavior is sane) and apt using the update_cache and other kwargs that it needs to mirror the dnf behavior

holta:
Obviously DNF is out of scope for now (:

(But sure if someone really wants to port IIAB back to Fedora, we should not make their lives difficult ;)

jvonau:
> > state 'latest' instead of 'present' would force the cache updating
> 
> looks like that depends too:
> 
> > > You can use other states like V(latest) ONLY if they are supported by the underlying package module(s) executed.
> > 
> > 
> > https://github.com/ansible/ansible/blob/devel/lib/ansible/modules/package.py
> 
> It looks like dnf supports update_cache but the actual behavior is completely different. Worse, `state: latest` is also different between them... apt won't update the cache without `update_cache: true`: https://github.com/ansible/ansible/blob/3827d66bfbdc568f58cce57b113131598c46cd5a/lib/ansible/modules/apt.py#L1448 so it isn't helpful in this situation... maybe we could make our own module wrapper that satisfies both of these needs though with dnf using no extra kwargs (the out of the box behavior is sane) and apt using the update_cache and other kwargs that it needs to mirror the dnf behavior

Using packages with 'latest' is an implied update_cache with dnf or apt backends, and update_cache is not mentioned in the docs as update_cache is not supported by all the backends reported by {{ ansible_pkg_mgr }} that packages uses to call the underling packaging program but respects update_cache when supported by the backend. 

Little history, the packages collection was chosen to be used as at the time (2013ish as [XSCE](https://github.com/XSCE/xsce) before the big gatekeeper got control) as fedora was moving from yum to dnf and debian was not even in the picture yet, it was an easy way to install the dependencies without knowing which packaging backend was in use. Next came along debian/ubuntu and some of the package names had naming differences as noted in https://github.com/iiab/iiab/blob/master/roles/2-common/tasks/yum-historical.yml.unused

At this point moving from packages to apt is not a bad thing if you are only going to support debian derivatives but only for currently supported roles. However calling update_cache (apt update) is a waste of bandwidth, should only be used when introducing a new repo or at the very beginning before anisble is installed.

Perhaps admin-console could grow an 'update OS' function to deal with stale installs noted in #4483

PR #4488, opened 2026-09-03
FQR bbox meta.json

orblivion:

When creating an FQR, store the region's `bbox` in `meta.json` in the FQR's directory. Use this `bbox` to determine what the region is for the purposes of displaying on the map, check for overlapping regions, etc. In the near future it'll be used for FQR updates. 

Also starting to add some tests. For now I have tests for detecting overlapping regions (since it's especially error prone). It might pass for now on a dev machine, but over time you may have to install it to run the tests it since `tile-extract.py` is made from a template. So when we put this in CI I think we should do it that way.

**This again invalidates all previous FQRs**. I tried my best to give useful error messages for IIAB-updating implementers to cope, namely to delete and download again. (While I was at it, I added this delete hint as a suggestion for users who are missing one or two of the 3 components [osm, satellite, terrain] in their FQR to get those items. However eventually that will be replaced with a suggestion to update.)

I didn't split this up quite as much to smaller commits. I felt like it was diminishing returns. LMK if it's hard to follow. I may add some clarifying comments here.

Ubuntu 26.04 on Multipass

@holta @chapmanjacobd

### DIFF
--- a/roles/maps/templates/tile-extract.py.j2
+++ b/roles/maps/templates/tile-extract.py.j2
@@ -39,7 +39,7 @@ def get_bounds(extract_name, fqr_dir):
         # refers to the transition of adding the meta.json file to the fqr directory.
         print (f"Warning: not displaying '{extract_name}' on the map - meta.json not found. This might be because of an incomplete download, or a recent IIAB update.")
         print ("You may need to delete this region and download it again. The command to delete it is:")
-        print() # TODO - recommend an "update" instead of delete when we implement that
+        print()
         print(delete_command(extract_name))
         print() # Extra line because we may list multiple incomplete FQRs here
     except (json.decoder.JSONDecodeError, KeyError):
--- a/roles/maps/templates/tile-extract.py.j2
+++ b/roles/maps/templates/tile-extract.py.j2
@@ -18,6 +18,14 @@ TMP_DIR = "/library/downloads/maps"
 # directory, you end up with weird errors
 os.chdir(HOSTING_DIR)
 
+def delete_command(extract_name):
+    """
+    Helper function that returns the text for the delete command, which we use in a few places.
+    """
+    return (
+        f"sudo {{ maps_tools.tile_extract.working_dir }}/tile-extract.py delete {extract_name}"
+    )
+
 # Any pmtiles file for a region that crosses 180/-180 longitude (aka the
 # antimeridian) will say its min and max longitudes are -180 and 180.
 # Technically true, but there's of course the massive gap in the middle.
@@ -501,12 +509,16 @@ def check_for_existing_name(extract_name):
         # automatically. It's a separate non-trivial task because the deletion
         # sequence itself has yes/no logic related to unexpected files. And so
         # the deletion may not happen, which we then have to account for here.
-        print(f'Already have a full quality region called \"{extract_name}\". If you would like to overwrite it, please delete it first:')
-        print("")
-        print(f"sudo {{ maps_tools.tile_extract.working_dir }}/tile-extract.py delete {extract_name}")
+        #
+        # I specifically say "looks like" because this comes up (as it should)
+        # even if you have an invalid fqr, i.e. a directory without a meta.json
+        # file. Such a directory would, all the same, block you from
+        # downloading a new region with that name.
+        print(f'It looks like you already have a Full Quality Region called \"{extract_name}\". To delete this region, run:')
+        print() # TODO - recommend an "update" in case that's what they want to do, when we implement that
+        print(delete_command(extract_name))
         return False
-    else:
-        return True
+    return True
 
 def check_for_overlaps(new_box):
     [new_min_lon, new_min_lat, new_max_lon, new_max_lat] = new_box
--- a/roles/maps/templates/tile-extract.py.j2
+++ b/roles/maps/templates/tile-extract.py.j2
@@ -146,7 +146,7 @@ def check_bbox_error(bbox):
         if lat > 90:
             return "latitude > 90"
         if lat < -90:
-            return "longitude < -90"
+            return "latitude < -90"
 
 def clean_bbox_from_ui_tool(extract_box_str):
     """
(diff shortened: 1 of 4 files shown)


roles/maps/templates/tile-extract.py.j2
```diff
@@ -547,32 +582,27 @@ def update_extracts_json():
     region_extracts = defaultdict(
         lambda: dict(dates=defaultdict(dict)),
     )
-    # region_extract_types[extract_name] = {extract_type_1, extract_type_2, etc}
-    region_extract_types = defaultdict(set)
```
orblivion, line 551: I realized that `region_extract_types[extract_name]` is redundant with `region_extracts[extract_name]["dates"].keys()`. The only difference is the warning message. I think it now says `openstreetmap-openmaptiles` instead of `vector`, etc.

roles/maps/templates/tile-extract.py.j2
```diff
@@ -547,32 +582,27 @@ def update_extracts_json():
...
 
-    for full_path, region_extract_type, date, extract_name in iter_region_files():
-        render_bounds, ui_bounds = get_bounds(full_path)
+    for extract_name, fqr_dir in get_fqrs().items():
+        render_bounds, ui_bounds = get_bounds(extract_name, fqr_dir)
         if render_bounds is None:
             continue # Something went really haywire. Skip this pmtiles file.
-
-        # Add coordinates unless it's for some reason inconsistent between satellite, vector, terrain.
-        if extract_name in region_extracts:
```
orblivion, line 559: We now have one unambiguous bbox from meta.json so we don't need to check if the pmtiles report different bboxes.

roles/maps/templates/tile-extract.py.j2
```diff
@@ -160,7 +110,49 @@ def validate_extract_name(extract_name):
...
+        if lon < -180:
+            return "longitude < -180"
+
+    if min_lat >= max_lat:
+        return "latitudes are equal or out of order"
+    for lat in [min_lat, max_lat]:
+        if lat > 90:
+            return "latitude > 90"
+        if lat < -90:
+            return "longitude < -90"
```
chapmanjacobd: should be "latitude < -90"?

roles/maps/templates/tile-extract.py.j2
```diff
@@ -501,30 +512,54 @@ def check_for_existing_name(extract_name):
...
-
-        if (
-            new_min_lon < existing_max_lon and
-            new_max_lon > existing_min_lon and
-            new_min_lat  < existing_max_lat and
-            new_max_lat  > existing_min_lat
-        ):
+
+    for existing_region in existing_region_extracts.values():
+        if _has_overlap(new_bbox, existing_region["ui_bounds"]):
```
chapmanjacobd: is this function still good for checking `ui_bounds` vs the original `bbox`? 

eg. the `bbox[2] + 360` thing, etc

seems tricky to do both? maybe more tests are needed for ui_bounds style inputs with values > 180...?

roles/maps/templates/tile-extract.py.j2
```diff
@@ -501,30 +512,54 @@ def check_for_existing_name(extract_name):
...
+        bbox_1[1] < bbox_2[3] and
+        bbox_2[1] < bbox_1[3]
+    )
+
+def check_for_overlaps(new_bbox):
+    """
+    Assumes new_box and existing boxes are validated
+    """
 
     existing_region_extracts = json.loads(open(EXTRACTS_JSON).read())["regions"]
```
chapmanjacobd: Ideally use context managers with `open()` as leaving the file descriptors open for a long time can lead to confusing results when there is an error (partial writes and/or corrupted files). On windows you could even be blocked from opening the file again if the program already has a file descriptor open somewhere (luckily no issue on linux but this is still something we should fix for the first reason)

maybe you'll like pathlib's `content = file_path.read_text(encoding="utf-8")` and `file_path.write_text("Hello, World!", encoding="utf-8")` though. That's also fine too

chapmanjacobd: eg. `json.loads(EXTRACTS_JSON.read_text())` if EXTRACTS_JSON is a pathlib path object

chapmanjacobd (changes requested):
looks promising so far... just a few issues we'll want to address before merging

roles/maps/templates/tile-extract.py.j2
orblivion: `new_bbox` comes from the return value from `clean_bbox_from_ui_tool` which already does the `lon_modulo` thing. `existing_region_extracts` comes from `extracts.json` which was made with `update_extracts_json` which reads all of the `meta.json` files and double checks that all of those values are valid. So, all of the bboxes here should be based on our decided standard, -180 to 180 longitude.

Granted, that's just how this function is currently used and it's not obvious and possibly error prone. But I want to clean this all up in the next couple PRs. One big thing would be to push all of the `+ 360` logic (both input and output), as well as the `render_bounds` logic, to the front end, since those are all display-related concerns. This extract utility should ideally just deal in -180 to 180 bboxes. Another thing would be better encapsulation.

orblivion: (You should be able to confirm that all bboxes in this function currently pass through `check_bbox_error`)

chapmanjacobd: okay I think I get it now :+1:

chapmanjacobd (approved):
looks good to merge once those dangling file descriptors are fixed

roles/maps/templates/tile-extract.py.j2
orblivion: Actually you were right to begin with. Even though it passes `check_bbox_error` I convert it to the `ui_bounds` format afterwards, and I directly use that in this function. I discovered this by manually testing overlaps at 180, actually.

I was going to just iterate over the `meta.json` files here at some point in the near future, so I'll just do that now.

orblivion: (I just didn't see what you were directly pointing at.)

orblivion:
I deleted the file descriptor you pointed to but I have three others in this file, let me fix those.

chapmanjacobd (approved):

PR #4489, opened 2026-09-04
Update the region extract script to only deal in our standard `bbox`

orblivion:
Sorry it's a lot of words here, but I think it's a relatively simple change.

Update the region extract script to only deal in our standard "-180 to 180" `bbox` (i.e. the one we save to `meta.json` for FQRs). Previously, the extract script would handle transformations that were useful for the UI (`ui_bounds`, `render_bounds`, etc). Now those transformations have been moved to the UI. I realized that it makes sense to push it closer to where it's actually needed.

#### Firstly, *inputs* to the extract script

When generating the CLI command for the extract script in the UI, we now transform the longitudes so that they are always between -180 and 180. Previously, the UI could pass in longitudes of 170 to 190 and the extract script would convert it to 170 to -170. Now the UI does this when generating the extract script command, and the script requires it.

While I was at it: if the user makes a region that takes up more than the width of the world, it now turns the longitudes into -179.99999999 to 180 so they get the whole width. Previously the extract script threw an error if the longitude difference was more than 360.

#### Secondly, *outputs* from the extract script

No longer saving `ui_bounds` and `render_bounds` to `extracts.json` (the file that the front end reads to learn about available FQRs). Instead, we directly save `bbox`, just like `meta.json`. (We actually had this before, but it was before we came up with our new longitude standard). `ui_bounds` and `render_bounds` are now calculated by the front end upon reading `extracts.json`.

*(If you're curious: `ui_bounds` is used to draw the rectangle for the FQR, and it seems to want the same coordinate system that the rectangle drawing tool wants, i.e. east longitude > west longitude instead of -180 to 180. `render_bounds` is how big of an area is covered by pmtiles. It actually wants longitude values within -180 to 180, but for regions that cross the antimeridian [i.e. 170 to -170], it needs the values to cover the whole width of the planet, i.e. -180 to 180)*

#### Discovered along the way: change of our `bbox` standard

Update our bbox standard to match MapLibre's `wrap()`, which does the desired longitude transformation (similar to the `lon_modulo` function I previously had in the extract script). Our standard was previously -180 to 179.9999.... MapLibre seems to go by -179.9999... to 180, and I think it would make our lives easier to match them.

I also document this (along with `meta.json` in general) by renaming `NAMING_CONVENTIONS.md` to `FORMAT.md` and adding a section. Not sure where else to put it.

Ubuntu 26.04 on Multipass

@chapmanjacobd @holta

### DIFF
--- a/roles/maps/FORMAT.md
+++ b/roles/maps/FORMAT.md
@@ -1,4 +1,4 @@
-# Naming Convention
+# Naming Conventions
 
 How map data files are named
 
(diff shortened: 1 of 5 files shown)


chapmanjacobd (approved):
Interesting! Looks like things are becoming more consistent. Nice improvements!

PR #4493, opened 2026-09-09
maps: encapsulate and further test FQR bbox logic

orblivion:

Make an `FQR` class in a new separate file `fqr.py`. Move some of the functions from `tile-extract.py` (bbox validation, overlap detection) to methods on `FQR`. The main reason is that I hope this makes it easier to maintain and follow what's going on. Unlike passing around `bbox`es which are just Python arrays, if an `FQR` object exists it means it's already passed the validations. Other nice things are `fqr.min_lon` instead of `bbox[0]`. And just splitting the code into files is nice.

Also test more of `FQR`'s functionality.

In the process I changed how the overlap function works. Initially due to some complication from how I made the `FQR` class, and thinking about a new edge case, but I think the current method is more straightforward anyway.

Again I tried to split the commits into chunks to make it a little easier to follow. The first commit is just moving function code as much as I could.

Ubuntu 26.04 Multipass

@chapmanjacobd @holta


chapmanjacobd (approved):
looks good!

Thinking about this more... If we wanted to normalize input -180 to just be 180 I think that could also be fine. But being strict and raising an error is even better (even if it is a bit arbitrary!) 

(I guess we won't actually have data like that in the wild if it is coming from maplibre and maplibre plugins)

PR #4494, opened 2026-09-09
maps: assorted small cleanup

orblivion:
### Fixes bug:

* Fix a small bug by switching from `os.walk()` to `Path.walk()`: we were not listing symlinks to directories within FQRs when deleting FQRs. (such symlinks would be user-created, and for reasons I can't anticipate. I justed want to warn about everything we're about to delete.). [See commit message](248e5e0a865814661774fb5b2c46e8b74314be28) for details.

* all-in on pathlib (had a tiny amount left, and it didn't require any adjustment)
* A couple tiny touchups in downloader script output
* Code cleanup, including pylint suggestions.

### DIFF
--- a/roles/maps/templates/tile-extract.py.j2
+++ b/roles/maps/templates/tile-extract.py.j2
@@ -37,7 +37,10 @@ def get_fqr(extract_name, fqr_dir):
     except FileNotFoundError:
         # TODO nov 2026 or so we can remove "updated your IIAB recently". This
         # refers to the transition of adding the meta.json file to the fqr directory.
-        return None, "meta.json not found. This might be because of an incomplete download, or a recent IIAB update."
+        return None, (
+            "meta.json not found. This might be because of an incomplete download, "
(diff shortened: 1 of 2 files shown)


roles/maps/templates/tile-extract.py.j2
```diff
@@ -221,7 +228,7 @@ def list_files(root):
     """
 
     listing = []
-    for directory, _, files in os.walk(root):
+    for directory, _, files in root.walk():
```
chapmanjacobd, line 231: FWIW: Path.walk() was only introduced in Python 3.12

just something to keep in mind if a user is running Python 3.11 or earlier. I think this should be pretty safe though

chapmanjacobd (approved):

roles/maps/templates/tile-extract.py.j2
holta: Python 3.15 becomes official Oct 1, so I'm not worried (:

PR #4495, opened 2026-09-09
Kolibri PPA: Ubuntu 26.10 disallows SHA1 apt pkg sigs, so do as on Debian

holta:
### Fixes bug:

- https://github.com/learningequality/kolibri/issues/15277

### Description of changes proposed in this pull request:

Do as on Debian 13+, as also explained here:
https://kolibri.readthedocs.io/en/latest/install/ubuntu-debian.html

### Smoke-tested on which OS or OS's:

Ubuntu 26.10 "Stonking Stingray" which is scheduled for final release on October 15, 2026.

### Mention a team member @username e.g. to help with code review:

@chapmanjacobd

PR #4496, opened 2026-09-09
Modernize Kolibri PPA apt pkg sig across all OS's [required on Debian 13+ & Ubuntu 26.10+]

holta:
Possible alternative/replacement for:

- PR #4495

(If this works across sufficient recent OS's, then let's make the jump!  Tested on Ubuntu 26.10 with GitHub Actions CI workflows about to provide more details on other OS's...)

@chapmanjacobd is get_url or get_curl preferred?


chapmanjacobd:
Yes I suppose this makes sense also

use get_curl ?

chapmanjacobd (approved):

PR #4497, opened 2026-09-09
add new uwsgi-initscripts package to restore prior behavior [for Captive Portal, on Ubuntu 26.10]

chapmanjacobd:
### Fixes bug:

systemd service uwsgi somehow being missing on Ubuntu 26.10?

add uwsgi-initscripts

https://packages.ubuntu.com/stonking/all/uwsgi-initscripts/filelist

We should test this on 26.10... should we add a CI test for _stonking_?

@holta

### DIFF
--- a/roles/nginx/tasks/install.yml
+++ b/roles/nginx/tasks/install.yml
@@ -33,6 +33,7 @@
 - name: Reload systemd after installing uWSGI packages
   systemd:
     daemon_reload: yes
+  when: not is_proot
 
 
 # 2021-08-07: Legacy from roles/httpd/tasks/install.yml
--- a/roles/nginx/tasks/install.yml
+++ b/roles/nginx/tasks/install.yml
@@ -24,16 +24,15 @@
 
 # Ubuntu 26.10 moved the legacy uwsgi.service and apps-enabled support into a separate package: uwsgi-initscripts
 # IIAB uses this aggregate service for Admin Console and Captive Portal
-- name: 'Install legacy uWSGI service support on Ubuntu 26.10+'
+- name: Install legacy uWSGI service support (if available; for changes in Ubuntu 26.10+)
   package:
     name: uwsgi-initscripts
     state: present
-  when: ansible_facts.distribution == 'Ubuntu' and ansible_facts.distribution_version is version('26.10', '>=')
+  ignore_errors: true
 
-- name: Reload systemd after installing legacy uWSGI service support
+- name: Reload systemd after installing uWSGI packages
   systemd:
     daemon_reload: yes
-  when: ansible_facts.distribution == 'Ubuntu' and ansible_facts.distribution_version is version('26.10', '>=')
 
 
 # 2021-08-07: Legacy from roles/httpd/tasks/install.yml
--- a/roles/nginx/tasks/install.yml
+++ b/roles/nginx/tasks/install.yml
@@ -28,7 +28,7 @@
   package:
     name: uwsgi-initscripts
     state: present
-  when: is_ubuntu and ansible_facts.distribution_version is version('26.10', '>=')
+  when: ansible_facts.distribution == 'Ubuntu' and ansible_facts.distribution_version is version('26.10', '>=')
 
 
 # 2021-08-07: Legacy from roles/httpd/tasks/install.yml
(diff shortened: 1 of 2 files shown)


holta:
> should we add a CI test for stonking?

Is that easily feasible?  (I thought GitHub Actions did not support non-LTS Ubuntu releases.)

roles/nginx/tasks/install.yml
```diff
@@ -22,6 +22,14 @@
       - uwsgi-plugin-python3    # these 2 packages on demand (not every IIAB needs these).
     state: present
 
+# Ubuntu 26.10 moved the legacy uwsgi.service and apps-enabled support into a separate package: uwsgi-initscripts
+# IIAB uses this aggregate service for Admin Console and Captive Portal
+- name: 'Install legacy uWSGI service support on Ubuntu 26.10+'
+  package:
+    name: uwsgi-initscripts
+    state: present
+  when: is_ubuntu and ansible_facts.distribution_version is version('26.10', '>=')
```
holta: `is_linuxmint` and `is_trisquel` are both subsets of `is_ubuntu` so be careful with numerical inequality operations here!  (Coincidentally Mint is about to release 23, and Trisquel should release 13, so coincidentally the '>= 26.10' would probably work... but sets a terrible + unsafe precedent ;-)

https://github.com/iiab/iiab/blob/3a969fdb42abf2e140dd7006b2e1b90eb7acbd44/roles/0-init/tasks/main.yml#L19-L25

holta:
> We should test this on 26.10

@chapmanjacobd can you do that quickly, using Multipass or similar?

chapmanjacobd:
ah I already tested that installing it in ubuntu 26.10 systemd-nspawn adds the missing service* so I think it should work fine

good point about the is_ubuntu subsets... lets see if I can tighten that up

\* or rather it relies on `systemd-sysv-generator` to convert the SysV service to systemd--same as before. But because the file isn't there on boot, we may need to explicitly run `systemctl daemon-reload` after installing if we want it immediately available

roles/nginx/tasks/install.yml
chapmanjacobd: should we do `ansible_facts.distribution == 'Ubuntu'` ? I wonder how we'll corral this when the derivatives start doing the same thing...

holta:
Good enough to merge?  Or does this need more testing / reviewing?

chapmanjacobd:
good to merge if the `ansible_facts.distribution == 'Ubuntu'` thing is okay with you. I'm sure we'll tweak this over time

holta:
> good to merge if the `ansible_facts.distribution == 'Ubuntu'` thing is okay with you. I'm sure we'll tweak this over time

Good point: Mint 23 will need this very shortly in not-so-many months/weeks/days from now...

- Can we try to harden this based on the actual availability of package `uwsgi-initscripts` ?
- Or even a quick+dirty approach apt installing `uwsgi-initscripts` with Ansible's `ignore_errors: true` might be lean and clean enough?

chapmanjacobd:
yeah I think `ignore_errors: true` is the right approach here. one sec

Ark74 (approved):
It's ok, proot will continue to use Debian 13 for a long time, so Ubuntu related changes won't affect much the current installation.

holta:
@chapmanjacobd is it possible this PR isn't enough to solve the [roles/captiveportal](https://github.com/iiab/iiab/tree/master/roles/captiveportal) problem on Ubuntu 26.10 ?

```
TASK [captiveportal : Start & Enable 'uwsgi' systemd service (may include 3 child processes for Captive Portal, and 3 child processes for Admin Console)] ***********************************************************************************
[ERROR]: Task failed: Module failed: Unable to enable service uwsgi: Failed to enable unit: Unit uwsgi.service does not exist

Origin: /opt/iiab/iiab/roles/captiveportal/tasks/enable-or-disable.yml:51:3

49 # Let's assume (for now) that Admin Console was/will be installed. Disabling of
50 # uWSGI's service can be implemented later as nec, e.g. if CP+AC both disabled?
51 - name: Start & Enable 'uwsgi' systemd service (may include 3 child processes for Captive Portal, and 3 child proc...
     ^ column 3

fatal: [127.0.0.1]: FAILED! => {"changed": false, "msg": "Unable to enable service uwsgi: Failed to enable unit: Unit uwsgi.service does not exist\n"}
```

chapmanjacobd:
ah interesting... maybe it isn't being installed if nginx is already installed earlier. Moving it from `roles/nginx/tasks/install.yml` to `roles/nginx/tasks/main.yml` should fix this:

https://github.com/iiab/iiab/pull/4501

Issue #4498, opened 2026-09-11
USB Upload not working


diensthunds:
Having read through the FAQ, and searched the already open, and closed issues, I can not find out why the USB upload is not working. 
I can download anything that's on the USB drive. But I can not upload anything. 
There is plenty of room on the drive.

Error message given:

ERROR: Please make sure one and ONLY one (no more, no less) removable USB stick is plugged into your Internet-in-a-Box. Please see IIAB FAQ, "[Can students upload their own work?](https://wiki.iiab.io/go/FAQ#Can_students_upload_their_own_work?)", for additional support.

Not able to create upload directory. 
Make sure 'usb_lib_writable_sticks' is set to 'True'.

I have confirmed that the usb_lib_writable_sticks is set to True.
I've tried putting a PUBLIC folder in the root of the USB and I've tried it without it being there. Same error either way.

holta:
Hi @diensthunds !

Are you able to run `sudo iiab-diagnostics` and share with us (post right here) the resulting URL?

diensthunds:
https://paste.centos.org/view/dc7bf0a6

holta:
@chapmanjacobd can you try to reproduce this?

chapmanjacobd:
> PHP Warning:  mkdir(): Permission denied in /library/www/html/upload2usb/upload2usb.php on line 57; PHP message: UPLOAD2USB ERROR: RuntimeException: {"usb_count":-1,"exception_msg":"Not able to create upload directory

Looks like the issue is people usually use exFAT or NTFS for the USB stick but ext4 isn't working out of the box because it actually does have Linux native permissions so the umask trick doesn't work.

You can verify this by checking the exit code of

    sudo -u www-data test -w /media/usb0
    echo $?

Prints `1` right? Does running this fix the problem?

    chmod u+rwx /media/usb0

diensthunds:
The reason I used ext4 was for the file size limits. I have some other zim files that I want to add to the collection that I don't see available through iiab. Some of which are rather large. I picked ext4 to format the disk because of the larger files. Didn't realize that extfat had such a large file size limit. I suppose I could go back and format the drive to NTFS as I can then use it on my other machines (a windows laptop and a macbook ((supported with additional software to read NTFS drives)).
Never thought about the ext4 on the drive being the issue. I even added in the -fstype=ext4 into the variable, but that didn't happen.

At any rate.
usb0 was already 755
doing the sudo -u www-data test -w /media/usb0 
[...]

Issue #3907, opened 2025-01-19
Rename usb_lib_umask0000_for_kolibri to be more generally applicable


avni:
- Rename usb_lib_umask0000_for_kolibri to ~~umask0000_for_writable_sticks~~ usb_lib_writable_sticks 
- ~~make roles/usb_lib refuse to install if umask0000_for_writable_sticks: True has not been set.~~

@holta, assign this to me and I'll work on this after #3875 (upload2usb v0.1) is merged.

holta:
High priority & easy, now that PR #3875 is merged!

@avni the name `umask0000_for_writable_sticks` is fine.  If others prefer `usb_lib_writable_sticks` that's also OK.

holta:
> * make roles/usb_lib refuse to install if umask0000_for_writable_sticks: True has not been set.

I'd suggest maybe dropping that idea for now?

(While not the norm, there might be a small demographic who legitimately want read-only USB sticks...)

avni:
> (While not the norm, there might be a small demographic who legitimately want read-only USB sticks...)

Thanks, that's a valid pt. (Also makes it a lot easier to close this out).

avni:
### Test Cases

**Setup:** 
- curl iiab.io/install.txt | bash -s 3913
- Reboot server. 

**Test case 1:** 
- Go to: http://box/local_content/upload/ and test a couple of the test cases from here: [LINK](https://github.com/iiab/iiab/pull/3875#issuecomment-2599628892)
- Confirm uploaded files are on USB stick.

**Test case 2:** 
- Go to: http://box/kolibri/en/device/#/content/export_channels and export a small channel such as "_Using Studio: Your Content Workspace for Kolibri_" (~200MB) 
- Confirm that the KOLIBRI_DATA folder was created on USB stick and has content from the channel in it.

EMG70:
Test case 1:

![Image](https://github.com/user-attachments/assets/0eb8a203-0386-4349-b30c-73f9b7cd4afa)
![Image](https://github.com/user-attachments/assets/9064c15e-3adb-46c5-a952-2136950dd216)

Test case 2:

![Image](https://github.com/user-attachments/assets/38f9d113-2e69-40f7-8177-a7c9a6959679)

![Image](https://github.com/user-attachments/assets/40172243-c6f2-49a2-9c42-c989eba3a964)

![Image](https://github.com/user-attachments/assets/c034c808-c407-466d-8b88-3df17346c4f3)
[...]

PR #4499, opened 2026-09-11
Also support USB stick filesystems ext2, ext3, ext4, btrfs, xfs, f2fs

chapmanjacobd:
### Fixes bug:

https://github.com/iiab/iiab/issues/4498#issuecomment-5636221750


holta:
@diensthunds would you be able to try this?

You can install a fresh (brand new) IIAB to test this on a clean/new Raspberry Pi OS as follows:

```
curl iiab.io/install.txt | bash -s 4499
```

diensthunds:
> @diensthunds would you be able to try this?
> 
> You can install a fresh (brand new) IIAB to test this on a clean/new Raspberry Pi OS as follows:
> 
> ```
> curl iiab.io/install.txt | bash -s 4499
> ```

I can try it with another sd card. 
It will take me awhile to set up the base OS and then do the IIAB install.

roles/usb_lib/templates/mount.d/70-usb-library.j2
```diff
@@ -44,16 +44,19 @@ elif [ "$UM_DEV" == "$BOOTFW_DEV" ]; then
...
 
 {% if usb_lib_writable_sticks %}
-# ext4 stores Unix permissions on disk and does not support umask mount options.
+# Native Linux filesystems store Unix permissions on disk and do not support
+# umask mount options.
 # Make the mount root writable by the web server without changing its contents.
-if [ "$UM_FILESYSTEM" = "ext4" ]; then
-    logger -t "usb_lib (70-usb-library)" "Making ext4 mount root writable by {{ apache_user }}"
+case "$UM_FILESYSTEM" in
+ext2|ext3|ext4|btrfs|xfs|f2fs)
```
holta: Lines 51-58 should probably be indented with 4 additional spaces?

roles/usb_lib/templates/mount.d/70-usb-library.j2
```diff
@@ -43,6 +43,22 @@ elif [ "$UM_DEV" == "$BOOTFW_DEV" ]; then
...
 
+{% if usb_lib_writable_sticks %}
+# Native Linux filesystems store Unix permissions on disk and do not support
+# umask mount options.
+# Make the mount root writable by the web server without changing its contents.
+case "$UM_FILESYSTEM" in
+    ext2|ext3|ext4|btrfs|xfs|f2fs)
+        logger -t "usb_lib (70-usb-library)" "Making $UM_FILESYSTEM mount root writable by {{ apache_user }}"
+        if ! chown "{{ apache_user }}:{{ apache_user }}" "$UM_MOUNTPOINT" ||
+        ! chmod u+rwx "$UM_MOUNTPOINT"; then
```
holta: Both sides of the "||" (if statement's or clauses) should probably stay lined up (:

```suggestion
           ! chmod u+rwx "$UM_MOUNTPOINT"; then
```

diensthunds:
Seems to have worked out. Flash a separate card with base os, then did the install of IIAB with this pull and plugged in another drive I had laying around, turned out it was already ext4 formatted from a different project. Was able to download and upload files through box.local/usb
I'd say make the merger from your pull to main branch.

chapmanjacobd:
Thanks for helping us test this out Chas!

roles/usb_lib/README.rst
```diff
@@ -26,7 +26,7 @@ Technical Details:
...
+* IIAB will generally mount USB sticks / drives 'rw' allowing root to both read and write to them.  In addition, in March 2021 (`PR #2715 <https://github.com/iiab/iiab/pull/2715>`_) Kolibri exports were enabled, by also giving non-root users read and write access to VFAT/FAT32, NTFS and exFAT USB sticks — using ``umask=0000`` (in /etc/usbmount/usbmount.conf) to override the ``umask=0022`` default.  This ``umask=0000`` is also required for students to upload to the teachers's VFAT/FAT32, NTFS and exFAT USB sticks, as introduced in January 2025 (`PR #3875 <https://github.com/iiab/iiab/pull/3875>`_).  For native Linux filesystems (ext2, ext3, ext4, btrfs, xfs and f2fs), which do not support ``umask`` mount options, USB_LIB instead makes the mounted filesystem root writable by the web server when ``usb_lib_writable_sticks: True``.  If, however, you prefer to restore usbmount's default, set ``usb_lib_writable_sticks: False`` in `/etc/iiab/local_vars.yml <http://FAQ.IIAB.IO/#What_is_local_vars.yml_and_how_do_I_customize_it%3F>`_ — please do this prior to installing IIAB — so you don't have to run: ``cd /opt/iiab/iiab ; ./runrole --reinstall usb_lib``
```
holta: "teacher's" indeed better than "teachers's"

```suggestion
* IIAB will generally mount USB sticks / drives 'rw' allowing root to both read and write to them.  In addition, in March 2021 (`PR #2715 <https://github.com/iiab/iiab/pull/2715>`_) Kolibri exports were enabled, by also giving non-root users read and write access to VFAT/FAT32, NTFS and exFAT USB sticks — using ``umask=0000`` (in /etc/usbmount/usbmount.conf) to override the ``umask=0022`` default.  This ``umask=0000`` is also required for students to upload to the teacher's VFAT/FAT32, NTFS and exFAT USB sticks, as introduced in January 2025 (`PR #3875 <https://github.com/iiab/iiab/pull/3875>`_).  For native Linux filesystems (ext2, ext3, ext4, btrfs, xfs and f2fs), which do not support ``umask`` mount options, USB_LIB instead makes the mounted filesystem root writable by the web server when ``usb_lib_writable_sticks: True``.  If, however, you prefer to restore usbmount's default, set ``usb_lib_writable_sticks: False`` in `/etc/iiab/local_vars.yml <http://FAQ.IIAB.IO/#What_is_local_vars.yml_and_how_do_I_customize_it%3F>`_ — please do this prior to installing IIAB — so you don't have to run: ``cd /opt/iiab/iiab ; ./runrole --reinstall usb_lib``
```

diensthunds:
> Thanks for helping us test this out Chas!

You are welcome. I'm glad I could help out.

holta:
@avni ok if we merge this now?

avni:
Yes, thank you, @diensthunds and @chapmanjacobd! 🎉

jvonau:
Would it not be easier to just give the mount point drwxrwxrwx permissions when the directories are created in https://github.com/iiab/iiab/blob/master/roles/usb_lib/tasks/install.yml#L71-L86? Would have to spit that stanza up to get the mode to be 0777 but should that now allow www-data to be able to write to any directory under /media/usbX/? 

Just tested changing only the mode, that change works, no need to chown.

chapmanjacobd:
Easier vs cleaner, I guess? I agree that this can be a bit subjective. I think either www-data or iiab-admin would be more natural mount owners than the default `root`. If we did 0777 then it doesn't really matter (like you say).

If there are specific IIAB uses of USBs which don't use $apache_user then I agree that 0777 could be an easy fix. But a cleaner fix (in my view) would probably be group ACL.

One of my biggest reasons for preferring chown over just chmod is that files and folders created within the folder won't have 0777 permissions (unless you do something like `umask 0000`)... it seems cleaner to me to keep things consistent throughout the directory tree.

jvonau:
The original intended use of usb_lib_writable_sticks was to allow kolibri's export function to work https://github.com/iiab/iiab/issues/3907 but testing I suspect was with MS filesystems only which need umask=0000 anyway. BTW ext4 gets mounted rw but the mountpoint permissions get in the way if you are not root doing the copying.

jvonau:
```
/dev/sda2 on /media/usb1 type ext4 (rw,nodev,noexec,noatime,nodiratime,sync)

jvonau@box:~ $ ls -la /media/usb1
drwxr-xr-x   2 www-data www-data  4096 Sep 14 20:55 UPLOADS.2026-09-14
```

PR #4500, opened 2026-09-12
Maps: Query for FQR files from the front end instead of using an index file

orblivion:

Right now the FQR downloader tool generates a file called `extracts.json` that lists all of the FQRs and their bboxes. That way the front end only has to get this one file and it gets all of that info.

This change will instead allow the front end to inspect the available FQR files on its own. Nginx will give a json-formatted directory listing at `/maps/data/` of all of the files under `/maps/`. Subdirectories work as well, which means fqrs can be inspected. So now the front end looks at `/maps/data/` for any directories called `*.fqr`. For each FQR it reads its `meta.json` for its bbox. It also does another directory listing on each `*fqr` directory itself to find its specific pmtiles file names (since they have an arbitrary date). This change has (2n + 1) queries instead of 1 query for this info, but the files are tiny. And more importantly, changes to the files will be visible right away.

So what's up with the `/maps/data/` subdirectory? I can't have `/maps/` list its files because it has an index.html. I could have put all of the data under `/maps/data/`, and maybe I should eventually. But for now, I just hacked it with a symlink from `/maps/data/` to `/maps/`. Nginx treats it as a different directory which allows us to give it different properties.

*  `extracts.json` no longer exists (and we no longer have to have our discussion of how to change its format)
* `update-json` / `reload` or whatever we were going to call it, no longer exists (though for now I turned it into a "debug" command that gives some warnings)
* Grassroots FQRs - copied by the implementer from a friend, etc - should be detected and show up immediately (assuming file permissions are set correctly?)

It works. I have a curiosity about performance; I thought I saw it load slower a couple times, but not always. Of course I test with a lot of FQRs so that could be not a priority for us anyway. I might have other thoughts about details before merging. I haven't had time to ponder it overall. But I think I'm pretty close to done.

#### In the future I could similarly implement:

* Automatic front-end "self-healing" of world maps after the implementer deletes files. This would replace the CLI utility `maps-update.py` that I made a few weeks back.
* "Debug" warnings about pmtiles files in the front end (so long as they don't require inspecting the file with the `pmtiles` CLI, though even then maybe we could figure something out)

Ubuntu 26.04 on Multipass

- [x] I'd like to try on RPi Zero 2W to see if the performance wrt Ngnix isn't somehow a problem.

@holta @chapmanjacobd

### DIFF
--- a/roles/maps/templates/index.html.j2
+++ b/roles/maps/templates/index.html.j2
@@ -804,7 +804,7 @@
                             // east > west, and `render_bounds` should be the entire width of the world because
                             // it's actually two partial regions on opposite sides of the map
                             dates,
-                              bbox: meta.bbox,
+                            bbox: meta.bbox,
                             ui_bounds: [meta.bbox[0], meta.bbox[1], meta.bbox[2] + 360, meta.bbox[3]],
                             render_bounds: [-179.99999999, meta.bbox[1], 180, meta.bbox[3]],
                           }
--- a/roles/maps/templates/index.html.j2
+++ b/roles/maps/templates/index.html.j2
@@ -119,13 +119,13 @@
                             const pmtilesDates = downloadedFQRegions.regions[fqrName].dates
 
                             const s2Id = "s2maps-sentinel2-2023"
-                            const s2FqrId = makeFqrSourceId(fqrName, s2Id, pmtilesDates.satellite)
+                            const s2FqrId = makeFqrSourceId(fqrName, s2Id, pmtilesDates[s2Id])
                             const osmId = "openstreetmap-openmaptiles"
-                            const osmFqrId = makeFqrSourceId(fqrName, osmId, pmtilesDates.vector)
+                            const osmFqrId = makeFqrSourceId(fqrName, osmId, pmtilesDates[osmId])
                             const terrariumId = "terrarium"
-                            const terrariumFqrId = makeFqrSourceId(fqrName, terrariumId, pmtilesDates.terrain)
+                            const terrariumFqrId = makeFqrSourceId(fqrName, terrariumId, pmtilesDates[terrariumId])
                             const hillshadeId = "hillshade"
-                            const hillshadeFqrId = makeFqrSourceId(fqrName, hillshadeId, pmtilesDates.terrain)
+                            const hillshadeFqrId = makeFqrSourceId(fqrName, hillshadeId, pmtilesDates[terrariumId])
                             const naturalEarth6Id = "naturalearth6-NE2_HR_SR_W_DR-WEBP"
                             const backgroundLayerId = "background"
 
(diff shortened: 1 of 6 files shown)


orblivion:
Okay I think it's ready for full review. In this case I'd recommend reviewing the whole thing, not individual commits.

roles/maps/templates/index.html.j2
```diff
@@ -682,53 +682,132 @@
...
+                          const dates = {}
+
+                          // Each valid pmtiles file in an FQR has the format:
+                          // [regionExtractType].[date].pmtiles
+                          // We want the latest date for each regionExtractType, assuming there are
+                          // (for some reason) multiple. If we sort it in alphabetical order, the
+                          // pmtiles for each regionExtractType should also be listed in chronological
+                          // order. So long as we overwrite, the latest date for each regionExtractType
+                          // should end up in `dates`.
+                          listing.sort()
```
chapmanjacobd: this seems fragile / might not be working correctly AS-IS because listing is an array of objects so each object is stringified into '[object Object]'

try something like this instead?

    listing.sort((a, b) => a.name.localeCompare(b.name))

roles/maps/templates/maps-nginx.conf.j2
```diff
@@ -19,5 +19,19 @@ location ~ ^/maps/ {
...
+    # This, combined with making data a symlink back to /, gives the browser a directory listing in json format of /maps/ and all subdirectories
+    location ~ ^/maps/data/ {
+        autoindex on;
+        autoindex_format json;
+
+        # Need to disable /maps/data/index.html to see the directory listing, since /maps/index.html exists
+        index does-not-exist-920f18f8cfc29bcbaa0c8430288fa4a919cd2be30a37dc4369b247193aa0d5ab.html;
+
+        # In case some wise guy tries to visit /maps/data/data/data/data/data/data/data/... and causes a problem.
+        location ~ ^/maps/data/data/ {
```
chapmanjacobd: hmmmm.... I wonder if an explicit `alias` might be better?

```nginx
     location ^~ /maps/data/ {
         alias {{ maps_serve_path }}/;  # /library/www/maps/
         autoindex on;
         autoindex_format json;
         index does-not-exist-920f18f8...html;

         location ~ \.(pmtiles)$ {
             chunked_transfer_encoding off;
             gzip off;
         }
     }
```

roles/maps/templates/index.html.j2
```diff
@@ -682,53 +682,132 @@
...
+                          listing.forEach(entry => {
+                              if (entry.type !== "file") {
+                                      return
+                              }
+
+                              const REGION_EXTRACT_TYPES = [
```
chapmanjacobd: move up out of a loop?

roles/maps/templates/tile-extract.py.j2
```diff
@@ -523,23 +501,20 @@ def update_extracts_json():
...
 
-    extracts = {"regions": region_extracts}
-    with EXTRACTS_JSON.open("w") as f:
-        json.dump(extracts, f)
 
 def main():
     match sys.argv[1]:
-        case "update-json":
+        case "debug":
             assert len(sys.argv) == 2, "update-json takes no extra arguments"
```
chapmanjacobd: > update-json

should we update this help text?

chapmanjacobd (approved):
Looks good! I like the overall direction

roles/maps/templates/maps-nginx.conf.j2
orblivion: Thanks. As I mentioned off-channel I was wondering about doing this without the symlink hack.

roles/maps/templates/index.html.j2
orblivion: Totally, I think this was an implicit TODO that I forgot to do anything about. Thanks. Thinking ahead though, I decided to take a more direct approach to sorting by date.

chapmanjacobd (approved):
lgtm 👍👍👍

PR #4501, opened 2026-09-13
misc fixes for Ubuntu 26.10

chapmanjacobd:
### Fixes bug:

https://github.com/iiab/iiab/pull/4497#issuecomment-5656061485

Ubuntu 26.10 includes systemd >=260 which removed SysV init scripts. Installing `uwsgi-initscripts` doesn't actually do anything interesting because `systemd-sysv-generator` is no longer included so we just make some directories and create our own service file for uWSGI.

Also a couple minor changes to make the IIAB install more predictable on 26.10-based containers

    go install github.com/chapmanjacobd/iiab-vm/v2@latest
    iiab-vm build large-ubuntu2610 --base ubuntu26.10 --repo ./iiab --local-vars ./iiab/vars/local_vars_large.yml --size 25000 --disk

@holta


.github/workflows/30min-large-on-U2610.yml
```diff
@@ -0,0 +1,74 @@
...
+      - '**/*.svg'
+      - '**/*.jpg'
+      - '**/*.png'
+      - '**/*.php'
+      - '**/*.js'
+  workflow_dispatch:
+
+jobs:
+  test-install:
+    runs-on: ubuntu-26.10
```
holta: I don't think GitHub Actions offers non-LTS Ubuntu runners (e.g. Ubuntu 26.10, even after it's released on Oct ~15 !)

cc: @Akatama who might know more (:

holta:
@Akatama am I right that GitHub Actions doesn't make it easy to run workflows on non-LTS versions of Ubuntu?

(e.g. Ubuntu 26.10 is just an example here, which Canonical should release on Oct 15!)

.github/workflows/30min-large-on-U2610.yml
chapmanjacobd: yeah... interesting how it just stays waiting instead of failing! Was kinda hoping it would be undocumented but still work

chapmanjacobd:
I'll do a 26.10 run locally to see if I can reproduce the error and see if this patch fixes it:

    go install github.com/chapmanjacobd/iiab-vm/v2@latest
    iiab-vm build large-ubuntu2610 --base ubuntu26.10 --repo ./iiab --local-vars ./iiab/vars/local_vars_large.yml --size 25000 --disk

https://github.com/chapmanjacobd/iiab-vm

chapmanjacobd:
hmmm I get the same error even with this PR...

chapmanjacobd:
oh... systemd 260 permanently removes SysV init scripts but also the systemd-sysv-generator. huh... I think it should be simple enough to just create our own service file then. Maybe we don't even need `uwsgi-initscripts`? I don't get why that package even exists now

chapmanjacobd:
looks like no more `captiveportal` errors after the latest changes

chapmanjacobd:
```

PLAY RECAP *********************************************************************
127.0.0.1                  : ok=1353 changed=586  unreachable=0    failed=0    skipped=314  rescued=0    ignored=8

./iiab-install  COMPLETED IN /opt/iiab/iiab

Admin Console:

PLAY RECAP *********************************************************************
127.0.0.1                  : ok=71   changed=60   unreachable=0    failed=0    skipped=4    rescued=0    ignored=0

```

Akatama:
Nope, 26.04 is still in public preview, which means it's not really even fully ready: https://docs.github.com/en/actions/reference/runners/github-hosted-runners

I doubt that they will bring 26.10, they haven't made non-LTS available before, and pretty much no one runs those in production anyway.

holta:
Works on the current Ubuntu 26.10!  (As confirmed by LARGE-sized IIAB install...)

https://paste.centos.org/view/eac9df19

jvonau:
https://packages.ubuntu.com/stonking/amd64/uwsgi/filelist
Any reason to not use the native uwsgi-app@.service? The only thing that would need to change is to install the ini files into the provided /etc/uwsgi/apps-available/ as apps-enabled went away but is provided by the add-on  uwsgi-initscripts that is ineffective now. Of course you would need to enable each new service individually.


From eac9df19
>Sep 14 19:46:33 u2610-large3 systemd[1]: Failed to start pollinate.service - Pollinate to seed the pseudo random number generator.
Sep 14 19:51:47 box dnsmasq[7599]: failed to create listening socket for port 53: Address already in use
Sep 14 19:51:47 box dnsmasq[7599]: FAILED to start up
Sep 14 19:51:47 box systemd[1]: Failed to start dnsmasq.service - dnsmasq - A lightweight DHCP and caching DNS server.
Sep 14 19:52:03 box avahi-daemon[8728]: dbus_bus_request_name(): Request to own name refused by policy
Sep 14 19:52:03 box systemd[1]: Failed to start avahi-daemon.service - Avahi mDNS/DNS-SD Stack.
Sep 14 19:54:33 box dbus-broker-launch[895]: Invalid user name 'colord' in service file '/usr/share/dbus-1/system-services/org.freedesktop.ColorManager.service'
Sep 14 20:00:30 box systemd[1]: Failed to start mosquitto.service - Mosquitto MQTT Broker.
Sep 14 20:00:31 box systemd[1]: Failed to start mosquitto.service - Mosquitto MQTT Broker.
Sep 14 20:00:31 box systemd[1]: Failed to start mosquitto.service - Mosquitto MQTT Broker.
Sep 14 20:00:31 box systemd[1]: Failed to start mosquitto.service - Mosquitto MQTT Broker.
Sep 14 20:00:31 box systemd[1]: Failed to start mosquitto.service - Mosquitto MQTT Broker.
Sep 14 20:00:31 box systemd[1]: Failed to start mosquitto.service - Mosquitto MQTT Broker.
Sep 14 20:04:57 box systemd[1]: Failed to start kolibri.service - LSB: kolibri daemon, an offline education platform.
Sep 14 20:04:58 box systemd[1]: Failed to start kolibri.service - LSB: kolibri daemon, an offline education platform.
Sep 14 20:06:10 box systemd[1]: Failed to start fwupd-refresh.service - Refresh fwupd metadata and update motd.
Sep 14 20:12:32 box resolvconf[54684]: Failed to set DNS configuration: Link lo is loopback device.
Sep 14 20:16:53 box nginx[60736]: box nginx: 2026/09/14 20:16:53 [error] 60736#60736: *2 directory index of "/library/www/html/home/" is forbidden, client: 127.0.0.1, server: box, request: "HEAD /home/ HTTP/1.1", host: "localhost"
Sep 14 20:21:20 box IIAB-CMDSRV[73045]: IIAB-CMDSRV : Kiwix Catalog json file not found

Some things seem to be broken

chapmanjacobd:
Great points! thanks for catching these two things. I'm not too familiar with uWSGI

Transitioning to uwsgi-app@.service would be a wider change as it runs each service as a different dynamic user under www-%s instead of reusing www-data. Probably a better choice long-term but it does mean that we need to handle things a bit differently here and in Admin Console...

Debian 12/13 and RasPiOS still use the classic apps-enabled + SysV-generated uwsgi.service Emperor. Adopting native units only to satisfy 26.10 forks IIAB's uwsgi handling into two architectures, and the admin-console repo can only serve one.

maybe we would be wanting the uwsgi-emperor package:

https://packages.ubuntu.com/stonking/amd64/uwsgi-emperor

https://packages.debian.org/trixie/uwsgi-emperor

?

I'll try a build with that instead

tim-moody:
>runs each service as a different dynamic user under www-%s instead of reusing www-data

I don't know where www-%s comes from, but Adm Cons needs its uwsgi wrapper to run as www-data, and its ini is so configured.

jvonau:
I'd make a clean break now and use the native unit file, debian ships the same unit file with the 2 edits, to the path and unit file name to admin-console and capriveportal.

jvonau:
> > runs each service as a different dynamic user under www-%s instead of reusing www-data
> 
> I don't know where www-%s comes from, but Adm Cons needs its uwsgi wrapper to run as www-data, and its ini is so configured.

"should be fine" given that www-data is defined in https://github.com/iiab/iiab-admin-console/blob/master/roles/console/templates/admin-console.ini.j2

PR #4503, opened 2026-09-16
Fix checking if the user is totally viewing within an FQR.

orblivion:
### Partially Fixes bug:

There is a bug wherein beige land areas in vector maps in FQRs, at least in islands, show up as water. I have a suspicion that this has to do with water from the global map overlapping it. This bug happens everywhere on the map, and this PR is not a general fix for it.

Back when I moved FQRs on top of the world map, I started to notice some performance issues, so I added an optimization that would hide the world map if you zoom in completely on an FQR. Hiding the world map in this way accidentally fixes the above bug with water overlapping the land. However it is incomplete, because if you zoom out enough you see the problem still.

With that background in mind: there is a second bug in the method I used to check for whether you are zoomed entirely on an FQR, particularly near the antimeridian. So FQRs of islands in Fiji have this land disappearing problem even when you are zoomed in. This PR fixes that particular issue, so you can at least inspect Fiji in vector mode. However, again, if you zoom out, the issue persists as it does elsewhere on the planet.

Ubuntu 26.04 on Multipass

@holta @chapmanjacobd


roles/maps/templates/index.html.j2
```diff
@@ -528,7 +528,15 @@
                 // Does SOME region contain EVERY corner of the map bounds?
                 return regionsOnScreen
                 .map(regionName => new maplibregl.LngLatBounds(downloadedFQRegions.regions[regionName].ui_bounds))
-                .some(regionBounds => mapCorners.every(lngLat => regionBounds.contains(lngLat)))
+                .some(regionBounds =>
+                    // Since ui_bounds can have eastern coordinates past 360 (when it crosses the antimeridian),
+                    // try shifting the mapCorners east by 360 to try all possibilities.
```
orblivion: If this comment is unconvincing I can elaborate it.

orblivion: Robot was unconvinced so I elaborated.

chapmanjacobd (approved):
Yes I think this is perfectly adequate for MapLibre

PR #4505, opened 2026-09-18
Feature: add forgejo role for multiple archs

Ark74:
### Fixes bug:

Adds the `forgejo` role for multiple archs, proot included.

Add a new `forgejo` role: an offline, self-hosted Git service (Forgejo, a community fork of Gitea). Sibling of the existing gitea role.

- Install: download the pinned Forgejo LTS binary from Codeberg, verify the   GPG signature, and link a stable symlink.
- Config: render `app.ini`. Secrets are generated once and referenced by file   (`SECRET_KEY_URI`, `INTERNAL_TOKEN_URI`, `JWT_SECRET_URI`).
- Database: SQLite by default (no extra service). MariaDB is optional   (`forgejo_db_engine: mariadb`).
- Platforms: systemd (x86_64, arm64) and proot (via pdsm). The pdsm wrapper   lives in the `proot_services` role.
- Network: HTTP behind nginx at `/forgejo`. The nginx config raises   `client_max_body_size` for large repos. Built-in SSH runs on systemd only;   proot is HTTP only for now.
- Update: a separate task (`tasks/update.yml`) backs up, swaps the binary, and   restarts. It never runs during install.
- Off by default in `default_vars` and all profiles. Registered in the `0-init`   checks and stage `6-generic-apps`.

Static checks: `ansible-lint` passes the `production` profile.

Not in this PR (planned next): SQLite -> MariaDB engine migration, git SSH under proot, and repo seeding (done downstream in iiab-android).

WIP. Draft PR. Only static checks pass so far (ansible-lint production profile, YAML and Jinja validation).
Device tests are planned on x86_64 and arm64 (systemd) and Termux/proot before this leaves draft.

...


holta:
@orblivion @chapmanjacobd

Do you have any suggestions?

chapmanjacobd:
> Update: a separate task (tasks/update.yml) backs up, swaps the binary, and restarts. It never runs during install.

Is there an existing precedent for this type of update task? Maybe this could be a shell script and systemd timer instead of in Ansible?

holta:
> > Update: a separate task (tasks/update.yml) backs up, swaps the binary, and restarts. It never runs during install.
> 
> Is there an existing precedent for this type of update task? Maybe this could be a shell script and systemd timer instead of in Ansible?

@Ark74 is the purpose of `tasks/update.yml` to update Forgejo software while trying to keeps users' artifacts intact?

1. Typically IIAB roles solve this as follows...

   ```
   cd /opt/iiab/iiab
   sudo ./runrole --reinstall forgejo
   ```

   In other words, it is in the end possible to instrument `roles/forgejo/tasks/install.yml` to achieve both functions — (A) initial install of Forgejo software and (B) subsequent update of Forgejo software, while delicately trying to preserve users' work 🙏

2. Or if really necessary, we can also consider something like `/usr/bin/iiab-update-forgejo` ?

Ark74:
> @Ark74 is the purpose of tasks/update.yml to update Forgejo software while trying to keeps users' artifacts intact?

That's the plan.

> Or if really necessary, we can also consider something like /usr/bin/iiab-update-forgejo ?

My first impression is that git management platforms might be more delicate than plain binary and regular db platforms. Maybe not, I don't have previous experience automating forgejo updated yet. So, some additional tests might be required :wink: 

Actually I already have a use case for it. So it's a matter of time to determine it.

Ark74:
All device test pass, it is still disabled by default.

```
forgejo_install: False
forgejo_enabled: False
```

Feel free to enable per local_vars use case basis.

PR #4506, opened 2026-09-18
maps: new satellite source

orblivion:

Adds a new satellite source (licensed [CC BY-NC 4.0](https://creativecommons.org/licenses/by-nc/4.0/)) for testing. It works on FQRs, world map z7, and world map z9. I haven't tried on world map z13. ~I haven't actually tried this catalog item on the world map yet. Maybe it works, but it's only available in full quality, i.e. zoom 0-13, i.e. \~250G. For now I add a warning not to use it for world map.~

To use it, set `maps_satellite_fqr_type` to `iiab` (the default is `s2maps`).

While I'm at it (since it was related code), in the front end, in the FQR setup process, I stopped adding sources and layers for missing pmtiles. That seems to remove a bunch of console noise about 404s that I get when I have such missing files.

I also add a "debug" option for the maps catalog (`maps_catalog_debug`) to use the local repository's current copy instead of the remote copy. It makes it easier for tasks such as this one that involves adding a new entry. It's a convenience for me as a developer. We could try to find a better option though if this isn't preferable.

Ubuntu 26.04 Multipass

@holta @chapmanjacobd


roles/maps/templates/index.html.j2
```diff
@@ -82,7 +81,7 @@
             // MapLibre sources have IDs and URLs. For FQRs we use the ID to generate the URL,
             // which is why it looks sort of like a URL path.
             // Make sure this matches get_extract_dir and get_new_extract_paths in tile-extract.py
-            makeFqrSourceId = (fqrName, baseId, dataDate) => `${fqrName}.fqr/${baseId}.${dataDate}`
+            const makeFqrSourceId = (fqrName, baseId, dataDate) => dataDate ? `${fqrName}.fqr/${baseId}.${dataDate}` : null
```
chapmanjacobd, line 84: should we throw here instead of setting null and having errors pop in random places?

chapmanjacobd (approved):
looks good

> SATELLITE_CATALOG_KEY = "{{ maps_satellite_fqr_max_zoom }}" + "-iiab"
    
> get_paths(catalog["satellite"][SATELLITE_CATALOG_KEY])

that means that only zoom level 13 is supported? could people make a smaller global version (like zoom level 7) with tile-extract.py?

roles/maps/templates/index.html.j2
orblivion: Under what condition, `undefined` arguments? What if I log a warning and return `null` in that case? I can't anticipate where this could happen but maybe it'll survive longer?

chapmanjacobd: ie.

```js
const makeFqrSourceId = (fqrName, baseId, dataDate) => {
  if (!dataDate) {
    throw new Error("dataDate is required to create fqrSourceId");
  }
  return `${fqrName}.fqr/${baseId}.${dataDate}`;
};
```

orblivion:
> could people make a smaller global version (like zoom level 7) with tile-extract.py?

In theory `pmtiles extract` could give us the lower zoom level world maps as well. But I think that especially for the big ones it could be memory intensive. Just now I used that method for the z7 and z9 maps. I could be wrong, but it seemed like it didn't write the file until the end. So it's holding the whole thing in memory?

I wonder if large FQRs have a similar issue going on, actually. But we haven't seen that be an issue on a Pi.

roles/maps/templates/index.html.j2
chapmanjacobd: well I'm not going to force you to do it but I think it's a good idea so you don't end up with cryptic errors in various places that try to access the string but get a null instead

orblivion:
Another data point: months ago, trying a z12 extract of a locally saved s2maps OOMed one of the machines I tried it on.

roles/maps/templates/index.html.j2
orblivion: That's a fair critique if you think there's a better way. But just to clarify why I'm doing this:

This has to do with how I'm using it. In short, it's not an error, it's a missing resource. It avoids a bunch of noise and if statements:

```
const satelliteFqrId = (
    makeFqrSourceId(fqrName, s2IiabKey, pmtilesDates[s2IiabKey]) ||
    makeFqrSourceId(fqrName, s2Id, pmtilesDates[s2Id])
)
```

And then of course I now check if it's `null` before creating the sources and layers. This way I just don't create them instead of having a bunch of 404s from `undefined` in the pmtiles URL as I had before, when I delete some of the files within the FQR.

I could rename it to `tryMakeFqrSourceId` or something to clarify that `null` is a valid return value. Would that help at all? And/or rename the variables to something like `maybeS2Id` etc?

(Ideally we could have some sort of `Maybe` type...)

chapmanjacobd: I guess it's fine as-is but it might be better to just inline the template string? an if statement would be clearer and then you don't accidentally use this function somewhere unhandled

orblivion: Could you clarify what you mean by inlining the template string?

chapmanjacobd: maybe something like this? 

```diff
diff --git c/roles/maps/templates/index.html.j2 i/roles/maps/templates/index.html.j2
index b39538cb9..ac2badf11 100644
--- c/roles/maps/templates/index.html.j2
+++ i/roles/maps/templates/index.html.j2
@@ -79,11 +79,10 @@
         <script src="maps.black-component.js" type="module"></script>
         <script>
             // MapLibre sources have IDs and URLs. For FQRs we use the ID to generate the URL,
             // which is why it looks sort of like a URL path.
             // Make sure this matches get_extract_dir and get_new_extract_paths in tile-extract.py
-            const makeFqrSourceId = (fqrName, baseId, dataDate) => dataDate ? `${fqrName}.fqr/${baseId}.${dataDate}` : null
 
             // we want style choice to survive the map reset that happens when a maps.black
             // feature or style changes, so we rely on this instead of the select tag value
             let mapStyleChoice
             // keep track of which layer ids refer to OSM layers. At some points we want to
@@ -122,19 +121,18 @@
                             // that might need to be changed. "iiab-sentinel2" is just the name of a key in `pmtilesDates`, and will
                             // be part of the name of a new source for the FQR (and eventually probably the whole name of the source
                             // for the world map) that we will create for "iiab-sentinel2" imagery.
                             const s2IiabKey = "iiab-sentinel2"
                             const satelliteFqrId = (
-                              makeFqrSourceId(fqrName, s2IiabKey, pmtilesDates[s2IiabKey]) ||
-                              makeFqrSourceId(fqrName, s2Id, pmtilesDates[s2Id])
+                              (pmtilesDates[s2IiabKey] && `${fqrName}.fqr/${s2IiabKey}.${pmtilesDates[s2IiabKey]}`) ||
+                              (pmtilesDates[s2Id] && `${fqrName}.fqr/${s2Id}.${pmtilesDates[s2Id]}`)
                             )
                             const osmId = "openstreetmap-openmaptiles"
-                            const vectorFqrId = makeFqrSourceId(fqrName, osmId, pmtilesDates[osmId])
                             const terrariumId = "terrarium"
-                            const terrainFqrId = makeFqrSourceId(fqrName, terrariumId, pmtilesDates[terrariumId])
+                            const terrainFqrId = pmtilesDates[terrariumId] && `${fqrName}.fqr/${terrariumId}.${pmtilesDates[terrariumId]}`
                             const hillshadeId = "hillshade"
-                            const hillshadeFqrId = makeFqrSourceId(fqrName, hillshadeId, pmtilesDates[terrariumId])
+                            const hillshadeFqrId = pmtilesDates[terrariumId] && `${fqrName}.fqr/${hillshadeId}.${pmtilesDates[terrariumId]}`
                             const naturalEarth6Id = "naturalearth6-NE2_HR_SR_W_DR-WEBP"
                             const backgroundLayerId = "background"
 
                             // Make an FQR version of each source
                             worldMapSourceIds.forEach(worldSourceId => {
@@ -148,14 +146,14 @@
                                         }
                                     } else {
                                         console.warn("Missing satellite source for", fqrName)
                                     }
                                 } else if (worldSourceId === osmId) {
-                                    if (vectorFqrId) {
-                                        style.sources[vectorFqrId] = {
+                                    if (pmtilesDates[osmId]) {
+                                        style.sources[`${fqrName}.fqr/${osmId}.${pmtilesDates[osmId]}`] = {
                                             ...style.sources[osmId],
-                                            url: style.sources[osmId].url.replace(osmId, vectorFqrId),
+                                            url: style.sources[osmId].url.replace(osmId, `${fqrName}.fqr/${osmId}.${pmtilesDates[osmId]}`),
                                             bounds: downloadedFQRegions.regions[fqrName].render_bounds,
                                             minzoom: WORLD_MAP_VECTOR_MAX_ZOOM + 1
                                         }
                                     } else {
                                         console.warn("Missing vector source for", fqrName)
@@ -197,15 +195,15 @@
                                     }
                                 } else if (worldLayer.source === osmId) {
                                     // This is unrelated to copying FQR layers. It could be in a differnet loop.
                                     worldMapOSMLayers.add(worldLayer.id)
 
-                                    if (vectorFqrId) {
+                                    if (pmtilesDates[osmId]) {
                                         return {
                                             ...worldLayer,
-                                            id: vectorFqrId + '.' + worldLayer.id,
-                                            source: vectorFqrId,
+                                            id: `${fqrName}.fqr/${osmId}.${pmtilesDates[osmId]}` + '.' + worldLayer.id,
+                                            source: `${fqrName}.fqr/${osmId}.${pmtilesDates[osmId]}`,
                                         }
                                     }
                                 } else if (worldLayer.source === hillshadeId) {
                                     if (hillshadeFqrId) {
                                         return {
@@ -599,11 +597,12 @@
                     }
                     if (regionsOnScreen.length) {
                         // awkward situation to be looking at multiple regions,
                         // so take the first one (which is the closest to the
                         // map center, see sorting above)
-                        return makeFqrSourceId(regionsOnScreen[0], 'terrarium', downloadedFQRegions.regions[regionsOnScreen[0]].dates.terrarium)
+                        const terrainDate = downloadedFQRegions.regions[regionsOnScreen[0]].dates.terrarium
+                        return terrainDate && `${regionsOnScreen[0]}.fqr/terrarium.${terrainDate}`
                     }
                     if (WORLD_MAP_TERRAIN_MAX_ZOOM > 0) {
                         // If no FQRs are on screen and we have a world map terrain, return to it
                         return 'terrarium'
                     }
@@ -672,12 +671,13 @@
                 const regionsSet = new Set(Object.keys(downloadedFQRegions.regions))
                 const regionsOnScreenSet = new Set(regionsOnScreen)
                 const regionsOffScreenSet = regionsSet.difference(regionsOnScreenSet)
 
                 // TODO
-                //   Define a func makeFqrLayerId next to makeFqrSourceId that's does something like this:
-                //     (worldLayerId, regionName, ...) => makeFqrSourceId(regionName, ...) + '.' + worldMapOSMLayers
+                //   Define a func makeFqrLayerId that uses the FQR source ID format and
+                //     appends '.' + worldMapOSMLayers:
+                //     (worldLayerId, regionName, ...) => `${regionName}.fqr/...` + '.' + worldMapOSMLayers
                 //   Use it where we define the layer ID. But also use it here. Something like:
                 //
                 //   regions(On/Off)ScreenSet.forEach(regionName => {
                 //     worldMapOSMLayers.forEach(worldLayerId => {
                 //       mb.map.setLayoutProperty(

```

---

here's another approach that might be preferable but is a bit more code churn

```diff
diff --git c/roles/maps/templates/index.html.j2 i/roles/maps/templates/index.html.j2
index b39538cb9..e2a50c2d5 100644
--- c/roles/maps/templates/index.html.j2
+++ i/roles/maps/templates/index.html.j2
@@ -79,11 +79,10 @@
         <script src="maps.black-component.js" type="module"></script>
         <script>
             // MapLibre sources have IDs and URLs. For FQRs we use the ID to generate the URL,
             // which is why it looks sort of like a URL path.
             // Make sure this matches get_extract_dir and get_new_extract_paths in tile-extract.py
-            const makeFqrSourceId = (fqrName, baseId, dataDate) => dataDate ? `${fqrName}.fqr/${baseId}.${dataDate}` : null
 
             // we want style choice to survive the map reset that happens when a maps.black
             // feature or style changes, so we rely on this instead of the select tag value
             let mapStyleChoice
             // keep track of which layer ids refer to OSM layers. At some points we want to
@@ -121,62 +120,56 @@
                             // "Key" instead of "Id" here because right now maps.black actually has a source "s2maps-sentinel2-2023"
                             // that might need to be changed. "iiab-sentinel2" is just the name of a key in `pmtilesDates`, and will
                             // be part of the name of a new source for the FQR (and eventually probably the whole name of the source
                             // for the world map) that we will create for "iiab-sentinel2" imagery.
                             const s2IiabKey = "iiab-sentinel2"
-                            const satelliteFqrId = (
-                              makeFqrSourceId(fqrName, s2IiabKey, pmtilesDates[s2IiabKey]) ||
-                              makeFqrSourceId(fqrName, s2Id, pmtilesDates[s2Id])
-                            )
+                            const satelliteKey = pmtilesDates[s2IiabKey] ? s2IiabKey : s2Id
                             const osmId = "openstreetmap-openmaptiles"
-                            const vectorFqrId = makeFqrSourceId(fqrName, osmId, pmtilesDates[osmId])
                             const terrariumId = "terrarium"
-                            const terrainFqrId = makeFqrSourceId(fqrName, terrariumId, pmtilesDates[terrariumId])
                             const hillshadeId = "hillshade"
-                            const hillshadeFqrId = makeFqrSourceId(fqrName, hillshadeId, pmtilesDates[terrariumId])
                             const naturalEarth6Id = "naturalearth6-NE2_HR_SR_W_DR-WEBP"
                             const backgroundLayerId = "background"
 
                             // Make an FQR version of each source
                             worldMapSourceIds.forEach(worldSourceId => {
                                 if (worldSourceId === s2Id) {
-                                    if (satelliteFqrId) {
+                                    if (pmtilesDates[satelliteKey]) {
+                                        const satelliteFqrId = `${fqrName}.fqr/${satelliteKey}.${pmtilesDates[satelliteKey]}`
                                         style.sources[satelliteFqrId] = {
                                             ...style.sources[s2Id],
                                             url: style.sources[s2Id].url.replace(s2Id, satelliteFqrId),
                                             bounds: downloadedFQRegions.regions[fqrName].render_bounds,
                                             minzoom: WORLD_MAP_SATELLITE_MAX_ZOOM + 1
                                         }
                                     } else {
                                         console.warn("Missing satellite source for", fqrName)
                                     }
                                 } else if (worldSourceId === osmId) {
-                                    if (vectorFqrId) {
+                                    if (pmtilesDates[osmId]) {
+                                        const vectorFqrId = `${fqrName}.fqr/${osmId}.${pmtilesDates[osmId]}`
                                         style.sources[vectorFqrId] = {
                                             ...style.sources[osmId],
                                             url: style.sources[osmId].url.replace(osmId, vectorFqrId),
                                             bounds: downloadedFQRegions.regions[fqrName].render_bounds,
                                             minzoom: WORLD_MAP_VECTOR_MAX_ZOOM + 1
                                         }
                                     } else {
                                         console.warn("Missing vector source for", fqrName)
                                     }
                                 } else if (worldSourceId === terrariumId) {
-                                    if (terrainFqrId) {
+                                    if (pmtilesDates[terrariumId]) {
+                                        const terrainFqrId = `${fqrName}.fqr/${terrariumId}.${pmtilesDates[terrariumId]}`
+                                        const hillshadeFqrId = `${fqrName}.fqr/${hillshadeId}.${pmtilesDates[terrariumId]}`
                                         style.sources[terrainFqrId] = {
                                             ...style.sources[terrariumId],
                                             url: style.sources[terrariumId].url
                                                 .replace(terrariumId, terrainFqrId)
                                                 .replace(/\-z0\-z10\.pmtiles$/, '.pmtiles'),
                                             bounds: downloadedFQRegions.regions[fqrName].render_bounds,
                                             minzoom: WORLD_MAP_TERRAIN_MAX_ZOOM + 1
                                         }
-                                        if (hillshadeFqrId) {
-                                            style.sources[hillshadeFqrId] = style.sources[terrainFqrId]
-                                        } else {
-                                            console.warn("Missing hillshade source for", fqrName)
-                                        }
+                                        style.sources[hillshadeFqrId] = style.sources[terrainFqrId]
                                     } else {
                                         console.warn("Missing terrain source for", fqrName)
                                     }
                                 } else if (worldSourceId === naturalEarth6Id) {
                                     // pass, only world map
@@ -186,34 +179,37 @@
                             })
 
                             // Make an FQR version of each layer
                             worldMapLayers.map(worldLayer => {
                                 if (worldLayer.source === s2Id) {
-                                    if (satelliteFqrId) {
+                                    if (pmtilesDates[satelliteKey]) {
+                                        const satelliteFqrId = `${fqrName}.fqr/${satelliteKey}.${pmtilesDates[satelliteKey]}`
                                         return {
                                             ...worldLayer,
                                             id: satelliteFqrId + '.' + worldLayer.id,
                                             source: satelliteFqrId,
                                         }
                                     }
                                 } else if (worldLayer.source === osmId) {
                                     // This is unrelated to copying FQR layers. It could be in a differnet loop.
                                     worldMapOSMLayers.add(worldLayer.id)
 
-                                    if (vectorFqrId) {
+                                    if (pmtilesDates[osmId]) {
+                                        const vectorFqrId = `${fqrName}.fqr/${osmId}.${pmtilesDates[osmId]}`
                                         return {
                                             ...worldLayer,
                                             id: vectorFqrId + '.' + worldLayer.id,
                                             source: vectorFqrId,
                                         }
                                     }
                                 } else if (worldLayer.source === hillshadeId) {
-                                    if (hillshadeFqrId) {
+                                    if (pmtilesDates[terrariumId]) {
+                                        const hillshadeFqrId = `${fqrName}.fqr/${hillshadeId}.${pmtilesDates[terrariumId]}`
                                         return {
                                             ...worldLayer,
-                                          id: hillshadeFqrId + '.' + worldLayer.id,
-                                          source: hillshadeFqrId,
+                                            id: hillshadeFqrId + '.' + worldLayer.id,
+                                            source: hillshadeFqrId,
                                         }
                                     }
                                 } else if (worldLayer.source === naturalEarth6Id) {
                                     // pass, only world map
                                 } else if (worldLayer.id === backgroundLayerId) {
@@ -599,11 +595,14 @@
                     }
                     if (regionsOnScreen.length) {
                         // awkward situation to be looking at multiple regions,
                         // so take the first one (which is the closest to the
                         // map center, see sorting above)
-                        return makeFqrSourceId(regionsOnScreen[0], 'terrarium', downloadedFQRegions.regions[regionsOnScreen[0]].dates.terrarium)
+                        const terrainDate = downloadedFQRegions.regions[regionsOnScreen[0]].dates.terrarium
+                        if (terrainDate) {
+                            return `${regionsOnScreen[0]}.fqr/terrarium.${terrainDate}`
+                        }
                     }
                     if (WORLD_MAP_TERRAIN_MAX_ZOOM > 0) {
                         // If no FQRs are on screen and we have a world map terrain, return to it
                         return 'terrarium'
                     }
@@ -672,12 +671,13 @@
                 const regionsSet = new Set(Object.keys(downloadedFQRegions.regions))
                 const regionsOnScreenSet = new Set(regionsOnScreen)
                 const regionsOffScreenSet = regionsSet.difference(regionsOnScreenSet)
 
                 // TODO
-                //   Define a func makeFqrLayerId next to makeFqrSourceId that's does something like this:
-                //     (worldLayerId, regionName, ...) => makeFqrSourceId(regionName, ...) + '.' + worldMapOSMLayers
+                //   Define a func makeFqrLayerId that uses the FQR source ID format and
+                //     appends '.' + worldMapOSMLayers:
+                //     (worldLayerId, regionName, ...) => `${regionName}.fqr/...` + '.' + worldMapOSMLayers
                 //   Use it where we define the layer ID. But also use it here. Something like:
                 //
                 //   regions(On/Off)ScreenSet.forEach(regionName => {
                 //     worldMapOSMLayers.forEach(worldLayerId => {
                 //       mb.map.setLayoutProperty(

```

orblivion: Okay I sort of thought that's what you might mean. I don't prefer the inlining because it's repeated code. I think I used to do something like that actually. However that doesn't mean I have to optionally return null as I do. We could come up with something else.

If you think this change is at least passable for now, I'd like to move on with this and figure out something better later. This part of the code is arranged in a confusing way to begin with. May be a good time to clean it all up and address your concern with the nulls as I do.

chapmanjacobd: yeah like I said it's not a blocking issue and it is subjective as long as the function is never exported

holta:
Looking Great! 🏞️

